PR #175, opened 2014-10-04
Adding start and len functionality for RTMP clipping

joeylgutierrez:
I've implemented the start and len features per http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/net/NetStream.html#play() for a personal project to allow a user to supply those parameters and only get that part of the stream.  The start param will cut off the stream content prior to that point and the len will clip the end of the stream after the duration length is met.  I saw that start and duration were already implemented to seek forward and really duration didn't seem to do anything for me.  I didn't know where to post this so i'm just submitting this as a pull-request if the project is interested in this functionality.

My example would be like
jwplayer("myElement").setup({
               primary: 'flash',
               playlist: [
                   {
                       start: start,
                       len: len,
                       sources: [
                           {file: "rtmp://myserver/void/mystream"}
                       ]
                   }
               ]
           });

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/media/RTMPMediaProvider.as
+++ b/src/flash/com/longtailvideo/jwplayer/media/RTMPMediaProvider.as
@@ -312,9 +312,18 @@
 						// Only send initial metadata
 						_metadata = true;
 						if (data.duration) {
+							
+							if(data.duration != _item.len && _item.len > -1)
+							{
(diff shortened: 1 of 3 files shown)


robwalch:
Closing as we do not have plans to implement or support playlist item start/duration/len params. Modify your streams and/or use seek onPlay.

PR #187, opened 2014-10-15
prevent duplicate captions if YouTube video has yt:cc=on

BenEfrati:
If YouTube video has yt:cc=on attribute when embedding the video in html5 mode, captions loads by default and can't be disabled. If you add caption via jwplayer there will be duplicate captions.
example:
http://jsfiddle.net/UAR3U/25/show/
The only way to disable YouTube captions is via javascript API:
    player.unloadModule("captions");

### DIFF
--- a/src/js/html5/providers/jwplayer.html5.youtube.js
+++ b/src/js/html5/providers/jwplayer.html5.youtube.js
@@ -239,7 +239,10 @@
                     return;
 
                 case youtubeStates.PLAYING: // 1: playing
-
+                
+//prevent duplicate captions when using JW Player captions and YT video has yt:cc=on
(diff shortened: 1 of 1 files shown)


DonatoB:
@BenEfrati, this fails jshint, please fix the quotes, indent the comment properly and give it a defensive wrapper like:
   
    if (_.isFunction(_youtubePlayer.unloadModule)) { ... }

    243 |                    _youtubePlayer.unloadModule("captions");
                                                       ^ Strings must use singlequote.

Then I will merge this.

DonatoB:
Moving this to https://github.com/jwplayer/jwplayer/pull/239. I set up a test page but cannot reproduce the original problem this was meant to solve. I will merge it after we can prove this fixes something.

PR #214, opened 2014-11-19
Audio Panning support (Flash only)

adinapoli:
Hello guys,

we are Iris Connect, a company which is using your player (with a Premium license) as part of our core business. We tried to contact the support team about adding audio panning for the flash player, but unfortunately it wasn't on the product roadmap.

As we needed it as part of an upcoming feature milestone we need to deliver, we forked the player and added support for panning (flash player only). It has ben tested with local video files, remote one and even RTMP. This patch exposes a new `jwplayer` method called `setPan`, which takes an integer from -100 being total left pan and +100 being right pan. 
As an example usage, here is how you set left panning:

``` javascript
var player = jwplayer()
player.setPan(-100)
```

We have no choice but maintain this fork to support our business, but it would be way better if we could make this merged upstream, to avoid keeping everything in sync.
Please let us know if you need us to amend/improve our code.

Thanks!
Alfredo

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ node_modules
 bin-debug
 bin-release
 examples/panning/cheetah.mp4
+examples/panning/server.go
--- a/.gitignore
+++ b/.gitignore
@@ -8,3 +8,5 @@ bin-debug
 bin-release
 examples/panning/cheetah.mp4
 examples/panning/server.go
+examples/panning/server
+examples/panning/.DS_Store
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@ node_modules
 *.iml
 bin-debug
 bin-release
+examples/panning/cheetah.mp4
--- a/.gitignore
+++ b/.gitignore
@@ -6,7 +6,4 @@ node_modules
 *.iml
 bin-debug
 bin-release
-examples/panning/cheetah.mp4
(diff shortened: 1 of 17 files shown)


DonatoB:
This is a neat commit, thank you for sharing it. 

We will not be merging it in, since it is a rather niche improvement, and we want to keep file size low for the rest of the world. This would make more sense as a 3rd-party plugin in the future.

adinapoli:
Hi @DonatoB , thanks for taking the time reviewing this.
Can you please tell me how I can reduce file size for this patch, is this is the only blocker?
We are premium customers of yours, and at the moment we cannot use the premium version of the player as we need to maintain a custom one, as we need this feature for our core business (cc @mrmattnewell for context).

I'm also happy to explore the plugin option, even though is my understanding this feature touches the internals of the video player, and as such is not feasible to implement as a plugin.

Thanks,

Alfredo Di Napoli

robwalch:
Hi @adinapoli,

We do not introduce new features in the player that touch the api without having parity between flash and html5. The api implementation here looks nice :ok_hand:. Implementing the feature in html5 however would involve diving into mediasource and webaudio apis. Not to mention adding support to our commercial media providers...

Customer asks and browser support play a big roll in dictating what we merge into and support in future releases. We hear you on this one and will work on finding it a place in the roadmap.

PR #217, opened 2014-11-25
[jwplayer] fixed infinite buffering issue on mobile [Delivers #83128276]

emaxsaun:
If you try to chain play() after onReady(), the player will get stuck in an infinite buffering state on mobile browsers. This small patch fixes it. By not putting the player into a buffering state during __completeLoad() for mobile, the player doesn't need to try to buffer, and do something that can't be done (autostart on mobile).

### DIFF
--- a/src/js/html5/providers/jwplayer.html5.video.js
+++ b/src/js/html5/providers/jwplayer.html5.video.js
@@ -331,10 +331,6 @@
 
             _source = _levels[_currentQuality];
 
-            if (!_isMobile) {
-                _this.setState(states.BUFFERING);
-            }
-
             clearInterval(_bufferInterval);
             _bufferInterval = setInterval(_sendBufferUpdate, 100);
 


fdambrosio:
iPad go in infinite loop buffering if I call a simple jwplayer().seek(10); 
without autostart on setup config

emaxsaun:
@ fdambrosio - Are you using the current 6.11 build?

fdambrosio:
Yes I use your hosted version on CDN for Premium user

fdambrosio:
Now it's fixed

fdambrosio:
BUT if I user playlistItem iPad again go in loop buffering opening player. 
this code doesn't work on iPad: 
var only_once =false;
jwplayer().playlistItem(0);
          jwplayer(').onPlay (function () {  
			  if (only_once == false) {
			  only_once = true;
			  jwplayer().seek(30); 
};
});

emaxsaun:
@fdambrosio - The CDN version does not have this fix applied. Please email support [at] jwplayer [dot] com for further support inquiries.

PR #238, opened 2015-01-05
This addresses a bug when we call setup multiple times before its ready [Delivers #83939432]

DonatoB:


src/js/api/jwplayer.api.js
```diff
@@ -333,21 +337,44 @@
...
+
+            if (toDestroy) {
+                if (_this.renderingMode === 'html5') {
+                    // calls jwPlayerDestroy()
+                    _this.destroyPlayer();
+                }
+
+                // If the tag is reused by another player, do not destroy the div
+                if (! sharedDOM) {
+                    var replacement = document.createElement('div');
```
robwalch: Rather than 'div' should we use the element name of `toDestroy` (eg. toDestroy is actually a span tag `<span id="player"/>`)?

src/js/embed/jwplayer.embed.js
```diff
@@ -161,12 +155,10 @@
             clearTimeout(_setupErrorTimer);
```
robwalch: if `_setupErrorTimer` can be removed.

DonatoB: done

src/js/api/jwplayer.api.js
DonatoB: It will always be a div, since the original element is trashed, copying over only the id. We could do this as a forward compatibility precaution, but it may be confusing.

PR #240, opened 2015-01-09
Defensively protect from pages which modify box-sizing [Delivers #83444132]

DonatoB:


src/js/html5/jwplayer.html5.js
```diff
@@ -28,6 +28,10 @@
         '-webkit-tap-highlight-color': 'rgba(255, 255, 255, 0)'
     });
 
+    // Reset box-sizing to default for player and all sub-elements
+    //  Note: If we use pseudo elements we will need to add *:before and *:after
+    _css(JW_CLASS, JW_CLASS + '*', { 'box-sizing': 'content-box'});
```
robwalch: Looks like one param too many.

robwalch: @DonatoB 
```
_css(JW_CLASS +', '+ JW_CLASS + '*', { 'box-sizing': 'content-box'});
```
ya?

DonatoB: You're exactly right! Thanks.

PR #251, opened 2015-02-02
Video player not loading after error

superman07:
After a jwplayer instance encounters an error during initialisation (for
example if remove is called before it is ready), then any further
jwplayer instances will not load.

### DIFF
--- a/src/js/html5/jwplayer.html5.model.js
+++ b/src/js/html5/jwplayer.html5.model.js
@@ -146,7 +146,7 @@
                 });
             } else {
                 _model.sendEvent(events.JWPLAYER_PLAYLIST_LOADED, {
-                    playlist: jwplayer(_model.id).getPlaylist()
+                    playlist: jwplayer(_model.id).getPlaylist && jwplayer(_model.id).getPlaylist() || []
                 });
                 _model.item = -1;
                 _model.setItem(0);


DonatoB:
Thanks @superman07, we will look into this and give you an update.

DonatoB:
@superman07 , we cannot reproduce the stated error, can you post reproduction steps, or host a demo .html file were we can see it in action?

superman07:
Hi @DonatoB,

I've been unable to successfully setup an example file for you presently, but the steps to reproduce are as follows:
- setup the player for an element with id "A"
- call jwplayer("A").remove() before the player is fully initialised or ready, noting a jwplayer exception  that was caught and logged in the console. This is not a "undefined is not a function" error, but rather an exception with the API information included
- setup the player for an element with id "B" and observe the player is not properly loaded, leaving a blank space where the player should have been

Cheers
superman07

dannyfinks:
@superman07 having no luck reproducing this. Would you be able to provide the relevant code so we can try to reproduce it locally?

emaxsaun:
@superman07 - Do you have a link that you can provide as well?

emaxsaun:
@superman07 I am going to close this request for the time being. Please email support [at] jwplayer [dot] com directly if you wish to continue the conversation about this issue.

PR #262, opened 2015-02-12
Fix IE8 issue embedding jwplayer

tyson-benson:
When JWPlayer is initialised as a descendant of a p tag the call to _container.outerHTML = html; causes an error to be thrown (Unknown runtime error).

See support question for details:
http://support.jwplayer.com/customer/en/portal/questions/11392466-ie8-and-p-tags?new=11392466

### DIFF
--- a/src/js/embed/jwplayer.embed.flash.js
+++ b/src/js/embed/jwplayer.embed.flash.js
@@ -150,14 +150,14 @@
             storedFlashvars[_container.id] = params;
 
             if (utils.isMSIE()) {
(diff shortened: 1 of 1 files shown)


tyson-benson:
I have created this pull request in order to track the code based resolution of this bug. I expect the code committed will not resemble the change in the pull request. But as of now, this code works (when jQuery is present).

robwalch:
Thanks for pointing out the bad quotes. We can't merge code that requires jQuery. I've tried to duplicate your issue by wrapping a player in a p tag, but did not see any errors.

robwalch:
Moved to #263

PR #270, opened 2015-02-23
AMD-ify the open source code base

DonatoB:
[Finishes #88909720]


src/js/__depricated__/jwplayer.embed.flash.js
```diff
@@ -247,7 +247,7 @@
         if (utils.isRtmp(file, type)) { return true; }
         if (type === 'hls') { return true; }
 
-        var mappedType = utils.extensionmap[type ? type : utils.extension(file)];
+        var mappedType = utils.extensionmap.getMappedType(type ? type : utils.extension(file));
```
robwalch: ternary conditionals like this should just use "or" operator. It's a dep script so ignoring.

PR #277, opened 2015-03-02
Set up example for templating html fragments using _.template and external styles

DonatoB:
[#89112506]

@jw-kaurand @robwalch


src/css/styles.less
```diff
@@ -0,0 +1,19 @@
+@JW_CSS_SMOOTH_EASE: opacity .25s ease;
+
+.jwplayer {
+    position: relative;
```
robwalch: For some reason `position: 'relative'` is not getting set, making the player fill the page.

PR #278, opened 2015-03-03
Bugfixes for playlist source filtering, from failing qunit tests

DonatoB:


src/js/playlist/playlist.js
```diff
@@ -30,34 +30,38 @@ define([
...
+    var _filterSources = Playlist.filterSources = function (sources, androidhls) {
+        sources = _.compact(_.map(sources, function(originalSource) {
+            if (! _.isObject(originalSource)) {
                 return;
             }
 
-            if (chooseProvider(source)) {
-                // We want sources of all the same type since they may be of different quality levels
-                selectedType = selectedType || source.type;
+            originalSource.androidhls =  androidhls;
```
robwalch: I added androidhls here, but this and other params from the top level config that providers need to know about should be handled in playlist normalization.

robwalch:
@DonatoB is there a commercial component to this you haven't committed yet? It's broken after this merge so I'm going to submit a fix, but I'd like you to extend Providers with an add method so that the Providers.providers array is private.

DonatoB:
I do have a fix, but I didn't find a way to make the array private without
adding an addProvider to open source. Or alternately duplicating the whole
providers file in commercial.

PR #280, opened 2015-03-05
Implement config.primary for preferring flash player

DonatoB:
[Finishes #88935292]

### DIFF
--- a/src/js/providers/providers.js
+++ b/src/js/providers/providers.js
@@ -6,33 +6,51 @@ define([
     ], function(html5, flash, youtube, _) {
 
 
-    function chooseProvider(source) {
-        // prevent throw on missing source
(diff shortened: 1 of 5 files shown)


src/js/providers/providers.js
```diff
@@ -6,33 +6,51 @@ define([
...
-    var priority = function(p) {
-        var idx = _.indexOf(providers, p);
-        if (idx < 0) {
-            // No provider matched
-            return Number.MIN_VALUE;
+    // When choosing a provider, go through this array
+    //   and select the first that works for the source
+    Providers.defaultList = [html5, flash, youtube];
+
+    _.extend(Providers.prototype, {
```
robwalch: Not sure what the experts say, but Function.prototype = {...} works, and on the surface is fewer operations.

PR #281, opened 2015-03-05
Fix errorscreen for failed embeds [Fixes #88511244]

DonatoB:


src/js/embed/embed.js
```diff
@@ -152,9 +151,9 @@ define([
 
         function _sourceError(evt) {
             if (evt && evt.message) {
-                _errorScreen('Error loading playlist: ' + evt.message);
+                _errorScreen('Error loading playlist', evt.message);
             } else {
-                _errorScreen(_errorText + 'No playable sources found');
+                _errorScreen('Error loading player: ', 'No playable sources found');
```
robwalch: Also a playlist error.

src/js/playlist/source.js
```diff
@@ -33,6 +33,10 @@ define([
             _source.type = extensionmap.extType(extension);
         }
 
+        if (!_source.type) {
+            return;
+        }
+
```
robwalch: Oh no! Worthy of a specific error "Could not determine media type", "{..file end string}"

PR #282, opened 2015-03-06
Decoupling the View from the Player

jw-kaurand:
Changes make it so that the view and its components do not rely on a reference to the player in order to function.  Instead they use references to the model and the api to get the necessary data and listen to the necessary events to function.  Some components (controlbar, display, and others) retain a vestigal reference to the skin,which is passed from the player.  This will only remain until we remove the skin code in favor of templates and css.

Moved a piece of muting logic into the model from the controlbar since it's a better fit.  [Finishes #89112506]


src/js/view/captions.js
```diff
@@ -75,17 +76,18 @@ define([
         function _init() {
 
             _display = document.createElement('div');
-            _display.id = _api.id + '_caption';
+            _display.id = _api.getContainer().id + '_caption';
```
DonatoB: good thinking!

src/js/view/controlbar.js
```diff
@@ -60,10 +60,9 @@ define([
 
 
     /** HTML5 Controlbar class * */
-    var Controlbar = function(_api, _config) {
-        _config = _config || {};
-        var _skin,
-            _dividerElement = _layoutElement('divider', CB_DIVIDER),
+    var Controlbar = function(_skin, _api, _model) {
+        _config = _model.componentConfig('controlbar') || {};
```
robwalch: ```_config``` is now undeclared; global namespace pollution. Make sure you lint these kinds of errors before a commit and PR.

src/js/view/display.js
```diff
@@ -53,19 +52,20 @@ define([
 
         function _init() {
             _display = document.createElement('div');
-            _display.id = _controller.id + '_display';
+            _display.id = _api.getContainer().id;
```
robwalch: deleted ```+ '_display';```!

PR #284, opened 2015-03-09
Bugfix for playlist item initialization with empty object for the source...

DonatoB:
...s or levels value

### DIFF
--- a/src/js/playlist/source.js
+++ b/src/js/playlist/source.js
@@ -28,16 +28,20 @@ define([
         // normalize for odd strings
         _source.file = strings.trim('' + _source.file);
 
+        // if type is given as a mimetype
+        if (_source.type && _source.type.indexOf('/') > 0) {
+            _source.type = extensionmap.mimeType(_source.type);
+        }
+
         // If type not included, we infer it from extension
         if (! _source.type) {
-            var extension = strings.extension(_source.file);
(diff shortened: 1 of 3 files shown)


src/js/playlist/source.js
```diff
@@ -31,6 +32,13 @@ define([
         if (! _source.type) {
```
robwalch: down below maybe we should handle the three cases where we set type in else if blocks ~like:
```
if youtube
else if rtmp
else
  check extension
```

src/js/playlist/source.js
```diff
@@ -31,6 +32,13 @@ define([
...
+
+            if (utils.isYouTube(_source.file, _source.type)) {
+                _source.type = 'youtube';
+            }
+            if (utils.isRtmp(_source.file, _source.type)) {
+                _source.type = 'rtmp';
+            }
         }
 
         if (!_source.type) {
```
robwalch: The normalize type cases:
1. ```if (_source.type && _source.type.indexOf('/') > 0) {``` only applies to the user setting the type as a mime type. (file extensions won't have slashes, so this could probably go before line 32.

2. m3u8 and smil are file extensions or the user might set these. Guess it's fine at the end.

robwalch:
@DonatoB have a look at test/embedding-extensions.html. I'm running into a playlist source check problem with this test.

PR #301, opened 2015-03-17
Rewrite of drag seeking functionality

DonatoB:
This change moves the logic for drag seeking (when you click down and slide the cursor along the seekbar)
out of the individual providers and into the javascript controller. This required a few functional changes.
    1. Providers will no longer begin playback after a seek. Instead the controller will decide the correct time to do this.
    2. The "dragging" state is no longer stored in the provider, it is stored in the model.
    3. Since dragging occurs during the "pause" state, we add a visual style flag "jw-flag-dragging" to hide the center play icon.

This fix was required for flash seek dragging to work, but also improves the UX for html5 due to improvements in how
events are being throttled.

[Fixes #89827172]

### DIFF
--- a/src/js/events/events.js
+++ b/src/js/events/events.js
@@ -57,6 +57,7 @@ define([], function() {
         // Controls show/hide 
         JWPLAYER_CONTROLS: 'jwplayerViewControls',
         JWPLAYER_USER_ACTION: 'jwplayerUserAction',
+        JWPLAYER_CONTROLBAR_DRAGGING: 'jwplayerControlbarDragging',
 
         // Instream events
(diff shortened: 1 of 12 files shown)


src/js/events/events.js
```diff
@@ -57,6 +57,7 @@ define([], function() {
         // Controls show/hide 
         JWPLAYER_CONTROLS: 'jwplayerViewControls',
         JWPLAYER_USER_ACTION: 'jwplayerUserAction',
+        JWPLAYER_CONTROLBAR_DRAGGING: 'jwplayerControlbarDragging',
```
robwalch: can we call this 'jwplayerScrubbing'?

PR #304, opened 2015-03-18
Update player version to 7.0

DonatoB:
Since our CDN does not have the correct resources yet, the repo
used for fetching logo and plugins will still be version 6.12
[Delivers #90154864]

### DIFF
--- a/src/js/utils/helpers.js
+++ b/src/js/utils/helpers.js
@@ -3,11 +3,13 @@ define([
     'events/events',
     'underscore'
 ], function(strings, events, _) {
-
-    // This is replaced by compiler
-    var _version = __BUILD_VERSION__;
-
     /*jshint maxparams:5*/
+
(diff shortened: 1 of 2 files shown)


src/js/utils/helpers.js
```diff
@@ -3,11 +3,13 @@ define([
     'events/events',
     'underscore'
 ], function(strings, events, _) {
+    /*jshint maxparams:5*/
 
+    // TODO:: the next lines are a holdover until we update our CDN w/ plugins for 7.0
     // This is replaced by compiler
-    var _version = __BUILD_VERSION__;
+    //var _version = __BUILD_VERSION__;
+    var _version = 6.12;
```
robwalch: woops! should be a string.

emaxsaun: @robwalch package.json needs to be updated for commercial as well.

PR #305, opened 2015-03-18
IE8 support [Finishes #89718610]

jw-kaurand:


src/js/utils/helpers.js
```diff
@@ -699,6 +699,10 @@ define([
     };
 
     utils.tryCatch = function(fn, ctx, args) {
+        // some browsers require these not be undefined
```
DonatoB: Which browsers?

PR #306, opened 2015-03-18
Create a timer class for QOE, and log setup time

DonatoB:
Here we create a new util which can help trace performance of individual methods across runs, or of events as they occur through the lifetime of the player.
[Delivers #88512202]

( Pair programmed w/ @jw-kaurand )

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -7,8 +7,9 @@ define([
     'utils/backbone.events',
     'utils/helpers',
     'utils/css',
+    'utils/timer',
     'underscore'
-], function(Embed, plugins, Instream, events, states, Events, utils, cssUtils, _) {
+], function(Embed, plugins, Instream, events, states, Events, utils, cssUtils, Timer, _) {
 
     function addFocusBorder(container) {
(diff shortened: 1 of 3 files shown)


src/js/api/api.js
```diff
@@ -494,7 +501,12 @@ define([
                 }
             });
 
-            _this.trigger(events.API_READY, {});
+            _qoe.tick(events.API_READY);
+
+            _this.trigger(events.API_READY, {
+                initTime  : _qoe.between(events.API_INITIALIZED, events.API_SETUP),
```
robwalch: "initTime" will confuse users and is not relevant to QoE. For example I can init the api ```var jw = jwplayer()``` and then wait before calling ```jw.setup()```. We may also choose to call setup multiple times on the same api instance.

I like that initialization time is tracked, but don't see a use for this measurement. I'd rather have a call that gives me a list of sorted events to visualize API lifespan... but that's another story.

What we might want is a library qoe instance that measures script execution on load. I have a feeling that's what users would think is "initTime". This would probably require some addition to webpack.

PR #308, opened 2015-03-18
Timeout setup after 10 seconds

jw-kaurand:
Adding a timeout to setup in case the setup process runs too long.  Will send the error if it takes over 10 seconds to run.  [Finishes #90197716]

### DIFF
--- a/src/js/controller/Setup.js
+++ b/src/js/controller/Setup.js
@@ -20,15 +20,14 @@ define([
 ], function(utils, Playlist, Skin, eventdispatcher, _, events) {
 
     var Setup = function(model, view) {
-        var _this = this,
-            _model = model,
+        var _model = model,
             _view = view,
             _skin,
             _cancelled = false,
             _eventDispatcher = new eventdispatcher(),
             _errorState = false,
             _setupFailureTimeout,
-            _errorTimeoutDelay = 10;
+            _errorTimeoutDelay = 10 * 1000;
 
         var PARSE_CONFIG = {
                 method: _parseConfig,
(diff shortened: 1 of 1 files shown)


src/js/controller/Setup.js
```diff
@@ -62,11 +66,18 @@ define([
         ];
 
         this.start = function () {
+            _setupFailureTimeout = setTimeout(_setupTimeoutHandler, _errorTimeoutDelay * 1000);
```
DonatoB: Put the * 1000 in the original place, ie

            _errorTimeoutDelay = 10 * 1000;

src/js/controller/Setup.js
```diff
@@ -20,11 +20,15 @@ define([
 ], function(utils, Playlist, Skin, eventdispatcher, _, events) {
 
     var Setup = function(model, view) {
-        var _model = model,
+        var _this = this,
```
DonatoB: Instead of using _this and _cancelled, you could bind 'this' into the setupTimeoutHandler on line 69

src/js/controller/Setup.js
```diff
@@ -62,11 +65,18 @@ define([
...
 
         this.start = function () {
+            _setupFailureTimeout = setTimeout(_setupTimeoutHandler.bind(this), _errorTimeoutDelay);
+
             _.defer(_nextTask);
         };
 
+        function _setupTimeoutHandler(){
+            this.destroy();
+            _error('Setup Timeout Error: Setup took longer than '+_errorTimeoutDelay+' seconds to complete.');
```
robwalch: That's going to read: "...Setup took longer than 10000 seconds to complete." :eyes: 

Make `_errorTimeoutDelay = 10` and call `setTimeout(fn, _errorTimeoutDelay * 1000)`

DonatoB: I guess this is an opinion type of thing. Kyle had done it your way originally but I asked him to put the calculation into the variables declaration.

From my point of view, it is confusing to put the multiplier separate from the definition of the delay. All timeouts in javascript are measured in ms, so I don't see any reason to represent them as seconds internally. I'd rather change the message to say 'ms' , or just /1000.

robwalch: Call it `_errorTimeoutSeconds` and it will make more sense. Do math in one place not two.

PR #311, opened 2015-03-19
Fix bug from setting mute before media item available

DonatoB:
This bug could be reproduced locally, but not from jenkins. It originated here :
https://github.com/jwplayer/jwplayer/commit/780fb5fd5c27b573c814cdce0ed221f202948b2e

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/controller/Controller.as
+++ b/src/flash/com/longtailvideo/jwplayer/controller/Controller.as
@@ -162,9 +162,11 @@ public class Controller extends GlobalEventDispatcher {
     }
 
     public function mute(muted:Boolean):Boolean {
-        if (muted !== _model.mute) {
-            _model.mute = muted;
-            return true;
+        if (_model.media) {
+            if (muted !== _model.mute) {
+                _model.mute = muted;
+                return true;
+            }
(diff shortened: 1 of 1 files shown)


src/flash/com/longtailvideo/jwplayer/controller/Controller.as
```diff
@@ -162,9 +162,11 @@ public class Controller extends GlobalEventDispatcher {
...
 
     public function mute(muted:Boolean):Boolean {
-        if (muted !== _model.mute) {
-            _model.mute = muted;
-            return true;
+        if (_model.media) {
+            if (muted !== _model.mute) {
+                _model.mute = muted;
+                return true;
+            }
```
robwalch: We should do this in the model

robwalch: public function set mute(b:Boolean):void {
    _config.mute = b;
    if (_currentMedia) _currentMedia.mute(b);
}
```

DonatoB: I agree, this is here to match what is currently being done for volume, we should follow up on this in the future.

PR #316, opened 2015-03-23
Stalled and buffering QOE for media items

DonatoB:
There are now two levels of QOE - player level and item level.
The player instance level is a Timer object managed by the API, whereas the item level instance is a Timer object managed by the model. They are called _qoe and _itemQoe respectively.

_itemQoe is reset with each new playlist item, whereas _qoe continues logging data throughout the lifetime of the player instance.

The data from these objects is surfaced via jwplayer().qoe() which returns an object of the form
{
    player : {qoe-data}
    item   : {qoe-data}
}
[Finishes #88512386]

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -187,6 +187,7 @@ define([
             }
 
             this.item = newItem;
+            this._qoeItem = new Timer();
             this.trigger(events.JWPLAYER_PLAYLIST_ITEM, {
                 index: this.item
             });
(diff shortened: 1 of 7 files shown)


src/js/controller/model.js
```diff
@@ -170,39 +182,44 @@ define([
...
+
+                _this.setVideoProvider(_currentProvider);
+            }
+
+            // this allows the Youtube provider to load preview images
+            if (_currentProvider.init) {
+                _currentProvider.init(item);
             }
+
+            this._qoeItem = new Timer();
```
robwalch: this should match the item property above (might return/throw before we get here)

PR #318, opened 2015-03-24
Seek faster while scrubbing through media

DonatoB:
A provider can trigger the JWPLAYER_MEDIA_SEEKED event to alert the player that a seek has completed. This allows us to seek more frequently when scrubbing through the media.
[#88941076]

### DIFF
--- a/src/js/view/controlbar.js
+++ b/src/js/view/controlbar.js
@@ -166,13 +166,17 @@ define([
             },
 
             _sliderMapping = {
-                time: _.throttle(_seek, 500),
+                time: _seekThrottled,
                 volume: _volume
             },
             _groups = {},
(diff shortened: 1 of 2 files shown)


src/js/view/controlbar.js
```diff
@@ -769,6 +782,11 @@ define([
             _model.setVolume(pct * 100);
         }
 
+        function _seekThrottled(pct) {
+            _seekTo = pct;
+            _seekThrottler(pct);
```
robwalch: So we're still doing a combination of throttled requests with requests made on slider time events. Is that to support js and flash? ideally the first request is immediate - 'dragging' begins, and while dragging seek requests are made on seeked events.

PR #319, opened 2015-03-24
Webpack IE8 Fonts issue fix

jw-kaurand:
Fixes IE8 issue where fonts weren't loading correctly in IE8 since the publicPath wasn't correctly set and the less code needed to be structured differently due to how the other fonts are being embedded so that IE8 sees them correctly.

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -159,15 +159,15 @@ module.exports = function(grunt) {
                         },
                         {
                             test: /\.woff(\?v=\d+\.\d+\.\d+)?$/,
-                            loader: 'url?limit=10000&minetype=application/font-woff'
+                            loader: 'url?limit=10000&mimetype=application/font-woff'
                         },
                         {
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -128,6 +128,7 @@ module.exports = function(grunt) {
                 },
                 output: {
                     path: 'bin-debug/',
+                    publicPath: '/jwplayer/bin-debug/',
```
robwalch: This is going to have to be an absolute path to our CDN before release?

Gruntfile.js
```diff
@@ -161,10 +166,6 @@ module.exports = function(grunt) {
                             loader: 'url?limit=10000&minetype=application/octet-stream'
```
robwalch: "mine" type? Should be mimetype no?

Gruntfile.js
DonatoB: We will revisit this soon, Kyle made a story in pivotal for this : https://www.pivotaltracker.com/story/show/91134114

PR #320, opened 2015-03-24
API event flow refactor

DonatoB:
The first thing we do is break out two logical sections of code into their own files.
1. callback method helper
2. mutator method helper

There are absolutely no logical changes to mutators, except location.

The callback system, was refactored such that any event triggered through the controller can be listened to via
```jwplayer().on(events.ANY_EVENT)`` and also ```jwplayer().on(states.ANY_STATE)```

The file 'callbacks-deprecate' allows us to also use the older jwplayer().onEvent and jwplayer.onState() until we are ready to remove this support.

[Delivers #90630030]

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -8,8 +8,10 @@ define([
     'utils/helpers',
     'utils/css',
     'utils/timer',
-    'underscore'
-], function(Embed, plugins, Instream, events, states, Events, utils, cssUtils, Timer, _) {
+    'utils/underscore',
+    'api/mutators',
+    'api/callbacks-deprecate'
+], function(Embed, plugins, Instream, events, states, Events, utils, cssUtils, Timer, _, mutatorsInit, legacyInit) {
 
     function addFocusBorder(container) {
         utils.addClass(container, 'jw-tab-focus');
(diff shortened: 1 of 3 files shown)

### CODE REVIEW

File: src/js/api/api.js [hunk 1]
```diff
@@ -388,8 +302,17 @@ define([
             plugins.registerPlugin(id, target, arg1, arg2);
         };
 
+        function _forwardStateEvent(evt) {
+            _this.trigger(evt.newstate, evt);
+        }
         _this.setController = function (player) {
+            if (_controller) {
+                _controller.off('all', _this.trigger);
+                _controller.off(events.JWPLAYER_PLAYER_STATE, _forwardStateEvent);
```
robwalch: can we just `_controller.off()`?

src/js/api/api.js
```diff
@@ -489,21 +347,18 @@ define([
...
-            _eventListener(events.JWPLAYER_MEDIA_SEEKED, utils.noop);
-
-            _eventListener(events.JWPLAYER_PLAYLIST_ITEM, function () {
+            this.on(events.JWPLAYER_PLAYLIST_ITEM, function () {
                 _itemMeta = {};
             });
 
-            _eventListener(events.JWPLAYER_MEDIA_META, function (data) {
+            this.on(events.JWPLAYER_MEDIA_META, function (data) {
                 _.extend(_itemMeta, data.metadata);
```
robwalch: oh didn't realize the js player did this. not at all the same as the Flash player. Not so good...

src/js/api/mutators.js
```diff
@@ -0,0 +1,64 @@
...
+            'setControls',
+            'setCurrentQuality',
+            'setVolume',
+            'setCurrentAudioTrack'
+        ];
+
+
+        // given a name "getBuffer", it adds to jwplayer.api a function which internally triggers jwGetBuffer
+        function generateInternalFunction(name) {
+            var internalName = 'jw' + name.charAt(0).toUpperCase() + name.slice(1);
```
robwalch: ew

src/js/api/api.js
DonatoB: We discussed this in person. 

The question is if we can safely assume that all references to the previous controller are dropped, and if so it would be safe to use .off().

However if the controller may be used again (perhaps it is swapped out due to a plugin like instream) then this would break other functionality.

For now, we are going to leave the conservative decision.

PR #321, opened 2015-03-24
Code Review : CSS and DOM notes

robwalch:
@jw-kaurand a quick run through of things I want us to clean up. You don't need to merge this. I just want us to define some rules about styling and DOM structure that will lead to cleaner code and simpler skinning.


DonatoB:
Closing since this isn't meant to be merged in. It can still be referenced and commented on.

PR #323, opened 2015-03-25
Fix for seeking outside of buffered zone

DonatoB:
[Delivers #91051588]

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -10,6 +10,9 @@ define([
 ], function(cssUtils, utils, stretchUtils, _, events, states, eventdispatcher, DefaultProvider) {
 
     var clearInterval = window.clearInterval,
+        stallInterval,
+        STALL_DELAY = 100,
+        STALL_TOLERANCE = STALL_DELAY/2,
         _isIE = utils.isMSIE(),
         _isMobile = utils.isMobile(),
         _isSafari = utils.isSafari(),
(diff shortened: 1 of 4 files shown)


src/js/providers/html5.js
```diff
@@ -10,6 +10,9 @@ define([
 ], function(cssUtils, utils, stretchUtils, _, events, states, eventdispatcher, DefaultProvider) {
 
     var clearInterval = window.clearInterval,
+        stallInterval,
+        STALL_DELAY = 100,
+        STALL_TOLERANCE = STALL_DELAY/2,
```
robwalch: 50 ms tolerance is too low (an animation lower than 20fps with no audio might not get time updates?). 125 ms (1/8 sec) would be much safer. It's slower than 1/16 -1/10 were the illusion of motion or delays are perceivable, but at least we won't send false stalls.

src/js/providers/html5.js
```diff
@@ -17,16 +20,37 @@ define([
...
+        };
+    };
+
     function _setupListeners(eventsHash, videoTag) {
         utils.foreach(eventsHash, function(evt, evtCallback) {
             videoTag.addEventListener(evt, evtCallback, false);
         });
+
+        var checker = stallCheckGenerator(videoTag, eventsHash.stalled);
+        stallInterval = setInterval(checker, STALL_DELAY);
```
robwalch: always call clearInterval(x) before x = setInterval()

src/js/providers/html5.js
```diff
@@ -17,16 +20,37 @@ define([
...
     }
 
     function _removeListeners(eventsHash, videoTag) {
         utils.foreach(eventsHash, function(evt, evtCallback) {
             videoTag.removeEventListener(evt, evtCallback, false);
         });
+
+        if (stallInterval) {
```
robwalch: don't need the if. assign a default value of 0 or -1 (pick your fave) to intervalIds so js knows it's a Number

DonatoB:
Will recreate this pull request against master

PR #326, opened 2015-03-26
RTMP Provider Ad Support

robwalch:
RTMP provider is prepared for ad playback control immediately when load(item) is called [Delivers #90936728]

This is done by setting media and sending a media loaded event followed by a buffering event (as other providers do).

Since all providers should send a media loaded event when the media property is set (updates the view and prevents controls from locking / buffering from being ignored), this is now enforced for all providers by building it into `MediaProvider.set media(DisplayItem)`.

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/media/RTMPMediaProvider.as
+++ b/src/flash/com/longtailvideo/jwplayer/media/RTMPMediaProvider.as
@@ -115,7 +115,7 @@
 				stagevideo: stagevideo,
 				renderstate: event['status']
 			}});
-		};
+		}
 
 
 		/** Load content. **/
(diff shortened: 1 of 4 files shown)


DonatoB:
Moving the MEDIA_LOADED event into mediaprovider.as is the definitely the right move for 7.
For this patch I'd prefer localizing the change to the RTMP provider.

src/flash/com/longtailvideo/jwplayer/media/RTMPMediaProvider.as
```diff
@@ -136,8 +136,14 @@
 				}
 				attachNetStream(_stream);
 			}
+
+			media = _video;
+			//If you send buffering events before the provider's media is set
+			//The view will think that the controls should be locked because the provider doesn't have a display.
+			setState(PlayerState.BUFFERING);
```
DonatoB: In the previous configuration, these 2 commands would not be executed when we fall through to the final else statement on line 170 : 
```js
    _loader.load(_item.file, XML);
```

Other than that, the logic appears the same. Is this part of the fix?

DonatoB: We spoke about this in person - yea, that is part of the fix.

PR #327, opened 2015-03-26
Don't confuse padded empty divs with audio mode [Delivers #90702642]

robwalch:
Check model height first when determining audio mode. Only check on resize after controls are setup.


DonatoB:
LGTM

PR #328, opened 2015-03-26
Fix for seeking outside of buffered zone [Delivers #91051588]

DonatoB:


DonatoB:
New version of https://github.com/jwplayer/jwplayer/pull/323

PR #336, opened 2015-03-30
getProvider API

jw-kaurand:
Retreives a string which represents the current provider API in use.  This string will appear in the right click menu on the player.  This is paired with a commercial PR of the same name to allow the right click in the commercial branch: https://github.com/jwplayer/jwplayer-commercial/pull/529.  [Finishes #90689434]

### DIFF
--- a/src/js/view/rightclick.js
+++ b/src/js/view/rightclick.js
@@ -21,7 +21,7 @@ define([
         JW_CSS_NONE = 'none',
         JW_CSS_WHITE = '#FFF';
 
-    var Rightclick = function(_playerElement, _model) {
+    var Rightclick = function(_playerElement, _api, _model) {
 
         var _config = {
                 aboutlink: LINK_DEFAULT + _version + '&m=h&e=o',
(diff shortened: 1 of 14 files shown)


src/js/view/rightclick.js
```diff
@@ -21,7 +21,7 @@ define([
         JW_CSS_NONE = 'none',
         JW_CSS_WHITE = '#FFF';
 
-    var Rightclick = function(_playerElement, _model) {
+    var Rightclick = function(_playerElement, _api, _model) {
```
DonatoB: Can we instead store the current provider in the model and fetch from there
model.get('provider')?

DonatoB:
With this solution, every time we want to know the provider, we would run from the API, through the controller, through the model to the provider where we query for it. I think we can simplify that a little bit.

Suppose we say *the purpose of the model is to represent the state of *everything*

It is not exactly true yet, but with this assumption, controller.getProvider would be simply
```js
    this.getProvider = _.partial(_model.get, 'provider');
```

It's a small change, but it means when we ask for the current provider we are checking a stored value of the object instead of querying the provider.

To make it work, we would modify ```model.setVideoProvider``` and add a line in there which says
```js
    this.set('provider', provider.getName())
```

I think this is valuable because it means our view doesn't need to access the API in order to render.

jw-kaurand:
Closed due to more up to date request

PR #338, opened 2015-03-31
Moving controls logic from view into controller

DonatoB:
Now the View will listen for changes from the Model related to controls, and when they occur, it will update its state.
[Finishes #90647130]

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -260,7 +260,12 @@ define([
             return this[attr];
         },
         'set' : function(attr, val) {
+            if (this[attr] === val) {
+                return;
+            }
             this[attr] = val;
+            this.trigger('change:'+attr, this, val);
+            this.trigger('change', this);
         }
     });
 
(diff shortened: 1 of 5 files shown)


src/js/controller/model.js
```diff
@@ -260,7 +260,12 @@ define([
             return this[attr];
         },
         'set' : function(attr, val) {
+            if (this[attr] === val) {
+                return;
+            }
             this[attr] = val;
+            this.trigger('change:'+attr, this, val);
+            this.trigger('change', this);
```
robwalch: why do we need both `change:x` and `change`? won't `on('change')` act as a catch all?

robwalch: It would be nice if the model only triggered attribute changes. We could then use 'all' for all changes.

DonatoB: True, I'll just document our discussion for the rest of the team.

My intention was to mimic the API of backbones implementation of Model, however we agreed that it wasn't the time to change our internal strategy, especially not if we would do so only partially.

PR #340, opened 2015-04-01
Separate controller instantiation from setup

DonatoB:
Part one of callInternal deprecation
[#90613644]


DonatoB:
I suggest viewing this merge without whitespace (add ?w=1 into the URL)

PR #343, opened 2015-04-01
Remove _callInternal from api

DonatoB:
This commit does a few things
1. Rename internal-api as internal-controller - this more accurately defines it's purpose. API related code was moved into a more appropriate location.
2. mutators.js will no longer pass through to jwGet/jwSet methods, it will use the controller directly
3. Remove all jwMethods into api-deprecate.js
4. Move all Queueing into the controller and remove callInternal from api

[#90613644]

Conflicts:
	src/js/api/internal-api.js


src/js/api/api-mutators.js
```diff
@@ -0,0 +1,78 @@
...
+            //'playlistItem',
+        ];
+        _.each(passthroughs, function(func) {
+            _api[func] = function() {
+                return _controller[func].apply(_controller, arguments);
+            };
+        });
+
+
+        // One additional for legacy
```
robwalch: one what?

src/js/api/api.js
```diff
@@ -87,20 +89,20 @@ define([
         _qoe.tick(events.API_INITIALIZED);
 
         this.setup = function (options) {
-                _qoe.tick(events.API_SETUP);
-                // Remove any players that may be associated to this DOM element
-                _this.remove();
+            _qoe.tick(events.API_SETUP);
+            // Remove any players that may be associated to this DOM element
+            // this.remove();
```
robwalch: need to do something about previous setup

src/js/api/api.js
```diff
@@ -309,43 +283,11 @@ define([
...
-                    funcName = args.shift();
-                if (!_.isFunction(_controller[funcName])) {
-                    return null;
-                }
-                return _controller[funcName].apply(_controller, args);
-            }
-            _queuedCalls.push(arguments);
-        }
-
-        _this.callInternal = _callInternal;
```
robwalch: add it to deprecate. We can have a shoddy implementation or throw an exception / console.error when used.

test/index.html
```diff
@@ -100,7 +100,9 @@
         })(simpleName);
     }
 
-    jwplayer('video-container').setup(config);
+    var jwp = jwplayer('video-container');
+    jwp.setVolume('40');
+    jwp.setup(config);
```
robwalch: if you play, then setup again, what will happen without api.remove() we probably want a brand new internal controller and model, but may be able to reuse api and controller if they are stateless (watch out for that queue).

robwalch:
Check out https://github.com/jwplayer/jwplayer/pull/344. This is in the same direction but focused on adding and removing the player instances.

I want to see if we can also use global-api to access instances, configs, etc... without using the global jwplayer().

PR #344, opened 2015-04-02
optimize jwplayer() by storing all instances in global-api before setup

robwalch:
- removes global-api from jwplayer.api (will move to commercial legacy.js)
- global-api creates and adds instances on first query
- jwplayer().remove() removes itself from the global list then is destroyed
- api is cleaned up and global-api addPlayer and new removePlayer methods are private

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -38,13 +38,13 @@ define([
         };
     }();
 
-    var Api = function (container) {
+    var Api = function (container, globalRemovePlayer) {
         var _this = this,
+            _instream,
             _originalContainer = container,
-            _controller = null,
+            _controller = new Controller(),
             _playerReady = false,
             _queuedCalls = [],
-            _instream,
             _itemMeta = {},
             _eventQueue = [],
             _callbacks = {};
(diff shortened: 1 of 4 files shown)


src/js/api/api.js
```diff
@@ -250,12 +264,21 @@ define([
             return _instream;
         };
         _this.destroyPlayer = function () {
```
DonatoB: I think your change removes the necessity for a "remove" separate from "destroy"

We can then delete this function and jam its logic into api.remove(). This allows us to delete the line _api.destroyPlayer() from globalApi's _removePlayer!

src/js/api/global-api.js
```diff
@@ -47,23 +47,26 @@ define([
...
+            if (_instances[i].uniqueId === api.uniqueId) {
+                _instances.splice(i, 1);
+                api.destroyPlayer();
+                break;
+            }
+        }
+    };
+
     return {
         _instances : _instances,
```
DonatoB: Don't need to return this

src/js/api/global-api.js
```diff
@@ -47,23 +47,26 @@ define([
...
         selectPlayer : selectPlayer,
-        playerById : playerById,
-        addPlayer : addPlayer
+        playerById : playerById
```
DonatoB: Don't need to return this either, i think

DonatoB:
Looks good!

PR #352, opened 2015-04-04
unit tests code coverage

robwalch:
- small changes to item and source based on test and branch analysis (TDD!)
- shortened bind polyfill type exception message (considered removal of exception, or testing this script. opted to remove polyfills from coverage preprocessing)
- chose not to omit utils/underscore from coverage since we can see how much of it we use
- attempted to clarify what is being tested in the playlist item tests by changing some language
- embed no longer modifies setup options after creating config
- grunt watch runs tests
- added tests:
  - playlist item input source type normalization
  - playlist item input.captions are converted to tracks
  - filterPlaylist edge cases and androidhls setting copied
  - embed/embed partial test with comments
  - player selection test (loads jwplayer module, less and templates)

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -45,7 +45,6 @@ define([
             _controller = new Controller(),
             _playerReady = false,
             _itemMeta = {},
-            _eventQueue = [],
             _callbacks = {};
 
         // Set up event handling
@@ -250,20 +249,6 @@ define([
             _instream.loadItem(item);
             return _instream;
         };
-        this.destroyPlayer = function () {
-            _reset();
-
-            // so players can be removed before loading completes
-            _playerReady = true;
-            _controller.playerDestroy();
-
-            // terminate state
-            _playerReady = false;
-            _controller = null;
-            _eventQueue = [];
-            _itemMeta = {};
-            _callbacks = {};
-        };
 
         this.playAd = function (ad) {
             var plugins = this.plugins;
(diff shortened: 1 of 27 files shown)


test/unit/embed.js
```diff
@@ -0,0 +1,67 @@
...
+    'test/underscore',
+    'jquery',
+    'embed/embed'
+], function (_, $, Embed) {
+    /* jshint qunit: true */
+
+    module('Embed');
+
+    /**
+     * Comments by robwalch on 2015/4/5
```
donato: These comments probably belong in pivotal

src/js/api/api.js
```diff
@@ -126,8 +125,8 @@ define([
                 }
             });
 
-            this._embedder = new Embed(options, this, _controller);
-            this._embedder.embed();
+            this._embedder = new Embed(this, _controller);
+            this._embedder.embed(options);
```
donato: Good call, this reduces the scope of the options object inside of Embed.

src/js/api/global-api.js
```diff
@@ -33,7 +34,9 @@ define([
             return _addPlayer(new Api(domElement, _removePlayer));
         }
         // invalid query
-        return null;
+        return {
+            registerPlugin: plugins.registerPlugin
```
donato: Based off of the usage, it appears to me that registerPlugin should be moved from the jwplayer instance API to the global API. 

ie, from ```jwplayer().registerPlugin``` to ```jwplayer.registerPlugin```

What do you think?

test/config.js
```diff
@@ -22,6 +21,13 @@ if (!window.__karma__) {
     callback = window.__karma__.start;
 }
 
+// Add qunit-fixture to page if not present
```
donato: Why remove this from the html and add it this way? I feel like the html solution is a little simpler.

robwalch: Because karma does not add it. So this way, we do it here for both.

src/js/api/global-api.js
robwalch: We can move the legacy query behavior to commercial but it's a little weird:

jwplayer(invalid) returns {registerPlugin: function...}

vs

jwplayer(invalid) returns null

registerPlugin should go in the same place as registerProvider - currently that's in jwplayer.api, but jwplayer[here] makes more sense.

PR #354, opened 2015-04-06
State Specific Styles Isolated to Separate Files

jw-kaurand:
Isolates state specific styling to separate files. This makes it easier to find and modify the styles that exist in specific states, as well as apply states consistently by applying a single class to the root of the player. also changed how we're setting up icon styles so that they can be used as mixins so that its easier to apply an icon's style information to an element.

### DIFF
--- a/src/css/states/play.less
+++ b/src/css/states/play.less
@@ -1,5 +1,0 @@
-.jwplayer.jw-state-play {
-    .jw-controlbar .jw-controlbar--left-group .jw-play-state-icon {
-        .jwplayer .jw-icon-pause;
-    }
-}
--- a/src/css/states/play.less
+++ b/src/css/states/play.less
@@ -1,0 +1,5 @@
+.jwplayer.jw-state-play {
+    .jw-controlbar .jw-controlbar--left-group .jw-play-state-icon {
+        .jwplayer .jw-icon-pause;
+    }
+}
--- a/src/css/states/replay.less
+++ b/src/css/states/replay.less
@@ -1,0 +1,27 @@
+.jwplayer.jw-state-replay {
+    .jw-display {
+        display: block;
+
+        .jw-preview {
+            display: block;
+        }
+    }
+
+    .jw-display-icon-cont {
+        .jw-icon-display {
+            .jwplayer .jw-icon-replay;
+        }
+
+        .jw-text {
+            display: none;
+        }
+    }
+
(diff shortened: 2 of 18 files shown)


donato:
Can we rename the "replay" state to "complete" state?

src/css/states/play.less
```diff
@@ -0,0 +1,5 @@
+.jwplayer.jw-state-play {
+    .jw-controlbar .jw-controlbar--left-group .jw-play-state-icon {
+        .jwplayer .jw-icon-pause;
```
donato: What does this do?

test/manual/css-skins/scripts.js
```diff
@@ -138,30 +124,26 @@ $(document).ready(function(){
...
-        $('.jwplayer').removeClass('play-state');
-        $('.jwplayer').addClass('pause-state');
-        $('.jwplayer').removeClass('buffering-state');
-        $('.jwplayer .jw-display-icon-cont .jw-icon-display').addClass('jw-icon-play');
-        $('.jwplayer .jw-display-icon-cont .jw-icon-display').removeClass('jw-icon-buffer');
-
-        $($('.jwplayer .jw-controlbar--center-group .jw-icon-inline')[0]).addClass('jw-icon-play');
-        $($('.jwplayer .jw-controlbar--center-group .jw-icon-inline')[0]).removeClass('jw-icon-pause');
+        $('.jwplayer').addClass('jw-state-pause');
+        $('.jwplayer').removeClass('jw-state-play jw-state-idle jw-state-buffer jw-state-replay');
```
donato: We could use a helper for these : removeStateClasses($obj)

src/css/states/replay.less
```diff
@@ -0,0 +1,27 @@
+.jwplayer.jw-state-replay {
+    .jw-display {
+        display: block;
+
+        .jw-preview {
+            display: block;
+        }
+    }
+
+    .jw-display-icon-cont {
```
donato: jw-display-icon-cont is an odd name, does "cont" stand for container?

src/css/states/play.less
jw-kaurand: This mixin adds the styles of .jwplayer .jw-icon-pause class to the play/pause toggle icon in the control bar.  In this case it's just setting the content of the after for the icon code of the last element to be the pause icon.  Doing this gets us a semantic way to say which icon to use for that element's :after value without resorting to setting the character code statically.

PR #357, opened 2015-04-06
State Specific Styles Isolated to Separate Files (rebase)

jw-kaurand:
Isolates state specific styling to separate files. This makes it easier to find and modify the styles that exist in specific states, as well as apply states consistently by applying a single class to the root of the player. also changed how we're setting up icon styles so that they can be used as mixins so that its easier to apply an icon's style information to an element.

Edits based on feedback: 
* Player state files and classes renamed based on https://github.com/jwplayer/jwplayer-commercial/wiki/Html-Structure-and-Classes
* Simplified the add/remove classes on the test page
* Renamed ".jw-display-icon-cont" ".jw-display-icon-container"


src/css/states/playing.less
```diff
@@ -0,0 +1,5 @@
+.jwplayer.jw-state-playing {
+    .jw-controlbar--left-group .jw-playback-toggle-icon {
+        .jwplayer .jw-icon-pause;
```
jw-kaurand: These mixins adds the styles of .jwplayer .jw-icon-pause class to the play/pause toggle icon in the control bar. In this case it's just setting the content of the after for the icon code of the last element to be the pause icon. Doing this gets us a semantic way to say which icon to use for that element's :after value without resorting to setting the character code statically.

test/manual/css-skins/index.html
```diff
@@ -150,22 +138,22 @@
...
-                            <div class="jw-dock-image" style="background-image: url(../../../jwplayer-commercial/test/assets/os/both.png)"></div>
+                            <div class="jw-dock-image" style="background-image: url(/jwplayer-commercial/test/assets/os/both.png)"></div>
                             <div class="jw-arrow"></div>
                             <div class="jw-overlay">
                                 <span class="jw-text">Button sucessfully added</span>
                             </div>
                         </div>
                         <span class="jw-dock-button">
-                            <div class="jw-dock-image" style="background-image: url(../../../jwplayer-commercial/test/assets/smiley.png)"></div>
+                            <div class="jw-dock-image" style="background-image: url(/jwplayer-commercial/test/assets/smiley.png)"></div>
```
robwalch: move these assets to a location relative to test/manual/css-skins/ like test/manual/css-skins/icons. Then your test page can just point to icons/file

PR #363, opened 2015-04-08
Event flow fixes [Delivers #91738776 #90930900 #88935528]

robwalch:


src/js/api/api.js
```diff
@@ -42,36 +43,42 @@ define([
...
-            args.type = args.type || type;
-            args = normalizeOutput(args);
+            // map events
+            if (type === 'ready') {
+                _playerReady = true;
+                _qoe.tick('ready');
+                args.setupTime = _qoe.between('setup', 'ready');
+            } else if (type === events.JWPLAYER_PLAYER_STATE) {
+                type = args.newstate;
+            }
```
donato: I don't think we should add this logic into the trigger function. An alternative solution to the problem of triggering "ready" events is to move the logic from API into controller. I put an example here https://github.com/jwplayer/jwplayer/commit/c913b5c9a5605244fd6bcdd62c31c849bb0cc488

PR #365, opened 2015-04-08
event flow fixes

robwalch:
fixes a variety of issues in commercial player

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -78,10 +78,10 @@ define([
                 case events.JWPLAYER_PLAYER_STATE:
                     // These two states exist at a provider level, but the player itself expects BUFFERING
                     if (evt.newstate === states.LOADING) {
-                        this.mediaController.trigger(events.JWPLAYER_PROVIDER_LOADING, evt);
+                        this.mediaController.trigger(events.JWPLAYER_PROVIDER_LOADING);
                         evt.newstate = states.BUFFERING;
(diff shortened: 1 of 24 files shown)


src/js/controller/model.js
```diff
@@ -266,8 +271,9 @@ define([
             if (this[attr] === val) {
                 return;
             }
+            var oldVal = this[attr];
             this[attr] = val;
-            this.trigger('change:' + attr, this, val);
+            this.trigger('change:' + attr, this, val, oldVal);
```
donato: The backbone way to handle this would be to store the previous attributes value in the model, which are query-able via "model.previous('state')"

For now, this is fine, but something to think about in the future.

test/unit/api.js
```diff
@@ -0,0 +1,159 @@
...
+    'test/underscore',
+    'jquery',
+    'api/api',
+    'utils/backbone.events'
+], function (_, $, Api, Events) {
+    /* jshint qunit: true */
+
+    module('Api');
+
+    test('extends Events', function(assert) {
```
donato: The test passes whether we extend or not, but it will verify for us that the methods actually exist. I propose we change this to "contains expected methods" and test it for these
```
setup
load
play
pause
seek
stop
remove
resize
playlistItem
playlistNext
playlistPrev
registerPlugin
addButton
removeButton
get
set
on
off
once
trigger
```

PR #366, opened 2015-04-08
Visual Quality Events listener [Finishes #88511936]

jw-kaurand:
Handles visualQuality events coming from the Flash player and the QOE module will set the current visual quality value on the model.  [Finishes #88511936]  This is paired with the jwplayer-commercial PR https://github.com/jwplayer/jwplayer-commercial/pull/542

### DIFF
--- a/src/js/controller/qoe.js
+++ b/src/js/controller/qoe.js
@@ -71,9 +71,6 @@ define([
             trackFirstFrame(model);
             trackStalledTime(model);
         });
-        model.mediaController.on(events.JWPLAYER_VISUAL_QUALITY, function(evt){
-            model.set('visualQuality', evt.level);
-        });
     }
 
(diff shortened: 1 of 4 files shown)


src/js/controller/qoe.js
```diff
@@ -71,6 +71,9 @@ define([
             trackFirstFrame(model);
             trackStalledTime(model);
         });
+        model.mediaController.on(events.JWPLAYER_VISUAL_QUALITY, function(evt){
+            model.set('visualQuality', evt.level);
+        });
```
robwalch: this should be done in the model's video event handler. It's not qoe's responsibility to set model state.

robwalch:
I've added some things to jwplayer7/visual-quality-api so it's now newer than this branch.

PR #370, opened 2015-04-09
improved flash to js event mapping

robwalch:
Customer events implement toJsObject() for js event mapping
- Player and Media events received in flash.js now only have the properties we need
- Special event name mapping (jw-error to error) happens in toJsObject()

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/events/MediaEvent.as
+++ b/src/flash/com/longtailvideo/jwplayer/events/MediaEvent.as
@@ -29,6 +29,18 @@ public class MediaEvent extends PlayerEvent {
     public static const JWPLAYER_MEDIA_LEVELS:String = "levels";
     public static const JWPLAYER_MEDIA_LEVEL_CHANGED:String = "levelsChanged";
 
+    private var _currentQuality:Number;
+    private var _bufferPercent:Number;
+    private var _position:Number;
+    private var _duration:Number;
+    private var _volume:Number;
+
+    public var levels:Array = null;
+    public var metadata:Object = null;
+    public var offset:Number = 0;
+    public var mute:Boolean = false;
+
+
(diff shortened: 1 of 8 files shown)


src/flash/com/longtailvideo/jwplayer/events/PlayerEvent.as
```diff
@@ -37,5 +37,16 @@ public class PlayerEvent extends Event {
         return new PlayerEvent(this.type, this.message);
     }
 
+    public function toJsObject():Object {
+        if (message) {
+            return {
+                type: type.replace(/^jw\-/, ''),
```
donato: What is this for?

src/flash/com/longtailvideo/jwplayer/events/MediaEvent.as
```diff
@@ -39,34 +51,117 @@ public class MediaEvent extends PlayerEvent {
...
+    override public function toJsObject():Object {
+        var js:Object = super.toJsObject();
+        switch (type) {
+            case JWPLAYER_MEDIA_TIME:
+            case JWPLAYER_MEDIA_BUFFER:
+                if (!isNaN(_bufferPercent)) {
+                    js.bufferPercent = _bufferPercent;
+                }
+                if (!isNaN(_position)) {
+                    js.position = Math.round(_position * 1000) / 1000;
```
donato: Why round numbers at this level?

src/flash/com/longtailvideo/jwplayer/model/Model.as
```diff
@@ -217,8 +217,8 @@ public class Model extends GlobalEventDispatcher {
                     return;
                 }
             }
-            dispatchEvent(evt);
         }
+        dispatchEvent(evt);
```
donato: Why is this moved?

PR #371, opened 2015-04-09
CSS Skinned Logo [Finishes #90058444]

jw-kaurand:
Added the logo component as a templated component.  Removed some related code from the commercial branch into here in the model since we've moved functionality out of that into the OS branch.  Stripped out the JWPlayer default logo and link logic since that's no longer part of the player.  This is related to the commercial PR https://github.com/jwplayer/jwplayer-commercial/pull/545 [Finishes #90058444]

### DIFF
--- a/src/css/imports/logo.less
+++ b/src/css/imports/logo.less
@@ -5,15 +5,11 @@
     visibility: hidden;
     opacity: 0;
     pointer-events: all;
+    top: 0.5em;
+    right: 0.5em;
 
     &.jw-logo--visible {
         visibility: visible;
         opacity: 1;
-    }
-
-    // Default config.  Other values are handled in the constructor.
-    &.jw-logo--top-right {
-        top: 0.5em;
-        right: 0.5em;
     }
 }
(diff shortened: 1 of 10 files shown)


src/css/imports/logo.less
```diff
@@ -0,0 +1,19 @@
...
+    opacity: 0;
+    pointer-events: all;
+
+    &.jw-logo--visible {
+        visibility: visible;
+        opacity: 1;
+    }
+
+    // Default config.  Other values are handled in the constructor.
+    &.jw-logo--top-right {
```
donato: Can we ditch this class? It seems too specific

src/js/view/logo.js
```diff
@@ -4,24 +4,13 @@ define([
     'utils/css',
```
donato: let's try to remove dependancies on this utils file

donato:
Let's use this opportunity to remove _api from logo.js.

Instead fire off events for actions, which view.js will listen for, and react to. Ultimately view.js will be view-controller, and this will get us closer to that!

PR #372, opened 2015-04-09
Implement new context menu [Delivers #85531320]

donato:


src/js/view/rightclick.js
```diff
@@ -1,158 +1,116 @@
...
+            utils.removeClass(this.el, 'open');
+        },
+
+        setup : function(_model, _playerElement) {
+            this.parent = _playerElement;
+            this.model = _model;
+            this.mouseOverContext = false;
+            this.el = this.buildMenu();
+
+            this.model.on('change:provider', this.updateHtml, this);
```
robwalch: only build/update when the menu is open

src/templates/rightclick.html
```diff
@@ -0,0 +1,14 @@
+<div class="jw-rightclick">
+    <ul>
+    {{#each items}}
+        <li{{#if class}} class="{{class}}"{{/if}}>
+            <a href="{{link}}">
+                {{#if img}}
+                <img src="{{img}}" width="20">
```
robwalch: Let's use a web font. @jw-kaurand can set you up with the setup.

PR #373, opened 2015-04-10
Player is now instanciated with default 16/9

privaloops:


donato:
Does this fix any bug or defect which you've noticed? We don't think so, since the size is updated as soon as the video source loads.

PR #374, opened 2015-04-10
visual quality API [#88511936]

robwalch:
Handle and forward visual quality events from providers (flash) and store in temp media model for new getVisualQuality() api method

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -185,7 +185,7 @@ define([
             var repeat = false;
             var playlist = _this.get('playlist');
             if (index === playlist.length || index < -1) {
-                newItem = 0;m
+                newItem = 0;
                 repeat = true;
             } else if (index === -1 || index > playlist.length) {
                 newItem = playlist.length - 1;
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -185,7 +185,7 @@ define([
             var repeat = false;
             var playlist = _this.get('playlist');
             if (index === playlist.length || index < -1) {
-                newItem = 0;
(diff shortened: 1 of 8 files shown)


src/js/controller/model.js
```diff
@@ -105,6 +106,11 @@ define([
                 case events.JWPLAYER_PROVIDER_CHANGED:
                     this.set('provider', _provider.getName());
                     break;
+                case 'visualQuality':
+                    var visualQuality = _.extend({}, evt);
+                    delete visualQuality.type;
+                    this.mediaModel.visualQuality =  visualQuality;
```
donato: we should reset this when 'change:item'

src/js/providers/flash.js
```diff
@@ -171,103 +176,51 @@ define([
...
-                        });
+                    _swf.on(events.JWPLAYER_MEDIA_VOLUME, function(e) {
+                        this.sendEvent(events.JWPLAYER_MEDIA_VOLUME, e);
                     }, this);
 
-                    _swf.on(events.JWPLAYER_PROVIDER_CHANGED, function(data) {
-                        _flashProviderType = data.message;
-                        this.sendEvent(events.JWPLAYER_PROVIDER_CHANGED, data);
+                    _swf.on('visualQuality', function(e) {
+                        e.reason = e.reason || 'api'; // or 'user selected';
```
donato: can this be done at a lower level?

src/js/controller/model.js
robwalch: yes mediaModel should be set back to {}

PR #376, opened 2015-04-13
test and jshint updates to grunt and test files

robwalch:


Gruntfile.js
```diff
@@ -29,18 +29,18 @@ module.exports = function(grunt) {
         pkg: packageInfo,
 
         jshint: {
+            options: {
+                jshintrc: '.jshintrc'
+            },
             player : [
-                'src/js/**/*.js'
+                'src/js/{,*/}*.js'
```
donato: why this? I'm pretty sure ** is the preferred way for glob https://github.com/isaacs/node-glob

Gruntfile.js
```diff
@@ -29,18 +29,18 @@ module.exports = function(grunt) {
...
             ],
             tests : [
                 'test/{,*/}*.js'
             ],
             grunt : [
                 'Gruntfile.js'
-            ],
-            options: {
-                jshintrc: '.jshintrc'
```
donato: Why remove this? I think it is needed : https://github.com/gruntjs/grunt-contrib-jshint

robwalch: I just moved it up. options should be listed before targets for readability.

Gruntfile.js
robwalch: That includes js files in src/js/ not only the ones one level down from there.

donato: I double checked, and it turns out this ** symbol is called a *globstar* which actually will match files in the same directory. I tested it using node

> glob = require('glob')
> glob('src/js/**/*.js', function(err, files) { console.log(files); });

robwalch: So we want /**.js? not /*.js or /**/*.js right? globstar used to be
commented as a potential performance problem with grunt-watch in yeoman
generated task files - that's where I adopted {,*/}*.js from.

On Tue, Apr 14, 2015 at 10:06 AM, Donato Borrello <notifications@github.com>
wrote:

> In Gruntfile.js
> <https://github.com/jwplayer/jwplayer/pull/376#discussion_r28329163>:
>
> >              player : [
> > -                'src/js/**/*.js'
> > +                'src/js/{,*/}*.js'
>
> I double checked, and it turns out this ** symbol is called a *globstar*
> which actually will match files in the same directory. I tested it using
> node
>
> glob = require('glob')
> glob('src/js/**/*.js', function(err, files) { console.log(files); });
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/jwplayer/jwplayer/pull/376/files#r28329163>.
>



-- 
*Rob Walch*
JW Player Software Engineer
*JWPlayer* | Your Way to Play
rob@jwplayer.com | jwplayer.com

PR #377, opened 2015-04-13
Move item title from displayicon to it's own object

donato:
[Delivers #90111718]


src/js/view/display.js
```diff
@@ -322,6 +315,7 @@ define([
 
         function _errorHandler(evt) {
             _errorState = true;
+            // TODO : revisit, it no longer accepts a message
```
robwalch: Good catch.

PR #379, opened 2015-04-14
Replace display icon with html/css templates

donato:
[Delivers #91597708]


src/css/states/buffering.less
```diff
@@ -6,6 +6,14 @@
     .jw-display-icon-container {
         .jw-icon-display {
             .jwplayer .jw-icon-buffer;
+
+            -webkit-animation:spin 2s linear infinite;
+            -moz-animation:spin 2s linear infinite;
+            animation:spin 2s linear infinite;
```
robwalch: these need to be faster

PR #380, opened 2015-04-14
Convert dock to html/css templating

donato:
At the same time we are moving the logic for adding/removing buttons out of the view and into the controller

[Delivers #91599538]

### DIFF
--- a/src/js/view/dock.js
+++ b/src/js/view/dock.js
@@ -1,365 +1,45 @@
 /*jshint evil:true*/
 define([
-    'view/overlay',
-    'view/touch',
-    'events/events',
+    'handlebars-loader!templates/dock.html',
     'utils/helpers',
-    'utils/css',
-    'underscore'
-], function(Overlay, Touch, events, utils, cssUtils, _) {
+    'utils/underscore'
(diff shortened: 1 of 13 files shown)


src/js/view/dock.js
```diff
@@ -1,365 +1,45 @@
...
 define([
-    'view/overlay',
-    'view/touch',
-    'events/events',
+    'handlebars-loader!templates/dock.html',
     'utils/helpers',
-    'utils/css',
-    'underscore'
-], function(Overlay, Touch, events, utils, cssUtils, _) {
+    'utils/underscore'
```
robwalch: os unit tests can't find utils/underscore

robwalch: ...fixed by removing alias so that we don't have issues or conflicts between requirejs and webpack https://github.com/jwplayer/jwplayer/commit/295b739eec48927239aa56b915e1f8f0df5d69a7

Looks like require wasn't event trying to load 'utils/underscore' because it already loaded 'underscore' which was setup to load from the same path in require's config.

donato: Good call on this

PR #382, opened 2015-04-15
View Template and CSS refactor [Delivers #90152754]

jw-kaurand:
Implements templates for the main player (.jwplayer) and the view (.jw-main)
Moves styles that are assigned to elements in view.js into classes.
Namespaces a bunch of classes
Implements hasClass and toggleClass into helpers.js
handles several styles using the player root flags
Moves styles applied to the view from jwplayer.less to view.less
Applies !important as necessary.
Changes some api calls to use the local api
Implements logic some features as classes: 
* Stretching
* Aspect Ratio
* Fullscreen
[Delivers #90152754]


src/css/imports/captions.less
```diff
@@ -13,6 +13,10 @@
         &.jw-captions-enabled {
             display: block;
         }
+
+        &.jw-captions-disabled {
+            display: none;
```
donato: Why not assume disabled unless enabled?

src/css/imports/view.less
```diff
@@ -0,0 +1,122 @@
+.jwplayer {
+    width: 100%;
```
donato: This scope is too big to be part of view.less, they should be in styles.less

src/css/imports/captions.less
jw-kaurand: This is for explicitly disabling subtitles in relation to casting.   May be able to revisit and remove this depending on the cast provider work

src/js/utils/helpers.js
```diff
@@ -581,6 +581,20 @@ define([
         return val;
     };
 
+    utils.hasClass = function (element, searchClass) {
```
donato: Include attribution :P https://github.com/jquery/jquery/blob/53aa87f3bf4284763405f3eb8affff296e55ba4f/src/attributes/classes.js#L147

src/js/utils/helpers.js
```diff
@@ -603,6 +617,22 @@ define([
         element.className = strings.trim(_.difference(originalClasses, removeClasses).join(' '));
     };
 
+    utils.toggleClass = function (element, c, toggleTo) {
+        if(_exists(toggleTo)) {
```
donato: I think it'd be worth adding a short-circuit for performance
    
    var hasClass = utils.hasClass(element, c);
    toggleTo = _.isBoolean(toggleTo) ? toggleTo : !hasClass;
    
    // short circuit if nothing to do
    if (toggleTo === hasClass) {
        return;
    }
    
    // now toggle

src/js/view/view.js
```diff
@@ -690,7 +647,7 @@ define([
                 id = _model.id + '_view';
             cssUtils.block(id);
 
-            // when jwResize is called remove aspectMode and force layout
+            // when jwResize is caandlled remove aspectMode and force layout
```
donato: .
                            ...
                          .......
                       .....:::.....
                   ....:::::::::::::....
                 ....:::::::::::::::::....
             ..::::::::::::  =  ::::::::::....
             ..::::::::::   ===   ::::::::::....
         ..::::::::::::    ======    ::::::::::....
         ..::::::::::    ===========    ::::::::...
       ..::::::::::   ================    ::::::::..
       ..::::::::::  ======= . =========   ::::::::...
     ..::::::::::  ========  |\  ========  :::::::::...
     ..::::::::::  ========  |#\ ========  :::::::::...
     .::::::::::  ========  /%#|  =======   :::::::::..
     .::::::::::  =======  /#%#|  =======   :::::::::..
     .::::::::::  ======  |#%#/  ========  :::::::::..
     ..:::::::::  ======   \%/  ========  :::::::::..
      ..:::::::::   ======  %  =======  ::::::::::..
       ..:::::::::    ===___%____===   ::::::::::..
        ..::::::::::    |  .::.  |   ::::::::::...
         .::::::::::    |  .::.  |   ::::::::::..
          ..::::::::    |  .::.  |   ::::::::...
           .::::::::    |  .::.  |   ::::::::..
             ...::::    |  .::.  |   ::::....
              ..::::    |  .::.  |   ::::...
                ....    |  .::.  |   ....
                  ..    |  .::.  |   ..

src/js/api/api.js
```diff
@@ -110,7 +111,8 @@ define([
         legacyInit(this);
 
         // These should be read-only model properties
-        this.container = document.createElement('div');
+        this.container = utils.createElement(playerTemplate({}));
+        this.container = this.container.parentElement.removeChild(this.container);
```
robwalch: Not a fan of this. it doesn't make sense to template everything - especially where speed is concerned. The benefit here is setting the element's classname but at the cost of adding a module, and calling and manipulating the result of utils.createElement.

Ultimately we should to do all setup in a document fragment, and append that to the page after the view is ready. We should also remove this responsibility from api.js and let view.js deal with the old container.

PR #383, opened 2015-04-15
Hide controls when controls(false)

donato:


src/js/view/view.js
```diff
@@ -501,7 +501,17 @@ define([
             _this.trigger(evt.type, evt);
         }
 
+        var toggleControls = function() {
+            if (_model.get('controls')) {
+                _controlsLayer.style.display = '';
+            } else {
+                _controlsLayer.style.display = 'none';
+            }
```
robwalch: can we toggle a flag (css class) that hides the controls layer?

donato: It is done. The rule is in "styles.less" but will be moved into view.less after Kyles related change.

PR #386, opened 2015-04-16
Separate "complete" state from "idle" state

donato:
This change causes providers to set their state to COMPLETE as opposed to
IDLE after playback has concluded.

This distinction is important because the player behaves differently before playback
has began, versus after it has finished. Currently this difference is limited to the
View, however it could become important for plugins related to sharing as well.

[Delivers #92440878]

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -222,6 +222,8 @@ define([
             }
 
             function _stop(internal) {
+                var fromApi = !internal;
+
                 // Something has called stop() in an onComplete handler
                 if (_isIdle()) {
                     _stopPlaylist = true;
@@ -238,7 +240,7 @@ define([
                     return false;
                 }
 
-                if (!internal) {
+                if (fromApi) {
                     _stopPlaylist = true;
                 }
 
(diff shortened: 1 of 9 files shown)


src/js/controller/controller.js
```diff
@@ -311,19 +314,24 @@ define([
...
+
+                var idx = _model.get('item');
+                if (idx === _model.get('playlist').length - 1) {
+                    // If it's the last item in the playlist
+                    if (_model.get('repeat')) {
+                        _next();
+                    } else {
                         _loadOnPlay = 0;
-                        _stop(true);
                         setTimeout(function() {
```
robwalch: does this still need to be deferred? We should try to get rid of as many of these as possible.

src/js/view/view.js
```diff
@@ -877,9 +877,12 @@ define([
         }
 
         function _showDisplay() {
+            var hasControls = _model.get('controls');
+            var state = _model.get('state');
+
             var model = _instreamMode ? _instreamModel : _model;
```
jw-kaurand: Doesn't look like you're using model anymore after the refactor.  Can probably get rid of it.

donato:
Going to keep looking into this.

PR #387, opened 2015-04-17
Filter out playlist sources when no provider can be chosen [Delivers #92362852 #92391708]

robwalch:


src/js/providers/providers.js
```diff
@@ -35,7 +35,7 @@ define([
             var idx = _.indexOf(this.providers, p);
             if (idx < 0) {
                 // No provider matched
-                return Number.MIN_VALUE;
+                return 0;
```
donato: Can we use a negative index here? This would follow the precedence of indexOf which is 0 or higher if found and less than 0 if not found

PR #388, opened 2015-04-17
Unit test setup module

robwalch:


src/js/controller/Setup.js
```diff
@@ -79,16 +76,11 @@ define([
...
         function _nextTask() {
-            if (_cancelled) {
-                return;
-            }
-
             for (var i = 0; i < _queue.length; i++) {
                 var task = _queue[i];
                 if (_allComplete(task.depends)) {
-                    _queue.splice(i, 1);
+                    _queue.splice(i--, 1);
```
donato: This is a little confusing, maybe a while loop is a better option here

PR #389, opened 2015-04-17
View template Aspect Ratio Fix for Firefox [Finishes #90152754]

jw-kaurand:
Fix by using insertRule which is more compatible cross-browser (was working in webkit but not Firefox).  There was also an issue where there were 2 '%' signs, which chrome counted as 1

[Finishes #90152754]

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -379,7 +379,9 @@ define([
 
         function _setAspectRatio(){
             var styleTarget = '#'+_playerElement.id+'.jw-aspect-mode:before';
-            document.styleSheets[0].addRule(styleTarget, 'padding-top: ' + _model.aspectratio + '%');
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -379,7 +379,9 @@ define([
 
         function _setAspectRatio(){
             var styleTarget = '#'+_playerElement.id+'.jw-aspect-mode:before';
-            document.styleSheets[0].addRule(styleTarget, 'padding-top: ' + _model.aspectratio + '%');
+            var targetStylesheet = document.styleSheets[0];
```
robwalch: You should not be modifying an arbitrary style sheet on the page. Query the element (if you don't already have a reference to it) and set inline styles.

PR #393, opened 2015-04-18
Add multi-line caption support

mattcoz:
Adding a "white-space" value of "pre-line" to support the proper display of multi-line captions.


donato:
This change will allow newline characters to break the line, while still collapsing spaces and tabs internally - as defined [here](https://developer.mozilla.org/en-US/docs/Web/CSS/white-space).

Looks good, thanks for the commet @mattcoz - just a heads up, we will likely be rewriting this module in the coming weeks, to use html templating and *.less files for styling.

PR #401, opened 2015-04-20
Show a preview image for playlist items

donato:
[#92873620]

### DIFF
--- a/src/css/imports/display.less
+++ b/src/css/imports/display.less
@@ -1,10 +1,19 @@
 @import "vars.less";
 .jw-display {
     position: absolute;
-    display: none;
+    //display: none;
     width: 100%;
     height: 100%;
 }
+
+/*
+.jwplayer .jw-controls {
+    pointer-events : none;
+}
+.jwplayer.jw-user-inactive .jw-controls {
+    pointer-events : all;
+}
+*/
 
 
 // Preview Image
--- a/src/js/view/preview.js
+++ b/src/js/view/preview.js
@@ -1,0 +1,36 @@
+define([
+    'utils/underscore'
+
+], function(_) {
+
+    var Preview = function(_model) {
+        this.model = _model;
+
+        this.setup();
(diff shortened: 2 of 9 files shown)


src/js/view/preview.js
```diff
@@ -0,0 +1,36 @@
...
+
+            this.loadImage(this.model, this.model.get('playlistItem'));
+        },
+        loadImage: function(model, playlistItem) {
+            var img = playlistItem.image;
+
+            if (_.isString(img)) {
+                this.el.style['background-image'] = 'url(' + img + ')';
+            } else {
+                this.el.style['background-image'] = '';
```
jw-kaurand: 'this' refers to the model when we get here as a result of the listener on the model.  It's causing errors when switching between playlist items.

donato: Yup, you're right, line 10 should bind "this" by using the third parameter.
I'll get that in tomorrow.

src/js/view/view.js
```diff
@@ -879,26 +880,13 @@ define([
...
-
             // debug this, find out why
-            if (!(_isMobile && model.fullscreen)) {
-                model.getVideo().setControls(false);
+            if (!(_isMobile && _model.fullscreen)) {
+                _model.getVideo().setControls(false);
             }
         }
 
         function _hideDisplay() {
```
jw-kaurand: Seems like we can get rid of this and a few of the calls to it now that there's nothing left in it

src/css/imports/display.less
```diff
@@ -1,11 +1,20 @@
 @import "vars.less";
 .jw-display {
     position: absolute;
-    display: none;
+    //display: none;
     width: 100%;
     height: 100%;
 }
 
+/*
```
jw-kaurand: Can these be removed or are they reminders for something?

jw-kaurand:
A newer version of this branch has been merged.

PR #402, opened 2015-04-21
disable cookie storage [Delivers #86021180]

robwalch:


src/js/controller/model.js
```diff
@@ -43,6 +44,11 @@ define([
             },
             _currentProvider = utils.noop;
 
+        if (config.cookies) {
```
donato: If we are using this, then we should add "cookies" to the _defaults map

src/js/controller/model.js
```diff
@@ -264,11 +268,11 @@ define([
         this.loadVideo = function() {
             this.mediaController.trigger(events.JWPLAYER_MEDIA_PLAY_ATTEMPT);
             var idx = this.get('item');
-            this.getVideo().load(this.get('playlist')[idx]);
+            _provider.load(this.get('playlist')[idx]);
         };
 
         this.playVideo = function() {
-            this.getVideo().play();
+            _provider.play();
```
donato: ```this.getVideo()``` is better because it abstracts away the fact that we have the current provider bound in the Model closure. I think this is important because we want to remove the _provider var from model at some point, and instead pass the commands to the mediaController

src/js/controller/model.js
robwalch: It's here https://github.com/jwplayer/jwplayer/pull/402/files#diff-50a064d2ccf19c5ee17fb7fc2e513525R7

src/js/controller/model.js
robwalch: Not valid at this point because _provider is being used more than getVideo(). I just went with the one we use more, preferring to use internal vars over public methods.

src/js/controller/model.js
donato: Gotcha, thanks

PR #414, opened 2015-04-29
Fix for stretching preview images [Finishes #92873620]

jw-kaurand:
This fixes an issue where the preview image wasn't receiving rules to style it when the scaling mode changed.  Stretching class applied to the jwplayer container where we call resize in view.js.  Scaling for the preview image defaults to uniform scaling.  Was checking whether we could apply all stretching via classes, but fill and exactfit would require inline styling.  Remains being done via JS for now, but may be  able to be done with less JS and more CSS if needed. [Finishes #92873620]

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -307,6 +307,13 @@ define([
                 utils.toggleClass(_controlsLayer, 'jw-cast-available', val);
             });
 
+            // set initial state
+            if(_model.get('stretching')){
+                _onStretchChange(_model, _model.get('stretching'));
+            }
+            // watch for changes
+            _model.on('change:stretching', _onStretchChange);
+
             _model.on('change:castState', function(evt) {
                 if (!_castDisplay) {
                     _castDisplay = new CastDisplay(_model.id);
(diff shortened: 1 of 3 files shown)


src/js/view/view.js
```diff
@@ -736,6 +736,12 @@ define([
                 return;
             }
             var transformScale = provider.resize(width, height, _model.stretching);
+            if(_model.stretching && !utils.hasClass(_playerElement, 'jw-stretch-' + _model.stretching)) {
+                utils.removeClass(_playerElement,
+                    'jw-stretch-none jw-stretch-uniform jw-stretch-fill jw-stretch-exactfit');
```
donato: Since this is meant to directly mirror the state of the model, we can break it out of the resize function.

In the View constructor
```
    // set initial state
    this.onStretchChange(_model, _model.get('stretching');
    // watch for changes
    _model.on('change:stretching', this.onStretchChange, this);
```

Then when you define "onStretchChange", you can use oldVal and newVal to modify the classes, instead of a whole list of possibilities

PR #419, opened 2015-05-01
Ads control bar styled and Fixes some VAST setup issues

jw-kaurand:
Creating flag classes and related stylesheets for Ads and Live state.
Fixes some issues with Vast ads where the player would not start up.
Alt text modified to have its class named changed slightly and toggled via the flag states.
[Finishes #91599370 #92236942]


src/css/states/ads.less
```diff
@@ -0,0 +1,23 @@
+.jwplayer.jw-flag-ads {
+    .jw-controlbar {
+        display: table !important;
```
donato: We need it to fade out, so this won't work

donato:
In our current design, "ads" and "live" are not states, they are flags. To me, this distinction means that the player is always in precisely one state (playing/paused/...) but may have multiple modifier flags (audiomode/ads/live/...) 

Do you think they should be states instead?

donato:
Great call on using class flags to show/hide the timebar and moving that into the view. Much better now.

PR #422, opened 2015-05-01
Move the logo to the right side of the page

donato:
[Delivers #93117792 #93061754]

### DIFF
--- a/src/css/imports/dock.less
+++ b/src/css/imports/dock.less
@@ -1,10 +1,7 @@
 @dock-corner-radius: 0.25em;
 
 .jw-dock {
-    position: absolute;
-    top: @ui-margin;
-    right: @ui-margin;
-    left: @ui-margin;
+    margin: @ui-margin/2;
     display: block;
     opacity: 1;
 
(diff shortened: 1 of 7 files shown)


src/css/imports/dock.less
```diff
@@ -12,15 +9,21 @@
         cursor:pointer;
         float: left;
         position: relative;
+
+        // This allows it to be positioned beneath the logo
+        clear : right;
```
jw-kaurand: IT can look odd with a larger logo and only one dock button.  If you float the buttons right an clear the dock left, then you can get the dock below the logo and right align the buttons.

screen shot 2015-05-04 at 12 03 55 pm

src/css/imports/dock.less
```diff
@@ -12,15 +9,21 @@
...
         height: 2em;
-        margin-right: 0.5em;
+        margin: 0.5em;
         background-color: @def-background-color;
         background: @def-transparent-background-style;
         background-size: @def-background-size;
         border-radius: @dock-corner-radius;
     }
 
+    .jw-dock-button:first-child {
```
jw-kaurand: Do we need this style?

PR #425, opened 2015-05-04
Audio-only player style update [Fixes #93659784]

jw-kaurand:
Audio only player update so that it is visible in audio-only mode.  Further development for this (such as having a poster while audio is playing) would need to be a separate ticket to implement.  [Fixes #93659784]

### DIFF
--- a/src/css/flags/audioplayer.less
+++ b/src/css/flags/audioplayer.less
@@ -1,5 +1,28 @@
-.jwplayer {
-    &.jw-flag-audio-player {
+.jwplayer.jw-flag-audio-player {
+    background-color: transparent;
+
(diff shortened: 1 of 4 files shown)


src/css/flags/audioplayer.less
```diff
@@ -1,5 +1,28 @@
-.jwplayer {
-    &.jw-flag-audio-player {
+.jwplayer.jw-flag-audio-player {
+    background-color: transparent;
+
+    .jw-preview {
         background-color: transparent;
     }
+    .jw-controlbar {
+        display: table;
```
donato: We should have this one from imports/controlbar.less, are there others that we can also prune out?

PR #426, opened 2015-05-04
Locking down class names

donato:
Internal structure is now in line with : https://github.com/jwplayer/jwplayer-commercial/wiki/Html-Structure-and-Classes
[#88947454]

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -248,7 +248,7 @@ define([
                 return;
             }
 
-            _container = _playerElement.getElementsByClassName('jw-main')[0];
+            _container = _playerElement;
             _videoLayer = _playerElement.getElementsByClassName('jw-media')[0];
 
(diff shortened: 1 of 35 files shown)


src/js/view/view.js
```diff
@@ -249,11 +249,21 @@ define([
             }
 
             _container = _playerElement.getElementsByClassName('jw-main')[0];
```
robwalch: @donato was `jw-main` removed? If so we need to create another container for plugins and a sweep needs to be done to cleanup orphaned code.

jw-kaurand: Seems like this is causing some issues with jwplayer-commercial/test/vast-linear.html due to it looking for the jw-main class

donato: Plugins will go into .jw-overlays, (this is annotated in lucidcharts using
a comment)

robwalch: ok. updated chart and commit. it read 'jw-overlay' which is taken by tooltips

PR #430, opened 2015-05-06
Various VAST fixes [Finishes #92236942]

jw-kaurand:
VAST fixes include

* Flash Postrolls not playing
* Prerolls not having a skip button
* Ad companions not firing
* 2nd+ ad pods not displaying ad text
[Finishes #92236942]

### DIFF
--- a/src/js/providers/flash.js
+++ b/src/js/providers/flash.js
@@ -158,8 +158,7 @@ define([
                     ];
 
                     var forwardEvents = [
-                        events.JWPLAYER_MEDIA_BUFFER_FULL,
-                        events.JWPLAYER_MEDIA_BEFORECOMPLETE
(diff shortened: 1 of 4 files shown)


src/js/providers/flash.js
```diff
@@ -186,10 +188,17 @@ define([
...
                         this.sendEvent(e.type);
 
-                    }, this).on(events.JWPLAYER_MEDIA_COMPLETE, function(e) {
-                        this.setState(states.COMPLETE);
+                    }, this).on(events.JWPLAYER_MEDIA_BEFORECOMPLETE, function(e){
+                        _beforecompleted = true;
                         this.sendEvent(e.type);
-
+                        if(_attached === true) {
+                            _beforecompleted = false;
```
robwalch: There are some other places where this should be reset to false, no?

PR #433, opened 2015-05-07
add onCuePoint event to flash player

ganondalf:
Sends a cue point event at the time a cue point occurs


robwalch:
@ganondalf please open a new PR merging into master - that's out development branch. If this checks out in master, it will get merged into release for our next release candidate.

PR #437, opened 2015-05-07
add onCuePoint event to flash player

ganondalf:
Sends a cue point event at the time a cue point occurs


donato:
If you want to get this into JW7, please make the request against the branch jwplayer7/master

donato:
For JW6 as a hotfix, it will be up to @esboyd if we do it or not

PR #439, opened 2015-05-08
Fixes to getSafeRegion

jw-kaurand:
_controlbar.showTemp and _controlbar.hideTemp just made the controller visible so its dimensions could be found.  instead we're using one function that gets the dimensions of the controlbar without relying on toggling it on and then off.  Fixes an case where dock buttons could be undefined and adds a clearfix psuedoelement to the dock so that the floated dock buttons will push out the dock's size. previously was only returning 0 height.

### DIFF
--- a/src/css/imports/dock.less
+++ b/src/css/imports/dock.less
@@ -7,12 +7,17 @@
 
     // This allows it to be positioned beneath the logo
     clear : right;
+
+    &:after {
+        content: '';
+        clear: both;
+        display: block;
+    }
     
     .jw-dock-button {
         cursor:pointer;
         float: right;
         position: relative;
-
 
         width: 2.5em;
         height: 2em;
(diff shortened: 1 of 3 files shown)


donato:
Wouldn't this make the controlbar flicker?

robwalch:
@donato Yeah it might. @jw-kaurand show/hide temp used to temporarily give the controller layout (display !none) so that we can get it's bounds, without making it visible. It used to set the opacity to 0 to do this. That may not be necessary. If it's visibility is already hidden we should still be able to get it's bounds as long as display is set to table.

```
        _this.showTemp = function() {
            if (!this.visible) {
                _controlbar.style.opacity = 0;
                _controlbar.style.display = 'inline-block';
            }
        };

        _this.hideTemp = function() {
            if (!this.visible) {
                _controlbar.style.display = 'none';
            }
        };
```

PR #440, opened 2015-05-11
Font updates and controlbar styling updates

jw-kaurand:
* Updates margins on left and right child components 
* got vertical sliders styled correctly
* moved thumbs around
* fixed vertical centering issue for the time bar. (+1 squashed commit)
* bounds for the thumb position rendering
* left align menu options
* centering thumbs for sliders
* removes padding and margin on sliders to make the sub bars lay on top of each other correctly
* margins for elements set instead of padding to make layout more consistent
* corrects classes on the casting icon
* time tooltip no longer has an undefined class

[Finishes #93718726] [Finishes #93712934]


src/js/view/components/slider.js
```diff
@@ -87,6 +87,8 @@ define([
             this.trigger('update', { percentage : percentage });
         },
         render : function(percentage) {
+            percentage = Math.max(0, Math.min(percentage, 100));
```
donato: percentage = utils.between(percentage, 0, 100);

PR #442, opened 2015-05-12
Adds a state for errors [Finishes #92560486]

jw-kaurand:
Squashed commits:
[e0c49a4] Font Color Changes and test updates

* Changed font colors for titles.
* Moved error screen style location out of the error state styles so that it doesn't apply any width or height styling while in the error state
* Added error state to the UI test pages.

[Finishes #92560486]
[920ffd7] Adds a state for errors

Adds an error state to the player that occurs when items can't be loaded or played.  The styles for this state are shared with the startup error state which will occur when the player cannot setup with the information it currently has.

[Finishes #92560486]
[5488a3d] Error screen in progress

Errorscreen changed to a flag so that view can use it to display error information
Error screen mocks the HTML structure and uses the same css to display the error info


src/js/providers/flash.js
```diff
@@ -222,7 +222,6 @@ define([
                         this.sendEvent(events.JWPLAYER_MEDIA_ERROR, {
                             message: 'Error loading media: File could not be played'
                         });
-                        this.setState(states.IDLE);
```
donato: Why not use 

    this.setState(states.ERROR);  //?

Without it, the mediaModel's state and the models state will fall out of sync

robwalch: Will that create a new event / player state in the public API? we should avoid that.

jw-kaurand: this.setState(states.IDLE); was removed because setting that state is handled in the controller.js with the JWPLAYER_MEDIA_ERROR handler on line 98.  That changes the state so we don't have to put that in every provider.   As far as the public API we're in about the same boat as we are with the complete state.  jwplayer().getState() returns 'error' in the error state as the jwplayer().getState() returns 'complete' in the complete state.  To match the api from jw6 we'd probably have to have both of them return 'idle' in their states.

donato: Thanks for the explanation, it makes a lot of sense to set the state in the controller. When this is mergable ping me and I'll put it in, or feel free to merge it yourself.

robwalch:
What's the hold up on this one? Can we clean it up or close it?

robwalch:
@jw-kaurand close this with a link to your new PR when ready

jw-kaurand:
An updated version of this PR is available here: https://github.com/jwplayer/jwplayer/pull/479

PR #443, opened 2015-05-13
New skins

kasperific:


jw-kaurand:
PR based on older code.  Would overwrite more recent changes.  https://github.com/jwplayer/jwplayer/pull/448 will be used instead.

PR #444, opened 2015-05-13
Load providers before continuing setup

donato:
In the future we will move this into commercial with our Dash providers code.

[Delivers #91404968]


donato:
Closing temporarily because webpack in os doesn't like loading providers from commercial.

PR #450, opened 2015-05-14
Slider moves based on page mouse position

jw-kaurand:
Slider Moves based on page mouse position relative to the railBounds


donato:
This is excellent, great work. Tested locally in Chrome/Firefox and Safari

PR #451, opened 2015-05-15
DFXP subtitle parser in HTML5-mode got begin time wrong if there is a duration attribute between begin and end.

johanberonius:
This is a minimal fix to get "begin" time correct if there are other attribute between "begin" and "end" (such as "duration")
It is still pretty naive and breaks if "begin" and "end" attributes come in the other source order. This really should be done with a proper XML-parser.

### DIFF
--- a/src/js/html5/parsers/jwplayer.html5.parsers.dfxp.js
+++ b/src/js/html5/parsers/jwplayer.html5.parsers.dfxp.js
@@ -59,9 +59,10 @@
             try {
                 var idx = data.indexOf('begin=\"');
                 data = data.substr(idx + 7);
-                idx = data.indexOf('\" end=\"');
+                idx = data.indexOf('\"');
                 entry.begin = _seconds(data.substr(0, idx));
-                data = data.substr(idx + 7);
+                idx = data.indexOf('end=\"');
+                data = data.substr(idx + 5);
                 idx = data.indexOf('\"');
                 entry.end = _seconds(data.substr(0, idx));
                 idx = data.indexOf('\">');


johanberonius:
Example:
This worked:
`<p begin="00:01:51.40" end="00:01:54.04" duration="00:00:02.64">Nu kör vi.</p>`

This did not, but now it does:
`<p begin="00:01:04.80" duration="00:00:02.00" end="00:01:06.80">Välkomna</p>`

donato:
Both of these are pretty cool, two quick things

1. Can you please show two pages, one that demonstrates the bug, and one which demonstrates the fix?
2. Can you please make the pull request against our branch jwplayer7/master instead of master. We have our master branch locked right now.

johanberonius:
> Can you please show two pages, one that demonstrates the bug, and one which demonstrates the fix?

This program is working:
http://www.ur.se/Produkter/183058-Kalkyl-Volym
with this subtitle file:
http://undertexter.ur.se/183000-183999/183058-15.tt
that look like this:
`<p begin="00:03:32.12" end="00:03:33.68" duration="00:00:01.56">Precis.</p>`

But this program doesn't work before this fix
http://www.ur.se/Produkter/181810-Bilderna-som-forandrade-vetenskapen-Apolloprogrammet
with this subtitle file:
http://undertexter.ur.se/181000-181999/181810-22.tt
that look like this:
`<p begin="00:03:16.60" duration="00:00:02.68" end="00:03:19.28">När de landade på månen</p>`

I do not have a public page with this fix implemented at the moment.

> Can you please make the pull request against our branch jwplayer7/master instead of master. We have our master branch locked right now.

I made a pull request to that branch too:
https://github.com/jwplayer/jwplayer/pull/452

Note that this is for HTML5-mode in JW Player 6, it works in Flash-mode since the subtitle files are parsed as XML.
In JW Player 7 this seems to affect both modes.

robwalch:
> In JW Player 7 this seems to affect both modes.

@johanberonius yup! In JW7 DFXP captions parsing and rendering is handled in JavaScript.

donato:
JW6 has been locked down for new feature development, this is a cool change and we will work on getting the fix into the jwplayer7/master branch.

PR #452, opened 2015-05-15
DFXP subtitle parser got begin time wrong if there is a duration attribute between begin and end.

johanberonius:
This is a minimal fix to get "begin" time correct if there are other attribute between "begin" and "end" (such as "duration")
It is still pretty naive and breaks if "begin" and "end" attributes come in the other source order. This really should be done with a proper XML-parser.

### DIFF
--- a/src/js/parsers/captions/parsers.dfxp.js
+++ b/src/js/parsers/captions/parsers.dfxp.js
@@ -56,9 +56,10 @@ define([
             utils.tryCatch(function() {
                 var idx = data.indexOf('begin=\"');
                 data = data.substr(idx + 7);
-                idx = data.indexOf('\" end=\"');
+                idx = data.indexOf('\"');
                 entry.begin = _seconds(data.substr(0, idx));
-                data = data.substr(idx + 7);
+                idx = data.indexOf('end=\"');
+                data = data.substr(idx + 5);
                 idx = data.indexOf('\"');
                 entry.end = _seconds(data.substr(0, idx));
                 idx = data.indexOf('\">');


johanberonius:
Example:
This worked:
`<p begin="00:01:51.40" end="00:01:54.04" duration="00:00:02.64">Nu kör vi.</p>`

This did not, but now it does:
`<p begin="00:01:04.80" duration="00:00:02.00" end="00:01:06.80">Välkomna</p>`

donato:
Same comment as on the other PR.

src/js/parsers/captions/parsers.dfxp.js
```diff
@@ -56,9 +56,10 @@ define([
             utils.tryCatch(function() {
                 var idx = data.indexOf('begin=\"');
                 data = data.substr(idx + 7);
```
donato: Overwriting "data" is what causes it to not work when end is before begin. To make it work in all cases, we want to avoid that. Would this work for you?


            utils.tryCatch(function() {
                // Find begin val
                var idxStart = data.indexOf('begin=\"');
                var idxEnd = data.indexOf('\"', idxStart);
                entry.begin = _seconds(data.substr(idxStart, idxEnd));

                // Find end val
                idxStart = data.indexOf('end=\"');
                idxEnd = data.indexOf('\"', idxStart);
                entry.end = _seconds(data.substr(idxStart, idxEnd));

                // Find text val
                idxStart = data.indexOf('\">');
                data = data.substr(idxStart + 2);
                entry.text = strings.trim(data).replace(/>\s+</g, '><');
            });

robwalch:
We could use `jwplayer.utils.parseXML(string)` to return an xml document, then step through child nodes and attributes.

src/js/parsers/captions/parsers.dfxp.js
johanberonius: That is better, but you still need to offset `idxStart`with the length of the attribute name. `idxStart + 7` for `begin="` and `idxStart + 5` for `end="`.

Or better yet, just match the whole thing with a RegExp and forget about all `indexOf`s and `substr`s.

But then again, maybe proper XML-parsing is the best way. The way it is done in Flash-mode in JW Player 6:
https://github.com/jwplayer/jwplayer/blob/master/src/flash/com/longtailvideo/jwplayer/parsers/DFXP.as#L45

donato:
Agreed on all of it. @johanberonius feel free to choose either path, if you submit a reasonable update we will be glad to merge it.

robwalch:
@johanberonius PR #478 parses DFXP as XML. Here's the [diff](https://github.com/jwplayer/jwplayer/commit/1574b0d7bc4b9cdeecbf5a85273cea5a1da4054c?diff=split#diff-6cc80f30e8fecb522d563e0fba5485eeL7).

PR #468, opened 2015-05-26
Jwplayer7/css skins update rebase

kasperific:


donato:
First things first, can you check your editor settings to make sure that the "tab" button does 4 spaces instead of a tab character? This is a stylistic rule that we try to enforce across the code base.

Overall it looks really cool! One thing that I noticed that can be improved upon, is the strategy of declaring a bunch of variables before calling the "basic-skin-styles" mixin. In programming, it is always better to explicitly pass values as parameters as opposed to setting global values which are referenced elsewhere. This is considered a better option because it is less vulnerable to side-effects which can occur when using global or shared variables.

kasperific:
Thanks! 

So, the variables declared within each skin is meant to override global variables in the cascade. They will only affect the selectors in which they are defined. The more specific .jw-six-skin variables overwrite global variables assigned to the .jwplayer class, but not do affect any other skin.

Switching it to a parametric mixin would mean something like this:

.mixin(@color; @padding: 2; @var3; @var4; @var5; @var6; @var7; @var8; @var9; @var10; @var11; @var12; @var13; @var14; @var15; @var16; @var17 ) {
  color-2: @color;
  padding-2: @padding;
....
}

And each skin would have to have a value for each of those parameters. They don't all share the same parameters, which is why some skins have many local variables (six) and some have very few (glow).

Kyle says to think of it more like underscore.js. Does that make sense?

jw-kaurand:
It's acting in a similar way to how we use same way we use _.extend() for overwriting default config variables.  I can see the value in dealing with assigning local values to override default values since this way it's a more concise, more readable and a much more LESS oriented solution to the problem.  Creating a mixin with a bunch of parameters that uses some of the global variables as parameters doesn't gain much more safety in comparison to this approach.

jw-kaurand:
Closing so that we can use https://github.com/jwplayer/jwplayer/pull/470 instead which has a few formatting updates and updates to make RECESS happy.  Otherwise identical.

PR #469, opened 2015-05-27
simple skin highlights for menus

jw-kaurand:
Assigns skin colors to menu items which were missed previously

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -278,15 +278,22 @@ define([
                 '.jw-thumb:after'
             ];
 
-            var iconHovers = ['.jw-icon:hover'];
+            var iconHovers = ['.jw-icon:hover', '.jw-option:hover'];
             var sliders = ['.jw-progress'];
             var containers = ['.jw-container'];
+
 
             addStyle('color', iconHovers, _model.get('skinColorActive'));
             addStyle('color', icons, _model.get('skinColorInactive'));
             addStyle('background', cues, _model.get('skinColorInactive'));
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -278,15 +278,22 @@ define([
...
             addStyle('color', iconHovers, _model.get('skinColorActive'));
             addStyle('color', icons, _model.get('skinColorInactive'));
             addStyle('background', cues, _model.get('skinColorInactive'));
             addStyle('background', sliders, _model.get('skinColorActive'));
             addStyle('background', containers, _model.get('skinColorBackground'));
+
+            addStyle('color', ['.jw-active-option'], _model.get('skinColorActive'));
+            addStyle('background', ['.jw-active-option'], _model.get('skinColorInactive'));
+
+            addStyle('color', ['.jw-icon-hd', '.jw-icon-cc'], _model.get('skinColorActive'));
```
donato: There is something loopy going on here, 

on line 287 we say "color icons the inactive color"
on line 295 we say "these two icons should be colored the active color"
on line 296 we say "color icons the inactive color if they have the class off"

One solution is to use an "on" class instead of an "off" class. What do you think?

jw-kaurand: We don't currently have a distinct "jw-on/jw-active" style for tooltip icons, since that's covered by our default styling.  Just targeting 'jw-icon-tooltip' instead of those two classes would introduce that logic on the volume tooltip as well.  This is logic targeted at our menu tooltips so maybe it makes sense to have that as a specific class.

PR #474, opened 2015-05-28
DRM protection rename [#95182330]

chazmatazz:


src/js/playlist/playlist.js
```diff
@@ -48,9 +49,12 @@ define([
             if (androidhls) {
                 originalSource.androidhls =  androidhls;
             }
-
-            if (originalSource.protection || itemProtection) {
-                originalSource.protection = originalSource.protection || itemProtection;
+            if(_hasDrm()) {
```
robwalch: Edition enforcement should be done in commercial where the feature is is implemented, not here in os playlist source normalization.

robwalch: @chazmatazz it's a good start, but can you remove the second commit and implement the edition check [here](https://github.com/jwplayer/jwplayer-commercial/blob/5f4bc2b84cd01974fefac14f8727d621a169945e/src/js/providers/shaka.js#L113)?

PR #476, opened 2015-05-29
Touch Event Refactor [Finishes #92560486]

jw-kaurand:
Streamlines interactions for desktop and mobile by handling touch and mouse events through unified event flows.  We can still differentiate the different event types where necessary (tapping the screen to display the UI on tablet vs pausing when clicking) and menus are navigateable via touch or mouse interaction.

[Finishes #92560486]

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -13,88 +13,73 @@ define([
 
     var ui = function (elem) {
         var _elem = elem,
-            _isListening = false,
+            _isPressed = false,
             _startEvent = null,
-            _gotMove = false,
+            _hasMoved = false,
             _lastClickTime = 0;
 
         _.extend(this, Events);
 
-        elem.addEventListener(TOUCH_START, touchHandler);
-        document.addEventListener(TOUCH_MOVE, touchHandler);
-        document.addEventListener(TOUCH_CANCEL, touchHandler);
-        document.addEventListener(TOUCH_END, documentEndHandler);
-        elem.addEventListener(TOUCH_END, documentEndHandler);
+        elem.addEventListener(TOUCH_START, interactStartHandler);
 
-        elem.addEventListener(MOUSE_DOWN, touchHandler);
-        elem.addEventListener(MOUSE_MOVE, touchHandler);
-        elem.addEventListener(MOUSE_UP, documentEndHandler);
+        elem.addEventListener(MOUSE_DOWN, interactStartHandler);
 
-        elem.addEventListener('click', clickHandler);
+        function interactStartHandler() {
+            _isPressed = true;
 
-        function clickHandler(evt){
-            triggerEvent(events.touchEvents.CLICK, evt);
+            elem.addEventListener(TOUCH_MOVE, interactDragHandler);
+            elem.addEventListener(TOUCH_CANCEL, interactEndHandler);
+            elem.addEventListener(TOUCH_END, interactEndHandler);
+
+            document.addEventListener(MOUSE_MOVE, interactDragHandler);
+            document.addEventListener(MOUSE_UP, interactEndHandler);
         }
 
-        function documentEndHandler(evt) {
-            console.log('End Handler');
-            if (_isListening) {
-                if (_gotMove) {
-                    triggerEvent(events.touchEvents.DRAG_END, evt);
-                } else {
-                    if(!(evt instanceof MouseEvent)){
-                        triggerEvent(events.touchEvents.TAP, evt);
-                    }
-                }
-            }
-            _gotMove = false;
-            _isListening = false;
-            _startEvent = null;
-        }
-
-        function touchHandler(evt) {
+        function interactDragHandler(evt) {
             var touchEvents = events.touchEvents;
-            console.log('evt', evt.type);
 
-            if (evt.type === TOUCH_START || evt.type === MOUSE_DOWN) {
-                _isListening = true;
-                _startEvent = createEvent(touchEvents.DRAG_START, evt);
+            if (_hasMoved) {
+                triggerEvent(touchEvents.DRAG, evt);
+            } else {
+                triggerEvent(touchEvents.DRAG_START, evt, _startEvent);
+                _hasMoved = true;
+                triggerEvent(touchEvents.DRAG, evt);
             }
-            else if (evt.type === TOUCH_MOVE || evt.type === MOUSE_MOVE) {
-                if (_isListening) {
-                    if (_gotMove) {
-                        triggerEvent(touchEvents.DRAG, evt);
-                    }
-                    else {
-                        triggerEvent(touchEvents.DRAG_START, evt, _startEvent);
-                        _gotMove = true;
-                        triggerEvent(touchEvents.DRAG, evt);
-                    }
-                }
-            }
-            else {
-                if (_isListening) {
-                    if (_gotMove) {
-                        triggerEvent(touchEvents.DRAG_END, evt);
+        }
+
+        function interactEndHandler(evt) {
+            var touchEvents = events.touchEvents;
+
+            if (_isPressed) {
+                if (_hasMoved) {
+                    triggerEvent(touchEvents.DRAG_END, evt);
+                } else {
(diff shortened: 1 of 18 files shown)


src/js/utils/ui.js
```diff
@@ -0,0 +1,142 @@
+define([
+    'utils/backbone.events',
+    'events/events',
+    'utils/underscore'
+], function(Events, events, _) {
+    var TOUCH_MOVE = 'touchmove',
```
donato: I don't think declaring the names as vars helps with readability, but that's subjective. We aren't doing it elsewhere for 'click' so not sure we need to do it here.

src/js/utils/ui.js
```diff
@@ -0,0 +1,142 @@
...
+
+        _.extend(this, Events);
+
+        elem.addEventListener(TOUCH_START, interactStartHandler);
+        elem.addEventListener(MOUSE_DOWN, interactStartHandler);
+
+        function interactStartHandler() {
+            _isPressed = true;
+
+            if(self._events[events.touchEvents.DRAG]){
```
donato: This seems functional, but really we want to pretend we don't know the _events object exists. It is an implementation detail of the backbone events object and is subject to change.

src/js/view/adskipbutton.js
```diff
@@ -2,16 +2,19 @@ define([
...
     'handlebars-loader!templates/adskipbutton.html'
-], function(utils, cssUtils, events, Events, _, AdSkipTemplate) {
+], function(utils, cssUtils, events, UI, Events, _, AdSkipTemplate) {
 
     var AdSkipButton = function(skipMessageCountdown, skipMessage) {
         this.skipMessage = skipMessage;
         this.skipMessageCountdown = skipMessageCountdown;
         this.skipMessage = skipMessage;
 
+        new UI (_instreamSkipContainer).on(events.touchEvents.CLICK, skipAd).on(events.touchEvents.TAP, skipAd);
```
donato: This pattern can be a little confusing, an extra set of parenthesis would disambiguate intention

    (new UI(..)).on(..)

src/js/view/controlbar.js
```diff
@@ -2,20 +2,24 @@ define([
...
     function button(icon, click) {
-        var element = document.createElement('span');
+        var element = document.createElement('span'),
+            handler;
         element.className = 'jw-icon jw-icon-inline ' + icon;
         element.style.display = 'none';
 
         if (click) {
-            element.onclick = function() { click(); };
+            handler = function() { click(); };
```
donato: Why is this wrapper function needed?

src/js/view/components/menu.js
```diff
@@ -2,34 +2,51 @@ define([
...
+    'events/events',
+    'utils/ui',
     'handlebars-loader!templates/menu.html'
-], function(Tooltip, utils, _, menuTemplate) {
+], function(Tooltip, utils, _, events, UI, menuTemplate) {
 
     var Menu = Tooltip.extend({
+        'constructor' : function(name) {
+            Tooltip.call(this, name);
+            this.iconUI = new UI(this.el).on(events.touchEvents.TAP, utils.noop);
```
donato: Why extend the constructor, if all of this can go into the setup method?

It seems like since sometimes these will never be set up at all (for example a CC button when there are no closed captions), it would be a waste of effort to set up all the UI events.

src/js/view/adskipbutton.js
robwalch: You can listen to both event at once using 'click tap'. The problem with this approach of registering two event listeners is that some devices will dispatch both events. Instead you want to test for the presence of the Touch class and if present listen for tap else listen for click.

src/js/utils/ui.js
```diff
@@ -0,0 +1,128 @@
...
+            elem.removeEventListener('touchend', interactEndHandler);
+
+            document.removeEventListener('mousemove', interactDragHandler);
+            document.removeEventListener('mouseup', interactEndHandler);
+
+            if (_hasMoved) {
+                triggerEvent(touchEvents.DRAG_END, evt);
+            } else {
+                // This allows the controlbar/dock/logo click events not to be forwarded to the view
+                evt.cancelBubble = true;
```
donato: Looks like we can't use this : https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/cancelBubble

why not do preventDefault()?

src/js/utils/ui.js
```diff
@@ -0,0 +1,128 @@
+define([
+    'utils/backbone.events',
+    'events/events',
+    'utils/underscore'
+], function(Events, events, _) {
+    var ui = function (elem, options) {
```
donato: An object which is meant to be "newed" should be named with a capital letter

src/js/utils/ui.js
```diff
@@ -0,0 +1,128 @@
...
+            if (evt.preventDefault) {
+                evt.preventDefault();
+            }
+        }
+
+        var self = this;
+        function triggerEvent(type, srcEvent) {
+            preventDefault(srcEvent);
+            if( _enableDoubleTap && (type === events.touchEvents.CLICK || type === events.touchEvents.TAP)){
+                if(Date.now() - _lastClickTime < _doubleClickDelay) {
```
donato: Date.now() doesn't work in IE8 - https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/now

_.now() is a good abstraction for this

src/js/view/components/chapters.mixin.js
```diff
@@ -58,6 +58,7 @@ define([
             var _this = this;
             _.each(this.cues, function (cue) {
                 cue.align(duration);
+                // mousemove/mouseout because this currently mouse specific functionality.
```
donato: I don't understand this comment

src/js/view/components/menu.js
```diff
@@ -2,40 +2,50 @@ define([
...
-], function(Tooltip, utils, _, menuTemplate) {
-
+], function(Tooltip, utils, _, events, UI, menuTemplate) {
     var Menu = Tooltip.extend({
         setup : function (list, selectedIndex) {
-            if(this.content){
-                this.content.removeEventListener('click', this.selectListener);
-                this.removeContent();
+            if(!this.iconUI){
+                this.iconUI = new UI(this.el).on('tap', utils.noop);
```
donato: I see this twice, what is the point of the listener which does nothing?

src/js/view/controlbar.js
```diff
@@ -15,7 +16,9 @@ define([
         element.style.display = 'none';
 
         if (click) {
-            element.onclick = function() { click(); };
+            // Some actions receive parameters which change functionality based on their presence (E.G. play)
```
donato: Will we need this comment if we rename the variable from "click" to "apiAction"?

src/js/view/display.js
```diff
@@ -1,32 +1,35 @@
...
     'utils/underscore'
-], function(utils, events, Events, states, stretchUtils, _) {
-
+], function(UI, utils, events, Events, states, stretchUtils, _) {
     var Display = function(_model) {
         var _display,
-            _alternateClickHandler,
-            _lastClick;
+            _alternateClickHandler;
+            //_lastClick;
```
donato: remove commented out code

src/js/view/components/chapters.mixin.js
jw-kaurand: I added that because I wasn't modifying mouseover/mouseout logic in the touch refactor because we aren't duplicating this functionality across mouse and touch interaction.  I'll remove it since its unnecessary.

src/js/view/controlbar.js
jw-kaurand: apiAction adds some clarity.  The second comment explains why we aren't doing
``` javascript
new UI(element).on('click tap', apiAction);
```
instead since having the handler like that would send along the event as a parameter to the api call.

src/js/view/components/menu.js
jw-kaurand: Unnecessary now.  There was an issue previously with how 'off' was working when everything was defined in the constructor.  Not an issue now so it's getting removed.

PR #477, opened 2015-05-29
Premium skin CSS bug fixes

kasperific:


src/css/imports/mixins.less
```diff
@@ -1,9 +1,10 @@
 @import "vars";
 
 .inset-controlbar() {
+    display: block;
     margin: 0 auto;
-    max-width: @controlbar-max-width;
-    bottom: @ui-margin;
+    max-width: 96%;
```
robwalch: Hmm don't we want to use margins and a fixed max width? percentages tend to be arbitrary / last ditch effort.

PR #481, opened 2015-06-02
Add audiotracks UI into controlbar [Delivers #95256388 #87139744]

chazmatazz:


src/js/view/components/listmenu.js
```diff
@@ -0,0 +1,57 @@
+define([
```
robwalch: Too much duplication here. Just make non-togglable a setup configuration option in menu.js.

robwalch: Ex:
``` javascript
setup : function (list, selectedIndex, options) {
    ///
    if (list.length > 2 || (options && options.toggle === false)) {
        ///
    } else if (list.length === 2) {
        ///
    }
```

src/js/view/controlbar.js
```diff
@@ -216,12 +222,26 @@ define([
...
+            this.elements.audiotracks.setup();
+
             this._model.mediaModel.on('change:levels', function(model, levels) {
                 this.elements.hd.setup(levels, model.get('currentLevel'));
             }, this);
             this._model.mediaModel.on('change:currentLevel', function(model, level) {
                 this.elements.hd.selectItem(level);
             }, this);
+            this._model.mediaModel.on('change:audioTracks', function(model, audioTracks) {
+                var list = [];
```
donato: var list = _.map(audioTracks, function(track) { return { label : track.name }; })

donato:
This looks pretty good! Just need to fix the PR

PR #485, opened 2015-06-03
makes scriptloader publicly available

alkupe:


donato:
It would be preferable to add the script loader to the helpers file, from within the helpers file

alkupe:
why

robwalch:
nice cyclical ref @alkupe. That's why.

donato:
It's simpler. Who is in control of the contents of the utils file? The utils file is!

If I want to know what's in it, I know I can look in just one place.

robwalch:
This must go in commercial legacy https://github.com/jwplayer/jwplayer-commercial/blob/dc40c2afb8b7703f235189bb488b143b56b665ce/src/js/api/legacy.js

I am against changes that go against modularity in this project, which this does. Additional requires slow down the code base.

PR #494, opened 2015-06-05
Feature/instream

chazmatazz:


robwalch:
#493 implements hiding of controls by instream

PR #505, opened 2015-06-08
Hide elapsed time and duration for livestreams. 

chazmatazz:


chazmatazz:
This also has the player detect live streams when item duration === Infinity. [Delivers #92363070]

robwalch:
:+1:

PR #507, opened 2015-06-08
Fix instream waterfall edgecases

robwalch:
- __DEBUG__ is not a build constant, and calling destroy more than once has no negative effect
- clear error event handler in the case that instream loads an item more than once (waterfall or adpod)
- do not destroy instream internally because of a media error. instream may be used to load another item (waterfall) and will be destroyed by vast after the error is dispatched if waterfall is not taking place.

### DIFF
--- a/src/js/api/instreamPlayer.js
+++ b/src/js/api/instreamPlayer.js
@@ -44,8 +44,6 @@ define([
             if (_instream) {
                 _instream.instreamDestroy();
                 _instream = null;
-            } else if (__DEBUG__ === true) {
-                console.log('Instream being destroyed a second time');
             }
         };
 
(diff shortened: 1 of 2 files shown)


donato:
Calling destroy more than once was actually an artifact of an old bug in the VAST plugin. That debug code is there to alert us when developing that this code-path was traversed which is a sure sign that something went awry.

The __debug__ constant is part of open source, but I mistakenly forgot to add it to the commercial project. Personally, I'd prefer to add it there, than to remove it here.

robwalch:
At this point, we need to identify actual runtime bugs after this is merged. That debug statement doesn't help us because the inner instream controller was already destroyed.

PR #509, opened 2015-06-09
High specificity for right-click jw icon styles

jw-kaurand:
High specificity for the icon styles so it isn't overridden.  [Finishes #95551132]


donato:
This one should come in after ShaRobs next commit.

robwalch:
Yep. jw-icon-jw-logo is now jw-rightclick-logo

PR #510, opened 2015-06-09
DVR controls [Delivers #96483792]

chazmatazz:


donato:
1. Group the constants into a hash named "AdaptiveTypes"
2. Fix the PR, an example is here : https://github.com/jwplayer/jwplayer/compare/bugfix/dvr-2

robwalch:
Don't need constants for these. It's pretty clear what 'DVR', 'LIVE' and 'VOD' mean. There's no advantage (compiler or autocomplete) to these in js. Just use strings.

PR #514, opened 2015-06-09
CSS cleanup [Delivers #96453472]

robwalch:
- Renamed classes:
  - jw-thumb => jw-knob
  - jw-icon-jw-logo => .jw-rightlick-logo
  - jw-rail-group => .jw-slider-container
- All elements use jw-reset to reset on the element level before other classes are applied
  - insures all elements are reset
  - reset is not more specific than skin styles
- Removed z-index use
  - We should only use DOM manipulation for layering. z-index interferes with host site's css.
- Added .jw-plugin class for plugin overlays
- Added .jw-toggle class for cc and hd buttons when they are not in menu mode
- Add css classes to simplify coloring player
  - .jw-background-color
  - .jw-button-color

### DIFF
--- a/src/css/imports/controlbar.less
+++ b/src/css/imports/controlbar.less
@@ -8,11 +8,14 @@
     bottom: 0;
     height: @controlbar-height;
     padding: 0 @ui-padding;
-    background-color: rgb(65, 64, 64);
 
     .jw-hidden {
         display: none;
     }
+}
+
+.jw-background-color {
+    background-color: rgb(65, 64, 64);
 }
 
 .jw-group {
--- a/src/css/imports/controlbar.less
+++ b/src/css/imports/controlbar.less
@@ -39,7 +39,7 @@
 
 // control bar content (icons/text)
 
-.jw-thumb,
+.jw-knob,
 .jw-icon-inline,
 .jw-icon-tooltip,
 .jw-icon-display,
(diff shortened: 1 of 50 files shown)


robwalch:
@jw-kaurand @donato Merge this before your css changes. This cleanup is essential.

donato:
Getting closer!
I like the jw-background-color, as for jw-button-color, shouldn't that just be jw-button?

1. I think the name jw-knob is a little odd, for reference : 
 * videojs : vjs-seek-handle
 * youtube : html5-scrubber-button
 * flowplayer : *none*

2. The jw-reset everywhere looks silly, I think there is a better why. 

  Why not have jw-reset as a mixin which is mixed into all of our other classes? 

   What if "jw-icon" and the rest simply include the styles of jw-reset, and then we ensure all elements have a jw-***anything*** class, instead of everything having a jw-reset and sometimes an additional jw-class?

kasperific:
About the reset, this shouldn't be thought of as in the same vein as a conventional browser reset. .jw-reset is not intended to reset browser defaults. We're preventing user styles from interfering with .jw- styles. So we need a solution that isn't as universal as a reset, and still loose enough to undo any CSS interference from the user.

It may look silly in the markup, but that's a far better scenario than repeating the same styles over and over again in the stylesheet, which is what we'd be doing if we use a mixin or include the .jw-reset styles in every CSS element. 

As for .jw-knob, I'm for whatever everyone's happy with. I do think .jw-nubbin would be fun.

PR #517, opened 2015-06-10
Implement DVR controller and seek support [Finishes #96483792]

chazmatazz:


src/js/view/controlbar.js
```diff
@@ -283,10 +283,13 @@ define([
             this.elements.cast.toggle(val);
         },
         onElapsed : function(model, val) {
-            this.elements.elapsed.innerHTML = utils.timeFormat(val);
+            var duration = this._api.getDuration();
```
robwalch: Get duration from the model. use of api should be avoided in the view.

PR #521, opened 2015-06-11
Should make mouse events work with flash. Needs to be combined with Kyle's code to test

chazmatazz:


donato:
Can't merge things that don't work!

I believe these lines
```js
        this.mouseEnabled = false;		
        this.mouseChildren = false;
```

are supposed to prevent flash right clicks. I found an alternate way to do this, but it requires flash 11.2

http://inflagrantedelicto.memoryspiral.com/2012/06/as3-quickie-removing-the-right-click-menu-from-flash-content/

PR #523, opened 2015-06-11
Don't round parameters that are not numbers [Finishes #94400362]

chazmatazz:
This fixes a specific issue, but has general implications because our data massaging was too generic.

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -22,22 +22,6 @@ define([
         utils.removeClass(container, 'jw-tab-focus');
     }
 
-    var normalizeOutput = function() {
-        var rounders = ['position', 'duration', 'offset'];
-
-        function round(val) {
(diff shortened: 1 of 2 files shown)


robwalch:
@chazmatazz please document what you are fixing next time.

"adRequest" events contain a position property whose value will be a string ("pre", "mid"...). API was rounding any event property named "position" (or "duration" or "offset").

We should remove this rounding feature. It's only purpose is to make values more "readable" in the console or other print out. Silly. DIY. I consider accurate time values a critical player feature.

robwalch:
@chazmatazz I'm seeing "adposition" used in Google IMA events, with some events including the standard time value in "position". Have you looked at other ad events to make sure the property names are right on this one?

chazmatazz:
adposition converts the position to a number. It is undocumented.

PR #524, opened 2015-06-11
Controlbar opacity fix in audio player, .jw-options styles

kasperific:
Increased opacity in audio player controlbar for controlbars that have <1 opacity. Also fixed wrong colors of cc option list items in five and vapor skins.

### DIFF
--- a/src/css/skins/five.less
+++ b/src/css/skins/five.less
@@ -49,7 +49,6 @@
 
         .jw-option {
             border-bottom: 1px solid #ECECEC;
-            color: @active-color;
             background-color: #fff;
             margin-right: 8px;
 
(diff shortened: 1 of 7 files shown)


src/css/skins/five.less
```diff
@@ -159,6 +157,17 @@
         }
     }
 
+    .jw-icon-cc {
```
robwalch: `.jw-icon-cc` should be `jw-icon-tooltip` so that all menus (hd and audiotracks) also get these colors right?

robwalch: @jw-kaurand can you confirm?

src/css/skins/vapor.less
```diff
@@ -149,7 +148,8 @@
 
         .jw-option {
             color: #fff;
-
+            
+            &:hover,
```
robwalch: @kasperific so now vapor has the same problem where options are only set for cc menu?

kasperific:
Additional commits for making .jw-option styles less specific, without affecting playlist options. Applies to Six, Vapor and Five.

PR #531, opened 2015-06-15
Hide controls a set time after user activity

donato:
This unites a few different strategies for hiding the controlbar into one.
Now it is hidden x seconds after the last user activity is recorded.
This fixes the use-case where the mouse would hover over the controlbar and slide
off of the player without hovering over the video element again.

[Delivers #96999912]

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -691,9 +691,7 @@ define([
         function _checkAudioMode(height) {
             _audioMode = _isAudioMode(height);
             if (_controlbar) {
-                if (_audioMode) {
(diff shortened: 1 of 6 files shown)


src/js/view/view.js
```diff
@@ -953,11 +918,9 @@ define([
             // player display
             switch (state) {
                 case states.PLAYING:
+                    _userActivity();
```
robwalch: This should be in the _isAudioFile case below. State changing to play should not result in controls being shown as if there was user activity.

PR #532, opened 2015-06-15
Optimize resizeMedia method

donato:
Since this method is triggered every 1/4 second, it is worth returning
early when we know there is no difference in the height/width and that
the strategy for stretching the video is unchanged.

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -395,6 +395,9 @@ define([
         };
 
         function _onStretchChange (model, newVal, oldVal) {
+            // Reset this var to ensure that _resizeMedia runs all the way
+            _lastMediaWidth = NaN;
+
             if(oldVal){
                 utils.removeClass(_playerElement, 'jw-stretch-' + oldVal);
             }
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -777,6 +782,13 @@ define([
             if (!provider) {
                 return;
             }
+
+            if (_lastMediaWidth === width && _lastMediaHeight === height) {
```
robwalch: If width or height are percentages, then we must not exit. stretching must be recalculated.

robwalch:
Ideal fix for modern browser / mobile would be to use object-fit. See my comments in iPad stretch-mode bug https://www.pivotaltracker.com/story/show/87335014

donato:
Then it's not worth it for now, can revisit for 7.1

robwalch:
Depending on the stretch mode, our media resize method should be optimized to the full extent it can be (given how transforms are used on video). If there are unnecessary style writes there, we should consider optimizing there.

PR #534, opened 2015-06-15
Add UI to jwplayer.utils (for plugins) [#96925108]

chazmatazz:


src/js/api/api.js
```diff
@@ -33,6 +34,7 @@ define([
 
         // This helps plugins, particularly analytics
         this.utils = utils;
+        this.utils.UI = UI;
```
robwalch: This should be done in helpers.js (aka utils). Import and add UI there.

PR #541, opened 2015-06-17
Flash Instream support for linear ads [Delivers #97046594]

robwalch:
Player's instream adapter now handles interfacing with the view, and dealing with ad state (ad pods, text options, etc...). It now also interfaces with either the html5 instream player or Flash's intream player depending on what the active provider is when ad playback is requested.


donato:
using test/vast-linear.html
1. No companion ads
2. State is paused during ad playback, and locked in "buffering" after first midroll
3. play/pause from controlbar do not do anything

PR #544, opened 2015-06-18
Update duration so that live is Infinity and unknown is 0 [#97017426]

chazmatazz:


src/js/utils/helpers.js
```diff
@@ -296,7 +296,7 @@ define([
             if(duration <= MIN_DVR_DURATION) {
                 return 'DVR';
             }
-            if (duration < 0 || duration === Infinity) {
+            if (duration === null || duration < 0 || duration === Infinity) {
```
donato: Can you add a comment here, for each possible duration and why that means live?

// Infinity is Safari, <0 is DVR, null is HLS?

donato:
ahhh, you guys kill me!

```sh
git rebase -i origin/master
```

PR #552, opened 2015-06-19
Trigger click events but do not perform action if controls are hidden [Delivers #97113410]

chazmatazz:


src/js/view/clickhandler.js
```diff
@@ -31,10 +31,6 @@ define([
                 return;
             }
 
-            if (!hasControls) {
```
donato: I agree with your decision to handle controls outside of this util. Do you think we should also remove the other reference to hasControls in this file as well?

chazmatazz: Yes

PR #563, opened 2015-06-23
Instream should not trigger play and pause events [Delivers #97561262]

robwalch:
- fix nullp in InstreamPlayer.as pause()
- detachMedia before media provider is paused
- instream players should not set player model state, causing play & plause api events to be fired
- flash provider should not dispatch events when detached
- view listens to instream model for state events in instream mode

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -115,7 +115,7 @@ define([
             // Last sent buffer amount
             _bufferPercent = -1,
             // Whether or not we're listening to video tag events
-            _attached = false,
+            _attached = true,
             // Quality levels
             _levels,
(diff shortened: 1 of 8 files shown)


robwalch:
@donato

src/js/providers/html5.js
```diff
@@ -115,7 +115,7 @@ define([
             // Last sent buffer amount
             _bufferPercent = -1,
             // Whether or not we're listening to video tag events
-            _attached = false,
+            _attached = true,
```
donato: Is this fixing something, or just neatening it?

robwalch: Just clean up.

PR #565, opened 2015-06-23
Dispatch events for adpods [Fixes #97299612]

robwalch:
VAST expects instream to dispatch a playlist item and a play events for each ad pod, as well as a playlist complete event after the last ad is complete.


Charles-I-Blanton:
Hi all, I'm new to github, so I hope I'm doing this right. We're looking to build and deliver ad pods withing JW Player, using wrapped ad tags. Any help will be appreciated!

PR #568, opened 2015-06-23
Fix doubleclicks [Delivers #97345526]

chazmatazz:


jw-kaurand:
Thanks for the acceptance criteria update on https://www.pivotaltracker.com/story/show/97345526  I'll check that out once i get home

jw-kaurand:
The logic of the PR is sound.  We just need to have a test case to work around for ads and double clicking.

PR #576, opened 2015-06-24
Enable autostart from API load

donato:
This feature was present in JW 6 when using flash mode. After loading a new file or playlist
from the API, it will autoplay when the config was set to autoplay.

[Delivers #97659106]


donato:
Updated

PR #577, opened 2015-06-24
Controlbar fades on desktop. Doesn't fade on mobile.

jw-kaurand:
Removes the display: table; on the jw-flag-ads since it was preventing the controller from naturally being removed when jw-flag-user-inactive was applied.  It was previously there to ensure that the control bar was displayed when we calculate the position of the ad skip button, but that is now the responsibility of the skin.
added a flag for mobile that will make sure that the control bar is displayed during ads.  this flag toggles on and off when its mobile and we're entering instream mode.  _userActivity fired at the beginning of instream mode so that the user can see the new controller state and know its an ad.
!isMobile is added to _componentFadeListeners as we aren't concerned about adding mouse listeners on mobile and it caused a rare issue if the user clicks the control bar a lot during an ad where the control bar wouldn't fade away.

[Finishes #97402648]

### DIFF
--- a/src/css/flags/ads.less
+++ b/src/css/flags/ads.less
@@ -5,8 +5,6 @@
     }
 
     .jw-controlbar {
-        display: table;
-
         .jw-icon-inline,
         .jw-icon-tooltip,
         .jw-text,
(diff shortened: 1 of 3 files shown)


src/css/flags/ads.less
```diff
@@ -38,6 +36,12 @@
 
 .jwplayer.jw-flag-ads-hide-controls {
     .jw-controls {
-      display : none !important;
+        display : none !important;
     }
 }
+
+.jwplayer.jw-flag-ads-mobile {
```
robwalch: I'd prefer a persistent mobile flag on the player. Then in this case you could use:
`.jwplayer.jw-flag-ads-hide-controls.jw-flag-mobile {`

PR #579, opened 2015-06-24
Custom labels [Finishes #97437680]

chazmatazz:


robwalch:
We could make this a feature in the model with support for all providers. Just map hlslabels to qualitylabels, and apply to all provider quality events.

PR #585, opened 2015-06-25
Use window.location.href for absolute path base [Delivers #91031760]

chazmatazz:


robwalch:
@chazmatazz please retest with my changes. get absolute path now gets the page href the same way js does - so the method is fixed, not just the one place we currently use it. Also this is the same method we use in the manifest loader for getting the href using ExternalInterface.

PR #588, opened 2015-07-01
Renamed classes for menu items for safety

jw-kaurand:
Renamed classes for menu items for safety. Now in the format of "jw-item-#" instead of "item-#" to protect against CSS pollution.


donato:
Bueno!

PR #592, opened 2015-07-01
Adds the scale fix when we're automatically scaling the video.

jw-kaurand:
If we're applying our own transforms then we will handle it correctly. Also only calls cssUtils.style once instead of twice on iOS 8

[Delivers #98206456]

### DIFF
--- a/src/js/utils/stretching.js
+++ b/src/js/utils/stretching.js
@@ -138,7 +138,8 @@ define([
             }
 
 
-            //// iOS 8 implemented object-fit poorly and needs additional styles to make it fit correctly when the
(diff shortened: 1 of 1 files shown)


src/js/utils/stretching.js
```diff
@@ -135,10 +136,10 @@ define([
                 scale = false;
                 cssUtils.transform(domelement);
             }
-            cssUtils.style(domelement, style);
 
-            //// iOS 8 implemented object-fit poorly and needs additional styles to make it fit correctly
-            if (utils.isIOS(8)){
+
+            //// iOS 8 implemented object-fit poorly and needs additional styles to make it fit correctly when the
```
chazmatazz: when the... ?

PR #596, opened 2015-07-06
added video tag ended test to stall check for IE [Delivers #97642274]

robwalch:
Video paused is not true in IE after playback ends. Added ended check so that stall checker doesn't change the state to buffering.

In the future we'll make sure to only run this check when playback is active. It would be idea to repurpose  _bufferInterval to perform this check as well, since we shouldn't expect stalls after the buffer is full.

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -26,7 +26,7 @@ define([
     var stallCheckGenerator = function(videotag, stalledHandler) {
         var lastChecked = -1;
         return function() {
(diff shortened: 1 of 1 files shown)


donato:
As noted in conversation, "ended" was chosen solution for the W3C spec, which IE11 follows ; https://connect.microsoft.com/IE/feedback/details/810454/html5-video-audio-doesnt-become-paused-after-playback-ends

PR #600, opened 2015-07-06
Make androidhls not show buffering

jw-kaurand:
androidhls never updates the current time so it will always fall into the stall state.  For now we will assume android hls streaming will send work well.

[Delivers #98233336]

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -52,6 +52,25 @@ define([
         }
     }
 
+    function _useAndroidHLS(source) {
+        if (source.type === 'hls') {
+            //when androidhls is set to true, allow HLS playback on Android 4.1 and up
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -52,6 +52,26 @@ define([
         }
     }
 
+    function _useAndroidHLS (source) {
+        if(source.type === 'hls') {
+            //when androidhls is set to true, allow HLS playback on Android 4.1 and up
+            if (source.androidhls) {
+                var isAndroidNative = utils.isAndroidNative;
+                if (isAndroidNative(2) || isAndroidNative(3) || isAndroidNative('4.0')) {
```
donato: these can be cached up top, look for _isAndroid, _isSafari, etc

PR #602, opened 2015-07-07
Fix various audio player issues

donato:
[Delivers #98561066 98562676 98562522]

### DIFF
--- a/src/css/skins/beelden.less
+++ b/src/css/skins/beelden.less
@@ -42,6 +42,9 @@
         box-shadow: inset 0px 7px 1px -5px rgba(128,128,128,1);
         border-radius: .3em;
     }
+    &.jw-flag-audio-player .jw-controlbar {
+        bottom: 0; // overwrite the inset-controlbar mixin
+    }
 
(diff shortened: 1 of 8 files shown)


src/css/skins/beelden.less
```diff
@@ -42,6 +42,9 @@
         box-shadow: inset 0px 7px 1px -5px rgba(128,128,128,1);
         border-radius: .3em;
     }
+    &.jw-flag-audio-player .jw-controlbar {
+        bottom: 0; // overwrite the inset-controlbar mixin
+    }
```
robwalch: can this be done once in jwplayer.less?

donato: True. I chose to put it here because it keeps it near the rule which originally changes the bottom value. My reasoning being, that if a skin is changing it, than the skin should be responsible for it.

I don't particularly like either solution, @jw-kaurand , how would you do it?

donato: With the goal of reducing duplication, while also keeping the code together, the best solution seems to be modifying the mixin.

PR #603, opened 2015-07-07
Bugfix/drag tolerance

donato:


donato:
@jw-kaurand @chazmatazz

robwalch:
Feels good on iPhone. Delivers [#98586186 #98563942]

PR #604, opened 2015-07-08
Stop bundling all skins into player

donato:
Instead they will be fetched from the CDN

[Delivers #97675542]

### DIFF
--- a/src/css/jwplayer.less
+++ b/src/css/jwplayer.less
@@ -32,12 +32,12 @@
 @import "flags/rightclick-open";
 
 // Skins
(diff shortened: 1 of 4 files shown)


jw-kaurand:
We're not including the other skins, however we're also not able to load them either.  http://localhost:3000/jwplayer/test/manual/ is set to load bekle and I'm not seeing the skin applied to the player.  Very nice reduction in overall size though.  262 kb to 225 kb in commercial.

jw-kaurand:
This looks like it is an issue with the content-type returned by the server.  Will merge this.

PR #608, opened 2015-07-09
Run stall check in same interval as buffer check [Delivers #98297658 #98786942]

robwalch:
Stall check is now performed in same interval as buffer check (during playback). Since the provider already stores currentTime in '_position' this property is used to check that currentTime is advancing. In order to support arbitrary frequency of intervals, the time at which position was last updated is stored in '_wasPlayingAt'. Thus stall check is only performed if more than 'STALL_DELAY' has passed since position last changed. play and time events also update '_position' and '_wasPlayingAt'.

Removed execution of _stalledHandler on video tag 'stalled' and 'waiting' events since we have not seen any example of these indicating under buffering from a user perspective. The interval check performs this task adequately when network is throttled or compromised.

Removed rounding of 'bufferPercent'. We've received good feedback on providing accurate time data so adding this as an additional enhancement to the code being changed makes sense.

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -526,7 +526,7 @@ define([
             if (!buffered || buffered.length === 0 || duration <= 0 || duration === Infinity) {
                 return 0;
             }
-            return Math.max(buffered.end(buffered.length-1) / duration, 1);
+            return utils.between(buffered.end(buffered.length-1) / duration, 0, 1);
         }
 
         function _endedHandler() {
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -501,7 +501,7 @@ define([
             }
 
             var buffered = _getBuffer();
-            if (buffered !== _buffered && buffered <= 1) {
+            if (buffered !== _buffered) {
                 _buffered = buffered;
                 _this.sendEvent(events.JWPLAYER_MEDIA_BUFFER, {
                     bufferPercent: buffered * 100
@@ -522,10 +522,11 @@ define([
 
         function _getBuffer() {
             var buffered = _videotag.buffered;
-            if (!buffered || !_videotag.duration || buffered.length === 0) {
+            var duration = _videotag.duration;
+            if (!buffered || buffered.length === 0 || duration <= 0 || duration === Infinity) {
                 return 0;
             }
-            return buffered.end(buffered.length-1) / _videotag.duration;
+            return Math.max(buffered.end(buffered.length-1) / duration, 1);
         }
 
         function _endedHandler() {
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -508,20 +495,29 @@ define([
...
-                clearInterval(_bufferInterval);
+        function _checkBufferAndPlayback() {
+            if (!_attached) {
+                return;
             }
 
-            if (newBuffer !== _bufferPercent) {
-                _bufferPercent = newBuffer;
+            var buffered = _getBuffer();
+            if (buffered !== _buffered && buffered <= 1) {
```
donato: is the "buffered <= 1" necessary?

src/js/providers/html5.js
```diff
@@ -11,8 +11,8 @@ define([
 ], function(cssUtils, utils, stretchUtils, _, events, states, eventdispatcher, DefaultProvider, video) {
 
     var clearInterval = window.clearInterval,
-        stallInterval,
         STALL_DELAY = 256,
+        BUFFER_INTERVAL = 50,
```
donato: Did you switch this from 100 for a specific reason? The less often we check the better imho

src/js/providers/html5.js
robwalch: Defensive move to insure bufferPercent never exceeds 100. It maintains the logic that was removed, but does not clear the interval:
```
if (newBuffer >= 1) {
   clearInterval(_bufferInterval);
```

src/js/providers/html5.js
robwalch: I think more buffer updates are good, and we're more likely to hit the stall delay earlier (might be better if we make stall delay a little less than a multiple or buffer interval e.g. 240)

donato: I agree that more info is better, I just brought it up because the ticket was originally labeled "optimize-performance"

src/js/providers/html5.js
donato: In that case, I'd rather see the _getBuffer method return Math.min(val, 100)

src/js/providers/html5.js
```diff
@@ -508,28 +495,38 @@ define([
...
 
         function _getBuffer() {
             var buffered = _videotag.buffered;
-            if (!buffered || !_videotag.duration || buffered.length === 0) {
+            var duration = _videotag.duration;
+            if (!buffered || buffered.length === 0 || duration <= 0 || duration === Infinity) {
                 return 0;
             }
-            return buffered.end(buffered.length-1) / _videotag.duration;
+            return Math.max(buffered.end(buffered.length-1) / duration, 1);
```
donato: Did you mean Math.min? Maybe utils.between(num, 0, 1) would be better?

PR #615, opened 2015-07-14
Adding the pointer cursor to the video [Delivers #94696812]

jw-kaurand:
Adding the pointer cursor to the video elements and providers.  This implies that the video is clickable for pause/playback toggling.  The pointer is toggled on/off as a result of 'controls' being set to true or false since that also dictates whether you can click to pause.
Flash has its pointer/arrow cursor set based on the controls value.
jw-controls-disabled changed to a flag so it can toggle cursor state.
Moved code that was dictating whether to show the HTML5 native controls to the HTML5 provider.

[Delivers #94696812]

### DIFF
--- a/src/css/flags/controls-disabled.less
+++ b/src/css/flags/controls-disabled.less
@@ -1,0 +1,9 @@
+.jw-flag-controls-disabled {
+    .jw-controls {
+        display: none;
+    }
+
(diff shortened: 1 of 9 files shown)


donato:
Looks really good, nice work

PR #625, opened 2015-07-17
Show title only if there's content [Delivers #99331362]

jw-kaurand:
Does not show the title container if there is no content to display in title and description.

[Delivers #99331362]

### DIFF
--- a/src/js/view/title.js
+++ b/src/js/view/title.js
@@ -42,7 +42,7 @@ define([
             this.description.innerHTML = (playlistItem.description && model.get('displaydescription')) ?
                 playlistItem.description : '';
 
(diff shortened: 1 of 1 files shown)


src/js/view/title.js
```diff
@@ -42,7 +42,11 @@ define([
             this.description.innerHTML = (playlistItem.description && model.get('displaydescription')) ?
                 playlistItem.description : '';
 
-            this.show();
+            if(this.title.innerHTML !== '' || this.description.innerHTML !== ''){
```
robwalch: Use firstChild. it's less computational than innerHTML when it comes to checking if the element is empty:

`if(this.title.firstChild || this.description.firstChild){`

PR #627, opened 2015-07-20
Display icon width and height equal [Delivers #98200440]

jw-kaurand:
Sets the display icon width and height to be equal in the core skin (3.5em) and removed padding-based centering.  centering now done via text-align.  Slight bump to the right in the idle mode of the player and slight bump upward on complete mode so that the icon looks a bit more visually centered in its container.

[Delivers #98200440]


src/css/states/idle.less
```diff
@@ -5,9 +5,14 @@
...
     }
 
-
     .jw-icon-display {
         .jw-icon-play;
+
+        &:before {
+            // Tweak to improve visual centering of the icon.
+            position: relative;
+            left: 0.075em;
```
robwalch: Would it be easier to maintain if this adjustment was made in the icon?

PR #631, opened 2015-07-22
Play and pause should work for ads as well

donato:
[Delivers #91358582]

### DIFF
--- a/src/js/api/api-mutators.js
+++ b/src/js/api/api-mutators.js
@@ -15,7 +15,6 @@ define([
             'fullscreen',
             'volume',
             'mute',
-            'state',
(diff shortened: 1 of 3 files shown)


jw-kaurand:
Player blows up if you pause(false) or play(true) during the playback of the ad.  Seems like it starts trying to load another ad when that's being done.

donato:
Thanks Kyle, looking into it

donato:
This requires these changes as well to work.

https://github.com/jwplayer/jwplayer-ads-vast/pull/31
https://github.com/jwplayer/jwplayer-ads-googima/pull/83

PR #636, opened 2015-07-28
	Allow the player to load polyfills before setup

donato:


test/unit/setup.js
```diff
@@ -128,7 +129,8 @@ define([
     });
 
     function testSetup(model, success, error, done) {
-        var setup = new Setup(api, model, view, model.config.setupTimeout);
+        var setup = new Setup(api, model, view, model.get('setupTimeout'));
+        console.log(model);
```
robwalch: halo `console.log(model);`

PR #639, opened 2015-07-30
API unit tests

robwalch:
Verifies API method presence and some functionality after setup, providing better code coverage of api and controller.


test/data/api-methods.js
```diff
@@ -0,0 +1,111 @@
...
+    once: null,
+    trigger: null,
+    onAdClick: null,
+    onAdCompanions: null,
+    onAdComplete: null,
+    onAdError: null,
+    onAdImpression: null,
+    onAdMeta: null,
+    onAdPause: null,
+    onAdPlay: null,
```
donato: These onXXX aren't supported anymore, are they?

donato:
In our other files we are using ```ok, equal, expect```, etc..., but here you are using ```assert.ok, assert.equal,...```, can we stick with the same naming?

test/unit/api.js
```diff
@@ -85,41 +88,122 @@ define([
...
+
+        // FIXME: These are not ready until after setup (controller.setup())
+        //api.qoe();
+        //api.createInstream();
+        //assert.strictEqual( api.getState(), undefined, 'getState returns undefined before setup');
+
+    });
+
+    test('sets up a player with setup(config)', function(assert) {
+        var done = assert.async();
```
donato: This is basically an integration test, it does way too much. A unit test should hit a single module, so that when something is broken we know exactly where too look. We already have broad coverage for this type of thing via the automated cucumber tests, so I don't think this offers anything new.

donato: We talked about this the goals are to
1. Get quicker feedback for problems than cucumber provides
2. Be able to visualize code paths from the API using karmas code coverage tools

test/data/api-methods.js
robwalch: maintained for backwards compatibility

test/unit/api.js
robwalch: These all hit code paths that cannot be reached before setup. Each step is clear enough to point to a failure.

Both goals are met. Run karma:local task and then look at the reports/coverage. Tests can also be run by opening /test in the browser.

PR #645, opened 2015-08-03
Fixes IE11 is centering issue [Delivers #96808212]

jw-kaurand:
Fixes an issue with how IE11 is centering issue.  Since we don't have any max-width logic relating to skins we're not gaining any advantage from using "margin: 0 auto;" instead of 2% to match our width.

[Delivers #96808212]

### DIFF
--- a/src/css/imports/mixins.less
+++ b/src/css/imports/mixins.less
@@ -25,7 +25,7 @@
 .inset-controlbar() {
     .jw-controlbar {
         display: block;
-        margin: 0 auto;
+        margin: 0 2%;
         max-width: 96%;
         bottom: .7em;
     }


src/css/imports/mixins.less
```diff
@@ -25,7 +25,7 @@
 .inset-controlbar() {
     .jw-controlbar {
         display: block;
-        margin: 0 auto;
+        margin: 0 2%;
         max-width: 96%;
```
robwalch: Doesn't this change mean that `max-width: 96%` can be deleted?

jw-kaurand:
This has too many interactions with #99798716 to act as a separate PR.  #96808212 will have to be solved in relation to the PR related to #99798716.  That story is currently in development.

PR #647, opened 2015-08-03
Centering and max-width for skins without full width [Delivers #99798716 #96808212]

jw-kaurand:
Assigns some styles for centering in IE11 as well as maxwidth in other browsers.  Some additional styles are added for IE based browsers to format correctly.  castAvailable is erroneously assigning values for IE browsers where the chromecast is not supported so a default value was added.

[Delivers #99798716 #96808212]

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -15,6 +15,7 @@ define([
         displaydescription: true,
         mobilecontrols: false,
         repeat: false,
+        castAvailable: false,
         skin: 'seven',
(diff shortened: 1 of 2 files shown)


src/js/api/config.js
```diff
@@ -15,6 +15,7 @@ define([
         displaydescription: true,
         mobilecontrols: false,
         repeat: false,
+        castAvailable: false,
```
robwalch: Can we do this in commercial api-config only?

PR #648, opened 2015-08-04
jw-flag-cast-available needs appid to apply the class [Delivers #100152244]

jw-kaurand:
Requires an cast appid to be set in the config for the cast button to display when chrome casts are available.

[Delivers #100152244]


robwalch:
This should be done in commercial cast controller that sets castAvailable to true. The view should not be concerned with this logic.

PR #649, opened 2015-08-04
Allow non-secure http pages to communicate with secure https hosted swf

robwalch:
Same update to vast.swf and googima.swf in their respective projects. While the behavior manifested itself a new bug, the same kind of setup in 6.12 would break all api calls that attempt to call internal flash methods.

This is a new feature that should stop support requests from users with improper setups that embed https swfs on http pages.

[Delivers #100055722]


robwalch:
This would open up new security holes during Flash playback. Closing until there's a concrete need and proof that the openings are not a severe concern.

PR #661, opened 2015-08-12
JW7-818 #Create PR Pointer Events

jw-kaurand:
PointerEvents implementation which handles PointerEvents from MS devices with correct pointer interaction depending on whether the user is using a tap or mouse interface.  Also makes the mouseover/pointerover interaction consistent by rolling up hover listeners into the UI Class

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -5,6 +5,7 @@ define([
     'utils/helpers'
 ], function(Events, events, _, utils) {
     var TouchEvent = window.TouchEvent || {};
+    var PointerEvent = window.PointerEvent || {};
 
     function getCoord(e, c) {
         return /touch/.test(e.type) ? (e.originalEvent || e).changedTouches[0]['page' + c] : e['page' + c];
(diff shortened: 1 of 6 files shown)


src/js/utils/ui.js
```diff
@@ -5,6 +5,7 @@ define([
     'utils/helpers'
 ], function(Events, events, _, utils) {
     var TouchEvent = window.TouchEvent || {};
+    var PointerEvent = window.PointerEvent || {};
```
donato: This isn't needed

src/js/utils/ui.js
```diff
@@ -65,38 +66,81 @@ define([
...
             _hasMoved = false,
             _startX = 0,
             _startY = 0,
             _lastClickTime = 0,
             _doubleClickDelay = 300,
             _touchListenerTarget,
-            _isDesktop = !utils.isMobile();
-
+            _isDesktop = !utils.isMobile(),
+            _hasPointerEvents = !_.isUndefined(window.PointerEvent),
```
donato: This can be done once, at a higher scope, instead of for each instance

src/js/utils/ui.js
```diff
@@ -65,38 +66,81 @@ define([
...
+            if( evt instanceof MouseEvent) {
+                if(!_hasPointerEvents || (_hasPointerEvents && evt.pointerType !== 'touch')) {
+                    triggerEvent(events.touchEvents.OVER, evt);
+                }
+            }
+        }
+
+        function outHandler(evt){
+            if( evt instanceof MouseEvent) {
+                if(!_hasPointerEvents || (_hasPointerEvents && evt.pointerType !== 'touch')) {
```
donato: (!x || (x && y)) === (!x || y)

donato:
Can we remove "_isDesktop" and use something feature-detection-y instead?

src/js/view/components/volumetooltip.js
```diff
@@ -19,9 +19,10 @@ define([
 
             utils.toggleClass(this.el, 'jw-hidden', false);
 
-            new UI(this.el).on('click', this.toggleValue.bind(this)).on('tap', this.toggleOpenState.bind(this));
-            this.el.addEventListener('mouseover', this.openTooltip.bind(this));
-            this.el.addEventListener('mouseout', this.closeTooltip.bind(this));
+            new UI(this.el, {'useHover': true}).on('click', this.toggleValue.bind(this))
```
donato: put "UI().on(...)" on a new line, to match the following 3 "on" methods

PR #663, opened 2015-08-12
Remove dependence on eventdispatcher [Delivers JW-1254]

pajong:
Removed eventdispatcher and updated providers and model accordingly


src/js/providers/html5.js
```diff
@@ -59,8 +59,19 @@ define([
...
+                callback(input);
+            });
+        };
+
+        _.extend(this, Events);
+
+        // legacy support
+        this.addEventListener = this.on;
+        this.sendEvent = this.trigger;
+        this.resetEventListeners = this.removeEventListener = this.off;
```
donato: Let these be gone!

donato:
Good start.

Now from here, delete all the "legacy" support code. Remove "addGlobalEventListener" from everywhere. The places that used to use "global event" are where you will move the "on('all',...)" code.

src/js/controller/instream-html5.js
```diff
@@ -103,7 +103,6 @@ define([
 
                 provider.resetEventListeners();
 
-                provider.addGlobalListener(_forward);
```
donato: I would expect this to be replaced by something else, not just removed. Will this work?

src/js/controller/instream-html5.js
```diff
@@ -101,12 +101,11 @@ define([
                     return;
                 }
 
-                provider.resetEventListeners();
+                provider.off();
 
-                provider.addGlobalListener(_forward);
```
donato: I would expect to see ```provider.on('all', ...)``` here, is it not needed?

src/js/providers/flash.js
```diff
@@ -348,18 +344,10 @@ define([
...
                     _item = null;
-                    _eventDispatcher.resetEventListeners();
-                    _eventDispatcher = null;
+                    this.off();
                 }
         });
 
-        // Overwrite the event dispatchers to block on certain occasions
-        this.sendEvent = function() {
-            if (!_attached) {
```
robwalch: The functionality of this `if (!_attached) {` statement is important, and it looks like it's been removed.

src/js/providers/html5.js
```diff
@@ -129,15 +128,6 @@ define([
 
             _fullscreenState = false;
 
-        // Overwrite the event dispatchers to block on certain occasions
-        this.sendEvent = function() {
-            if (!_attached) {
```
robwalch: Again, this functionality has been removed. Events should not be triggered if _attached is false.

PR #665, opened 2015-08-13
Filter by first source instead of provider priority

donato:
JW7-822


donato:
JW7-822 #comment Test comment

PR #670, opened 2015-08-13
JW7-1254 Remove eventdispatcher and clean up other files accordingly

pajong:
Removed eventdispatcher, removed some unused functions, and cleaned up some functions


src/js/controller/instream-html5.js
```diff
@@ -59,15 +63,18 @@ define([
...
                 _currentProvider.detachMedia();
-                _currentProvider.resetEventListeners();
+                _currentProvider.off();
                 _currentProvider.destroy();
             }
 
             // Return the view to its normal state
             _adModel = null;
+
+            _controller.off(null, null, this);
```
donato: do these arguments do anything?

donato: From the backbone.events documentation I see
```js
// Removes all callbacks for `context` for all events.
object.off(null, null, context);
```

This could benefit from a comment for other poor devs like myself :)

pajong:
@robwalch This is the right branch, can you please have a look? I added attached flag back to html5.js and flash.js

src/js/providers/flash.js
```diff
@@ -348,18 +344,18 @@ define([
...
                     _item = null;
-                    _eventDispatcher.resetEventListeners();
-                    _eventDispatcher = null;
+                    this.off();
                 }
         });
 
         // Overwrite the event dispatchers to block on certain occasions
-        this.sendEvent = function() {
+        this.sendEvent = function(type, data) {
```
robwalch: Let's not use `sendEvent` anymore since it can be confusing if both `sendEvent` and `trigger` can be used on an object. You can do:
```
this.trigger = function(type, args) {
    if (!_attached) {
        return;
    }
    return Events.trigger.call(this, type, args);
};
```

src/js/providers/html5.js
```diff
@@ -59,8 +59,15 @@ define([
...
-        _.extend(this, _dispatcher);
+        _.extend(this, Events);
+
+        // Overwrite the event dispatchers to block on certain occasions
+        this.sendEvent = function(type, data) {
+            if (!_attached) {
+                return;
+            }
+            this.trigger(type, data);
+        };
```
robwalch: Let's not use `sendEvent` anymore since it can be confusing if both `sendEvent` and `trigger` can be used on an object. You can do:
```
this.trigger = function(type, args) {
    if (!_attached) {
        return;
    }
    return Events.trigger.call(this, type, args);
};
```

PR #672, opened 2015-08-13
Hide the preview image when the player is in audio only mode. [JW7-1285]

emaxsaun:
Hide the preview image when the player is in audio only mode. [JW7-1285]

### DIFF
--- a/src/css/flags/audioplayer.less
+++ b/src/css/flags/audioplayer.less
@@ -43,3 +43,9 @@
 .jwplayer.jw-flag-audio-player .jw-preview {
     display: none;
 }
+
+.jwplayer.jw-flag-media-audio.jw-flag-audio-player {
+    .jw-preview {
+        display: none; 
+    }
+}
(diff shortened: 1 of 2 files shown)


emaxsaun:
@robwalch We went over this earlier.

src/css/flags/audioplayer.less
```diff
@@ -43,3 +43,9 @@
 .jwplayer.jw-flag-audio-player .jw-preview {
     display: none;
 }
+
+.jwplayer.jw-flag-media-audio.jw-flag-audio-player {
```
donato: We only care if the player is on audio-mode (ie , the audio-player flag), thus the media-audio flag should not be here.

emaxsaun: That doesn't work in both cases. If we look above the stuff I added, we have this:

 .jwplayer.jw-flag-audio-player .jw-preview {
     display: none;
 }

But this doesn't work for when the player is at a small size.

emaxsaun: Yeah, I just tested locally with .jwplayer.jw-flag-audio-player by itself and it didn't work, but with .jwplayer.jw-flag-media-audio.jw-flag-audio-player { it works fine.

PR #674, opened 2015-08-17
JW7-818 Pointer Events Support and UI.js Refactoring

jw-kaurand:
JW7-818 #Create PR PointerEvents implementation which handles PointerEvents from MS devices with correct pointer interaction depending on whether the user is using a tap or mouse interface.  Also makes the mouseover/pointerover interaction consistent by rolling up hover listeners into the UI Class.  Also, Simplifying and formatting changes.  document.elementFromPoint to fix an issue with setPointerCapture for PointerEvents

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -95,6 +95,7 @@ define([
             }
         }
 
+        // Always add this, in case we don't properly identify the device as mobile
         elem.addEventListener('touchstart', interactStartHandler);
 
         // overHandler and outHandler not assigned in touch situations
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -4,7 +4,9 @@ define([
     'utils/underscore',
     'utils/helpers'
 ], function(Events, events, _, utils) {
-    var TouchEvent = window.TouchEvent || {};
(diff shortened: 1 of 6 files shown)


src/js/utils/ui.js
```diff
@@ -4,7 +4,9 @@ define([
     'utils/underscore',
     'utils/helpers'
 ], function(Events, events, _, utils) {
-    var TouchEvent = window.TouchEvent || {};
+    var _usePointerEvents = !_.isUndefined(window.PointerEvent);
```
donato: Love this.

src/js/utils/ui.js
```diff
@@ -63,42 +69,75 @@ define([
...
+            }
+        } else if(_useMouseEvents){
             elem.addEventListener('mousedown', interactStartHandler);
+            if(options.useHover) {
+                elem.addEventListener('mouseover', overHandler);
+                elem.addEventListener('mouseout', outHandler);
+            }
         }
+
         elem.addEventListener('touchstart', interactStartHandler);
```
donato: should this ben in an 

```js
else if (_useTouchEvents) {
```
?

PR #675, opened 2015-08-17
JW7-832 JW7-1278 Mobile UI Icon Size Increase

jw-kaurand:
Increases the size of the ui on mobile devices an appropriate amount by adding a flag to the player.  Some extra attention given to the large-ui version of the seven skin so it doesn't look dumb with its extra padding in the small mode.  Also includes a test page with a viewport set to device-width for testing the Mobile UI in a setting more appropriate for the form factor.

### DIFF
--- a/src/css/flags/ads.less
+++ b/src/css/flags/ads.less
@@ -41,7 +41,7 @@
     }
 }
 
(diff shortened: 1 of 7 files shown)


donato:
Could we re-use ```jw-flag-touch-screen``` instead of adding a new flag?

PR #676, opened 2015-08-18
JW-1262 Optimize CSS Utils

pajong:
Optimized css utils by removing block/unblock, and unused functions

### DIFF
--- a/src/js/utils/css.js
+++ b/src/js/utils/css.js
@@ -1,14 +1,12 @@
 define([
     'utils/helpers',
-    'utils/strings',
-    'utils/underscore'
-], function(utils, Strings, _) {
+    'utils/strings'
+], function(utils, Strings) {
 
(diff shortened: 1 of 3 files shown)


src/js/utils/css.js
```diff
@@ -72,49 +53,22 @@ define([
...
-    var _block = function (id) {
-        // use id so that the first blocker gets to unblock
-        if (_cssBlock === null) {
-            // console.time('block_'+id);
-            _cssBlock = {
-                id: id,
-                styleSheets: {},
-                elements: []
-            };
+        for (style in styles) {
```
donato: Last I checked, for..in loops are considered a little slow, and prone to mis-use. People tend to prefer normal for loops https://jsperf.com/for-loop-vs-for-in-loop/4

donato:
Love this change!

PR #678, opened 2015-08-24
JW7-1254 Add Events to player instance

pajong:
Added Events to player instance for eventdispatcher remove merge


pajong:
@donato Can you please take a look? Thank you

PR #680, opened 2015-08-25
JW7-815 - Control Compact Mode and Menu Drawer

jw-kaurand:
This manages setting the player into compact mode, which will limit the number of icons which will appear in the controlbar.  In compact mode, we will place extra menu icons into a drawer, which will allow the user to toggle the visibility of the icons when necessary.  Individual commit messages describe each step.

Delivers subtasks JW7-1279 JW7-1281 JW7-1282 JW7-1280 JW7-1283

### DIFF
--- a/src/css/imports/tooltip.less
+++ b/src/css/imports/tooltip.less
@@ -1,5 +1,5 @@
 @import "vars";
-@import "../imports/icons";
+@import "icons";
 
 .jw-icon-tooltip.jw-open .jw-overlay {
     opacity: 1;
--- a/src/css/imports/tooltip.less
+++ b/src/css/imports/tooltip.less
@@ -1,4 +1,5 @@
 @import "vars";
+@import "../imports/icons";
 
 .jw-icon-tooltip.jw-open .jw-overlay {
     opacity: 1;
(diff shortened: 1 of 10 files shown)


donato:
Cool stuff!

src/css/imports/tooltip.less
```diff
@@ -1,4 +1,5 @@
 @import "vars";
+@import "../imports/icons";
```
robwalch: same folder. `@import "icons"` doesn't work?

src/js/view/view.js
```diff
@@ -805,6 +814,8 @@ define([
             _showing = true;
             utils.removeClass(_playerElement, 'jw-flag-user-inactive');
 
+            _controlbar.checkCompactMode(_videoLayer.clientWidth);
+
```
robwalch: Now that we're sampling clientWidth in this method, we may want to debounce the 'mousemove' event handler that triggers this method. Or, find a better place to run this check and/or sample clientWidth.

robwalch: Turn on "Show paint rectangles" in chrome or do some display profiling to see if mouse movement thrashes performance. That's my concern with the previous comment.

jw-kaurand: I'm not seeing any problems with it thrashing performance or causing a lot of repaints.  I'll check to see if it has issues in other browsers as well.  It's being called here to make sure that it's called at least once when the UI goes from being not displayed to being visible.  It could be run just once after if we call that only when `_showing` was false when we run `_userActivity`.

PR #682, opened 2015-08-26
JW7-1328 Fix bug with youtube provider not working

pajong:
Fixed youtube provider having issues with on not being defined


pajong:
@donato can you please look at this? it's a quick fix

PR #683, opened 2015-08-26
JW7-815 JW7-1282 JW7-1283 Bugfixes for display consistency

jw-kaurand:
Touch only has the "X" close button for the drawer.
clearCompactMode called when something which updates the content of the control bar is called so that its always up-to-date
_userActivity only removes classes and checks for compact mode if we were previously not showing the UI.
Pre-calculate the breakpoint for quicker execution and greater consistency for switching between display modes.

### DIFF
--- a/src/js/view/controlbar.js
+++ b/src/js/view/controlbar.js
@@ -66,9 +66,7 @@ define([
         this._model = _model;
         this._isMobile = utils.isMobile();
         this._compactModeMaxSize = 400;
-
-        this._leftGroupExpandedSize = false;
-        this._rightGroupExpandedSize = false;
+        this._maxCompactWidth = -1;
 
         this.setup();
     }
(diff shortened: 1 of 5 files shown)


src/js/view/controlbar.js
```diff
@@ -388,27 +388,38 @@ define([
                 this.elements.time.drawCues();
             }
         },
+        clearCompactMode : function() {
+            this._maxCompactWidth = -1;
+            this._model.set('compactUI', false);
+        },
+        // Sets this._maxCompactWidth and this._minCompactWidth so we calculate less per call of isCompactMode
```
donato: minCompactWidth was removed?

src/js/view/view.js
```diff
@@ -811,10 +811,12 @@ define([
         }
 
         function _userActivity() {
-            _showing = true;
-            utils.removeClass(_playerElement, 'jw-flag-user-inactive');
+            if(!_showing){
```
donato: Can we avoid this bifurcation? If so I'll wait to merge, if not let me know.

jw-kaurand: It's more efficient to use this with the _showing flag since we will only run that once when content becomes visible.   While we don't take a big performance hit on Chrome from `_controlbar.checkCompactMode(_videoLayer.clientWidth);` being run multiple times, it could be less efficient on other browsers.

PR #684, opened 2015-08-26
JW7-883 Allow apostrophes in image URL

pajong:
Changed serialize function in parser to change apostrophes to %27

### DIFF
--- a/src/js/utils/parser.js
+++ b/src/js/utils/parser.js
@@ -100,7 +100,7 @@ define([
                     return Number(val);
                 }
             }
-            val = val.replace("'", "%27");
+            val = val.replace('\'', '%27');
         }
         return val;
     };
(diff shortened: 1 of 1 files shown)


src/js/utils/parser.js
```diff
@@ -81,22 +81,26 @@ define([
...
+                    return true;
+                }
+                if (lowercaseVal === 'false') {
+                    return false;
+                }
+                if (!isNaN(Number(val)) && !isNaN(parseFloat(val))) {
+                    return Number(val);
+                }
             }
+            val = val.replace('\'', '%27');
```
robwalch: We shouldn't change this function. Not all values passed to config should be transformed this way. The issue is when we [construct an image url in css.js](https://github.com/jwplayer/jwplayer/blob/master/src/js/utils/css.js#L112). At that point we should `escape()` the value to handle apostrophes and quotes and wrap it in quotes to protect agains parenthesis in the value.

robwalch: Here's a little more info syntax http://stackoverflow.com/questions/2168855/is-quoting-the-value-of-url-really-necessary

PR #686, opened 2015-08-27
JW7-1340 Remove event listener we don't need that was breaking Win7 IE 11

jw-kaurand:
JW7-1340 Remove event listener we don't need that was breaking Win7 IE 11.  The listener is pretty much just listening to when releasePointerCapture is being called.


dannyfinks:
@jw-kaurand @robwalch should this be merged to master or release?

robwalch:
@dannyfinks master. hotfixes that need to go into upcoming releases should be tagged in JIRA (fix in version + comment at me there.) You can also add version milestones in here to communicate this publicly.

PR #688, opened 2015-08-28
JW7-834 Fix iOS youtube start button

pajong:


pajong:
I went back to builds long time ago, and confirmed that our youtube start button never worked. To fix this, I made a separate display icon container for mobile, and copied the same css as the original display icon container, then allowed click through to the mobile display icon container, so that it will handle the click/tap event the same way as a poster click.

PR #690, opened 2015-08-28
Improve Flash plugin loader error message and simplify queue handling

robwalch:
- Add url to AssetLoader to make it easy to access the url in loader callbacks
- Simplify plugin loading by keeping an count of active loaders rather than dictionary shenanigans
- Include plugin url in plugin loader error, and pass original Flash error message to JS JWPLAYER_MEDIA_ERROR event

These changes to help developers understand and deal with crossdomain setups and Flash errors in general. Any error in the Flash player currently gets masked as 'Error loading media: File could not be played'. That is not helpful at all, when flash hosted on one domain can't access vast or googima.swf on another (caused by some of out custom test setups).

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/controller/PluginLoader.as
+++ b/src/flash/com/longtailvideo/jwplayer/controller/PluginLoader.as
@@ -27,17 +27,19 @@ import flash.utils.Dictionary;
  */
 public class PluginLoader extends EventDispatcher {
 
+    public var plugins:Object;
+
+    private var _remainingLoaders:uint;
+    private var _errorState:Boolean;
+
     public function PluginLoader() {
-        loaders = new Dictionary();
+        _remainingLoaders = 0;
         plugins = {};
     }
-    public var plugins:Object;
-    private var loaders:Dictionary;
-    private var errorState:Boolean = false;
 
-    public function loadPlugins(plugins:Array):void {
-        if (plugins.length) {
-            for each(var plugin:Object in plugins) {
+    public function loadPlugins(pluginsToLoad:Array):void {
+        if (pluginsToLoad.length) {
+            for each(var plugin:Object in pluginsToLoad) {
                 if (plugin) {
                     loadLocalPlugin(plugin.swf);
                 }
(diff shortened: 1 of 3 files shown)


src/flash/com/longtailvideo/jwplayer/controller/PluginLoader.as
```diff
@@ -27,17 +27,19 @@ import flash.utils.Dictionary;
...
 public class PluginLoader extends EventDispatcher {
 
+    public var plugins:Object;
+
+    private var _remainingLoaders:uint;
+    private var _errorState:Boolean;
+
     public function PluginLoader() {
-        loaders = new Dictionary();
+        _remainingLoaders = 0;
```
donato: Can this ``` = 0 ``` be in the class definition instead of the constructor?

src/flash/com/longtailvideo/jwplayer/controller/PluginLoader.as
```diff
@@ -52,38 +54,35 @@ public class PluginLoader extends EventDispatcher {
...
         var loader:AssetLoader = evt.target as AssetLoader;
-        var url:String = loaders[loader] as String;
+        loader.removeEventListener(Event.COMPLETE, loadSuccess);
+        loader.removeEventListener(ErrorEvent.ERROR, pluginLoadFailed);
+        var url:String = loader.url;
         var pluginId:String = url.substr(url.lastIndexOf("/") + 1).replace(/(.*)\.swf$/i, "$1").split("-")[0];
         plugins[pluginId] = loader.loadedObject as DisplayObject;
-        loader.removeEventListener(Event.COMPLETE, loadSuccess);
-        delete loaders[loader];
+        _remainingLoaders--;
```
donato: Would we need to also decrement this counter if a load failed? Otherwise checkComplete will never be true.

src/flash/com/longtailvideo/jwplayer/controller/PluginLoader.as
robwalch: Yes. Is that better?

src/flash/com/longtailvideo/jwplayer/controller/PluginLoader.as
robwalch: That's how it worked before, and the reason for `_errorState`. It does not complete if an error occurred... but maybe it should?

src/flash/com/longtailvideo/jwplayer/controller/PluginLoader.as
donato: It's technically the same, but imho it's tidier. One fewer line and the definition includes the initial state.

PR #691, opened 2015-08-31
JW7-1082 JW7-1002 JW7-1339  Menu logic improvements and bugfixes

jw-kaurand:
JW7-1082 Menus should close after selecting an option from them
JW7-1002 Fixes large menus having unselectable bottom items
JW7-1339 Active Toggle buttons should also be collapsed into the drawers

Other fixes - 
Only one menu can be open at a time
Do not collapse into drawer if it would hide only 1 item
The timeout for determining the user to inactive refreshes each time the user interacts with part of the controller
Menus and drawer close when the ui hides from inactivity or is toggled closed
Added 'directSelect' parameter to UI for interactions where the event should only be fired if the target is the listened element and not its children
Some comment fixes
Fixes issues with hovering the drawer on PointerEvents devices and issues with closing menus

### DIFF
--- a/src/js/view/components/drawer.js
+++ b/src/js/view/components/drawer.js
@@ -13,7 +13,7 @@ define([
         },
         setup : function (list, isCompactMode) {
             if(!this.iconUI){
-                this.iconUI = new UI(this.el);
+                this.iconUI = new UI(this.el, { 'directSelect': true });
 
                 this.toggleOpenStateListener = this.toggleOpenState.bind(this);
                 this.openTooltipListener = this.openTooltip.bind(this);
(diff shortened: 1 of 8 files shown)


src/js/view/components/drawer.js
```diff
@@ -24,31 +24,31 @@ define([
...
+            if (isCompactMode && this.activeContents.length > 1) {
                 utils.removeClass(this.el, 'jw-off');
 
-                this.iconUI.on('tap', function (evt) {
-                    if (evt.target === this.el) {
+                this.iconUI
+                    .on('tap', function () {
                         this.isOpen = !this.isOpen;
                         utils.toggleClass(this.el, this.openClass, this.isOpen);
                         this.trigger('drawer-open', {'isOpen': this.isOpen});
```
donato: Maybe we should have drawer-close, instead of drawer-open with a false param

src/js/view/controlbar.js
```diff
@@ -430,7 +457,8 @@ define([
 
             this.elements.drawer.setup(this.layout.drawer, isCompact);
 
-            if(!isCompact){
+            // If we're not in compact mode or we're not hiding icons, then put them back where they came from.
+            if(!isCompact || this.elements.drawer.activeContents < 2){
```
donato: activeContents.length

src/js/view/controlbar.js
```diff
@@ -388,11 +403,23 @@ define([
...
+            _.each(this.menus, function(ele){
+                if(!evt || evt.target !== ele.el) {
+                    ele.closeTooltip(evt);
+                }
+            });
+        },
+        hideComponents : function(){
+            this.closeMenus();
+            this.elements.drawer.closeTooltip();
+            utils.toggleClass(this.el, 'jw-drawer-expanded', false);
```
donato: utils.removeClass ?

PR #696, opened 2015-09-02
Optimize CSS routines for JW7-831

robwalch:
Using a template for the color config options simplifies this routine in the view and allows us to finally get rid of old css utils making the player smaller.

I also noticed a bunch of places where we were using hyphen-delimited syntax to set inline styles in JavaScript. Our style utility has to replace all of these with camel case on the fly because setting properties on a style object must be done in "camelCase". Instead of making it do extra work and mixing up syntaxes just always use camel case for style json.

### DIFF
--- a/src/js/utils/css.js
+++ b/src/js/utils/css.js
@@ -3,46 +3,8 @@ define([
     'utils/strings'
 ], function(utils, Strings) {
 
-    var MAX_CSS_RULES = 50000,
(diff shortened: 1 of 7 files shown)


robwalch:
@pajong @jw-kaurand

robwalch:
Adding another template adds a significant number of kB to the player. We should reevaluate how we do templates and css loading with webpack.

PR #704, opened 2015-09-04
Fix control not fading on autostart

pajong:
Added call to userActivity when state changes to playing

JW7-978


donato:
If we go from buffering to playing, it will now show the controlbar.

donato:
Looks good!

PR #705, opened 2015-09-04
JW7-1350 Hide logo during ad playback

pajong:


src/css/flags/ads.less
```diff
@@ -33,6 +33,10 @@
         }
 
     }
+
+    .jw-logo {
+        display: none !important;
```
donato: We try not to use !important, can you find an alternative way?

robwalch:
@pajong I reversed this commit. Don't use `!important`. What if someone wants to customize their player to show the logo? This makes it very difficult. You want to make the rule more specific than it what is overriding it, but not by overriding everything else with `!important` or inline styles.

http://www.standardista.com/css3/css-specificity/

PR #707, opened 2015-09-04
JW7-1350 Hide logo during ad playback

pajong:
Hided logo without using !important


pajong:
@robwalch I did not need !important, but I put it in the previous commit just to be sure - I thought we had to hide the logo whenever the ad playback happens

PR #712, opened 2015-09-08
JW7-1343 Add x-flv support to actionscript

pajong:


robwalch:
Isn't there another place in the code parses extensions from mime types? x- should be stripped as part of that.

I'm assuming that in this case we split a mime type by / and didn't sanitize the right part ex `video/x-flv`. This is not a case where the extension is in the file name correct? ex `file.x-flv`.

pajong:
@robwalch Yes, this is not the case where the extension is in the file name.
I will look into other places to see where the parsing and removing x- needs to happen

PR #719, opened 2015-09-10
Add preload feature to html5 provider JW7-838

pajong:
Tested for autostart/ non auto start on:
- Android
- Safari
- Firefox
- Chrome
- IE (preload tag is not supported, but it does not break anything)
- Edge

Works the same way as before if preload config is not there, and preloads if there is preload in config.

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -436,6 +436,7 @@ define([
                 _source = _levels[_currentQuality];
                 _videotag.src = _source.file;
                 _videotag.preload = preloadValue;
+                _bufferInterval = setInterval(_checkBufferAndPlayback, BUFFER_INTERVAL);
             } else {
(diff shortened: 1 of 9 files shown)


src/js/providers/html5.js
```diff
@@ -428,6 +428,15 @@ define([
             _setLevels(item.sources);
             this.sendMediaType(item.sources);
 
+            // canSeek to true to prevent canPlayHandler from starting the video
+            _canSeek = true;
```
donato: Good comment, it helped me understand your goal.

I think we can solve the problem a different way, since _canSeek is technically false (because we aren't ready to seek yet). You identified that the issue is in _canPlayHandler. Maybe if we don't call _canPlayHandler inside of _onLoadedMetaData that would suffice for preventing the video starting.

donato:
Looks good to me!

@robwalch @jw-kaurand want to chime in?

PR #723, opened 2015-09-12
Model expects attributes to be set with get/set methods

robwalch:
- `getVisualQuality` returns undefined (JW7-1438)
- `controller.id` is undefined and no longer used
- `controller.showView` had some logic that passed undefined to `document.getElementById`
- `instream-adapter` had invalid `model.state` checks, preventing double-click to fullscreen from working. It should have implications with ad-pod functionality too, but seems to work as-is and after these fixes.


donato:
Nice cleanup

PR #727, opened 2015-09-15
Change html5 provider state to error on error

pajong:


src/js/providers/html5.js
```diff
@@ -443,8 +446,11 @@ define([
                 return;
             }
 
-            _setLevels(item.sources);
-            this.sendMediaType(item.sources);
+            // check if _setLevels has already been called
```
donato: why?

pajong: https://jwplayer.atlassian.net/browse/JW7-1460
Due to preloading, this is being called twice

donato: I am hesitating on this one, because sometimes we unknowingly rely on the player to re-set values - this may be here for a reason. For example, a midroll may sendMediaType of "audio" which adds a flag in the view. When returning to the original provider, it may not re-send the media type of "video" because the sources are equal. (This is just a possible scenario but it may be worth testing)

I see two options
1. The easy way - we fix the story exactly as stated by suppressing the event from the public API, and leaving the provider as it was
2. We examine all use-cases of init and load from within the provider and from the model and ad plugins. Then we make a change - probably very similar to this one, but after exhaustive research.

pajong: What do you mean by suppressing the event from the public API?

donato: Find the place where the event is exiting into the public API - ie through
the on('levels') and filter it out there

pajong: The on('levels') count increases in videoEventHandler in model (this.mediaController.trigger(type, evt);), so I am not sure how we can filter this.
Would it cause problems if we don't send levels event on init? It will be called in load anyways when the user tries to play the video.

PR #734, opened 2015-09-18
JW7-1312 Send event so that the plugin can know how it is activated.

jw-kaurand:
Send event so that the plugin can know how it is activated.  Related to sharing test updates in Commercial (https://github.com/jwplayer/jwplayer-commercial/pull/853) and Plugins (https://github.com/jwplayer/jwplayer-plugins-commercial/pull/17)


donato:
Good call

PR #735, opened 2015-09-18
Ignore displaytitle config setting when displaying errors

robwalch:
Error messages will now display even when displaytitle is set to false (that setting is only intended to disable playlist item titles). Now we'll also show the error name and message. Message is the error message (not `error.toString()` which usually reads "Error: Error: ...", and a new property `error` gives access to the original exception error (to look at the stack...).

JW7-1488

### DIFF
--- a/src/js/utils/trycatch.js
+++ b/src/js/utils/trycatch.js
@@ -20,9 +20,10 @@ define([
         }
     };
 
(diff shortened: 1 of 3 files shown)


donato:
I made a PR against this PR @robwalch , please review : https://github.com/jwplayer/jwplayer/pull/739

PR #736, opened 2015-09-18
Handle Flash setup error

robwalch:
Handle the exception that occurs when calling setup on a bad flash embed.

``` javascript
    jwplayer("container").setup({
        file: 'file.mp4',
        primary: 'flash',
        flashplayer: 'https://xx/jwplayer.flash.swf'  // loading from a domain that restricts script access
    });
```

The swf embeds but cannot be setup. Rather than buffer indefinitely, forward the error for display.

JW7-1489


donato:
test this please

PR #738, opened 2015-09-21
Add preload functionality to flash provider

pajong:
@robwalch @donato Can you please take a look at this?

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/controller/Controller.as
+++ b/src/flash/com/longtailvideo/jwplayer/controller/Controller.as
@@ -228,8 +228,8 @@ public class Controller extends GlobalEventDispatcher {
     }
 
     public function init(item:PlaylistItem):void {
-        var _loaded:Boolean = load(item);
-        if (_loaded) {
+        var loadSuccessful:Boolean = load(item);
+        if (loadSuccessful) {
             _model.media.init(item);
         }
     }
--- a/src/flash/com/longtailvideo/jwplayer/controller/Controller.as
+++ b/src/flash/com/longtailvideo/jwplayer/controller/Controller.as
@@ -227,6 +227,13 @@ public class Controller extends GlobalEventDispatcher {
         return false;
     }
 
+    public function init(item:PlaylistItem):void {
+        var _loaded:Boolean = load(item);
+        if (_loaded) {
+            _model.media.init(item);
+        }
+    }
+
     public function load(item:PlaylistItem):Boolean {
         if (_model.item && _model.item.file === item.file) {
             // resume current item
--- a/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
+++ b/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
@@ -50,6 +50,10 @@ public class VideoMediaProvider extends MediaProvider {
     /** Is buffering due to load/seek or underflow? **/
     private var seeking:Boolean;
 
+    private var _preload:Boolean;
+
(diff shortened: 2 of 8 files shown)


src/flash/com/longtailvideo/jwplayer/controller/Controller.as
```diff
@@ -227,6 +227,13 @@ public class Controller extends GlobalEventDispatcher {
         return false;
     }
 
+    public function init(item:PlaylistItem):void {
+        var _loaded:Boolean = load(item);
```
donato: Using underscore as a prefix for variable names is generally used to imply scope - I think the convention used most often in this codebase is to use _ for instance variables which are treated as "private."

So perhaps an alternative would be just ```loaded``` or even better, ```loadSuccessful```

src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
```diff
@@ -50,6 +50,10 @@ public class VideoMediaProvider extends MediaProvider {
     /** Is buffering due to load/seek or underflow? **/
     private var seeking:Boolean;
 
+    private var _preload:Boolean;
```
donato: Can we solve this without relying on flag variables?

pajong: I tried to find a way to remove flags, but some functions in here (e.g. loadStream) need to know whether if we are preloading while they do not have any information about the item. Do you have any suggestions?

src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
```diff
@@ -88,45 +92,19 @@ public class VideoMediaProvider extends MediaProvider {
...
-            for (i = 0; i < _item.levels.length; i++) {
-                if (_item.levels[i].label == config.qualitylabel) {
-                    _currentQuality = i;
-                    break;
-                }
-            }
+    override public function init(itm:PlaylistItem):void {
+        setupVideo(itm);
+        // start loading the content only if preload is 'auto'
+        if (itm.preload === "auto") {
```
robwalch: Here `preload` can be any value that is not "none". We want to preload when item `preload` is set to "meta".

src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
```diff
@@ -313,13 +338,24 @@ public class VideoMediaProvider extends MediaProvider {
...
+        // set state to loading if this is not preload
+        if (!_preload) {
+            setState(PlayerState.LOADING);
+        }
         sendBufferEvent(0);
 
         // TODO: do this on enter frame like HLS
         clearInterval(_interval);
         this.seeking = true;
         _interval = setInterval(positionHandler, 100);
```
robwalch: `_interval = setInterval` should only be done after preload.

src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
robwalch: 1. Call `_stream.pause()` at the end of `init` after `loadQuality` (Then you don't have to call it at the end `loadStream`)
2. Call `setState(PlayerState.LOADING)` in `load` and `seek`, instead of `loadStream`
:boom: no more `_preload ` flag.

You can do the same with `_afterPreload`. Why do loadQuality and seek call loadStream? Why not just call play instead of calling loadStream if the item passes to init matches the item passed to load?

PR #740, opened 2015-09-21
Fix buffer not fired, levels fired twice, preload attribute undefined

pajong:


src/js/providers/html5.js
```diff
@@ -374,6 +375,13 @@ define([
                     _delayedSeek = -1;
                     _this.seek(startTime);
                 }
+
+                // if preloaded, set state to loading to set buffer event
+                if (_afterPreload) {
+                    _afterPreload = false;
+                    _this.setState(states.LOADING);
```
robwalch: Why not just always call this? What do we need `_afterPreload` for?

robwalch:
@dannyfinks this will also fix the preload="undefined" issue.

PR #742, opened 2015-09-22
JW7-1307 Adds class to button targeted to place icons in.  Fix for plugin issues

jw-kaurand:
Added a class that will be applied to dock buttons so that we can place icons inside them in a targeted manner.
Made ready handler fire when the player becomes ready so that plugins are added to the overlays correctly.
changed parameters sent into _resizePlugin call that were undefined

### DIFF
--- a/src/templates/dock.html
+++ b/src/templates/dock.html
@@ -1,7 +1,7 @@
 <div class="jw-dock jw-reset">
     {{#each this}}
-    <div class="jw-dock-button jw-background-color jw-reset" button="{{id}}">
-        <div class="jw-icon jw-dock-image jw-reset" style="background-image: url({{img}})"></div>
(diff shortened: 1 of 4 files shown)


src/templates/dock.html
```diff
@@ -1,7 +1,7 @@
 <div class="jw-dock jw-reset">
     {{#each this}}
-    <div class="jw-dock-button jw-background-color jw-reset" button="{{id}}">
-        <div class="jw-icon jw-dock-image jw-reset" style="background-image: url({{img}})"></div>
+    <div class="jw-dock-button jw-background-color {{#if class}}{{class}}{{/if}} jw-reset" button="{{id}}">
```
donato: Total nitpick here, but I usually prefer to put the extra space inside of the {{if}}

PR #744, opened 2015-09-22
Flash preload address code review feedback

pajong:
Only preload when "metadata" or "auto", change var name, remove _preload tag, remove unnecessary call to loadStream in seek

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
+++ b/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
@@ -50,6 +50,10 @@ public class VideoMediaProvider extends MediaProvider {
     /** Is buffering due to load/seek or underflow? **/
     private var seeking:Boolean;
 
+    private var _preload:Boolean;
+
+    private var _afterPreload:Boolean;
+
     /** Set the current quality level. **/
     override public function set currentQuality(quality:Number):void {
         if (!_item) return;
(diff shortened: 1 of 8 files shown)


src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
```diff
@@ -88,47 +88,29 @@ public class VideoMediaProvider extends MediaProvider {
...
-        // Set initial quality and set levels
-        _currentQuality = 0;
-
-        for (var i:Number = 0; i < _item.levels.length; i++) {
-            if (_item.levels[i]["default"]) {
-                _currentQuality = i;
-                break;
-            }
+    override public function init(itm:PlaylistItem):void {
+        if (itm.preload !== "metadata" && itm.preload !== "auto") {
```
donato: Check here

src/js/providers/flash.js
```diff
@@ -103,7 +103,25 @@ define([
...
+                    if (item.preload === 'none' || _playerConfig.autostart) {
+                        return;
+                    } else {
+                        _item = item;
+
+                        // if we just call flashCommand with _swf undefined, init will go into commandQueue
+                        if (_swf) {
+                            _flashCommand('init', item);
+                        } else {
+                            this.once('swfDefined', function() {
```
donato: try to reproduce this case, I'm not sure it's possible

src/flash/com/longtailvideo/jwplayer/player/Player.as
```diff
@@ -26,6 +26,8 @@ public class Player extends Sprite implements IPlayer {
 
     protected var _instream:InstreamPlayer;
 
+    private var _preloaded:Boolean;
```
donato: Test without using this, relying on the controllers file check

src/js/providers/flash.js
pajong: This actually happens, and init goes into "queueCommand", which is ran when we are trying to load.

PR #746, opened 2015-09-22
JW7-1486 JW7-1460 Fix levels eventfired twice and buffer event not fired

pajong:


src/js/providers/html5.js
```diff
@@ -368,6 +369,9 @@ define([
                 _setVideotagSource();
                 _videotag.load();
             } else {
+                if (_forceStateToLoad) {
```
pajong: @robwalch We need to check this because if we just call this, it will send 'buffer' event when we try to replay the same video. I talked to Danny, and decided that we do not want to change the behavior.

robwalch: Can you just check the current state instead?

if _this.state === IDLE

robwalch: We also talked about doing this in the model instead of each provider.

pajong: The state becomes IDLE when we try to replay the same video, which will call the "buffer" event.
I also tried to do it in the model, but it will have to be with when the JWPLAYER_MEDIA_BUFFER_FULL event happens, which will again call the "buffer" event when replayed.

src/js/providers/html5.js
```diff
@@ -368,6 +369,9 @@ define([
...
+                    _this.setState(states.LOADING);
```
donato: Let us try to put the setState(loading) in the two places where completeLoad is called. For quality level don't check for ismobile.

In the future, isMobile should be "is initiated by a touch event on mobile, or is desktop"

PR #748, opened 2015-09-22
JW7-1502 Do not preload when preload is none

pajong:


src/js/playlist/playlist.js
```diff
@@ -27,6 +27,9 @@ define([
             // include selected file in item for backwards compatibility
             item.file = item.sources[0].file;
 
+            // set preload for the item
+            item.preload = preload;
```
donato: item.preload = item.preload || preload

src/js/playlist/playlist.js
```diff
@@ -27,6 +27,9 @@ define([
             // include selected file in item for backwards compatibility
             item.file = item.sources[0].file;
 
+            // set preload for the item
+            item.preload = item.preload || preload;
```
donato: This works, can be even better though! 

             if (originalSource.drm || itemDrm) {
                 originalSource.drm = originalSource.drm || itemDrm;
             }

donato: This is an example from line 55, same file

Issue #701, opened 2015-09-03
jwplayer 7 - Chrome + Disabled Plugin completely disables ability to play using Flash


spodlecki:
Google Chrome _Version 44.0.2403.157 (64-bit)_
Mac OSX _10.10.3_

Reproduce by:
- Visit chrome://settings/
- Click "show advanced settings"
- Under "Privacy" click "Content Settings"
- Under "Plugins" choose "Let me choose when to run plugin content"
- Click done and revisit your builder

https://www.dropbox.com/s/qug11cg4up284ou/Broke.mov?dl=0

robwalch:
Hi @spodlecki. We are working on fixing bugs and improving the experience when Flash is blocked or paused by browsers. Our next release 7.1 is scheduled for mid sept. It may not have all these changes in but you can expect them there or 7.1.1 shortly after.

spodlecki:
Understood.

- What versioning system does JW Player use? It seems like JW Player races as quickly as possible to X.10.X and then jumps new major version. In any case, I'll be stuck with JW Player 5.10 until either 6 or 7 gets stable. 6 works ok, but some major performance problems -- too many customer tickets come in to justify "upgrading." Am rootin' for JW7, but we may have to throw in the towel.
- When GitHub is updated, does that mean www.jwplayer.com is released as well? I assume this is a trimmed down version of JW Player
- Is this the official repo/bug tracker for JW Player?

Apologies for questions on a Github Issue; JW Support has yet to respond to my account's ticket. "Priority E-mail" isn't so priority when its been 48 hours ;)

robwalch:
In JW Player 7 we are trying to adhere to [semver](http://semver.org/).

Have you tried 6.12? The html5 Flash players are very performant.

Flash has moved into the background to play video - it's not a fully fledged player. That makes dealing with Flash blocking more complex and we haven't planned an prioritized the changes needed until recently.

This is the open-source project for JW Player. Our commercial product is based on this player and you can find [release notes for the product on our support site](http://support.jwplayer.com/customer/portal/articles/1403726-jw-player-7-release-notes). We opened issues here as a way developers to communicate and contribute to the player directly.

donato:
@spodlecki - 
1. As of JW 7.0.0 we are using semantic versioning
2. The version on our website uses this source code and adds features on top of it - advertising, HLS, DASH, chromecast, etc...
3. This is the official repo and bug tracker

robwalch:
[...]

PR #751, opened 2015-09-22
Handle Flash throttling as flash blocked

robwalch:
Throttled event is fired when tab is backgrounded, as well as when Chrome puts the Flash plugin into 
"Power Save" mode. This shows the user that the player is in a psuedo-paused state and allows them to resume plugin playback* by right clicking and selecting "Run This Plugin" or by simply clicking.

*When clicking the player state will likely be toggled to pause and another click is required to start playing. We don't have a way to resolve this that would not interfere with players that were simply throttled by being backgrounded so we'll revisit this issue in 7.2.


robwalch:
This fixes #701 in https://github.com/jwplayer/jwplayer/tree/v7.1.1

PR #761, opened 2015-09-23
Enable preload for flash provider

pajong:
JW7-842

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
+++ b/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
@@ -88,47 +88,29 @@ public class VideoMediaProvider extends MediaProvider {
         }
     }
 
+    override public function init(itm:PlaylistItem):void {
+        setupVideo(itm);
+        // start loading the content only if preload is 'auto'
+        if (itm.preload === "auto") {
+            loadQuality();
+            _stream.pause();
+        }
+    }
+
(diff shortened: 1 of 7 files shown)


src/js/controller/model.js
```diff
@@ -288,7 +288,7 @@ define([
 
         // The model is also the mediaController for now
         this.loadVideo = function(item) {
-            this.mediaModel.set('playAttempt', true);
+            this.set('playAttempt', true);
```
donato: Why change this from mediaModel to regular model?

donato: Ps, we can use PROMISES now!

robwalch: `mediaModel` would be better than `model` because it is reset when the item changes.

src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
```diff
@@ -88,47 +88,29 @@ public class VideoMediaProvider extends MediaProvider {
...
+        if (_item !== itm) {
+            setupVideo(itm);
+            loadQuality();
+        } else if (itm.preload === "metadata") {
+            loadQuality();
+        } else {
+            play();
         }
 
+        // need to call this every time we load, but after setupVideo has been called
```
donato: why?

pajong: _currentQuality is changed in setupVideo

PR #762, opened 2015-09-23
Update duration on provider bufferChange and meta events

robwalch:
If providers send their duration in meta and/or bufferChange events while preloading the update will be set in the model and reflected in the view. getDuration() will also return the value set in the model before playback begins.

JW7-1427

I've also updated the html5 video provider to update time properties on video tag progress, durationchange and timeupdate events, so our model is always as up-to-date as possible.

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -9,9 +9,8 @@ define([
     'utils/backbone.events'
 ], function(cssUtils, utils, stretchUtils, _, events, states, DefaultProvider, Events) {
 
-    var clearInterval = window.clearInterval,
+    var clearTimeout = window.clearTimeout,
         STALL_DELAY = 256,
-        BUFFER_INTERVAL = 100,
         _isIE = utils.isMSIE(),
         _isMobile = utils.isMobile(),
         _isSafari = utils.isSafari(),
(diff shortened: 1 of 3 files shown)


src/js/providers/html5.js
```diff
@@ -159,44 +156,49 @@ define([
...
-        function _durationUpdateHandler() {
+        function _timeUpdateHandler(event) {
             if (!_attached) {
                 return;
             }
-            var newDuration = _videotag.duration;
-            if (_duration !== newDuration) {
-                _duration = newDuration;
+            // timeupdate event means we can seek and are playing
+            if (event.type === 'timeupdate') {
```
donato: I don't think we should combine them into one method if we have to branch later

.jshintrc
```diff
@@ -1,6 +1,6 @@
 {
     //"indent": 4,
-    "maxlen": 120,
+    "maxlen": 160,
```
donato: NNOOOOOOPE

src/js/providers/html5.js
```diff
@@ -212,41 +233,23 @@ define([
                 return;
             }
 
-            if (!_canSeek) {
-                _canSeek = true;
+            _canSeek = true;
+            if (_this.state === states.IDLE || _this.state === states.LOADING) {
```
donato: Maybe this isn't needed, but worth a check

src/js/view/components/timeslider.js
```diff
@@ -95,15 +96,20 @@ define([
...
             this.updateBuffer(pct);
         },
-        onPosition : function(model, pos) {
+        onPosition : function(model, position) {
+            this.onTime(position, model.get('duration'));
+        },
+        onDuration : function(model, duration) {
+            this.onTime(model.get('position'), duration);
+        },
+        onTime : function(position, duration) {
```
donato: updateMetadata?

PR #764, opened 2015-09-24
Solidify error handling

donato:
The view should not be listening for error events, it should be displaying
the state of the model!

JW7-1426


donato:
*Note - using this branch to test on other devices, don't merge just yet!

donato:
test this please

PR #768, opened 2015-09-25
Remove transition to complete state after instream complete

donato:
JW7-1517

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -128,8 +128,6 @@ define([
         }
 
         function _instreamItemComplete(e) {
-            // Allow 'play' state change to trigger next adPlay event in ad pods
-            _instream._adModel.set('state', 'complete');
 
             if (_array && _arrayIndex + 1 < _array.length) {
                 // destroy skip button


src/js/controller/instream-adapter.js
```diff
@@ -128,8 +128,6 @@ define([
         }
 
         function _instreamItemComplete(e) {
-            // Allow 'play' state change to trigger next adPlay event in ad pods
```
pajong: do we know what happens to the adPlay event in ad pods after this?

donato: Good point, I tested ad pods, but you are right this would cause a regression for ad play event between ad pods.

The relevant bits of code are here : https://github.com/jwplayer/jwplayer/commit/1cc409dcb649be85c5f66df3aba9a236490eae8a#diff-79d1ee28e72759418dcb9c1e8855dd7cR98
https://github.com/jwplayer/jwplayer/commit/05dcb16b4a15fb0f698b0a2ce9cbbbb8290d80b0#diff-79d1ee28e72759418dcb9c1e8855dd7cR61

I will continue searching :+1:

Charles-I-Blanton:
Hi guys, 

Could you provide a code example of ad pods working, using wrapped ad tags? thanks

PR #775, opened 2015-09-28
JW7-1531 Fix poster image not displayed during audio playback

pajong:
View is listening to the change in mediaType. init happens after than view load, so it does not call toggleClass for audio when we send mediaType on load.

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -876,10 +876,13 @@ define([
             if (_castDisplay) {
                 _castDisplay.setState(_model.get('state'));
             }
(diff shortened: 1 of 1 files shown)


donato:
I think a better fix would be to make the view work no matter which one happens first!

For example this is what we do in view.js for changing skins

```
            this.onChangeSkin(_model, _model.get('skin'), '');
            _model.on('change:skin', this.onChangeSkin, this);
```

in other words, we trigger the change event with current value, and then listen for changes to that value.

PR #780, opened 2015-09-29
Fix replay not working for flash provider JW7-842

pajong:
After "NetStream.Play.Complete" is called, playing the same stream does not work. It only works in repeat, because repeat actually resumes the play before the complete event is fired. Therefore, for replay, we need to reload the video. Setting _item to null will reload the video on load, because we do a comparison between the passed in item and _item before executing video load.

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
+++ b/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
@@ -370,6 +370,7 @@ public class VideoMediaProvider extends MediaProvider {
         switch (evt.info.code) {
             case "NetStream.Play.Stop":
                 complete();
+                _item = null;
                 break;
             case "NetStream.Play.StreamNotFound":
                 error('Error loading media: File not found');


robwalch:
What if we call seek() play it instead?

pajong:
It is already calling seek(0) when it stops, but does not work when "NetStream.Play.Complete" is called. It does the exact same thing as when "repeat" is set to true, but it does not work because Complete gets called before we resume the video.

pajong:
I deleted the old comment by accident - but it is already calling seek(0) when we are stopping. It is doing the exact same thing as if "repeat" is set to true, but it does not work because resuming the video happens after the complete event happened in netStream.

donato:
@robwalch - this seems like your expertise, I'll let you handle this one :+1:

robwalch:
@pajong that's only necessary is `netstream.close()` was called. It looks like there is some janky logic calling close in certain cases when 'stop()' is called.

http://help.adobe.com/en_US/FlashPlatform/reference/actionscript/3/flash/net/NetStream.html#close()

Avoiding reload of data on replay saves $$ so consider it worth focusing on.

PR #782, opened 2015-09-30
Fix disappearing volume control on "six" skin

jlowcs:
On "six" skin: fix disappearing volume control when moving cursor from the volume icon to the volume control.


jlowcs:
Used a padding so that there is no gap between the icon and the control.

PR #786, opened 2015-10-01
JW7-1342 JW7-1158 Fix ad skip button and overlay ads on small player, and fix caption position

pajong:
Caption and overlay ads have been changed to always be 0.5em above the controlbar.
When inactive and controlbar is hidden, they move down to be 0.5em above the bottom of the player.
Skip button height does not change when controlbar is hidden.


donato:
Will come back to this after getSafeRegion

PR #788, opened 2015-10-02
Show time tooltip for DVR streams

donato:
This involves calculating and formatting the time offset based on the
negative duration given.

JW7-1368

### DIFF
--- a/src/js/utils/parser.js
+++ b/src/js/utils/parser.js
@@ -116,16 +116,20 @@ define([
     };
 
     /** Format the elapsed / remaining text. **/
-    parser.timeFormat = function(sec) {
-        if (sec > 0) {
-            var hrs = Math.floor(sec / 3600),
-                mins = Math.floor((sec - hrs * 3600) / 60),
-                secs = Math.floor(sec % 60);
-
-            return (hrs ? hrs + ':' : '') + (mins < 10 ? '0' : '') + mins + ':' + (secs < 10 ? '0' : '') + secs;
-        } else {
+    parser.timeFormat = function(sec, allowNegative) {
+        if (sec <= 0 && !allowNegative) {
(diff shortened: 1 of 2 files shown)


src/js/view/components/timeslider.js
```diff
@@ -157,11 +157,22 @@ define([
...
             var timetipText;
             if (this.activeCue) {
                 timetipText = this.activeCue.text;
             } else {
-                timetipText = utils.timeFormat(time);
+                var allowNegativeTime = true;
+                timetipText = utils.timeFormat(time, allowNegativeTime);
+
+                // If DVR and within one second of live
+                if (duration < 0 && (Math.abs(time) < 1) ) {
```
robwalch: Does this work with DVR stream time events?

"within one second of live" `if (duration < 0 && (Math.abs(time) < 1)` 

I would expect it to be more like within 25-30 seconds is "Live" and 1 second of latency is very rare or unlikely.

donato: As discussed in person, this is just for the text inside of the tooltip.

src/js/utils/parser.js
```diff
@@ -116,16 +116,20 @@ define([
...
+    parser.timeFormat = function(sec, allowNegative) {
+        if (sec <= 0 && !allowNegative) {
             return '00:00';
         }
+
+        // If negative add a minus sign
+        var prefix = (sec < 0) ? '-' : '';
+        sec = Math.abs(sec);
+
+        var hrs  = Math.floor(sec / 3600),
```
donato: *technically* we should be doing Math.ceil if the time is negative

donato:
Waiting to merge for @pajong 's unit test change

donato:
test this please

PR #794, opened 2015-10-05
Fix a bug with stretching

pajong:


robwalch:
This doesn't make `isInt` private

pajong:
@robwalch I separated that because it was breaking the build due to the test looking for isInt defined

PR #797, opened 2015-10-06
JW7-1342 Fix ad skip button and overlay ads on small player

pajong:


src/css/imports/jwplayerelement.less
```diff
@@ -116,6 +116,19 @@
 
 .jw-plugin {
     position: absolute;
+    bottom: 2.5em;
+
+    .jw-banner {
+        max-width: 100%;
+        opacity: 0;
```
donato: why opacity of zero? Isn't this for nonlinears?

pajong: the opacity is changed to 1 by show function when we want to show the nonlinear ad

PR #799, opened 2015-10-06
JW7-842 Treat preload metadata as auto in flash provider

pajong:
We decided to preload as if it was auto whenever  preload is not none, because we can't just download metadata without start playing the video in flash.


src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
```diff
@@ -90,11 +90,8 @@ public class VideoMediaProvider extends MediaProvider {
 
     override public function init(itm:PlaylistItem):void {
         setupVideo(itm);
-        // start loading the content only if preload is 'auto'
-        if (itm.preload === "auto") {
```
pajong: We do not need to check if 'auto' or 'metadata' anymore, because we want to treat them as the same. We do not need to check for 'none', because init is only called if preload !== none from flash.js

PR #800, opened 2015-10-07
allow pre-processing of captions via a configuration setting

chrisdeely:
This is a feature request to allow preprocessing of caption lines. I have encountered a situation where characters within caption lines have become corrupted (unrelated to JW Player) and need to be replaced on the client side.

Adding a ```captions.captionParser``` configuration option seems to be a very clean solution, and opens other possibilities including displaying related content based on the content of a live broadcast.

### DIFF
--- a/src/js/view/captionsrenderer.js
+++ b/src/js/view/captionsrenderer.js
@@ -19,9 +19,7 @@ define([
         edgeStyle: null,
         windowColor: '#FFF',
         windowOpacity: 0,
-        captionParser: function (input) {
-            return input;
-        }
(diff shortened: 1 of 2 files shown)


donato:
@chrisdeely - I can imagine a few really cool uses for this, what do you think about naming the option something like ```preprocessor```?

Also, might as well use _.identity for the default method.

chrisdeely:
@donato agreed on all points - I'll update the PR. I knew underscore had an appropriate method, just couldn't find it!

donato:
@chrisdeely - thanks looks good!

PR #801, opened 2015-10-07
1. Refactor setup steps

donato:
The primary goal of this is to remove model.setItem from setup-steps.js
This will make it easier to remove setItem from the model and move it into the controller.

This change means that you are no longer guaranteed to have a playlistItem when setting up the view,
so there is much more safety around setup methods in the view.

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -23,9 +23,10 @@ define([
 
         this.mediaController = _.extend({}, Events);
         this.mediaModel = new MediaModel();
-        this.set('mediaModel', this.mediaModel);
 
         QOE.model(this);
+
+        this.set('mediaModel', this.mediaModel);
 
         this.setup = function(config) {
             if (config.cookies) {
@@ -34,6 +35,8 @@ define([
             }
 
             _.extend(this.attributes, config, _cookies, {
+                // always start on first playlist item
+                item : 0,
                 // Initial state, upon setup
                 state: states.IDLE,
                 // Initially we don't assume Flash is needed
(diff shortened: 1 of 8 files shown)


src/js/view/title.js
```diff
@@ -25,7 +25,9 @@ define([
             this.description = arr[1];
 
             this.model.on('change:playlistItem', this.playlistItem, this);
-            this.playlistItem(this.model, this.model.get('playlistItem'));
```
robwalch: We should add a helper function for these like:

```
this.model.watch('playlistItem', this.playlistItem, this);
```

src/js/controller/model.js
```diff
@@ -161,20 +164,23 @@ define([
...
                 }
             }
 
             _currentProvider = new Provider(_this.get('id'), _this.getConfiguration());
 
+            var container = this.get('mediaContainer');
             if (container) {
                 _currentProvider.setContainer(container);
+            } else {
+                this.once('change:mediaContainer', function(model, container) {
```
robwalch: These `once` listeners are dangerous. This one needs to be cleared when _currentProvider changes, incase that happens before the event is fired.

src/js/view/title.js
donato: agreed! I wonder if there is an industry standard for this already?

src/js/view/preview.js
```diff
@@ -4,15 +4,16 @@ define([
...
-
-        this.model.on('change:playlistItem', this.loadImage, this);
     };
 
     _.extend(Preview.prototype, {
         setup: function(element) {
             this.el = element;
 
-            this.loadImage(this.model, this.model.get('playlistItem'));
+            this.model.on('change:playlistItem', this.loadImage, this);
```
robwalch: Do we insure that setup is only called once? It might be useful to have a shortcut for `off(event, fn, context).on('event, fn, context)` to avoid duplicate listeners. Maybe that could be part of the `watch` helper I suggested below:

src/js/controller/model.js
donato: Good catch, thanks

src/js/view/title.js
donato: Chatting here ; https://gitter.im/marionettejs/backbone.marionette

donato:
@robwalch - I addressed your feedback about the once call and binding events in the constructor instead of the setup

PR #802, opened 2015-10-07
JW7-1559 Make preload bubble down to source level

pajong:


pajong:
https://jwplayer.atlassian.net/browse/JW7-838, in the description: 1. Add a player level config option, which bubbles down to the individual source level

PR #804, opened 2015-10-07
3. Remove setPlaylist from the model

donato:
Please review/merge other two PR's first, as it will simplify this one


donato:
Need to fix the unit tests.

PR #806, opened 2015-10-07
JW7-1258 Display the display icon when paused on mobile

pajong:


src/css/flags/touch.less
```diff
@@ -21,5 +21,8 @@
     .jw-display-icon-container {
         pointer-events: none;
     }
-}
 
+    &.jwplayer&.jw-state-paused .jw-display-icon-container {
```
donato: Why the ampersands?

pajong: because we want to select .jwplayer.jwflag-touch.jw-state-paused.
ampersands refers to parent selectors.

donato: Understood!

I was hinting that it's more complex than it probably ought to be.

If for some reason it needs to be that "specific" then the alternative is making the rule you are overwriting less "specific" instead.

donato: Another little trick that will help us in the long run is to keep track of "why" things are so specific. For example see this -> https://github.com/jwplayer/jwplayer/blob/master/src/css/flags/audioplayer.less#L33

PR #808, opened 2015-10-09
JW7-1562 Fix postroll replaying single playlist item

pajong:
Changing mediaModel's state to buffering also changes model's state. Then, on the next lines, when we check the type of ads (pre/mid/post roll), we are not getting model state === complete on postrolls, so they are being treated as midrolls.

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -83,8 +83,6 @@ define([
             // Make sure the original player's provider stops broadcasting events (pseudo-lock...)
             _oldProvider.detachMedia();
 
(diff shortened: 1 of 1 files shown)


donato:
This is a finicky part of the code. It's worth digging in a little. 

What commit caused this bug to start happening?
Using "git blame" I found this
https://github.com/jwplayer/jwplayer/commit/20f194a3153386c4beea1e3fc63f1ef0a1011a42

pajong:
@donato We actually found an issue with the test page, which uses "onComplete" instead of "onBeforeComplete" to fire postroll. I will close this PR.

PR #809, opened 2015-10-09
JW7-1258 JW7-1259 Make touchHandler to resume play on paused

pajong:


donato:
This story is changing what a tap on mobile does. In the past it toggled visibility of the controls over the player. Now, it shouldn't toggle controls, it should start playback when the player is paused. The toggling only occurs when paused. 

Look at the rest of the _touchHandler and fix it up.

PR #811, opened 2015-10-09
3. Remove setPlaylist from model

donato:


src/js/api/api-mutators.js
```diff
@@ -70,6 +70,7 @@ define([
             'setFullscreen',
             'setVolume',
             'setMute',
+            'setPlaylist',
```
robwalch: Adding an API method? If so that should be clearer in the PR title, and the implications should be discussed - like how is this different than `load()`?

If it's not needed publicly, don't add it here.

src/js/controller/setup-steps.js
```diff
@@ -25,15 +25,15 @@ define([
             },
             LOAD_YOUTUBE : {
                 method: _loadYoutube,
-                depends: ['LOAD_PLAYLIST']
+                depends: ['LOAD_POLYFILLS']
             },
             LOAD_SKIN : {
                 method: _loadSkin,
-                depends: []
+                depends: ['LOAD_POLYFILLS']
```
robwalch: Comment on which polyfills Youtube and skins require.

src/js/controller/setup-steps.js
```diff
@@ -25,15 +25,15 @@ define([
...
             },
             LOAD_PLAYLIST : {
                 method: _loadPlaylist,
-                depends: ['LOAD_PLUGINS']
+                depends: ['LOAD_PLUGINS', 'LOAD_YOUTUBE']
```
robwalch: RSS Playlist need to be loaded before LOAD_YOUTUBE's `hasYoutube`.

`_completePlaylist` should be a new step here, and the `_.isString(playlist)` handle loading earlier on.

src/js/api/api-mutators.js
donato: You're right, good call. 

I don't particularly like the alternative of passing in references to the controller into setup, but it's better than adding to the public API.

src/js/controller/setup-steps.js
donato: I'm pushing for the idea that the whole app waits for polyfills - that way we don't need to worry about moving around dependencies as we change around functionality.

PR #812, opened 2015-10-09
Fix for mediaModel getting reset unexpectedly

donato:
This bug manifested as the HD menu not appearing because the controlbar
listens for a "change:levels" event from the mediaModel in order to populate.
It was, however, listening to the wrong instance of mediaModel, because we bound
the listener when we see a "change:playlistItem" event.

We assumed that every time we create a new mediaModel that we would also call
the "change:playlistItem" in order to rebind listeners. This is not true in some
situations, such as when the new item is the same as the old one.

The better solution is to use an event which specifically tells you when the item changes
which I call here "itemSet"

JW7-1563

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -120,7 +120,7 @@ define([
                     });
                 });
                 // For onItem callback
-                _model.on('change:playlistItem', function(model, playlistItem) {
+                _model.on('itemSet', function(model, playlistItem) {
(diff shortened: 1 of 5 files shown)


robwalch:
It would make more sense to listen to `change:mediaModel`. Having an event that fires when we set, but the value does not change suggests that the scheme we're using is broken.

robwalch:
@donato let's get #811 in first and then revisit this one. If the view is notified when media model changes this will be a smaller change.

PR #813, opened 2015-10-12
JW7-1272 Fix logo to show when controls set to hidden

pajong:


jw-kaurand:
This works as expected, however it can cause the player to be interacted with in potentially undesirable ways when controls is set to false.  On http://player-develop-test-jenkins.longtailvideo.com/builds/lastSuccessfulBuild/archive/test/squash/bootstrap.html?primary=html5&edition=ads&feature=editions_logo_options we can play/pause the video player by clicking on the logo in the last test and we still link out to other pages when we click on the logo while controls is false.  We can talk to Andrew to confirm intended behavior in this situation.

PR #815, opened 2015-10-12
JW7-1422 Move overlay:before up so we get mouseout event on controlbar

pajong:


jw-kaurand:
This issue actually raises a fairly interesting problem.  Mousing over the time bar uses some of the time tooltip as part of the mouse over target for hovering the controlbar.  By moving the :before up we're making the hover target for the controlbar significantly smaller and making it so that the user can easily mouse off of the time bar by moving even a little above the rail of the time slider.  This issue could probably be better addressed by deciding exactly how the mouse over functionality on a the control bar should work and then modifying the "jw-tooltip-time" div to be positioned correctly as the hover target on the slider. From there, the overlay in the tooltip and all content should be made to not take pointer events.

PR #817, opened 2015-10-13
Fix for HD menu not showing up 

donato:
JW7-1563

### DIFF
--- a/src/js/view/controlbar.js
+++ b/src/js/view/controlbar.js
@@ -226,7 +226,7 @@ define([
 
             if(this.elements.playlist) {
                 this.elements.playlist.on('select', function (value) {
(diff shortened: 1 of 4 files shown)


src/js/view/controlbar.js
```diff
@@ -224,7 +226,7 @@ define([
 
             if(this.elements.playlist) {
                 this.elements.playlist.on('select', function (value) {
-                    this._model.once('setItem', function () {
+                    this._model.once('itemReady', function () {
```
robwalch: We should add a comment to clean this up. There's not a reliable way to remove the event. Ideally we could just call load and play in one command or load followed by play() would do the queueing reliably like autostart does above.

PR #820, opened 2015-10-14
Display error message when flash is throttled

robwalch:
JW7-1512

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -88,7 +88,13 @@ define([
             // If we attempt to load flash, assume it is blocked if we don't hear back within a second
             _model.on('change:flashBlocked', function(model, isBlocked) {
                 if (isBlocked) {
-                    this.trigger(events.JWPLAYER_ERROR, { message: 'Flash plugin is blocked'});
+                    var evt = {
+                        message: 'Flash plugin is blocked'
+                    };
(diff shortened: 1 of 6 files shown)


src/js/view/view.js
```diff
@@ -862,10 +864,10 @@ define([
             _this.setAltText((live) ? 'Live Broadcast' : '');
         }
 
-        function _errorHandler() {
-            var evt = _model.get('errorEvent');
-
-            if (evt.name) {
+        function _errorHandler(model, evt) {
+            if (!evt) {
+                _title.updateText('', '');
```
donato: It'd be ideal if we can re-set this to be playlist items title and description

src/js/controller/controller.js
```diff
@@ -88,7 +88,13 @@ define([
             // If we attempt to load flash, assume it is blocked if we don't hear back within a second
             _model.on('change:flashBlocked', function(model, isBlocked) {
                 if (isBlocked) {
-                    this.trigger(events.JWPLAYER_ERROR, { message: 'Flash plugin is blocked'});
+                    var evt = {
+                        message: 'Flash plugin is blocked'
+                    };
+                    this._model.set('errorEvent', evt);
```
donato: I would prefer to have the view react to the change:flashBlocked event

PR #824, opened 2015-10-15
JW7-1494 Increase flash timeout to 4 seconds and fire clearError on unblock flash

pajong:


robwalch:
@jong just change the timeout and have @dannyfinks change his test to deal with longer timeouts and ignoring the error. Let's not send new events from the player for a test.

PR #827, opened 2015-10-15
Fix quality level and remove cookies

donato:
JW7-1071

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -10,7 +10,6 @@ define([
         //androidhls: true,
         autostart: false,
         controls: true,
-        cookies: true,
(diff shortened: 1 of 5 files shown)


src/js/api/config.js
```diff
@@ -10,7 +10,6 @@ define([
         //androidhls: true,
         autostart: false,
         controls: true,
-        cookies: true,
```
robwalch: Do we still want this config option to control whether settings are stored? It's a feature.

robwalch: @androoP @esboyd ? Is this feature no longer needed if the settings are stored in local storage?

PR #829, opened 2015-10-15
Add a flag for denoting that an overlay is open

jw-kaurand:
Add a flag for denoting that an overlay is open.  while an overlay is open, hide the title.


src/css/flags/overlay.less
```diff
@@ -0,0 +1,5 @@
+.jwplayer.jw-flag-overlay-open {
+    .jw-title {
+        visibility: hidden;
```
robwalch: Why are we using visibility for this instead of display: none?

PR #832, opened 2015-10-16
JW7-894 Replace p. with ssl.

pajong:


src/js/utils/playerutils.js
```diff
@@ -12,7 +12,7 @@ define([
         var semver = version.split('+')[0];
         var repo = Constants.repo + semver + '/';
         if (validator.isHTTPS()) {
-            return repo.replace('http://', 'https://ssl.');
+            return repo.replace(/^http:/, 'https://ssl.');
```
robwalch: Wouldn't this make the repo https://ssl.//ssl.p.jwpcdn.com ? Is this tested?

PR #833, opened 2015-10-16
JW7-1566 Check googima adpod completed before destroying instream

pajong:


pajong:
Googima sends adpod complete information after this PR: https://github.com/jwplayer/jwplayer-ads-googima/pull/95

PR #835, opened 2015-10-17
Listen to video tag for paused state and don't set state while detached

### LINKED ISSUE #834: Forward pause events from the video tag
The html5 provider is not forwarding pause events from the video tag. This is necessary to determine when video is paused on iPhone.

robwalch:
This is in response to a question on stackoverflow http://stackoverflow.com/questions/33089838/detect-pause-and-done-buttons-click-on-iphone-for-jwplayer7

The player state should change to paused if the video tag is paused without jw api interaction. The provider state should remain unchanged when the provider is not attached.

Clearing the stalled timer and maintaining canSeek state is moved up before exiting when not attached.

Fixes #834
JW7-1605

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -68,6 +68,13 @@ define([
             return Events.trigger.call(this, type, args);
         };
 
+        this.setState = function(state) {
+            if (!_attached) {
+                return;
+            }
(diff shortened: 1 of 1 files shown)


donato:
This looks right, but has a huge scope. Might as well get it in before the regression testing starts!

PR #836, opened 2015-10-19
JW7-1585 - Fix polyfills prematurely resolving.

jw-kaurand:
Fix polyfills prematurely resolving.

### DIFF
--- a/src/js/controller/setup-steps.js
+++ b/src/js/controller/setup-steps.js
@@ -79,9 +79,17 @@ define([
             }
         }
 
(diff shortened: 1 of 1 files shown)


robwalch:
@donato @jw-kaurand I think the right thing to do here would be to make to steps - one to resolve each required polyfill.

donato:
Good fix.

@robwalch - I agree in terms of code organization, but in terms of functionality, they should be able to load async from each other, and only load the ones necessary.

PR #837, opened 2015-10-19
JW7-1585 - Broke out polyfills into steps. Completes another step when resolved.

jw-kaurand:
Broke out polyfills into steps. Completes another step when resolved.

Should be merged at the same time as https://github.com/jwplayer/jwplayer-commercial/pull/944


robwalch:
Awesome! I like the 'group' step that requires all polyfills and is required by subsequent steps. @donato does this look good to you?

PR #839, opened 2015-10-19
Use trycatch util when executing player event callbacks

robwalch:
This allows disabling of try catch on event handlers using `jwplayer.debug = true` which was the main goal of adding this util to the player.

The method also returns early if the callback is not a function or String.

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -4,6 +4,7 @@ define([
     'utils/backbone.events',
     'utils/helpers',
     'utils/timer',
+    'utils/trycatch',
     'utils/underscore',
     'controller/controller',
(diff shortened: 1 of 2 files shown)


src/js/api/api.js
```diff
@@ -50,13 +51,16 @@ define([
             if (_.isString(callback)) {
```
donato: maybe we delete the if(_.isString) ?

robwalch: We're still testing that this exception gets thrown. I like the idea of keeping *some* error or warning. I only added the isFunction check to pass a unit test where we're not passing in a function. Should we be as strict as possible or not at all?

PR #841, opened 2015-10-19
JW7-1590 Fix display icon triggering chrome power save

pajong:


pajong:
@jw-kaurand Tell me if you like this fix or not!

PR #842, opened 2015-10-20
JW7-1317 Fix android youtube being stuck on cued or unstarted

pajong:
If we try to play the youtube video too quickly on Android, the youtube API sends us unstarted event, or cue it until we start playing. To fix this issue, we want to play the video when we get these states on Android.

### DIFF
--- a/src/js/providers/youtube.js
+++ b/src/js/providers/youtube.js
@@ -247,6 +247,10 @@ define([
             switch (_youtubeState) {
 
                 case youtubeStates.UNSTARTED: // -1: //unstarted
+                    // play video on android to avoid being stuck in this state
+                    if (utils.isAndroid()) {
+                        _youtubePlayer.playVideo();
+                    }
                     return;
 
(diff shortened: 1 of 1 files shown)


src/js/providers/youtube.js
```diff
@@ -247,6 +247,10 @@ define([
             switch (_youtubeState) {
 
                 case youtubeStates.UNSTARTED: // -1: //unstarted
+                    // play video on android to avoid being stuck in this state
+                    if (utils.isAndroid()) {
+                        _youtubePlayer.playVideo();
```
robwalch: How are you sure we want to play when the YT player state changes to unstarted or cued? Couldn't YT enter this state during our player's complete, idle or preload phase (no autostart)?

pajong: I have tested the cases, but it does not happen. Preload does not change the YT player state, and for complete we get ENDED state. The only time we get unstarted and cued event is when we tried to play the video.

PR #846, opened 2015-10-22
JW7-1614 Fix youtube poster not being displayed

pajong:


robwalch:
The poster is supposed to be displayed in the YouTube iFrame. Not in the player.

src/js/providers/youtube.js
```diff
@@ -374,6 +374,7 @@ define([
 
             if (!item.image) {
```
robwalch: This `image` property could be added on setup so that we don't need new event handling:

https://github.com/jwplayer/jwplayer/blob/master/src/js/controller/setup-steps.js#L218

PR #849, opened 2015-10-26
JW7-1638 Fix duration NaN issue

pajong:


pajong:
@donato I tried to set default by doing _videotag.setAttribute('duration', 0), but it returns NaN when we try to get _videotag.duration right after that (I think we can't just change to fake duration of the video tag with unknown source)

PR #860, opened 2015-11-03
JW7-1679 Fix first frame fired with negative value with preload auto

pajong:


robwalch:
Let's try to distinguish between preload load and play load in flash so that we don't track play attempt in this case.

PR #863, opened 2015-11-04
JW7-1671 Fix flash replay not working

pajong:


donato:
Why do you think complete was there to begin with? 

This commit message does not say what the fix actually is, or why it was broken.

PR #865, opened 2015-11-04
JW7-1679 Do not send firstFrame event when preload is auto on flash

pajong:


donato:
Okay, so now we aren't sending it with visualQuality, then when are we sending it?

Add more details to commit and pull request so that others can understand what's going on.

pajong:
@donato Sorry for the lack of explanation. In qoe.js, there is a function called onTimeIncreasesGenerator, which calls the first frame event if the provider does not give one. So, for preload auto, where we do not have a playAttempt when the first frame occurs, we do not send the first frame event. Then, qoe will handle and treat the first increase in the time position as the first frame event.

robwalch:
visualQuality should not be fired from Flash until "real" playback starts.

PR #867, opened 2015-11-04
JW7-1671 Do not call complete function when flash video completes

pajong:
complete function dispatched an event that calls stop function.
The stop function in VideoMediaProvider resets _buffered and _keyframes variables,
causing the player to think that it is not buffered enough to play when we try to replay.


donato:
All of the flash "media" classes call complete on completion, so we should probably keep doing that.

This feels like it could cause other problems related to the "stop()" not being called when we really want it to be.

PR #868, opened 2015-11-04
JW7-1723 JW7-1710 - Adds a test to detect Edge

jw-kaurand:
Adds a test that will detect IE Edge to fix other issues.  For Trident tests, Edge counts as a trident browser because they have the same sorts of issues.  isChrome is changed to test for chrome in the userAgent && !isEdge because Edge will represent itself as chrome.

This will have a PR paired with it in commercial that maintains DASH support on Edge

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -28,7 +28,7 @@ define([
     browser.isIPod = _browserCheck(/iP(hone|od)/i);
     browser.isIPad = _browserCheck(/iPad/i);
     browser.isSafari602 = _browserCheck(/Macintosh.*Mac OS X 10_8.*6\.0\.\d* Safari/i);
-    browser.isEdge = _browserCheck(/edge/i);
+    browser.isEdge = _browserCheck(/\sedge\/\d+/i);
 
     var _isIETrident = browser.isIETrident = function(browserVersion) {
         if(browser.isEdge()){
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -25,12 +25,15 @@ define([
     };
 
     browser.isFF = _browserCheck(/firefox/i);
-    browser.isChrome = _browserCheck(/chrome/i);
     browser.isIPod = _browserCheck(/iP(hone|od)/i);
     browser.isIPad = _browserCheck(/iPad/i);
     browser.isSafari602 = _browserCheck(/Macintosh.*Mac OS X 10_8.*6\.0\.\d* Safari/i);
+    browser.isEdge = _browserCheck(/edge/i);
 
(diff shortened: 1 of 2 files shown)


jw-kaurand:
Related to the commercial PR here: https://github.com/jwplayer/jwplayer-commercial/pull/1001

src/js/utils/browser.js
```diff
@@ -25,12 +25,15 @@ define([
...
 
     browser.isFF = _browserCheck(/firefox/i);
-    browser.isChrome = _browserCheck(/chrome/i);
     browser.isIPod = _browserCheck(/iP(hone|od)/i);
     browser.isIPad = _browserCheck(/iPad/i);
     browser.isSafari602 = _browserCheck(/Macintosh.*Mac OS X 10_8.*6\.0\.\d* Safari/i);
+    browser.isEdge = _browserCheck(/edge/i);
 
     var _isIETrident = browser.isIETrident = function(browserVersion) {
+        if(browser.isEdge()){
```
donato: I'm not a big fan of pretending Edge is Trident, does anyone have a concrete example of a bug for IE9-11 which also applies to Edge?

robwalch: There are lots of IE11 bugs reported for this beta that occur in Edge. No one has verified if these issues also occur in IE9-10.

src/js/utils/browser.js
```diff
@@ -25,12 +25,15 @@ define([
     };
 
     browser.isFF = _browserCheck(/firefox/i);
-    browser.isChrome = _browserCheck(/chrome/i);
     browser.isIPod = _browserCheck(/iP(hone|od)/i);
     browser.isIPad = _browserCheck(/iPad/i);
     browser.isSafari602 = _browserCheck(/Macintosh.*Mac OS X 10_8.*6\.0\.\d* Safari/i);
+    browser.isEdge = _browserCheck(/edge/i);
```
robwalch: That's it? Seems a little... short and likely to match parts of other words or strings in other user-agents.

donato: This may help : https://msdn.microsoft.com/en-us/library/hh869301(v=vs.85).aspx

donato: /Edge\/12\./

jw-kaurand: I was thinking something like `/\sedge\/\d+/` would work pretty well in case they change the version number in the future and a leading space to make sure that we don't hit a substring of some other word.  We lowercase everything before we check it so caps can't detect the start of the word.

donato: bueno

jw-kaurand: test this please

PR #870, opened 2015-11-05
JW7-1671 Reload flash video on replay

pajong:
Flash video does not reload on replay because it has the same item.
When complete function is called after playback, it calls stop function.
Could not figure out why, but after stop function, it is not possible to resume the netstream.
Forcing the stream to reset fixes the issue.

Tried to set _item to null when complete is called, but that results a flash error.

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
+++ b/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
@@ -364,7 +364,8 @@ public class VideoMediaProvider extends MediaProvider {
     protected function statusHandler(evt:NetStatusEvent):void {
         switch (evt.info.code) {
             case "NetStream.Play.Stop":
-                complete();
+                sendMediaEvent(MediaEvent.JWPLAYER_MEDIA_COMPLETE);
+                clearInterval(_interval);
                 break;
             case "NetStream.Play.StreamNotFound":
                 error('Error loading media: File not found');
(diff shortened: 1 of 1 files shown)


src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
```diff
@@ -97,9 +99,10 @@ public class VideoMediaProvider extends MediaProvider {
     /** Load new media file; only requested once per item. **/
     override public function load(itm:PlaylistItem):void {
         setState(PlayerState.LOADING);
-        if (_item !== itm) {
+        if (_item !== itm || _forceReload) {
```
robwalch: complete() sets the state to IDLE so:

`if (_item !== itm || state === IDLE) {`

pajong: @robwalch  That will make the item load again after preload

src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
```diff
@@ -97,9 +99,10 @@ public class VideoMediaProvider extends MediaProvider {
     /** Load new media file; only requested once per item. **/
     override public function load(itm:PlaylistItem):void {
         setState(PlayerState.LOADING);
-        if (_item !== itm) {
+        if (_item !== itm || _complete) {
             setupVideo(itm);
             loadQuality();
+            _complete = false;
```
robwalch: Set this to false in `loadStream()`, since there are a couple of other ways that `_complete` can be reset (any call to `_stream.play` reloads the video).

PR #872, opened 2015-11-06
JW7-1740 JW7-1741 - Set loading state on mobile when the tag unpauses

jw-kaurand:
Set loading state on mobile when the tag unpauses

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -393,7 +393,7 @@ define([
                 // results in html5.controller calling video.play()
                 _sendBufferFull();
                 // If we're still paused, then the tag isn't loading yet due to mobile interaction restrictions.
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -392,6 +392,10 @@ define([
             if (_isMobile) {
                 // results in html5.controller calling video.play()
                 _sendBufferFull();
+                // If we're still paused, then the tag isn't loading yet due to mobile interaction restrictions.
+                if(!_videotag.paused){
+                    _this.setState(states.LOADING);
+                }
```
robwalch: Can you put this before `_sendBufferFull`? It would be a shame to set the state to loading after playback is stated.

donato:
What does this fix? Why does it fix that?

PR #873, opened 2015-11-06
Pause trident on stop to ensure it doesn't continue playing

jw-kaurand:
Pause trident after load to ensure that it doesn't start playing again once it loads content.
Renamed a variable to clarify the difference between MSIE and IETrident handling

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -11,7 +11,7 @@ define([
 
     var clearTimeout = window.clearTimeout,
         STALL_DELAY = 256,
-        _isIE = utils.isMSIE(),
+        _isMSIE = utils.isMSIE(),
         _isMobile = utils.isMobile(),
         _isSafari = utils.isSafari(),
         _isAndroid = utils.isAndroidNative(),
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -419,9 +420,13 @@ define([
                 return;
             }
             _videotag.removeAttribute('src');
-            if (!_isIE) {
+            if (!_isMSIE) {
                 _videotag.load();
             }
+            // IE may continue to play a video after changing source and loading a new media file.  JW7-1705
+            if(utils.isIETrident()) {
```
robwalch: `&& isEdge()` ?

donato: @jw-kaurand - good initiative to create a test-page which shows this behavior in Edge. 
We should post it as a question here : https://connect.microsoft.com/IE
and link the question here (as opposed to linking to a jw story, which is not visible to open-source community)

dannyfinks: @esboyd also has connections with their Browser team that he could possibly expedite this to.

jw-kaurand: The isIETrident check is checking if we're Edge or Trident, since this issue exists in both i think we can just use utils.isIETrident.  I'll post the question and remove the story link.

robwalch: @jw-kaurand @donato logged the bug with M$ https://connect.microsoft.com/IE/feedbackdetail/view/2000141/htmlmediaelement-autoplays-after-src-is-changed-and-load-is-called

PR #874, opened 2015-11-06
JW7-1539 Make audio object smaller than 5px by 5px

pajong:
Making the object tag smaller than 5px by 5px prevents Power Save throttle.
We do not want to power save in audio mode, because we cannot start the playback.

Asked @dannyfinks for a test case in squash.
Tested that this works in harness test page.

Tested with 0px by 0px, but it throttles.

### DIFF
--- a/src/css/flags/audioplayer.less
+++ b/src/css/flags/audioplayer.less
@@ -3,6 +3,12 @@
 
     .jw-media {
         visibility: hidden;
+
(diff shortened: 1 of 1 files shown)


robwalch:
@pajong you should make the test. Give @dannyfinks a break. Just modify add a player config json file and add a scenario that uses it.

PR #875, opened 2015-11-06
Fix to chrome detection

jw-kaurand:
Fix to because it wasn't calling the check before.


robwalch:
Fix to?

donato:
I updated the title and description

PR #876, opened 2015-11-06
Set state to loading when toggling quality and always seek

robwalch:
Fixes issue where changing quality leaves player while paused leaves the player in a paused state, after playback resumes in the new state (JW7-1675). Changing the state to loading when we set currentQuality ensures that the state goes back to playing.

Seeking also doesn't work with progressive streams because we don't allow seeking to unbuffered areas. Not anymore. Flash can seek to these locations it just takes more time.

Finally prefer the use of netstream state over state captured on previous interval. It's more accurate (position and buffered time).


robwalch:
Closed until seeking is improved.

PR #878, opened 2015-11-09
Change text on flash blocked error

donato:
Since flash may fail to load for reasons other than flash-blocking
(such as slow internet or missing .swf files on cdn), the text should
be more generic.


donato:
test this please

donato:
test this please

PR #879, opened 2015-11-09
Set state to loading when toggling quality and always seek

robwalch:
Fixes issue where changing quality leaves player while paused leaves the player in a paused state, after playback resumes in the new state (JW7-1675). Changing the state to loading when we set currentQuality ensures that the state goes back to playing.

Seeking also doesn't work with progressive streams because we don't allow seeking to unbuffered areas. This change allows that when quality switching, and when not, allows it up to 5 seconds past the buffered time (this is much better than an arbitrary delay of 20 milliseconds in which time the buffer does who-knows-what).


donato:
test this please

donato:
test this please

PR #881, opened 2015-11-09
Save all sources before filtering to bestType

pajong:
Since playlist item sources were being filtered out if the type is different from bestType,
we could not find castable hls source when dash was chosen as bestType.
Saving the sources before filtering will allow us to search for castable sources in chromecast controller.
JW7-1751

### DIFF
--- a/src/js/playlist/playlist.js
+++ b/src/js/playlist/playlist.js
@@ -18,7 +18,9 @@ define([
 
         _.each(playlist, function(item) {
             item = _.extend({}, item);
-            item.sources = _filterSources(item.sources, providers, androidhls,
+            item.allSources = _formatSources(item.sources, androidhls,
+                item.drm || configDrm, item.preload || preload);
+            item.sources = _filterSources(item.allSources, providers, androidhls,
                 item.drm || configDrm, item.preload || preload);
 
             if (!item.sources.length) {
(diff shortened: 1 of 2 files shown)


donato:
We shouldn't have to call mapSources twice.

Also that name could be improved, maybe something like "_formatSources"

src/js/playlist/playlist.js
```diff
@@ -39,15 +41,8 @@ define([
...
-    // A playlist item may have multiple different sources, but we want to stick with one.
-    var _filterSources = function(sources, providers, androidhls, itemDrm, preload) {
-
-        // legacy plugin support
-        if (!providers || !providers.choose) {
-            providers = new Providers({primary : providers ? 'flash' : null});
-        }
-
-        sources = _.compact(_.map(sources, function(originalSource) {
+    var _mapSources = function(sources, androidhls, itemDrm, preload) {
```
donato: In programming, a "map" is a term used to describe the application of a function against every element of a collection.

With a name like "mapSources" you really don't learn anything about what the function does, as a matter of fact, you may assume that it takes a function and returns the map of that function against sources. 

Which is of course false. 

Try to think of a better name for the function based on what it's purpose is, or what it actually does.

src/js/playlist/playlist.js
```diff
@@ -18,7 +18,9 @@ define([
 
         _.each(playlist, function(item) {
             item = _.extend({}, item);
-            item.sources = _filterSources(item.sources, providers, androidhls,
+            item.allSources = _mapSources(item.sources, androidhls,
+                item.drm || configDrm, item.preload || preload);
+            item.sources = _filterSources(item.allSources, providers, androidhls,
```
donato: Double check this line

PR #882, opened 2015-11-10
OSX Firefox Flash fix for click handling on the video/poster image/flash ad

jw-kaurand:
This enables a workaround for detecting clicks on a flash object in OS X Firefox.  This adds a special, regular click listener to the UI object for the clickHander when using OS X Firefox.  Clicks weren't working in the bug because OS X FF wasn't handling mouse up listeners when a flash object was clicked, thus making it impossible for the UI to detect clicks.  OS X FF however will dispatch a regular click event from a flash object, so if we listen for that in those circumstances we can handle ad clicks correctly on the platform.  This strategy previously didn't work because the nonlinear ads existing within the flash object would have disrupted this strategy.  This also enables the user to click on the poster image to start playing a video on OS X FF.

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -28,6 +28,7 @@ define([
     browser.isIPod = _browserCheck(/iP(hone|od)/i);
     browser.isIPad = _browserCheck(/iPad/i);
     browser.isSafari602 = _browserCheck(/Macintosh.*Mac OS X 10_8.*6\.0\.\d* Safari/i);
+    browser.isOSX = _browserCheck(/Mac OS X/i);
     browser.isEdge = _browserCheck(/\sedge\/\d+/i);
 
     var _isIETrident = browser.isIETrident = function(browserVersion) {
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -7,6 +7,7 @@ define([
     var _usePointerEvents = !_.isUndefined(window.PointerEvent);
     var _useTouchEvents = !_usePointerEvents && utils.isMobile();
     var _useMouseEvents = !_usePointerEvents && ! _useTouchEvents;
+    var _isOSXFirefox = utils.isFF() && utils.isOSX();
 
     function getCoord(e, c) {
         return /touch/.test(e.type) ? (e.originalEvent || e).changedTouches[0]['page' + c] : e['page' + c];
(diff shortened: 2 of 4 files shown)


cpeterso:
@jw-kaurand: does this click event problem only happen on OS X? Have you tested Firefox on Windows or Linux? Do you have a test case you can share?

dannyfinks:
@cpeterso the issue is only present in OS X Firefox and not Windows. This is a defect discovered within our internal QA Process. You can attempt to click on the poster image in the player displayed at the top of http://www.jwplayer.com/

jw-kaurand:
@cpeterso I created a test page to dsplay the issue we've run into a new branch.  You can find it here: https://github.com/jwplayer/jwplayer/tree/bugfix/ff-osx-test-page.  The relevant test page is test/manual/firefox-osx-mouseup.html

In order to manage user interface interaction in a consistent manner across browsers and interaction models (clicks, taps, pointerevents) we use a class, called UI, to manage event listening and report on how elements are interacted with.  In this relevant case, we don't listen for a click event, but rather listen for mouse down and mouse up events instead.  When we hear both, and depending on how far the mouse has moved, we interpret this as a click.  What we're seeing is that the listeners that are placed on the element that contains the flash object that plays the video are not receiving the mouse up event like we'd expect from mouse interaction.  Instead, we're just seeing mouse down and mouse move.  This is interpreted as a drag action, which in this case does nothing.  We don't see this issue on any other platform other than OSX Firefox with a Flash video playing the content.  The fix in this pull request to use a special case to handle click events on the jw-media container, since those are still received correctly despite the mouseup events not being available.

The test page allows you to see the events reported by the UI object of each item that has listeners assigned.  The relevant element in this case is the jw-media container which contains the flash video.  Try playing the video by clicking on the play button and then clicking on the video as its playing.  Only mousedown and mouseup events are reported in the console logs.  If the UI were reporting click events (the click events reported in the log are just from a standard click listener), then the player would toggle between play and pause states when clicking on the video area like it does in Firefox on Windows.

Let me know if there's anything I can explain better.

robwalch:
Making a few changes. Will reopen after they're checked in.

robwalch:
@jw-kaurand moved the click/firefox/osx logic to the ui mousedown handler. Also cleaned up rightclick handling which really does not need to use UI and feels a lot better triggered on the first down (mouse, pointer or touchstart).

PR #885, opened 2015-11-10
Save all sources before filtering in playlist

pajong:
Since playlist item sources were being filtered out if the type is different from bestType,
we could not find castable hls source when dash was chosen as bestType.
Saving the sources before filtering will allow us to search for castable sources in chromecast controller.
JW7-1751

### DIFF
--- a/src/js/playlist/playlist.js
+++ b/src/js/playlist/playlist.js
@@ -18,8 +18,9 @@ define([
 
         _.each(playlist, function(item) {
             item = _.extend({}, item);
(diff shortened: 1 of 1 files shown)


donato:
Looks good, this is actually a great improvement to the readability of this section of code. It separates the concerns of parsing sources from filtering sources. Big win! I can foresee the next iteration where the "formatSource" bits of code move into the playlist/source.js file, and this one get's even more simplified.

I will wait for the PR builder to finish before merging.

PR #886, opened 2015-11-10
Change oldProvider state when instream destorys

pajong:
When instream is created, we change model's mediaModel state.
When the ad errors out, the instream destroys and tries to start the video.
Since the mediaModel's state is buffering, it changes the state to playing, but the provider's state is still playing.
This causes the mediaModel's state to stay in buffering, causing buffer icon to display on mobile devices.
JW7-1739

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -260,6 +260,9 @@ define([
                         item.starttime = _oldpos;
                         _model.loadVideo(item);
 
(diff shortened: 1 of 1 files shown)


pajong:
test this please

donato:
This section of code is extremely volatile, I would be surprised if this doesn't cause many side-effects.

Try to reduce the scope by narrowing in on the specific path that the code runs when there is an error, but which doesn't run when it is working properly.

PR #887, opened 2015-11-10
Improved handling of Flash throttle events JW7-1603

robwalch:
Flash throttled state now displays "Click to run Flash" instead of the same error as a blocked plugin ("Flash plugin failed to load").

Flash will not send a throttled event where `event.state` is `"throttled"` or `"paused"` in audio players or players larger than 400x300. Events where `event.state` is `"resume"` are always sent to reset any previous throttled state. Flash will not send throttled events after the first time event with a position != 0. At this point we assume playback has begun and throttling can be ignored.

Throttle events are now forwarded to the API should developers want to listen to them. These events are useful in out tests as well to observe offscreen player behavior.

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/player/Player.as
+++ b/src/flash/com/longtailvideo/jwplayer/player/Player.as
@@ -1,5 +1,6 @@
 ﻿package com.longtailvideo.jwplayer.player {
 import com.longtailvideo.jwplayer.controller.Controller;
+import com.longtailvideo.jwplayer.events.MediaEvent;
 import com.longtailvideo.jwplayer.events.PlayerEvent;
 import com.longtailvideo.jwplayer.model.Model;
 import com.longtailvideo.jwplayer.model.PlayerConfig;
@@ -34,7 +35,6 @@ public class Player extends Sprite implements IPlayer {
 
         RootReference.init(this);
         this.addEventListener(Event.ADDED_TO_STAGE, stageReady);
-        this.addEventListener('throttle', flashThrottled);
 
         this.tabEnabled = false;
         this.tabChildren = false;
@@ -55,6 +55,7 @@ public class Player extends Sprite implements IPlayer {
 
     private function stageReady(e:Event):void {
         this.removeEventListener(Event.ADDED_TO_STAGE, stageReady);
+        this.addEventListener('throttle', flashThrottled);
         RootReference.init(this);
         _view.setupView();
     }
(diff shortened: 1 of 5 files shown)


src/js/controller/model.js
```diff
@@ -237,6 +242,8 @@ define([
             this.mediaModel.off();
             this.mediaModel = new MediaModel();
             this.set('mediaModel', this.mediaModel);
+            this.set('flashThrottle', undefined);
```
donato: Let's think about the flash to flash case

src/js/controller/model.js
```diff
@@ -72,6 +72,11 @@ define([
 
         function _videoEventHandler(type, data) {
             switch (type) {
+                case 'flashThrottle':
+                    this.set('flashThrottle', data);
```
donato: It's unclear at this point in the code what "data" actually is

src/flash/com/longtailvideo/jwplayer/player/Player.as
```diff
@@ -63,16 +64,36 @@ public class Player extends Sprite implements IPlayer {
...
         // in Chrome we only get 'throttle' and 'resume' for offscreen and power-save throttling
         var state:String = e['state'] as String;
-        if (state === 'resume') {
-            // After resume, we're past Chrome Power Save
-            // stop listening to the off-screen throttle events
-            this.removeEventListener('throttle', flashThrottled);
+
+        if (state !== 'resume') {
+            _model.removeEventListener(MediaEvent.JWPLAYER_MEDIA_TIME, removeThrottleListener);
+            // Ignore throttle events for players ignored by Chrome's heuristics
```
donato: We should be consistent about what level of abstraction to use inside a specific statement. I'd prefer to see this chrome stuff higher up, perhaps in providers/flash.js where we write 

```js
                    // We cannot rely on firefox to properly return results
                    if (utils.isChrome()) {
                        _swf.on('throttle', onThrottle, this);
                    }
```

src/flash/com/longtailvideo/jwplayer/player/Player.as
```diff
@@ -63,16 +64,36 @@ public class Player extends Sprite implements IPlayer {
...
+            _model.addEventListener(MediaEvent.JWPLAYER_MEDIA_TIME, removeThrottleListener);
         }
+
         SwfEventRouter.triggerJsEvent('throttle', {
             state: state
         });
     }
 
+    private function removeThrottleListener(e:MediaEvent):void {
+        // After a time event has been received, we're past Chrome Power Save
```
donato: Not sure this is true, I think we can receive off-screen throttle events, and then when the player is scrolled on screen we would see the actual swf-throttle occur.

Sample page would involve a flash player off screen, and then api.play() on that player after scrolling it into view.

src/js/controller/model.js
robwalch: No it's not. It's the event object.

src/flash/com/longtailvideo/jwplayer/player/Player.as
robwalch: Power saving will occur on the first rendered frame if at all. This time event is the closest thing we have to a consistent first frame.

PR #888, opened 2015-11-10
 Try/Catch on api.trigger instead of api.on

donato:
Our first approach to safely handling user callbacks was to wrap
the users code in a try/catch, and calling that our new safeCallback.
This solution does not work because when the user tries api.off('event', fn)
the fn no longer matches the function in api._events and so the off command doesn't
remove the bound event.

This fixes that by try/catching around when the callback is called in the trigger method.
A neat side-effect of this is that we don't need to overwrite the .on method at all, so
I went ahead and cleaned out some older debug code as well.

JW7-1764

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -44,28 +44,7 @@ define([
                 args = {};
             }
             args.type = type;
-            return Events.trigger.call(_this, type, args);
-        };
-
-        this.on = function(name, callback) {
-            if (!_.isFunction(callback)) {
-                if (_.isString(callback)) {
-                    throw new TypeError('eval callbacks depricated');
-                }
-                utils.log('Expected a function', name, callback);
-                return this;
-            }
-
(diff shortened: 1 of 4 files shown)


src/js/api/api.js
```diff
@@ -44,28 +44,7 @@ define([
...
-                var status = trycatch.tryCatch(callback, this, arguments);
-
-                if (status instanceof trycatch.Error) {
-                    utils.log('There was an error calling back an event handler for "'+
-                        name+'". Error: '+ status.message);
-                }
-            };
-
-            return Events.on.call(_this, name, safeCallback);
+            return Events.triggerSafe.call(_this, type, args);
```
robwalch: Please add the same jwplayer.debug check here that we use in `trycatch.tryCatch`. If `jwplayer.debug` is true then just use `Events.trigger` rather than `Events.triggerSafe`.

PR #893, opened 2015-11-11
Added a config variable to preserve autostart setting in model.setup(…

paulmandal:
…) when running in a mobile SDK

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -53,7 +53,8 @@ define([
             });
 
             // Mobile doesn't support autostart
-            if (utils.isMobile()) {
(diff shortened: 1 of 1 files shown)


robwalch:
Your comment on ln 56 breaks js hint too long. Run the build before submitting PRs.

paulmandal:
Sorry about that!

robwalch:
@paulmandal have you tested this change in a build to verify that autostart works on Android or iPhone? Another way you could get autostart to work, since you have your own provider is to map autostart to 'sdkAutostart' in your code, and then look for that in the config of your provider.

PR #904, opened 2015-11-16
Support JSON formatted playlist

pajong:
When rss parse fails, we want to see if the playlist is json formatted.
JW7-1206

### DIFF
--- a/src/js/playlist/loader.js
+++ b/src/js/playlist/loader.js
@@ -24,6 +24,7 @@ define([
             var status = utils.tryCatch(function() {
                 var childNodes = loadedEvent.responseXML.childNodes;
                 var rss = '';
+                var pl;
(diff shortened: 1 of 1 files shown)


src/js/playlist/loader.js
```diff
@@ -35,12 +36,21 @@ define([
                     rss = rss.nextSibling;
                 }
 
+                // If the response is not a valid RSS, check if it is a JSON
                 if (parsers.localName(rss) !== 'rss') {
-                    _playlistError('Not a valid RSS feed');
-                    return;
+                    try {
```
robwalch: I'd make this `===` and flip the case statement.

robwalch:
@pajong create tests - unit and squash. Work with @dannyfinks and @arup87 on feature file(s) and squashing them. I can assist with unit test development.

PR #905, opened 2015-11-16
Default preload setting matches browser or behaves as "none"

robwalch:
Revert the default preload behavior to match earlier players and do set video tag preload attribute unless a preload option is passed to setup.

Also disable "metadata" preload for Flash mp4 since we can only support "auto" in VideoMediaProvider.

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -14,7 +14,6 @@ define([
         displaydescription: true,
         mobilecontrols: false,
         repeat: false,
-        preload: 'metadata',
         castAvailable: false,
         skin: 'seven',
         stretching: stretchUtils.UNIFORM,
--- a/src/js/providers/flash.js
+++ b/src/js/providers/flash.js
@@ -104,8 +104,8 @@ define([
 
         _.extend(this, Events, {
                 init: function(item) {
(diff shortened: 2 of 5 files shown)


src/js/api/config.js
```diff
@@ -14,7 +14,7 @@ define([
         displaydescription: true,
         mobilecontrols: false,
         repeat: false,
-        preload: 'metadata',
+        preload: '',
```
pajong: Should we just do preload: 'none' for default?

src/js/providers/flash.js
```diff
@@ -105,7 +105,7 @@ define([
         _.extend(this, Events, {
                 init: function(item) {
                     // if preload is none or autostart is true, do nothing
-                    if (item.preload === 'none' || _playerConfig.autostart) {
```
pajong: We could leave this as it is if we set preload to none by default

src/js/providers/html5.js
```diff
@@ -415,7 +415,9 @@ define([
             _isAndroidHLS = _useAndroidHLS(_source);
             _videotag.src = _source.file;
             // set preload from the source, or default to metadata
```
pajong: This comment needs to be updated

PR #907, opened 2015-11-17
Support JSON formatted playlist

pajong:
When rss parse fails, we want to see if the playlist is json formatted.
Unit test is added to test this functionality.
JW7-1206


pajong:
Squash test is in PR: https://github.com/jwplayer/jwplayer-commercial/pull/1055

PR #911, opened 2015-11-20
Updates for Google IMA to act as a provider and use player UI

jw-kaurand:
Allows the instream adaptor to be assigned a provider, which is expected to output all appropriate playback information. Google IMA uses this to communicate with the player.
Styles added to support google IMA playback.

JW7-1530

Related to the PR at: https://github.com/jwplayer/jwplayer-ads-googima/pull/97

### DIFF
--- a/src/css/flags/ads.less
+++ b/src/css/flags/ads.less
@@ -39,6 +39,18 @@
     }
 }
 
+.jwplayer.jw-flag-ads-googleima {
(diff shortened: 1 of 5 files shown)


jw-kaurand:
Related to the PR at https://github.com/jwplayer/jwplayer-ads-googima/pull/97

PR #913, opened 2015-11-23
Expose reason for a play event

pajong:
We want to store a reason for a play event in API,
so that we can expose that in analytics.
0 - Unknown reason
1 - Interaction with JW Player UI
2 - Autostart from config
3 - Repeat
4 - External api
5 - Internal api (related)
6 - Internal api (playlist progression)
JW7-1645

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -36,8 +36,13 @@ define([
         this._ = _;
         this.Events = Events;
         this.version = version;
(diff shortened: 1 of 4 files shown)


pajong:
PR for related plugin: https://github.com/jwplayer/jwplayer-plugin-related/pull/29
PR for analytics: https://github.com/jwplayer/jwplayer-analytics/pull/157

PR #916, opened 2015-11-24
Fix mp4 quality switch in Firefox

robwalch:
Firefox is stuck buffering when switching qualities because our provider requires a 'canplay' event to fire for delayed seek to work. This event does not fire after changing sources in Firefox because the video tag goes into a paused state and does not fire 'canplay' until play() is called.

The regression in 7.2 comes from removing `_canPlayHandler()` from within `_loadedMetadataHandler` https://github.com/jwplayer/jwplayer/commit/eb6a6251f13f91226c454631595e9ffeac934bca.

JW7-1813

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -514,6 +514,10 @@ define([
                 }
             } else {
                 _delayedSeek = seekPos;
+                // Some browsers won't fire canplay event in a paused state (Firefox)
+                if (_videotag.paused) {
+                    _videotag.play();
+                }
             }
         };
 
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -14,6 +14,7 @@ define([
         _isMSIE = utils.isMSIE(),
         _isMobile = utils.isMobile(),
         _isSafari = utils.isSafari(),
+        _isFirefox = utils.isFF(),
         _isAndroid = utils.isAndroidNative(),
         _isIOS7 = utils.isIOS(7),
         _name = 'html5';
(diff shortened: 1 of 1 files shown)


dannyfinks:
test this please

src/js/providers/html5.js
```diff
@@ -514,6 +514,10 @@ define([
                 }
             } else {
                 _delayedSeek = seekPos;
+                // Some browsers won't fire canplay event in a paused state (Firefox)
+                if (_videotag.paused) {
```
donato: Okay, so if this affects Firefox, why is this change for all browsers?

robwalch: Branching on user agent is bad practice. In my testing this change resulted in quality change and midroll resumes that performed better in Chrome and iOS because it no longer depends on the round about "buffer full" event triggering play for selling to occur.

donato: Let's talk!

cpeterso:
@robwalch, do you consider this canplay behavior a Firefox bug? If so, I can ask @jyavenard to debug from the Firefox side.

robwalch:
Hi @cpeterso. Yes. We'll make a standalone test to duplicate the behavior. This fix is good for us until we address the larger issue of cleaning up delayed seeking.

cpeterso:
Sounds good. Thanks!

jyavenard:
Likely the same as what dashif player workaround:
https://github.com/Dash-Industry-Forum/dash.js/pull/818

Firefox tracking for canplay event not fired as per spec is https://bugzilla.mozilla.org/show_bug.cgi?id=1194624

donato:
We are going to gate this fix for Firefox only, and then do some performance enhancements for quality switching in v7.3.0

cpeterso:
This Firefox issue should be fixed in Firefox 45 by @jyavenard's fix for [Firefox bug 1194624](https://bugzilla.mozilla.org/show_bug.cgi?id=1194624).

robwalch:
@cpeterso Thanks Chris!

PR #922, opened 2015-11-25
Hotfix/air sdk readme

cjmyles:
Updated the Readme to illustrate that the grunt command won't work unless it knows where to find the AIR SDK


donato:
Thanks @cjmyles , can you try rebasing these into one commit, and removing the change from html5.js?

donato:
try
```sh
git rebase -i origin/master
```

PR #925, opened 2015-12-01
Remove JW Player style tab focus

pajong:
We want to remove JW Player style tab focus, and just use browser's style.
We also want to prevent focus by interaction (mouse click), and only focus with tab.
JW7-1793

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -60,8 +60,7 @@ define([
             _exitFullscreen,
             _elementSupportsFullscreen = false,
 
-            // Used to differentiate tab focus events from click events, because when
-            //  it is a click, the mouseDown event will occur immediately prior
+            // Used to differentiate tab focus events from click events
             _focusFromClick = false,
 
             _this = _.extend(this, Events);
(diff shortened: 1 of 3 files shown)


src/js/view/view.js
```diff
@@ -174,38 +170,17 @@ define([
...
 
-        function handleMouseDown() {
-            _focusFromClick = true;
-
-            // After a click it no longer has 'tab-focus'
-            _this.trigger(events.JWPLAYER_VIEW_TAB_FOCUS, {
-                hasFocus: false
-            });
+        function handleMouseDown(e) {
+            e.preventDefault();
```
pajong: This is to prevent the player from being focused with a mouse click - we only want to focus when we are tabbing into the player.

src/js/view/view.js
```diff
@@ -602,9 +577,7 @@ define([
             _controlsLayer.appendChild(_controlbar.element());
 
             _playerElement.onfocusin = handleFocus;
-            _playerElement.onfocusout = handleBlur;
             _playerElement.addEventListener('focus', handleFocus);
```
robwalch: Let's only listen to focusin for showing the controlbar:

` _playerElement.addEventListener('focusin', handleFocus);`

PR #927, opened 2015-12-02
Show playlist previous/next buttons in compactUI

pajong:
We still want to be able to navigate into previous or next playlist item in compact mode.
JW7-1649

### DIFF
--- a/src/css/flags/compact-player.less
+++ b/src/css/flags/compact-player.less
@@ -1,7 +1,5 @@
 .jw-flag-compact-player {
     .jw-icon-playlist,
-    .jw-icon-next,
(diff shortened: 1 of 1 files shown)


jw-kaurand:
Looks pretty good though we may want to modify the padding on the content.  The current level of padding between the next and prev buttons is somewhat small with a lot of padding on the outside so it looks somewhat awkward and could cause the button to be harder to tap accurately on mobile.  Might be something to look into as we test.

pajong:
@jw-kaurand I talked to @androoP , and we want to leave the padding as it is for now.
He said that modifying padding is in the plan for later.

PR #931, opened 2015-12-03
Sets volume in Flash for Google IMA when volume levels change

jw-kaurand:
Sets volume in Flash for Google IMA when volume levels change

JW7-1046


jw-kaurand:
Related to the PR here - https://github.com/jwplayer/jwplayer-ads-googima/pull/99

PR #932, opened 2015-12-03
Support HLS DVR on Safari/iOS, and handle android HLS duration

pajong:
Checked that HLS DVR works in iOS and Safari.
Handling android HLS getting incorrect duration before time event.
JW7-1560 JW7-1789

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -95,7 +95,7 @@ define([
 
                 pause: _pauseHandler,
                 //ratechange: _generalHandler,
-                //readystatechange: _generalHandler,
+                readystatechange: _readyStateHandler,
                 seeked: _seekedHandler,
                 //seeking: _seekingHandler,
                 //stalled: _stalledHandler,
@@ -160,6 +160,14 @@ define([
 
         function _clickHandler(evt) {
             _this.trigger('click', evt);
+        }
+
+        function _readyStateHandler() {
+            if (_videotag.readyState === 4) {
+                if (_delayedSeek > 0 && _duration > _delayedSeek) {
+                    _this.seek(_delayedSeek);
+                }
+            }
         }
 
         function _durationChangeHandler() {
@@ -502,14 +510,20 @@ define([
                 });
             }
 
+            if (_videotag.seekable && _videotag.seekable.length &&
+                _videotag.seekable.end(_videotag.seekable.length - 1) >= seekPos) {
+                // We can seek if the end seekable range is at or past the seek position
+                _canSeek = true;
+            }
+
             if (_canSeek) {
                 _delayedSeek = 0;
-                // handle readystate issue
-                var status = utils.tryCatch(function() {
+                // setting currentTime can throw an exception if video.readyState != 4
+                try {
                     _this.seeking = true;
                     _videotag.currentTime = seekPos;
-                });
-                if (status instanceof utils.Error) {
+                } catch(e) {
+                    _this.seeking = false;
                     _delayedSeek = seekPos;
                 }
             } else {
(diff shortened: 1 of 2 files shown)


robwalch:
@egreaves there are some good notes in this ticket with regards to TimeRanges. Not exactly related to the tracks work you are doing, but good to know.

src/js/providers/html5.js
```diff
@@ -116,7 +116,7 @@ define([
             // Current position
             _position,
             // Whether seeking is ready yet
-            _canSeek = false,
+            _loadInProgress = false,
```
robwalch: :disappointed:

src/js/providers/html5.js
```diff
@@ -220,7 +220,15 @@ define([
             _position = currentTime;
         }
 
-        function _setDuration(duration) {
+        function _updateDuration() {
+            var duration = _videotag.duration;
+            var end = _getSeekableEnd();
+            if (duration === Infinity && end) {
+                var seekableDuration = end - _videotag.seekable.start(0);
+                if (seekableDuration !== Infinity && seekableDuration > 120) {
```
robwalch: Add a comment here about how the player interprets negative duration as "DVR" mode.

src/js/providers/html5.js
```diff
@@ -228,12 +236,15 @@ define([
         }
 
         function _sendMetaEvent() {
+            if (_isAndroidHLS) {
+                return;
+            }
```
robwalch: Not ever sending the meta event could be a problem for integrations that depend on it. How about something like this? On Android I think _duration would still be 0.
```
function _sendMetaEvent() {
    _updateDuration();
    _this.trigger(events.JWPLAYER_MEDIA_META, {
        duration: _duration,
        height: _videotag.videoHeight,
        width: _videotag.videoWidth
    });
}
```

src/js/providers/html5.js
```diff
@@ -502,14 +526,19 @@ define([
                 });
             }
 
-            if (_canSeek) {
+            if (_getSeekableEnd() >= seekPos) {
+                // We can seek if the end seekable range is at or past the seek position
+                _loadInProgress = true;
+            }
+
+            if (_loadInProgress) {
```
robwalch: :sob: `_loadInProgress` does nothing or is it trying to deal with Android duration ranges coming and going?

src/js/providers/html5.js
```diff
@@ -502,14 +526,19 @@ define([
...
                 _delayedSeek = 0;
-                // handle readystate issue
-                var status = utils.tryCatch(function() {
+                // setting currentTime can throw an exception if video.readyState != 4
```
robwalch: `// setting currentTime can throw an invalid DOM state exception if the video is not ready`

src/js/providers/html5.js
pajong: I tried to handle maximum seekable range, but it looks like this is not necessary. I'll remove this and change the variable name back to canSeek, since loadInProgress is misleading.

src/js/providers/html5.js
```diff
@@ -220,7 +220,16 @@ define([
             _position = currentTime;
         }
 
-        function _setDuration(duration) {
+        function _updateDuration() {
+            var duration = _videotag.duration;
+            var end = _getSeekableEnd();
+            if (duration === Infinity && end) {
```
robwalch: We can set `_canSeek` to true here since there is a seekable TimeRange.

src/js/providers/html5.js
```diff
@@ -228,12 +237,19 @@ define([
         }
 
         function _sendMetaEvent() {
+            var duration = _videotag.duration;
+            if (_isAndroidHLS) {
```
robwalch: && duration === Infinity

PR #934, opened 2015-12-03
Add focus outline style for IE11

pajong:
Since IE does not display focus outline, we add that for IE.
JW7-1793

### DIFF
--- a/src/css/imports/jwplayerelement.less
+++ b/src/css/imports/jwplayerelement.less
@@ -38,8 +38,12 @@
         display: none;
     }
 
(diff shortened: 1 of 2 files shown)


pajong:
@robwalch This is a very bad way, but this is the only way to get IE to show outline around focus. I tried to copy the focus style of other elements, but it does not work. I tried to make the focus into the video element when playerElement gets the focus, but it will disable backwards tabbing, since focus on the div will keep on passing the focus to the video.

If we do not like this fix, we might not have a fix for IE.

Issue #908, opened 2015-11-18
Fetch and AJAX utils


robwalch:
Our AJAX utils are geared toward xml loading and not well suited to loading JSON (getting better #904). We should consider using/polyfilling Fetch, and make sure the util(s) allows us to specify the type of content we're requesting to optimize response handling.

donato:
var fetchPromise = utils.fetch('playlist.rss');
fetchPromise.then(handleRSS);

robwalch:
#### Fetch API Streaming Response Sample (Available in Chrome 43+)
https://googlechrome.github.io/samples/fetch-api/fetch-response-stream.html
```
        fetch(url).then(function (res) {
          var contentLength = res.headers.get('Content-Length');
          var reader = res.body.getReader();
          reader.read().then(function (result) {
          if (result.done) {
            // handle complete
            return;
          }
          // handle progress
          // cancel fetch
          reader.cancel();
        })
        .catch(function (e) {
          // handle error
        });
```
XHR pseudo-streaming `xhr.addEventListener('progress', ...)` could be a fallback for text only content.

robwalch:
Ajax refactor is complete. Fetch util could use more work but isn't required for anything at this point.

PR #935, opened 2015-12-04
Ajax util refactor and tests

robwalch:
This refactor fixes a couple of bugs in the ajax util and adds new arguments for handling things like timeouts and JSON parsing which we were doing externally.

Bugs addressed:
1. `_isCrossdomain` did not compare protocols. http/https cross protocol requests should be handled as crossdomain. This change adds `utils.crossdomain(uri)` to jwplayer.
2. The "Invalid XML" error was not being fired for invalid XML in most browsers. Now it will be fired in all browsers when the `requireValidXML` is set to `true`.

An options argument can be specified as the first parameter (or fourth for backwards support) with the following arguments:

options | description
------------ | -------------
url | The uri to request
withCredentials | Set to `true` to enable  `xhr.withCredentials` if available
retryWithoutCredentials | Set to `true` reattempt crossdomain requests that fail with `withCredentials`
timeout | The number of milliseconds after which the request should be aborted if not complete (defaults to one minute)
oncomplete | Success callback
onerror | Error callback
mimeType | If set this will be passed to `xhr.overrideMimeType`. Defaults to 'text/xml' when `responseType` : Is not set or empty.
requireValidXML | Defaults to false. If true, parser errors will result in an "Invalid XML" error.
responseType | If set assign the value to xhr.responseType if available. When set to "json", the parsed response will  `xhr.response`
plainText | Sets `responseType` to "text" if not set or empty

Example:

```
jwplayer().utils.ajax({
  url: "https://content.jwplatform.com/feeds/SkYJuQTJ.json",
  responseType: "json",
  oncomplete: function(xhr) {
    console.log(xhr.response);
  }
});
```

There are some options I have _not_ implemented but are common in other ajax utils. If there's a need we can add:
* `headers` : An object with key value pairs that should each be applied to `setRequestHeader`
* `beforeSend`: A function to execute (ex: `beforeSend(xhr)`) before `xhr.send`.
* `onprogess` : Get progress updates and handle data chunks
* `returnPromise` : return a Promise object instead of the XHR instance. Use `ajax().then().catch()` instead of oncomplete and onerror.


robwalch:
@donato @pablos1 please have a look. :eyes:

robwalch:
Closes #908

jw-kaurand:
Code looks pretty good, though captions seems to be exploding due to the access of xmlEvent.responseXML in captions.js _xmlReadHandler  Chrome doesn't like it when we access the variable since it causes a state error.

src/js/utils/ajax.js
```diff
@@ -1,140 +1,220 @@
...
+        try {
+            b.href = uri;
+            b.href = b.href; /* IE fix for relative urls */
+            return a.protocol + '//' + a.host !== b.protocol + '//' + b.host;
+        } catch(e) {}
+        return true;
+    };
+
+    var ajax = exports.ajax = function (url, completeCallback, errorCallback, args) {
+        if (_.isObject(url)) {
```
donato: We may want to get rid of this legacy fallback in the future, once we update vast.
Add a comment for it?

    // Legacy support for vast plugin, remove at later point

src/js/utils/ajax.js
```diff
@@ -1,140 +1,220 @@
...
         }
+        var xhr;
+        var options =_.extend({
+            xhr: null,
+            url: url,
+            withCredentials: false,
+            retryWithoutCredentials: false,
+            timeout: 60000,
+            timeoutId: -1,
+            oncomplete: completeCallback || noop,
```
donato: camelCase? onComplete, onError

robwalch: I kept these lowercase to mimic DOM event properties.

robwalch:
@jw-kaurand Captions are now requested expecting a mixed response. Thanks!

PR #937, opened 2015-12-04
enabled CC menu for HLS videos

egreaves:
Added CC support in the html5 provider for HLS videos.


src/js/controller/captions.js
```diff
@@ -213,14 +212,14 @@ define([
...
-        function _selectDefaultIndex() {
+        function _selectDefaultIndex(tracksAvailable) {
             var captionsMenuIndex = 0;
             var label = _model.get('captionLabel');
 
             // Because there is no explicit track for "Off"
             //  it is the implied zeroth track
             if (label === 'Off') {
-                _model.set('captionsIndex', 0);
+                _model.set('captionsIndex', captionsMenuIndex);
```
robwalch: Off always means `0`. The first index in the menu (or toggle) is "Off".

robwalch: Minified no difference in size. I prefer the hard 0 so there's no way moving this would introduce a bug.

src/js/controller/captions.js
```diff
@@ -237,7 +236,12 @@ define([
             }
 
             // set the index without the side effect of storing the Off label in _selectCaptions
-            _model.set('captionsIndex', captionsMenuIndex);
+            if(tracksAvailable) {
```
robwalch: Instead of passing `tracksAvailable = true` into this function, can you look at whether or not tracks were added by checking `_tracks.length`, or is that not the same?

src/js/controller/captions.js
egreaves: Good point - I concur.

src/js/controller/captions.js
egreaves: Yeah, that achieves the same outcome. I'll make the change.

src/js/providers/html5.js
```diff
@@ -87,7 +87,7 @@ define([
 
                 //play: _onPlayHandler, // play is attempted, but hasn't necessarily started
                 //loadstart: _generalHandler,
-                //loadeddata: _onLoadedData, // we have duration
+                loadeddata: _onLoadedData, // we have duration
```
robwalch: Can we update the comment for this event. What are we really looking for on this event? Tracks for one. Anything else?

src/js/providers/html5.js
```diff
@@ -782,6 +790,45 @@ define([
...
+            }
+        }
+
+        function _getTextTracks() {
+            return _textTracks;
+        }
+
+        function _setSubtitlesTrack(index) {
+            //index off by 1 because of 'off' option
+            if(_currentTextTrackIndex !== undefined) {
```
robwalch: Can you make the default `-1` like `_currentQuality`? I'd like us to stick to assigning variables a value when their declared.

src/js/providers/html5.js
egreaves: Just tracks (text, audio, metadata) at the moment.

PR #940, opened 2015-12-07
Fix issues with incorrect position for hls dvr

pajong:
For hls dvr on html5 provider, the position was not getting a correct value.
Getting relative position from seekable range to handle hls dvr positions.
JW7-1560

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -168,8 +168,8 @@ define([
                 return;
             }
 
-            _setBuffered(_getBuffer(), _position, _videotag.duration);
             _updateDuration();
+            _setBuffered(_getBuffer(), _position, _duration);
         }
 
         function _progressHandler() {
@@ -218,6 +218,9 @@ define([
         }
 
         function _setPosition(currentTime) {
+            if (_duration < 0) {
+                currentTime = -(_getSeekableEnd() - currentTime);
+            }
             _position = currentTime;
         }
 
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -168,8 +168,8 @@ define([
                 return;
             }
 
-            _setBuffered(_getBuffer(), _position, _videotag.duration);
             _updateDuration();
+            _setBuffered(_getBuffer(), _position, _duration);
```
pajong: Call update duration first so that we get correct duration value for hls dvr.

src/js/providers/html5.js
```diff
@@ -525,6 +528,10 @@ define([
                 return;
             }
 
+            if (seekPos < 0) {
+                seekPos += _videotag.seekable.start(0) + _videotag.seekable.end(0);
```
robwalch: I think you should use `_getSeekableStart() + _getSeekableEnd()` here.

src/js/providers/html5.js
```diff
@@ -436,6 +439,18 @@ define([
             }
         }
 
+        function _getSeekableStart() {
+            var index = _videotag.seekable ? _videotag.seekable.length : 0;
+            var start = Infinity;
+            if (index) {
+                index--;
+                start = _videotag.seekable.start(index);
```
robwalch: I don't think you need the above two lines. A length of 1 would still run once.

src/js/providers/html5.js
```diff
@@ -436,6 +439,18 @@ define([
...
+                while(index--) {
+                    start = Math.min(start, _videotag.seeakble.start(index));
+                }
+            }
```
robwalch: `return start;` buddy

PR #945, opened 2015-12-08
Support for audio tracks in HLS video

egreaves:
Added support for audio tracks in the html5 provider.


src/js/providers/html5.js
```diff
@@ -140,10 +140,10 @@ define([
             _beforecompleted = false,
 
             _fullscreenState = false,
-
-            // Video Text Tracks
+            // Video Tracks
             _textTracks,
-            _currentTextTrackIndex = -1;
+            _currentTextTrackIndex = -1,
+            _currentAudioTrackIndex = -1;
```
robwalch: Are these media tracks state properties being reset when loading new media?

src/js/providers/html5.js
```diff
@@ -837,7 +838,25 @@ define([
         this.getName = function() {
             return { name : _name };
         };
+        this.setCurrentAudioTrack = _setCurrentAudioTrack;
 
+        function _setAudioTracks(tracks) {
+            if(tracks && tracks.length > 0) {
+                _currentAudioTrackIndex = 0;
+                tracks[_currentAudioTrackIndex].enabled = true;
```
robwalch: Is it necessary to enable the first track? The media might define any track as the default track, and we should not change that when getting the list of available tracks.

robwalch: So instead look for the first enabled track, and set `_currentAudioTrackIndex ` to that.

src/js/view/controlbar.js
```diff
@@ -305,7 +305,7 @@ define([
             if (this.elements.playlist) {
                 this.elements.playlist.selectItem(itemIdx);
             }
-
+            //TO-DO: determine if this is the best spot for audio tracks setup
```
robwalch: Is this comment necessary?

src/js/providers/html5.js
egreaves: Not currently. Doing so after the loadeddata event fires should address this.

src/js/providers/html5.js
egreaves: If the media item always sets a default, then this isn't necessary.

src/js/providers/html5.js
robwalch: Only when there are tracks in the new source. You should clear these on init/load.

src/js/providers/html5.js
robwalch: It should or enable the first by default.

src/js/view/controlbar.js
egreaves: It isn't anymore; I'll remove it.

PR #952, opened 2015-12-10
Bugfix/update tracks after exit fullscreen ios

egreaves:
Updated the provider to listen for audio/text track change events. This keeps the model in sync when videos are in full screen mode on iOS devices where the native player controls are not available.

JW7-1884

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -39,6 +39,10 @@ define([
             this.setCaptionsList(captionsMenu);
             _selectDefaultIndex();
         }, this);
+        // Notified by provider if track changed while in fullscreen mode
+        _model.mediaController.on('subtitlesTrackChangedInFullScreen', function(e) {
+            _setCurrentIndex(e.index);
+        }, this);
 
         // Append data to subtitle tracks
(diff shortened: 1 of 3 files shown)


src/js/providers/html5.js
```diff
@@ -168,6 +168,8 @@ define([
         function _onLoadedData() {
             _setAudioTracks(_videotag.audioTracks);
             _setTextTracks(_videotag.textTracks);
+            _videotag.audioTracks.onchange = _audioTrackChangeHandler;
+            _videotag.textTracks.onchange = _textTrackChangeHandler;
```
robwalch: audioTracks textTracks are undefined in IE9

src/js/controller/captions.js
```diff
@@ -39,6 +39,10 @@ define([
             this.setCaptionsList(captionsMenu);
             _selectDefaultIndex();
         }, this);
+        // Notified by provider if track changed while in fullscreen mode
+        _model.mediaController.on('subtitlesTrackChangedInFullScreen', function(e) {
```
robwalch: This is no longer dispatched from the provider right?

PR #955, opened 2015-12-14
Send playReason with playAttempt event

pajong:
Instead of sending playReason with play event, we send it with playAttempt event, so that we know the playReason faster.
JW7-1860

### DIFF
--- a/src/js/events/change-state-event.js
+++ b/src/js/events/change-state-event.js
@@ -25,7 +25,7 @@ define([
                 oldstate: oldstate,
                 reason: model.mediaModel.get('state')
             };
-            // add reason for play of the event type is play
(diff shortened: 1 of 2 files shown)


src/js/events/change-state-event.js
```diff
@@ -25,10 +25,6 @@ define([
                 oldstate: oldstate,
                 reason: model.mediaModel.get('state')
             };
-            // add reason for play of the event type is play
-            if (eventType === 'play') {
-                evt.playReason = model.get('playReason');
-            }
```
robwalch: Why remove it from 'play'? Don't we want it in both events?

pajong: You are right, we should keep both. I'll update this.

Issue #919, opened 2015-11-24
Playlist thumbnails with special characters in URL are mutated


erikdonohoo:
All of our thumbnail URLs are broken because of this line...

https://github.com/jwplayer/jwplayer/blob/eacf851cafe407edcbc91d5117804ee2ff804161/src/js/view/preview.js#L24

We have urls to thumbnails in S3 that have a signature and key in them with special characters that are there by design, that `encodeUri()` destroys.  Any workaround for this?

donato:
Can you give us an example? Which special characters?

erikdonohoo:
A '%' symbol gets turned into '%25' even though the '%' was intentional as part of the S3 signature

donato:
@erikdonohoo if you have a specific example we can include it in our tests, otherwise we will just be guessing what you mean.

erikdonohoo:
Ok, for example:

`https://someurl.amazonaws.com/image.jpg?AWSAccessKeyId=SKIAJTKJN7H35T4LHI6Q&Expires=1449608325&Signature=Br6oJPAOa1Pf5ZDWv5l6DuxpOCE%3D`

becomes

`https://someurl.amazonaws.com/image.jpg?AWSAccessKeyId=SKIAJTKJN7H35T4LHI6Q&Expires=1449608325&Signature=Br6oJPAOa1Pf5ZDWv5l6DuxpOCE%253D`

michelegera:
+1

I’m working around this by calling `decodeURIComponent` on the thumbnail URL, but I agree with @erikdonohoo that it should be my responsibility to pass a well-formed URL and the player should not make any assumptions on it.

erikdonohoo:
@donato Any update on this?  Did you understand my example?

donato:
@erikdonohoo @michelegera - thanks for the examples, I totally agree.

donato:
This is targeted for jwplayer v7.3.0

PR #958, opened 2015-12-16
Stop encoding urls for preview images

donato:
This was originally done to allow a publisher who passed in non-encoded urls
to still work (specifically, they were passing a url which has a double-quote character).

This causes a problem for other users who have already encoded the url, since double-encoding
breaks pathing. A better solution is to require users to pass in a well-formed url.

https://github.com/jwplayer/jwplayer/issues/919
JW7-883


michelegera:
Thanks guys :clap:

PR #960, opened 2015-12-16
Display captions on iOS devices using tracks api

egreaves:
Display WebVTT captions natively on iOS devices. 

JW7-1828

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -458,7 +458,7 @@ define([
             }
 
             // if playlist item contains .vtt tracks, load them
-            if (_itemTracks) {
+            if (utils.isIOS() && _itemTracks) {
                 _setupSideloadedTracks();
             }
         }
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -144,8 +144,7 @@ define([
             _textTracks = null,
             _audioTracks = null,
             _currentTextTrackIndex = -1,
-            _currentAudioTrackIndex = -1,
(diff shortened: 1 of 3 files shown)


src/js/providers/html5.js
```diff
@@ -455,6 +456,41 @@ define([
...
+
+            // if playlist item contains .vtt tracks, load them
+            if (utils.isIOS() && _itemTracks) {
+                _setupSideloadedTracks();
+            }
+        }
+
+        function _setupSideloadedTracks() {
+            // cleanup dom
+            if (_videotag.childNodes) {
```
robwalch: I don't think you need this if statement. `while (_videotag.firstChild)` does it for you.

src/js/providers/html5.js
```diff
@@ -520,6 +557,7 @@ define([
             _source = _levels[_currentQuality];
             _position = item.starttime || 0;
             _duration = item.duration || 0;
+            _itemTracks = item.tracks;
```
robwalch: Look we're passing item `_setVideotagSource`. Why not use it? We should stop adding additional vars to this module's scope and if we really need to just store `_item`. There is no reason to make this a variable in this scope. You can do everything by passing references - functional and can't be tampered with outside of each functions' scope.

egreaves: I actually had it that way for that reason using ```arguments```. I agree with storing ```_item``` if we need to, but at this point, we don't. I'll remove ```_itemTracks```.

PR #963, opened 2015-12-17
Include play reason with beforePlay event

pajong:
Instead of firing play reason with playAttempt, fire it with beforePlay event to know the reason faster.
JW7-1860

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -295,7 +295,7 @@ define([
                 }
                 if (!_preplay) {
                     _preplay = true;
-                    _this.trigger(events.JWPLAYER_MEDIA_BEFOREPLAY, {});
+                    _this.trigger(events.JWPLAYER_MEDIA_BEFOREPLAY, {'playReason': _model.get('playReason')});
                     _preplay = false;
                     if (_interruptPlay) {
                         _interruptPlay = false;


src/js/events/change-state-event.js
```diff
@@ -25,10 +25,7 @@ define([
                 oldstate: oldstate,
                 reason: model.mediaModel.get('state')
             };
-            // add reason for play if the event type is play
-            if (eventType === 'play') {
-                evt.playReason = model.get('playReason');
-            }
+
```
robwalch: Why remove this?

src/js/controller/model.js
```diff
@@ -319,7 +319,7 @@ define([
         this.loadVideo = function(item) {
 
             this.mediaModel.set('playAttempt', true);
-            this.mediaController.trigger(events.JWPLAYER_MEDIA_PLAY_ATTEMPT, {'playReason': this.get('playReason')});
+            this.mediaController.trigger(events.JWPLAYER_MEDIA_PLAY_ATTEMPT, {});
```
robwalch: Why remove this?

src/js/controller/controller.js
```diff
@@ -274,10 +274,7 @@ define([
 
             function _play(meta) {
                 var status;
-
-                if (meta) {
-                    _model.set('playReason', meta.reason);
-                }
+                var playReason = meta ? meta.reason : 'unknown';
```
robwalch: Let's leave playReason in the other events. Model playReason could be set to unknown here.

PR #968, opened 2015-12-22
HTML5 support for ID3 metadata in HLS

egreaves:
JW7-1865


donato:
Can you rebase these into one commit?

PR #974, opened 2015-12-29
Only set crossorigin attribute on video tag when side loading tracks

robwalch:
When the crossorigin attribute is set in Chrome and Firefox, the video tag enforces CORS on video file equests. CORS is not enforced by default on video files, only on child text tracks linking to vtt files.

In Firefox when trying to load a video from a server without CORS headers with the attribute set, the console log states that the server hosting the video does not specify a cross-origin policy. There is no error logged in Chrome. For this reason we're only setting the attribute when loading text tracks. In 7.3 this will be limited to iOS where the crossorigin does not affect video.

JW7-1828
Fixes JW7-1922

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -164,7 +164,6 @@ define([
         // Enable AirPlay
         _videotag.setAttribute('x-webkit-airplay', 'allow');
         _videotag.setAttribute('webkit-playsinline', '');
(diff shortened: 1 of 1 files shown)


robwalch:
We should see if the video tag fires an error event after load() is called in Chrome for this CORS issue. It won't come up in 7.3, but when we make `<video crossorigin=""/>` configurable this will be helpful in dealing with user error when it comes to content server or player configuration.

PR #976, opened 2015-12-29
Map ID3 metadata for compatibility and ease of use

robwalch:
These changes make the ID3 metadata available in a way that is consistent with JW Player 6 and our current Flash implementation*, while also exposing ID3 key/data pairs and key: info/data collections. Here's an example:

```
{
  metadata: {
    title: "Who Knew - From Kiis 102.7 Los Angeles",    // JW6 Friendly Name
    TIT2: "Who Knew - From Kiis 102.7 Los Angeles",     // ID3 Key
    PRIV: {                                             // ID3 Key collection
      "com.apple.streaming.transportStreamTimestamp":{}  // ID3 PRIV info : [ArrayBuffer]
    },
    url: song_spot=x,                                   // JW6 WXXX friendly name
    WXXX: {                                             // ID3 Key collection
      "URL": "song_spot=x"                               // ID3 WXXX info : String
    },
    artist: "Pink",                                     // JW6 Friendly Name
    TPE1: "Pink"                                        // ID3 Key
  }
}
```
* One breaking change is that Flash 7.1-7.2 output a single key/info pair for some ID3 "PRIV" fields. This was incorrect since the actual data was omitted and multiple PRIV fields would overwrite each other. This will be addressed in 7.3 so that our Flash HLSID3 parser matches this output.

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -755,17 +755,15 @@ define([
                 TAL: 'album'
             };
             var id3Data = _.reduce(activeCues, function(data, cue) {
-                // These frienly name mapping provide compatibility with our Flash implementation
+                // These friendly names mapping provides compatibility with our Flash implementation prior to 7.3
                 if(friendlyNames.hasOwnProperty(cue.value.key)) {
                     data[friendlyNames[cue.value.key]] = cue.value.data;
                 }
-                /* Flatten the data by using the info value as the key if available
-                 * And provide a collection for keys with info values
-                 *   TLEN: 03:50
-                 *   artworkURL: http://domain.com/cover.jpg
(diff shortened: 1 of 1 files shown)


robwalch:
@egreaves I think this will work for everyone - those with existing implementations switching from Flash to HTML5 as well as those coming in wanting to get at general data or very specific data.

The only difference I've noticed in Safari from our Flash implementation is that we only get cues and cue change events when playing past the start of a segment in HLS. We do not get cues for the first segment or when seeking past the start of a segment. It's unfortunate, but that means when starting playback or seeking there could be up to a 10 second delay before any ID3 metadata is available. This appears to be how the spec was implemented in Safari and not something we can work around.

PR #980, opened 2016-01-04
Seeking to "Live" in DVR mode will seek to end minus 25 seconds

robwalch:
The last 25 seconds of a DVR stream in the time slider will show the "Live" tooltip. Clicking in that range or on the "Live" button to the right of the slider will seek to -25 seconds (the same time the stream would start playing at). The player will only seek forward in this case; If the position is closer to 0, the player will not seek back to -25 but continue playing in the "Live" range. The slider has been updates so that it will not render on mouse up at the position clicked if the DVR time is changed by the slider to meet these rules.

JW7-1744

Also applied changes to keyboard seeking, fixing DVR seeking with keyboard.
JW7-950

### DIFF
--- a/src/js/utils/parser.js
+++ b/src/js/utils/parser.js
@@ -142,9 +142,9 @@ define([
      * Determine the adaptive type
      */
     parser.adaptiveType = function(duration) {
-        if (duration !== -1) {
+        if (duration !== 0) {
             var MIN_DVR_DURATION = -120;
-            if(duration <= MIN_DVR_DURATION) {
+            if (duration <= MIN_DVR_DURATION) {
                 return 'DVR';
             }
             if (duration < 0 || duration === Infinity) {
--- a/src/js/view/components/slider.js
+++ b/src/js/view/components/slider.js
@@ -72,8 +72,10 @@ define([
                 }
             }
 
-            this.render(percentage);
-            this.update(percentage);
+            var updatedPercent = this.limit(percentage);
+            this.render(updatedPercent);
+            this.update(updatedPercent);
+
 
             return false;
         },
@@ -82,6 +84,10 @@ define([
             this.dragMove(evt);
         },
 
+        limit : function(percentage) {
+            // modules that extend Slider can set limits on the percentage (TimeSlider)
+            return percentage;
+        },
         update : function(percentage) {
             this.trigger('update', { percentage : percentage });
         },
--- a/src/js/view/components/timeslider.js
+++ b/src/js/view/components/timeslider.js
@@ -1,11 +1,12 @@
 define([
     'utils/underscore',
-    'view/components/slider',
     'utils/helpers',
+    'utils/constants',
+    'view/components/slider',
     'view/components/tooltip',
     'view/components/chapters.mixin',
     'view/components/thumbnails.mixin'
-], function(_, Slider, utils, Tooltip, ChaptersMixin, ThumbnailsMixin) {
+], function(_, utils, Constants, Slider, Tooltip, ChaptersMixin, ThumbnailsMixin) {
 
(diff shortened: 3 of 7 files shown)


src/js/view/components/timeslider.js
```diff
@@ -196,5 +207,7 @@ define([
 
     _.extend(TimeSlider.prototype, ChaptersMixin, ThumbnailsMixin);
 
+    TimeSlider.liveBuffer = liveBuffer;
```
robwalch: Use constants for liveBuffer

src/js/view/components/slider.js
```diff
@@ -72,8 +72,10 @@ define([
                 }
             }
 
-            this.render(percentage);
-            this.update(percentage);
+            // the update method can set limits and return an updated value (TimeSlider)
+            var updatedPercent = this.update(percentage);
```
robwalch: Create a new method in Slider/TimeSlider for capping percentage value before passing it to render and update.

src/js/view/components/timeslider.js
```diff
@@ -70,14 +72,22 @@ define([
...
         },
-        update: function(pct) {
+        update: function(percent) {
             if (this.activeCue && _.isNumber(this.activeCue.pct)) {
-                this.seekTo = this.activeCue.pct;
-            } else {
-                this.seekTo = pct;
+                percent = this.activeCue.pct;
+            }
+            var position = this.getSeekPos(percent);
```
robwalch: Return a percent, setting seekTo and including all this code in that function.

src/js/view/components/timeslider.js
```diff
@@ -133,20 +143,21 @@ define([
...
 
-        // These are new methods
-        performSeek : function () {
+        getSeekPos: function(percent) {
+            var currentPosition = this._model.get('position');
             var duration = this._model.get('duration');
             var adaptiveType = utils.adaptiveType(duration);
-            var position;
             if (adaptiveType === 'LIVE' || duration === 0) {
-                this._api.play();
```
robwalch: Remove `=== 'LIVE'`. If `duration === 0` we cannot seek. Continue to call `this._api.play();`.

src/js/utils/parser.js
```diff
@@ -142,9 +142,9 @@ define([
      * Determine the adaptive type
      */
     parser.adaptiveType = function(duration) {
-        if (duration !== -1) {
+        if (duration !== 0) {
```
donato: You may know this but, before this change, a duration of 0 returned undefined, now it returns "vod"


| Current | Value |
|---------|---|
| [-Infinity, -120]  | DVR |
| (-120, 0) | LIVE |
| [0, Infinity) | VOD |
| [Infinity, Infinity] | LIVE |

What do you think about this?

| Suggested | Value |
|---------|---|
| [-Infinity, 0)  | DVR |
| [0, 0] | Unknown |
| (0, Infinity) | VOD |
| [Infinity, Infinity] | LIVE |

donato:
Eric had an idea that after clicking live, we should we visually show the thumb as if we are fully caught-up, even though we have more buffered behind the scenes. Effectively locking the thumb to the right hand side.

src/js/utils/parser.js
robwalch: No. line 154 returns "VOD". The -1 behavior was for jw6. It should have been changed when providers were modified to return 0 instead of -1 as the default unknown duration.

robwalch:
Re: @esboyd That's a feature. These changes fix JW7-1744 and  JW7-950. I don't want to modify the sliders further to not display the amount of time we require to buffer as part of the fix. It would add a lot more complexity and potential for breaking changes.

PR #989, opened 2016-01-07
Update duration before triggering meta event

egreaves:
By updating the duration before triggering the `meta` event, duration will be displayed correctly on desktop and iOS. Android devices will show 'Live Broadcast' because duration isn't available for DVR streams.

JW7-1560

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -207,7 +207,6 @@ define([
             if (_isAndroidHLS && (_videotag.duration === Infinity) && (_videotag.currentTime === 0)) {
                 return;
             }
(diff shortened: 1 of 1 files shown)


robwalch:
@egreaves I modified this a little so that we can getDuration without any side-effects (seeking) and only do the seeking when the `_duration` property is updated. This allows us to get duration and send it in the meta event and then update the property and seek if appropriate.

Without that we'd potentially change the order of events involved in resuming playback after a midroll ad which is risky.

egreaves:
@robwalch that makes sense. Good catch.

Issue #965, opened 2015-12-18
JW7 support for RTMP source file via RSS playlist


pixelchutes:
I may have stumbled upon a bug with JW Player 7.2.4 and RTMP source files in RSS playlists...

I am facing the following error, however it's only occurring on RTMP URLs within an RSS playlist:
> Error loading media: File not found

The playlists in question adhere to the [RSS Playlist embed documentation](https://support.jwplayer.com/customer/portal/articles/1406722-rss-playlist-embed) for JW 7.

When monitoring the browser console, a `404 error` is being thrown. The _entire_ RTMP URL is being incorrectly fetched local to the website domain, and not from the CDN *(e.g. Ignoring the RTMP Application Path / streamer)*

In the 404 error log, it seems the double slash in the RTMP URL has been converted to a single slash...
e.g:
> domain.com/**rtmp:/**xyz.fcod.llnwd.net/a1234/b56/sub.folder/video.mp4

Beyond that, I'm not sure why it is even attempting to load over http(s) when the protocol is *explicitly* RTMP. I have tried both `flash` and `html5` mode to see if that made a difference, but it is resulting in the same error.

See forum thread for more detail: https://support.jwplayer.com/customer/en/portal/questions/15980772-does-jw7-still-support-rtmp-source-file-via-rss-playlist-

pixelchutes:
Alex helped to identify this was being caused by `type="video/mp4"` on the **jwplayer:source** node:
```
<jwplayer:source file="rtmp://example.com/vod/mp4:myVideo.mp4" type="video/mp4"/>
```

Removing the `type="video/mp4"` declaration did indeed get the video to play. However, what I don't understand is why this should be required? It seems like a workaround to a greater issue. Not only is this completely valid XML, it also works on JW5 and JW6.

Shouldn't we be able to declare our mp4 videos as `video/mp4` type in the RSS / XML markup without it breaking the actual RTMP attempt?

robwalch:
The player uses either the type or the file extension to determine which provider will load and play the media. Some types like mp4 have flash and html5 providers. RTMP is only supported in Flash and the player does not look at the protocol first to infer the type. It sees an explicit type of video/mp4 or the extension and expects a progressive download. Once down that route, it chokes on the url provided. It's likely the browser changing the url after setting video.src.

pixelchutes:
[...]

PR #994, opened 2016-01-11
Filter out other providers when primary is set

pajong:
Even when html5 is set to primary, we try to use all other providers when we try to filter the playlist items.
We want to only the provider that is set to primary when filtering playlist item.
JW7-1782

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -410,7 +410,14 @@ define([
 
             function _setPlaylist(p) {
                 var playlist = Playlist(p);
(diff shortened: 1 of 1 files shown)


robwalch:
If you filter out the provider needed to play the player can't setup. With these changes what happens on iOS when primary is set to "flash"?

robwalch:
@pajong The problem isn't how providers filter sources. This would actually break fallback on devices that can't use html5 or flash to play items.

The issue is that the rtmp url has an mp4 extension. See #965 . You need to check if the URL `isRTMP` before the html5 video canplay check which uses the mimetype we infer from the extension.

PR #995, opened 2016-01-11
Hide captions when displaying ads

egreaves:
JW7-1906

### DIFF
--- a/src/css/flags/ads.less
+++ b/src/css/flags/ads.less
@@ -31,10 +31,10 @@
         .jw-icon-tooltip.jw-icon-volume {
             display : none;
         }
-
     }
 
-    .jw-logo {
+    .jw-logo,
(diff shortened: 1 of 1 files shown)


robwalch:
@egreaves some ads display controls (skip button etc...) Does this change, using `.jw-flag-ads-hide-controls`, work for those ads? The `.jw-flag-ads` class includes any time that ads are active over the main program.

src/css/flags/ads.less
```diff
@@ -67,7 +67,7 @@
 }
 
 .jwplayer.jw-flag-ads-hide-controls {
```
robwalch: This will only work for ads that have their own controls. Linear video ads use jwplayer's controls, and would still display captions in this case. Please find a solution using the `jw-flag-ads` class.

PR #996, opened 2016-01-11
Fix reloading same playlist item for flash

pajong:
When stop is called, we need to reload the item even if it is the same as the stored item.

JW7-1821

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
+++ b/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
@@ -179,6 +179,7 @@ public class VideoMediaProvider extends MediaProvider {
         _buffered = 0;
         _starttime = 0;
         _offset = {time: 0, byte: 0};
+        _complete = true;
         super.stop();
     }
 


robwalch:
@pajong why not clear `_item` like in AdaptiveProvider?

pajong:
@robwalch We tried that before (not sure if you remember), but that breaks the provider. That's why we made the _complete flag last time.

robwalch:
Then wouldn't the change to AdaptiveProvider that we merged also break that provider?

robwalch:
How does it break the provider? Please include the test so that we can verify this and the change to HLS do not introduce new bugs.

Also what about the other providers in Flash? Don't they all require the same changes now too? Can we handle this at a higher level or in the MediaProvider base class?

pajong:
@robwalch It doesn't break in AdaptiveProvider. We were already setting _item to null on complete for AdaptiveProvider: https://github.com/jwplayer/jwplayer-commercial/blob/master/src/flash/com/longtailvideo/jwplayer/media/AdaptiveProvider.as#L190
I am not sure why flash provider is breaking if we set _item to null yet - I will look more into this.

Other providers in Flash does not check if stored item is equal to the new item being loaded, so they do not have this bug. AdaptiveProvider and VideoMediaProvider are the only two that have this issue.

PR #997, opened 2016-01-12
Check for rtmp and youtube before mimetype

pajong:
When mimetype is given, we do not check whether if the file is rtmp or youtubue.
This results html5 provider to try to play rtmp with mp4 extension.
Checking rtmp before looking at mimetype to prevent this from happening.
JW7-1782

### DIFF
--- a/src/js/playlist/source.js
+++ b/src/js/playlist/source.js
@@ -24,22 +24,21 @@ define([
         // normalize for odd strings
         _source.file = strings.trim('' + _source.file);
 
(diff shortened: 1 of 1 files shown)


junieth9:
Stop sending me this crap! I don't need it

robwalch:
@junieth9 just click "Unwatch" https://help.github.com/articles/unwatching-repositories/

robwalch:
This change fixes #965

PR #998, opened 2016-01-12
Hide preview for autostart on desktop

pajong:
We want to prevent preview image from flashing just before the playback begins when autostart is true.
We still want to show preview image when on complete state, so we are only hiding if the state is idle with autostart true.
JW7-1904

### DIFF
--- a/src/css/states/idle.less
+++ b/src/css/states/idle.less
@@ -3,6 +3,9 @@
 .jw-state-idle {
     .jw-preview {
         display: block;
+        &.jw-autostart {
+            display: none;
+        }
     }
 
     .jw-icon-display {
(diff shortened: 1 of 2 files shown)


robwalch:
@pajong can you achieve this without adding a flag by not calling `loadImage` when the playlistItem changes? We only want to show the preview image on complete for when the replay button is shown - so we also shouldn't call `loadImage` for every playlist item*.

*The only time we do want to call `loadImage`for the playlist item is when:
 - the media type is audio
 - the state is idle and autostart is false
 - there is no image so we want to clear `style.backgroundImage`
 - it's the last playlist item and the internal state is complete

Maybe instead of ` if (_.isString(img)) {` you could check these rules to only assign a poster style if appropriate.

PR #999, opened 2016-01-12
Trigger mediaType event from provider to notify model of audio-only HLS streams

egreaves:
The `adaptiveEventAudio` event notifies the flash provider that an HLS stream is audio-only. For audio only streams, the preview image disappears briefly while buffering before playback begins.  

In the html5 provider, `_videotag.videoTracks.length` will be 0 when HLS streams are audio-only. The `mediaType` event notifies the model before the provider's `load` function is called, so it's no longer necessary to call `sendMediaType` on load for HLS streams.

JW7-1914

### DIFF
--- a/src/js/providers/flash.js
+++ b/src/js/providers/flash.js
@@ -338,6 +338,10 @@ define([
                         this.trigger(events.JWPLAYER_MEDIA_ERROR, event);
                     }, this);
 
+                    _swf.on('adaptiveEventAudio', function() {
+                        this.trigger('mediaType', {mediaType: 'audio'});
+                    }, this);
+
                     if (flashThrottleTarget(_playerConfig)) {
                         _swf.on('throttle', function(e) {
                             clearTimeout(_flashBlockedTimeout);
--- a/src/js/providers/flash.js
+++ b/src/js/providers/flash.js
@@ -125,7 +125,10 @@ define([
                     _beforecompleted = false;
                     this.setState(states.LOADING);
                     _flashCommand('load', item);
-                    this.sendMediaType(item.sources);
+                    // HLS mediaType comes from the AdaptiveProvider
+                    if(item.sources.length && item.sources[0].type !== 'hls') {
+                        this.sendMediaType(item.sources);
+                    }
                 },
                 play: function() {
                     _flashCommand('play');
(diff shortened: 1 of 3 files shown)


src/js/providers/flash.js
```diff
@@ -338,6 +338,10 @@ define([
                         this.trigger(events.JWPLAYER_MEDIA_ERROR, event);
                     }, this);
 
+                    _swf.on('adaptiveEventAudio', function() {
+                        this.trigger('mediaType', {mediaType: 'audio'});
+                    }, this);
```
robwalch: If you send this, then I also expect you to send `{mediaType: 'video'}`. Every provider in Flash needs to be updated to output video or audio when a new item is played. Otherwise, in mixed playlist, after an audio only stream, 'mediaType' will be stuck on 'audio' and never update.

src/js/providers/html5.js
```diff
@@ -575,7 +579,10 @@ define([
             }
 
             _setLevels(item.sources);
-            this.sendMediaType(item.sources);
+
+            if(!item.sources.length || item.sources[0].type !== 'hls') {
```
robwalch: Why `!item.sources.length`? `sendMediaType` requires an array containing at least one item:
```
sendMediaType : function(levels) {
            var type = levels[0].type;
```

robwalch: I would expect this to say

```
if(item.sources.length && item.sources[0].type !== 'hls') {
```

src/js/providers/html5.js
```diff
@@ -1064,6 +1071,8 @@ define([
 
         this.getSubtitlesTrack = _getSubtitlesTrack;
 
+        this.getMediaType = _getMediaType;
```
robwalch: Why was this added? Don't add methods that are not used.

src/js/providers/html5.js
```diff
@@ -1122,6 +1131,15 @@ define([
         function _getSubtitlesTrack() {
             return _currentTextTrackIndex;
         }
+
+        function _setMediaType(videoTracks) {
+            _mediaType = videoTracks && videoTracks.length ? 'video' : 'audio';
```
robwalch: `mediaType` can be a local variable.

src/js/providers/flash.js
egreaves: `this.sendMediaType(item.sources);` is called in both providers' `load` function. This occurs before the `adaptiveEventAudio` event and always sends `{mediaType: 'video'}` for HLS streams.

src/js/providers/flash.js
```diff
@@ -338,6 +341,10 @@ define([
                         this.trigger(events.JWPLAYER_MEDIA_ERROR, event);
                     }, this);
 
+                    _swf.on('mediaType', function(e) {
+                        this.trigger('mediaType', {mediaType: e.mediaType});
```
robwalch: You should be able to just forward the event object `e` here, instead of creating a new object and copying the property we're looking for off it.

PR #1004, opened 2016-01-14
Change the behavior to load  preview image

pajong:
We want to load preview image only when:
1. Audio mode
2. State is idle with autostart false
3. State is idle with mobile
4. State is complete on last playlist item

JW7-1904

### DIFF
--- a/src/js/view/preview.js
+++ b/src/js/view/preview.js
@@ -1,35 +1,49 @@
 define([
     'utils/underscore',
     'utils/helpers'
-], function(_, util) {
+], function(_, utils) {
 
     var Preview = function(_model) {
         this.model = _model;
 
         this.model.on('change:playlistItem', onPlaylistItem, this);
+        this.model.on('change:mediaModel', onMediaModel, this);
     };
 
-    function loadOnComplete() {
-        if (this.model.get('state') === 'complete') {
+    function onChangeState(model, state) {
+        if (state === 'complete') {
+            this.loadImage(model, model.get('playlistItem'));
+        }
+    }
+
+    function onMediaModel() {
+        this.model.mediaModel.off('change:mediaType', onMediaType);
+        this.model.mediaModel.on('change:mediaType', onMediaType, this);
+    }
+
+    function onMediaType(mediaModel) {
+        var audio = (mediaModel.get('mediaType') === 'audio');
+        if (audio) {
             this.loadImage(this.model, this.model.get('playlistItem'));
         }
     }
 
     function onPlaylistItem(model, playlistItem) {
-        var audio = (this.model.mediaModel.get('mediaType') === 'audio');
-        var idleManualStart = (this.model.get('state') === 'idle') && (!this.model.get('autostart'));
-        var idleMobile = (this.model.get('state') === 'idle') && (util.isMobile());
+        var loadImage = (this.model.get('state') === 'idle') && (this.model.get('item') === 0) &&
+            (!this.model.get('autostart') || utils.isMobile());
         var endOfPlaylist = (this.model.get('item') === (this.model.get('playlist').length - 1));
 
-        this.model.off('change:state', loadOnComplete);
-        if (audio || idleManualStart || idleMobile) {
-            // load image if audio mode, or mobile/non-autostart with on idle state
+        this.model.off('change:state', onChangeState);
+        if (loadImage) {
+            // load image if mobile/non-autostart with on idle state, when the current item is the first playlist item
             this.loadImage(model, playlistItem);
         } else {
-            this.el.style.backgroundImage = '';
+            utils.style(this.el, {
+                backgroundImage: ''
+            });
             if (endOfPlaylist) {
                 // if the last playlist item, load image on complete
-                this.model.on('change:state', loadOnComplete, this);
+                this.model.on('change:state', onChangeState, this);
             }
         }
     }
@@ -46,9 +60,13 @@ define([
             var img = playlistItem.image;
 
             if (_.isString(img)) {
-                this.el.style.backgroundImage = 'url("' + img + '")';
+                utils.style(this.el, {
+                    backgroundImage: 'url("' + img + '")'
+                });
             } else {
-                this.el.style.backgroundImage = '';
+                utils.style(this.el, {
+                    backgroundImage: ''
+                });
             }
         },
         element : function() {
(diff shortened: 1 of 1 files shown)


src/js/view/preview.js
```diff
@@ -18,11 +26,21 @@ define([
         },
         loadImage: function(model, playlistItem) {
```
donato: loadImage, is no longer a function which loads images :)

To keep this code tight, we should separate out concerns so that each function has only one responsibility.

1. onPlaylistItem - does this item have an image? Should we load it?
1. loadImage - loads an image

The confusing part is that autostart **does need to load an image, it just doesn't want to display it on startup.

pajong:
Involving @robwalch , as he wanted to load the image only on complete, instead of loading and not displaying. Do we want to load on startup, or on complete?

src/js/view/preview.js
```diff
@@ -1,13 +1,39 @@
...
+        var idleManualStart = (this.model.get('state') === 'idle') && (!this.model.get('autostart'));
+        var idleMobile = (this.model.get('state') === 'idle') && (util.isMobile());
+        var endOfPlaylist = (this.model.get('item') === (this.model.get('playlist').length - 1));
+
+        this.model.off('change:state', loadOnComplete);
+        if (audio || idleManualStart || idleMobile) {
+            // load image if audio mode, or mobile/non-autostart with on idle state
+            this.loadImage(model, playlistItem);
+        } else {
+            this.el.style.backgroundImage = '';
```
robwalch: Why clear the style here? Can we avoid a style write?

src/js/view/preview.js
```diff
@@ -1,13 +1,39 @@
...
+    function loadOnComplete() {
+        if (this.model.get('state') === 'complete') {
+            this.loadImage(this.model, this.model.get('playlistItem'));
+        }
+    }
+
+    function onPlaylistItem(model, playlistItem) {
+        var audio = (this.model.mediaModel.get('mediaType') === 'audio');
+        var idleManualStart = (this.model.get('state') === 'idle') && (!this.model.get('autostart'));
+        var idleMobile = (this.model.get('state') === 'idle') && (util.isMobile());
```
robwalch: Can you combine idleMobile logic into idleManualStart?

src/js/view/preview.js
pajong: If we do not clear this, it would show the previous preview image just before buffering if the previous file was audio. Is there another way to handle this?

src/js/view/preview.js
pajong: Sure, I will do that.

robwalch: Also this is true for the start of any playlist item, you'll need to also check if this is the first playlistItem with 'item' === 0.

src/js/view/preview.js
```diff
@@ -1,13 +1,39 @@
...
     };
 
+    function loadOnComplete() {
+        if (this.model.get('state') === 'complete') {
+            this.loadImage(this.model, this.model.get('playlistItem'));
+        }
+    }
+
+    function onPlaylistItem(model, playlistItem) {
+        var audio = (this.model.mediaModel.get('mediaType') === 'audio');
```
robwalch: Has this been tested with a mixed playlist? I don't think 'mediaType' is set until after 'playlistItem' changes, so the value of 'mediaType' is not up-to-date with the current playlist item at this point.

src/js/view/preview.js
```diff
@@ -1,13 +1,39 @@
...
 
-        this.model.on('change:playlistItem', this.loadImage, this);
+        this.model.on('change:playlistItem', onPlaylistItem, this);
     };
 
+    function loadOnComplete() {
+        if (this.model.get('state') === 'complete') {
+            this.loadImage(this.model, this.model.get('playlistItem'));
+        }
+    }
```
robwalch: This would be better minified and written as:

```
    function onChangeState(model, state) {
        if (state === 'complete') {
            this.loadImage(model, model.get('playlistItem'));
        }
    }
```

src/js/view/preview.js
robwalch: You're right. Now that you have added `utils`, you should use `utils.style` because it will avoid unnecessary DOM writes when the style value is unchanged.

```
utils.style(this.el, {
  backgroundImage: ''
});
```

src/js/view/preview.js
robwalch: Be careful that preview is not loaded when the state is IDLE in-between playlist items. You probably need to have a second listener for 'change:mediaType' to deal with audio, as well as one for 'complete' which you have, and one only for 'idle start' that avoids idle in between playlist items (similar to the controlbar issue).

src/js/view/preview.js
```diff
@@ -1,13 +1,53 @@
...
+    'utils/underscore',
+    'utils/helpers'
+], function(_, utils) {
 
     var Preview = function(_model) {
         this.model = _model;
 
-        this.model.on('change:playlistItem', this.loadImage, this);
+        this.model.on('change:playlistItem', onPlaylistItem, this);
+        this.model.on('change:mediaModel', onMediaModel, this);
```
pajong: We need to do this because mediaModel changes when we get a new playlist item, so if we just listen to this.model.mediaModel.on('change:mediaType'), we do not get anything.

pajong:
Test this please

egreaves:
retest this please

src/js/view/preview.js
```diff
@@ -1,13 +1,53 @@
...
     };
 
+    function onChangeState(model, state) {
+        if (state === 'complete') {
+            this.loadImage(model, model.get('playlistItem'));
```
robwalch: What if `repeat` is true?

src/js/view/preview.js
```diff
@@ -1,13 +1,53 @@
...
+    function onMediaModel() {
+        this.model.mediaModel.off('change:mediaType', onMediaType);
+        this.model.mediaModel.on('change:mediaType', onMediaType, this);
+    }
+
+    function onMediaType(mediaModel) {
+        var audio = (mediaModel.get('mediaType') === 'audio');
+        if (audio) {
+            this.loadImage(this.model, this.model.get('playlistItem'));
+        }
```
robwalch: We want to alway change the backgroundImage property. When it's audio set it to the poster, otherwise clear it.

src/js/view/preview.js
```diff
@@ -20,9 +60,13 @@ define([
             var img = playlistItem.image;
```
robwalch: Just pass in img. Then you can reuse loadImage to clear the preview.

robwalch:
Please test this

PR #1007, opened 2016-01-14
Set mediaType from html5 provider only when source is HLS

egreaves:
JW7-1914


dannyfinks:
test this please

egreaves:
test this please

Issue #896, opened 2015-11-12
Determine if video tag requires interaction to start playback


robwalch:
Currently the player branches in different areas using `isMobile()` based on the user agent to prevent playback from being attempted until the user interacts with the player. (#893 model disables autostart, html5 provider buffering).

Instead, the player should have a check if interaction is required to start playback. This should be set through feature detection if possible or user agent (isMobile), and the setting should be able to be overridden in the config.

Once playback has started in a video tag, interaction is no longer required. So the branching that we do every time on mobile only really needs to happen once, after the video tag has played new sources do not need to wait and play() from the API will work for new sources.

The config option can have 3 settings, and will be set by the player by default based on what we can determine about the current platform:

`config.playRequiresInteraction` will be one of:
- `"always"` playback can autostart (*or begin by calling video.play())
- `"first"`  playback cannot autostart*, changes to "always" after first interaction
- `"none"` playback cannot autostart*

Since this behavior applies to the video tag, we can track the current state of this setting on the tag itself. Let's see if this or anything similar has been proposed to the w3c media group. 

Looked into what Apple is doing in WebKit. See `removeBehaviorsRestrictionsAfterFirstUserGesture` in [HTMLMediaElement](http://www.opensource.apple.com/source/WebCore/WebCore-7601.1.46.10/html/HTMLMediaElement.cpp)
http://trac.webkit.org/changeset/108831

http://trac.webkit.org/browser/trunk/LayoutTests/media/video-play-require-user-gesture.html?rev=108831

The video.play() now returns a promise, which gets rejected if a gesture is required to play:
https://webkit.org/blog/6784/new-video-policies-for-ios/

PR #1010, opened 2016-01-19
Update the instream provider's state when an ad is skipped

egreaves:
The html5 provider's load function handles setting the player's state to `loading` before a video plays. This is gated to non-mobile devices because user interaction is sometimes needed to play a video on mobile.

In the case of an ad pod on a mobile device - where each ad needs to autoplay - a skip event leaves the provider in a `playing` state which never gets changed before the next ad starts playing. This results in the model being stuck in a `buffering` state until the provider's state is changed when a subsequent ad completes or we switch back to the playlist item's provider.

Setting the state to `loading` in the instream provider ensures the model gets updated.

JW7-1843

### DIFF
--- a/src/js/controller/instream-html5.js
+++ b/src/js/controller/instream-html5.js
@@ -145,10 +145,6 @@ define([
                 provider.mute(_model.get('mute'));
 
                 _adModel.on('change:state', changeStateEvent, _this);
-            } else if(provider.state === 'playing') {
-                // state is still 'playing' after an ad is skipped. Need to update it here
-                // so the model gets the new play event when the next ad in an ad pod plays
-                provider.setState('loading');
             }
         }
 
--- a/src/js/controller/instream-html5.js
+++ b/src/js/controller/instream-html5.js
@@ -145,6 +145,10 @@ define([
                 provider.mute(_model.get('mute'));
 
                 _adModel.on('change:state', changeStateEvent, _this);
+            } else if(provider.state === 'playing') {
+                // state is still 'playing' after an ad is skipped. Need to update it here
+                // so the model gets the new play event when the next ad in an ad pod plays
+                provider.setState('loading');
             }
         }
 
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -305,7 +305,7 @@ define([
 
         function _playingHandler() {
             _this.setState(states.PLAYING);
-            if(!_videotag.attributes.__hasplayed) {
+            if(!_videotag.hasAttribute('__hasplayed')) {
                 _videotag.setAttribute('__hasplayed','');
             }
             _this.trigger(events.JWPLAYER_PROVIDER_FIRST_FRAME, {});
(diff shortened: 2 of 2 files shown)


egreaves:
Test this please

egreaves:
Test this please

egreaves:
retest this please

src/js/controller/instream-html5.js
```diff
@@ -145,6 +145,10 @@ define([
                 provider.mute(_model.get('mute'));
 
                 _adModel.on('change:state', changeStateEvent, _this);
+            } else if(provider.state === 'playing') {
```
pajong: Changing the provider state may cause other issues, although it will fix this particular bug.
Got a second opinion from Donato - When load is called, the state changes to 'loading'. We want to know why the state is playing in this bug, and fix it there.

pajong:
@robwalch Can we get a third opinion on this?

src/js/controller/instream-html5.js
egreaves: The state is still `playing` on mobile because the html5 provider does not change the state to `loading` on mobile devices:
```
if (!_isMobile) {
                // don't change state on mobile because a touch event may be required to start playback
                _this.setState(states.LOADING);
            }
```
In the case of ads, a touch event is not required to start playback of the next ad in an ad pod, so it's safe to assume the provider is in a `loading` state when `_checkProvider` is called in the `load` function in instream-html5.js. 

An alternative would be to set the provider's state to something like `skipped` or `stopped` in instream-adapter.js when an ad is skipped, but I did not want to introduce a new state to the provider if I could avoid it. The current states the provider cycles through for ad playback are:
* idle
* loading
* playing
* complete

When an ad plays to completion and the next ad starts, the provider goes through the following state changes: `playing > complete > loading > playing`. For skipped ads, it's stuck on `playing` even though we've stopped playback and started loading the next ad. The model's state - set to `buffering` between each ad in an ad pod - receives no notification that the new ad is loaded and has started playing, so the buffering icon displays until the provider's state changes to `complete` or the player switches from the instream provider back to the main provider that resumes normal media playback.

donato: @egreaves @robwalch - what do you think of this

```js
if (_isMobile && _isFirstPlay) {
    // do nothing, since we need to wait for a touch event
} else {
     _this.setState(states.LOADING);
}
```

robwalch: Shouldn't that be:
```
// do not change the state if we need to wait for a touch event
if (!_isMobile || !_isFirstPlay) {
     _this.setState(states.LOADING);
}
```

robwalch: @egreaves here's the tech-debt ticket https://github.com/jwplayer/jwplayer/issues/896

src/js/providers/html5.js
```diff
@@ -540,6 +543,7 @@ define([
                 _videotag.pause();
             }
             _currentQuality = -1;
+            _videotag.removeAttribute('__hasplayed');
```
robwalch: Let's leave the attribute there for the next play attempt or provider.

PR #1011, opened 2016-01-19
Expose drm used to decrypt the video

pajong:
This PR should only be merged if the following PR is merged: https://github.com/jwplayer/jwplayer-commercial/pull/1236

JW7-1887

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -176,6 +176,10 @@ define([
             return 'html5';
         };
 
+        this.getDrmUsed = function() {
+            return _controller._model.getVideo().drmUsed || null;
+        };
+
         this.load = function (toLoad) {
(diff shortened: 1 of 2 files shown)


robwalch:
@pajong please update unit tests when adding a new method:

https://github.com/jwplayer/jwplayer/blob/master/test/data/api-methods.js
https://github.com/jwplayer/jwplayer/blob/master/test/unit/api-test.js

I was expecting this test to fail when adding a new method that is not in api-methods:

robwalch:
Given that only shaka uses drm I don't think JW7-1887 merits adding a new method to the player. Events from the provider should include custom provider info, and that info should be forwarded through public events ('play' or 'meta') for tracking purposes. Fix jwplayer/jwplayer-commercial#1236 first then come back to this one.

PR #1012, opened 2016-01-19
Change preview load behavior

pajong:
Addressed feedback from: https://github.com/jwplayer/jwplayer/pull/1004/files

JW7-1904

### DIFF
--- a/src/js/view/preview.js
+++ b/src/js/view/preview.js
@@ -53,15 +53,13 @@ define([
             }
         },
         loadImage: function(model, img) {
+            var backgroundImage = '';
             if (_.isString(img)) {
-                utils.style(this.el, {
-                    backgroundImage: 'url("' + img + '")'
-                });
-            } else {
-                utils.style(this.el, {
-                    backgroundImage: ''
-                });
+                backgroundImage = 'url("' + img + '")';
             }
+            utils.style(this.el, {
+                backgroundImage: backgroundImage
+            });
         },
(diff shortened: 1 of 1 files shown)


src/js/view/preview.js
```diff
@@ -1,13 +1,53 @@
...
+        }
+    }
+
+    function onMediaModel() {
+        this.model.mediaModel.off('change:mediaType', onMediaType);
+        this.model.mediaModel.on('change:mediaType', onMediaType, this);
+    }
+
+    function onMediaType(mediaModel) {
+        var audio = (mediaModel.get('mediaType') === 'audio');
```
robwalch: Use the second parameter instead of calling `model.get`
```function onMediaType(mediaModel, mediaType) {```

File: src/js/view/preview.js [hunk 2]
```diff
@@ -1,13 +1,53 @@
...
+        var imageLoaded = (this.model.get('item') === 0) && (!this.model.get('autostart') || utils.isMobile());
+        if (audio) {
+            // only load the image if not already loaded
+            if (!imageLoaded) {
```
robwalch: You can leave out this check if you use `utils.style` to set `backgroundImage`.

pajong: @robwalch I added this check because you mentioned that image will be loaded twice for first playlist audio item - once from onPlaylistItem, once in here. Do we want to remove this check?

PR #1013, opened 2016-01-20
Hide controls when showing overlay

egreaves:
Hide `controls` via CSS when `sharing` or `related` overlays are showing. This circumvents the need to override the `controls` config value. JW7-1510

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "jwplayer",
-  "version": "7.3.0-beta.1",
+  "version": "7.3.0-beta.2",
   "description": "The JW Player is free for non-commerical use. To buy a license for commercial use, please visit \r http://www.jwplayer.com/pricing/",
   "repository": {
     "type": "git",
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "jwplayer",
-  "version": "7.3.0-beta.2",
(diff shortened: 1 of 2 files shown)


src/css/flags/overlay.less
```diff
@@ -5,4 +5,9 @@
     .jw-controls-right .jw-logo {
         display: none;
     }
+    &-sharing, &-related {
```
pajong: I think we only have sharing and related overlays, so we do not need to specify &-sharing, &-related

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "jwplayer",
-  "version": "7.3.0-beta.1",
+  "version": "7.3.0-beta.2",
```
pajong: We don't want to commit this together.

egreaves: Agreed.

src/css/flags/overlay.less
egreaves: This is currently true. I was initially going to do it that way, but after discussing with @robwalch, I concur that it's safer to limit the behavior to just the plugins for now, allowing something else to implement overlays slightly differently if the need arises. We could alternatively make it `&-plugin`, though, and apply that class to both.

pajong: Cool, I will merge sharing and related PRs. Just need this one to get the package.json commit out.

PR #1018, opened 2016-01-21
Do not trigger flash blocked on inactive tab

pajong:
When tab is inactive, do not send flash blocked error until the tab is in focus.
JW7-1915

### DIFF
--- a/src/js/providers/flash.js
+++ b/src/js/providers/flash.js
@@ -111,6 +111,23 @@ define([
             };
         }
 
+        function checkFlashBlocked() {
+            var _this = this;
+            _flashBlockedTimeout = setTimeout(function() {
+                Events.trigger.call(_this, 'flashBlocked');
+            }, 4000);
+            _swf.once('embedded', function() {
+                clearTimeout(_flashBlockedTimeout);
+                window.removeEventListener('focus', onFocus);
+                Events.trigger.call(_this, 'flashUnblocked');
+            }, this);
+        }
+
+        function onFocus() {
+            checkFlashBlocked();
+            window.removeEventListener('focus', onFocus);
+        }
+
         _.extend(this, Events, {
                 init: function(item) {
                     // if not preloading or autostart is true, do nothing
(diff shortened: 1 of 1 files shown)


src/js/providers/flash.js
```diff
@@ -206,13 +222,13 @@ define([
...
-                    _swf.once('embedded', function() {
-                        clearTimeout(_flashBlockedTimeout);
-                        Events.trigger.call(_this, 'flashUnblocked');
-                    }, this);
+
+                    // Wait until the window gets focus to see check flash is blocked
+                    if (document.hasFocus()) {
+                        checkFlashBlocked();
+                    } else {
+                        window.addEventListener('focus', onFocus);
```
robwalch: @pajong you have to be more careful when adding event listeners that you remove them more aggressively. Here's some scenarios you need to cover:

- This provider is removed
- Flash successfully sets up in a background tab (you need to test the same steps with "Run all plugin content" selected). This means on swf 'ready' or anywhere we also call `clearTimeout(_flashBlockedTimeout)`

src/js/providers/flash.js
```diff
@@ -111,6 +111,27 @@ define([
             };
         }
 
+        function checkFlashBlocked() {
+            var _this = this;
```
robwalch: @pajong because you are calling `checkFlashBlocked` from `onFocus`, the context is `window` not the provider instance.

You need to test this feature in 3 modes. Currently the throttled case works, but it's broken when blocking the plugin outright:

screen shot 2016-01-22 at 12 04 17 pm

src/js/providers/flash.js
```diff
@@ -111,6 +112,26 @@ define([
...
         }
 
+        function checkFlashBlocked() {
+            _flashBlockedTimeout = setTimeout(function() {
+                Events.trigger.call(_this, 'flashBlocked');
+            }, 4000);
+            _swf.once('embedded', function() {
+                removeBlockedCheck();
+                Events.trigger.call(_this, 'flashUnblocked');
+            }, this);
```
robwalch: `_this`

PR #1024, opened 2016-01-22
Fix youtube player control not showing on activity

pajong:
Youtube player does not receive mousemove event, because the child element is stealing the event.
We cannot make mousemove work, but we can add mouse over/out event so that control shows up when mouse moves in/out of the player.
JW7-924

### DIFF
--- a/src/js/view/clickhandler.js
+++ b/src/js/view/clickhandler.js
@@ -5,21 +5,24 @@ define([
     'utils/underscore'
 ], function(UI, events, Events, _) {
 
-    var ClickHandler = function(_model, _ele) {
+    var ClickHandler = function(_model, _ele, options) {
         var _display,
             _alternateClickHandler,
             _alternateDoubleClickHandler;
 
+        var _options = {enableDoubleTap: true, useMove: true};
         _.extend(this, Events);
 
         _display = _ele;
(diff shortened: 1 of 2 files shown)


src/js/view/view.js
```diff
@@ -522,7 +522,7 @@ define([
             var overlaysElement = _playerElement.getElementsByClassName('jw-overlays')[0];
             overlaysElement.addEventListener('mousemove', _userActivity);
 
-            _displayClickHandler = new ClickHandler(_model, _videoLayer);
+            _displayClickHandler = new ClickHandler(_model, _videoLayer, {useHover: true});
```
robwalch: Not used

src/js/view/clickhandler.js
```diff
@@ -5,21 +5,24 @@ define([
...
 
+        var _options = {enableDoubleTap: true, useMove: true};
         _.extend(this, Events);
 
         _display = _ele;
 
         this.element = function() { return _display; };
 
-        var userInteract = new UI(_display, {enableDoubleTap: true, useMove: true});
+        var userInteract = new UI(_display, _.extend(_options, options));
```
robwalch: options not used.

PR #1029, opened 2016-01-26
When testing provider priority, use relative comparisons

donato:
The previous tests were checking that the exact index of the provider in the list
of providers was the expected value. This doesn't scale well when adding
new providers, so instead we will ensure that they are in the correct ordering
relative to eachother.

### DIFF
--- a/test/unit/provider-test.js
+++ b/test/unit/provider-test.js
@@ -73,6 +73,15 @@ define([
         };
     }
 
+    function getKeyIndexes(arr) {
+        var o = {};
+        for (var i= 0; i < arr.length; i++) {
+            var name = getName(arr[i]);
+            o[name] = i;
+        }
+        return o;
+    }
+
     module('provider primary priority');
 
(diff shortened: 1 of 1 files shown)


test/unit/provider-test.js
```diff
@@ -80,18 +89,22 @@ define([
         assert.expect(4);
 
 		assert.ok(providerList.length >= 3, 'There are at least 3 providers listed');
-		assert.equal(getName(providerList[1]), 'html5', 'HTML5 is the number 1 provider');
-		assert.equal(getName(providerList[2]), 'flash', 'Flash is the number 2 provider');
-		assert.equal(getName(providerList[0]), 'youtube', 'YouTube is the number 3 provider');
+
+        var keys = getKeyIndexes(providerList);
+        console.log(keys.html5, keys.flash, keys.youtube);
```
pajong: Do we need this console?

PR #1033, opened 2016-01-28
Ensure meta event listener is on when there are no WebVTT captions

egreaves:
JW7-1984

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -17,12 +17,11 @@ define([
         // Listen for provider subtitle tracks
         //   ignoring provider "subtitlesTrackChanged" since index should be managed here
         _model.mediaController.on('subtitlesTracks', function(e) {
-            _model.mediaController.off('meta');
             if(! e.tracks.length) {
-                // If we don't get WebVTT captions, listen for metadata captions
-                _model.mediaController.on('meta',_metaHandler, this);
                 return;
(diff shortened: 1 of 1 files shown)


src/js/controller/captions.js
```diff
@@ -17,11 +17,12 @@ define([
         // Listen for provider subtitle tracks
         //   ignoring provider "subtitlesTrackChanged" since index should be managed here
         _model.mediaController.on('subtitlesTracks', function(e) {
+            _model.mediaController.off('meta');
```
robwalch: It would be good to add the function here in case other modules are listening for the meta event:

`_model.mediaController.off('meta', _metaHandler);`

robwalch:
@egreaves I'm still seeing issues when changing between playlist items in on [this page](http://player-pr-test-jenkins.longtailvideo.com/builds/722/archive/test/squash/bootstrap.html?primary=html5&edition=ads&feature=captions_streaming). If you go to #6 and then back to #4, CC no longer shows up on #4.

You might consider always listening for meta textdata events instead of this change, where it's only sometimes. Then ignore/remove the meta textdata track if subtitle tracks are present/added.

src/js/controller/captions.js
egreaves: Good point. I'll make that change.

PR #1036, opened 2016-01-28
Update dependencies

donato:


donato:
*** Re-open after we are finished with v7.3.0

PR #1043, opened 2016-02-01
Hide title when sharing/related overlay is visible

egreaves:
JW7-2005


pajong:
You forgot to have the ticket number in the commit message

egreaves:
Yeah, I realized that right after committing. Thanks for the reminder

PR #1045, opened 2016-02-02
Make a function to get media id from file path

pajong:
We want to be able to find media id based on playlist item file path, so that we can use that for analytics purposes.
JW7-1942

### DIFF
--- a/src/js/utils/parser.js
+++ b/src/js/utils/parser.js
@@ -54,6 +54,12 @@ define([
         return '';
     });
 
+    parser.getMediaId = function(url) {
+        var regex = /.*\/(?:manifests|videos)\/([a-zA-Z0-9]{8})[\.-].*/;
+        var mediaId = regex.exec(url);
+        return (mediaId && (mediaId.length === 2)) ? mediaId[1] : null;
+    };
+
     function containsParserErrors(childNodes) {
         return _.some(childNodes, function(node) {
             return node.nodeName === 'parsererror';
--- a/test/unit/parser-test.js
+++ b/test/unit/parser-test.js
@@ -67,6 +67,23 @@ define([
             'returns a directory url ending with a forward slash "'+ scriptPath +'"');
     });
 
+    test('parser.getMediaId', function(assert) {
+        var path = parser.getMediaId(null);
+        assert.equal(path, null, 'returns null when no file name is provided');
+
+        var mediaId = parser.getMediaId('/test/with/manifests/mediaid.js');
+        assert.equal(mediaId, null, 'returns null when media is not 8 characters long');
+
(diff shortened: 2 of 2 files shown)


src/js/utils/parser.js
```diff
@@ -54,6 +54,20 @@ define([
         return '';
     });
 
+    parser.getMediaId = function(url) {
+        var filenameRegex = /[^\/]+$/;
+        var mediaIdRegex = /[^.-]*/;
```
robwalch: Let's try to use one RegEx. For example this would match the video urls we know of (See examples in next comment):
```js
/.*\/(?:manifests|videos)\/([a-zA-Z0-9]{8})[\.-].*/
```

test/unit/parser-test.js
```diff
@@ -67,6 +67,20 @@ define([
...
+        assert.equal(path, null, 'returns null when no file name is provided');
+
+        var mediaId = parser.getMediaId('/test/with/short/mediaid.js');
+        assert.equal(mediaId, null, 'returns null when media is not 8 characters long');
+
+        mediaId = parser.getMediaId('/test/with/dash/mediaid1-dash.js');
+        assert.equal(mediaId, 'mediaid1', 'returns media id correctly with dash');
+
+        mediaId = parser.getMediaId('/test/with/mediaid2.js');
+        assert.equal(mediaId, 'mediaid2', 'returns media id correctly');
```
robwalch: Use real examples, with and without http(s):

//content.jwplatform.com/manifests/IMUjQRAB.m3u8
//content.jwplatform.com/videos/IMUjQRAB-injeKYZS.mp4
//content.jwplatform.com/videos/IMUjQRAB-rjiewRbX.ogg

test/unit/parser-test.js
```diff
@@ -67,6 +67,20 @@ define([
...
+        var path = parser.getMediaId(null);
+        assert.equal(path, null, 'returns null when no file name is provided');
+
+        var mediaId = parser.getMediaId('/test/with/short/mediaid.js');
+        assert.equal(mediaId, null, 'returns null when media is not 8 characters long');
+
+        mediaId = parser.getMediaId('https://content.jwplatform.com/videos/IMUjQRAB-injeKYZS.mp4');
+        assert.equal(mediaId, 'IMUjQRAB', 'returns media id correctly with dash');
+
+        mediaId = parser.getMediaId('http://jwpsrv.a.ssl.fastly.net/content/conversions/zWLy8Jer/videos/TwYS9seF-364765.mp4?token=0_56b131af_0x06f6f154c730a714e1707c208f6d5457fafcd496');
```
pajong: Add manifest test

PR #1046, opened 2016-02-02
Support HLS ID3 metadata in Edge

IsaacSchemm:
Since at least EdgeHTML 13.10586, Edge reads ID3 metadata from HLS streams, but the format exposed to JavaScript is different from that which Safari uses. This commit adds code to detect Edge's format and convert it to match Safari's before doing further processing.

The "if (array[i] >= 32)" statement is there because the data field of my cue points always starts with a 0x03 character. I think this signals a UTF-8 encoding for the text stream. I can't currently generate metadata in other encodings so I can't test this at the moment.

Edge will also sometimes expose more than one active cue point to JavaScript. This commit does not address that, so if two cue points are active with the same data field, JW Player may not use the more recent of the two.

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -775,6 +775,37 @@ define([
                 TAL: 'album'
             };
             var id3Data = _.reduce(activeCues, function(data, cue) {
+                if (!('value' in cue)) {
+                    // Cue is not in Safari's key/data format
+                    if ('data' in cue && cue.data instanceof ArrayBuffer) {
+                        // EdgeHTML 13.10586 cue point format - contains raw data in an ArrayBuffer.
+                        
+                        var oldCue = cue;
+                        var array = new Uint8Array(oldCue.data);
+                        
+                        cue = { value: { key: '', data: '' } };
+                        
+                        var i = 10;
+                        while (i < 14 && i < array.length) {
+                            if (array[i] === 0) {
+                                break;
+                            }
+                            cue.value.key += String.fromCharCode(array[i]);
+                            i++;
+                        }
+                        
+                        i = 20;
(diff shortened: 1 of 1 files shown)


robwalch:
@egreaves can you check that these updates don't break at runtime on older browsers? We might need to use `window.ArrayBuffer` for example to prevent runtime errors in IE9.

egreaves:
@robwalch will do

IsaacSchemm:
I've updated the branch to properly support both UTF-8 and UTF-16 by checking the byte at the start of the data field and using the appropriate decoding function.

robwalch:
Thanks Isaac! Do you have any test streams we could use verify UTF-8 and UTF-16 encoded metadata in HLS streams? Even just a few TS segments would allow us to setup tests to verify across iOS/Safari, Edge, Flash, and our upcoming HLS JS player in Chrome.

Thanks!

IsaacSchemm:
I can send you some .ts files. Is there any particular text strings you'd like to see encoded?

Note that I didn't add a decoder for ISO 8859-1 (byte == 0x00) -- I'm just decoding it as UTF-8. I'm not sure whether people would use it as ISO 8859-1, since according to the Wikipedia page on ID3, using UTF-8 was common practice before the 0x03 byte was introduced.

IsaacSchemm:
Here's a test VOD stream I put together with both UTF-8 and UTF-16 metadata:
http://www.lakora.us/media/playlist.m3u8
You might want to mirror that locally in case I have to change my website in the future for some reason.

robwalch:
Made a local mirror. Your PR was merged and released in 7.3.1. This sample will help us test across browsers and HLS implementations.

Thanks!

PR #1048, opened 2016-02-03
Use metadata source for determining embedded 608 captions position

egreaves:
This addresses captions rendering issues caused by the model's static position for Live DVR streams. 
`metadata[source]` will be used to determine the alignment position for Live DVR streams with a negative duration. 
JW7-1984


egreaves:
test this please

PR #1051, opened 2016-02-05
Only listen for changes to captions/subtitles tracks

egreaves:
Monitor changes to `subtitles` or `captions` textTracks. `metadata` textTracks are unaffected by user/api changes to the CC menu. There's also no need to reset the `_activeCuePosition` in `_textTrackChangeHandler ` since it pertains to `metadata` tracks only. 

JW7-2023

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -756,7 +756,8 @@ define([
             var _selectedTextTrackIndex = -1, i = 0;
             _activeCuePosition = -1;
             for (i; i < _videotag.textTracks.length; i++) {
-                if (_videotag.textTracks[i].mode === 'showing') {
+                if ((_videotag.textTracks[i].kind === 'subtitles' || _videotag.textTracks[i].kind === 'captions')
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -756,7 +756,8 @@ define([
             var _selectedTextTrackIndex = -1, i = 0;
             _activeCuePosition = -1;
             for (i; i < _videotag.textTracks.length; i++) {
-                if (_videotag.textTracks[i].mode === 'showing') {
+                if ((_videotag.textTracks[i].kind === 'subtitles' || _videotag.textTracks[i].kind === 'captions')
+                    && _videotag.textTracks[i].mode === 'showing') {
```
robwalch: Putting && at the start of the line introduced a jshint error breaking the build. Fix that and put the track _videotag.textTracks[i] into a var when you need to get it 3 or more times.

robwalch: You can use _.findIndex to get the selected text track index. And you only need to do it when a selected track is found.

PR #1057, opened 2016-02-17
Improvements to visual quality detection

egreaves:
`visualQuality.reason` is set to `'initial choice'` when the videoTag's source has changed or when the provider picks an initial quality. It's `'auto'` for HLS streams and `'manual'` otherwise.

JW7-1977

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -227,18 +227,19 @@ define([
         }
 
         function _checkVisualQuality() {
-            if (_visualQuality.level.width !== _videotag.videoWidth ||
-                _visualQuality.level.height !== _videotag.videoHeight) {
-                _visualQuality.level.width = _videotag.videoWidth;
-                _visualQuality.level.height = _videotag.videoHeight;
-                if (!_visualQuality.level.width || !_visualQuality.level.height) {
+            var level = _visualQuality.level;
+            if (level.width !== _videotag.videoWidth ||
+                level.height !== _videotag.videoHeight) {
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -227,18 +227,19 @@ define([
         }
 
         function _checkVisualQuality() {
-            if (_visualQuality.width !== _videotag.videoWidth || _visualQuality.height !== _videotag.videoHeight) {
-                _visualQuality.width = _videotag.videoWidth;
-                _visualQuality.height = _videotag.videoHeight;
-                if (!_visualQuality.width || !_visualQuality.height) {
```
robwalch: `var levels = ` to shorten this a bit:

src/js/providers/html5.js
```diff
@@ -477,13 +483,13 @@ define([
...
             _currentTextTrackIndex = -1;
             _activeCuePosition = -1;
-            _visualQuality = {
-                index: 0,
-                label: '',
-                width: 0,
-                height: 0,
-                bitrate: 0
-            };
+            if(!_visualQuality.reason) {
```
robwalch: `delete _visualQuality.reason` in init() of this will only run once.

src/js/providers/html5.js
robwalch: Also up top make sure to add levels to the initial definition of _visualQuality so we never get a 'levels undefined' error
`var _visualQuality = {
  levels: {}
},`

PR #1062, opened 2016-02-20
Display VTT natively when not using the Flash Provider

egreaves:
`captions.js` must wait on the `itemReady` event to determine which provider is being used. When not using flash, we want the browser to handle VTT display.

JW7-1651

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -14,6 +14,9 @@ define([
         // Listen for captions menu index changes from the view
         _model.on('change:captionsIndex', _captionsIndexHandler, this);
 
+        // Listen for item ready to determine which provider is in use
+        _model.on('itemReady', _itemReadyHandler, this);
+
         // Listen for provider subtitle tracks
         //   ignoring provider "subtitlesTrackChanged" since index should be managed here
         _model.mediaController.on('subtitlesTracks', _subtitlesTracksHandler, this);
@@ -138,8 +141,6 @@ define([
             _tracksById = {};
             _metaCuesByTextTime = {};
             _unknownCount = 0;
-
-            _model.on('itemReady', _itemReadyHandler, this);
         }
 
         function _itemReadyHandler(item) {
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -139,14 +139,13 @@ define([
             _metaCuesByTextTime = {};
             _unknownCount = 0;
 
-            // meta event listener may have been turned off in subtitlesTracks event
-            _model.mediaController.off('meta', _metaHandler);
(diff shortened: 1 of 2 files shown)


src/js/providers/html5.js
```diff
@@ -501,7 +501,7 @@ define([
             }
 
             // if playlist item contains .vtt tracks, load them
-            if (utils.isIOS() && item) {
+            if (item) {
```
robwalch: Since we know we need this to happen when the source is set, let's get rid of this if statement, making sure an item is always passed to `_setVideotagSource`.

src/js/controller/captions.js
```diff
@@ -132,18 +132,21 @@ define([
         }
 
         /** Listen to playlist item updates. **/
-        function _itemHandler(model, item) {
+        function _itemHandler(item) {
```
robwalch: This is still a change event callback so doesn't model need to be the first parameter? Now this method is setting `_item` to the model.

src/js/controller/captions.js
```diff
@@ -132,18 +132,21 @@ define([
...
             _item = item;
             _tracks = [];
             _tracksById = {};
             _metaCuesByTextTime = {};
             _unknownCount = 0;
 
-            // meta event listener may have been turned off in subtitlesTracks event
+            _model.on('itemReady', _itemReadyHandler, this);
```
robwalch: You can use the first param `model` here rather than `_model`, or register this listener once up top rather than again and again in each item event.

robwalch:
Let's reopen after all comments are addressed and 7.3.2 fixes are in.

src/js/controller/captions.js
```diff
@@ -138,12 +141,13 @@ define([
...
             _unknownCount = 0;
+        }
 
-            // meta event listener may have been turned off in subtitlesTracks event
+        function _itemReadyHandler(item) {
             _model.mediaController.off('meta', _metaHandler);
             _model.mediaController.off('subtitlesTracks', _subtitlesTracksHandler);
 
-            if (!utils.isIOS()) {
+            if (_model.get('provider').name === 'flash') {
```
robwalch: We really want all providers to implement a method that tells us if they support captions tracks, and if not continue here. Since currently the only provider with this kind of support is html5, this line should be changed to:
```js
if (_model.get('provider').name !== 'html5') {
```

src/js/providers/html5.js
```diff
@@ -480,6 +480,10 @@ define([
         }
 
         function _setVideotagSource(item) {
+            if(!item) {
+                return;
```
robwalch: Not good. If this happens we're in big trouble. I'd rather an exception get thrown so we can find the cause and fix it.

PR #1068, opened 2016-02-23
Use captions renderer when not rendering VTT captions in HTML5

egreaves:

Sideloaded VTT captions in HTML5 should be handled by the browser. Other types of captions (data containing cues, srt, etc) should be handled by the captions renderer. 

Fixes #
JW7-1651

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -147,19 +147,27 @@ define([
             _model.mediaController.off('meta', _metaHandler);
             _model.mediaController.off('subtitlesTracks', _subtitlesTracksHandler);
 
-            if (_model.get('provider').name !== 'html5') {
-                var tracks = item.tracks,
-                    track, kind, i;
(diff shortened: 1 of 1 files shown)


src/js/controller/captions.js
```diff
@@ -147,19 +147,27 @@ define([
...
-                        } else if (track.data) {
-                            _addTrack(track);
-                        }
+            var tracks = item.tracks,
+                track, kind, isVTT, i;
+            var isHTML5 = _model.get('provider').name === 'html5';
+
+            for (i = 0; i < tracks.length; i++) {
+                track = tracks[i];
+                isVTT = track.file && (/\.vtt$/i.test(track.file));
```
robwalch: That regex won't catch urls with ".webvtt" as an extension or url params ("caps.vtt?nonce=123"). Try this:
```js
/\.(?:web)?vtt(?:\?.*)?$/
```

Issue #1071, opened 2016-02-24
[7.3.2-production] Safari HLS playback issue with preload: none


pixelchutes:
### Expected Behavior

Playback of HLS media should work _automatically_ in Safari / OS X, even when using `preload: "none"`.

**NOTE:** This works as expected in `v7.2.4`, I believe this behavior was introduced in `7.3.0`

### Actual Behavior

Forever stuck in `buffering` state, playback requires manual intervention.
Removing `preload` directive allows for automatic playback...

**NOTE:** Only reproducible when HLS is not 1st item in playlist.

### Steps to reproduce

1. Load an HLS URL (in my case, over `https`) to JWPlayer `7.3.2` via `playerInstance.playlistItem(1)`;
    * This is in Safari 9.0.2 on OS X 10.11.2, where HLS is the 2nd item in the playlist.
2. Player enters `buffering` state, but playback never begins _(progress bar shows 00:00 of 00:00)_.
3. Controls show "pause" icon, clicking it results in `NaN:NaN`, like this: ![nan_safari](https://cloud.githubusercontent.com/assets/352182/13295181/d57ff162-dae4-11e5-95ec-32862d90a039.jpg)
4. Controls now show "play" icon, clicking it loads the HLS stream and playback begins. *(Manual intervention `playReason: "interaction"` required)*

### Code to reproduce
```html
<div id="player"></div>
<script>
    var demoPlaylist = [
        {
          title: 'First playlist item',
          file: 'https://content.jwplatform.com/videos/HkauGhRi-640.mp4',
          image: 'https://content.jwplatform.com/thumbs/HkauGhRi-640.jpg'
        },
        {
          title: 'HLS Video 1 (2nd playlist item)',
          description: 'This fails to start automatically in Safari / OS X',
          file: 'https://content.jwplatform.com/manifests/IMUjQRAB.m3u8',
          image: 'https://assets-jpcust.jwpsrv.com/thumbs/IMUjQRAB-720.jpg'
        }
    ];
    
    var playerInstance = jwplayer('player');
    playerInstance.setup({        
        playlist: demoPlaylist,
        preload: 'none', // Causes HLS to fail in Safari (if not first playlist item)        
        width: '100%',
        aspectratio: '16:9',
        visualplaylist: true,
        displaytitle: true,
        repeat: false        
    })
</script>
```
### Environment

- MacBook Pro
- OS X 10.11.2 / Safari 9.0.2
- JW Player 7.3.2

[...]

PR #1073, opened 2016-02-24
Fix playlist item loading with html5 provider

robwalch:

These changes remove browser specific logic from `_completeLoad` so that video.load() will always get called after the video source has changed, and buffer full is only sent for mobile when the hasplayed attribute is not yet present (#896).

Browsers require that video.load() be called after changing video.src when preload is set to none. We can determine that the src has changed and no data has been preloaded when the src changes or the loadstarted event has fired, and the meta and data events have not yet fired. These states are being tracked in an attribute on the video element called 'jw-loaded'. When the attribute is set to 'none' or 'started' video.load() must be called (#1071).

`_setupSideloadedTracks` is moved to provider.load so that texttracks are only created once per item change, not on changes to item source quality levels. This allows us to remove `_source`. 

JW7-2102 JW7-2103

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -1,19 +1,19 @@
 define([
     'utils/css',
     'utils/helpers',
+    'utils/dom',
     'utils/underscore',
     'events/events',
     'events/states',
     'providers/default',
     'utils/backbone.events'
-], function(cssUtils, utils, _, events, states, DefaultProvider, Events) {
+], function(cssUtils, utils, dom, _, events, states, DefaultProvider, Events) {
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -52,6 +51,12 @@ define([
         }
     }
 
+    function _removeChildren(element) {
```
donato: We have a util for this, called "emptyElement" in /utils/dom.js

donato:
please test this

jw-kaurand:
test this please

Issue #1075, opened 2016-02-25
Thumbnail tracks are side loaded in iOS


robwalch:
### Expected Behavior

Thumbnail tracks should not be loaded or displayed as captions in iOS.

### Actual Behavior

When a playlist item is loaded with thumbnail vtt files, the vtt file is sideloaded. Because "thumbnails" is not a known type of track, Safari treats it as captions

#### Steps to repro

Setup the 7.3.0-7.3.3 with thumbnail tracks in Safari or iOS.
https://content.jwplatform.com/feeds/QZt0Hg9v.json

robwalch:
@dannyfinks this can be duplicated in 7.3 using this config:

https://content.jwplatform.com/feeds/QZt0Hg9v.json

PR #1076, opened 2016-02-25
HTML5 media type and thumbnail tracks in Safari

### LINKED ISSUE #1052: Use videotag.videoWidth to determine mediaType in html5
If videotag.videoTracks might return 0 while while loading on iOS and is not supported at all in many browsers. 

videotag.videoWidth is widely supported. As long as there is an event we can depend on it having a valid width for video or be 0 for audio we can use it to determine mediaType more reliably than the current method:
https://github.com/jwplayer/jwplayer/blob/master/src/js/providers/html5.js#L1155

For visualQuality we now use the first time event to ensure that size is set. There are documented bugs in older browsers where size was still unknown on loadeddata or loadedmetadata https://bugzilla.mozilla.org/show_bug.cgi?id=926753

robwalch:
- Prevent mediaType "audio" events from being sent when videoTracks is 0 but there is video. Use videoWidth instead.
- Prevent thumbnail tracks from being side loaded and displayed as captions in Safari

Fixes #1052 #1075

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -550,8 +550,13 @@ define([
             for (var i = 0; i < tracks.length; i++) {
                 // only add .vtt tracks
                 if(tracks[i].file.indexOf('.vtt') === -1) {
-                    break;
+                    continue;
                 }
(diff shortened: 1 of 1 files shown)


egreaves:
test this please

PR #1083, opened 2016-02-26
Update how primary works

donato:
This will help the HLS JS implementation whose priority is between html5 and flash.

HTML5 Primary / Default
    [ Youtube, HTML5, HLS-JS, Flash, Dash]
Flash Primary
    [ Youtube, Flash, HTML5, HLS-JS, Dash]

JW7-2072


donato:
test this please

Issue #1067, opened 2016-02-23
Captions under controlbar with HLS audio stream


fdambrosio:
Hi, 
there's a problem with 7.3.0 with captions and audio HLS stream: the vtt captions slide under the control bar when the mouse cursor go outside the player screen.

robwalch:
Thanks @fdambrosio,

We're adding this scenario to our test cases and discussing solutions.

Can you provide a little more info on which devices and browsers you're seeing this issue on?

kylesessions:
I'm experiencing a similar issue that I believe might be related. I'm using version 7.3.2. The problem I'm seeing is that, when the video controls are not displayed, the captions are displayed too low in the video frame, and the bottom of them gets cut off:

![img_0013](https://cloud.githubusercontent.com/assets/11525617/13301520/2b4e2414-dafa-11e5-859b-86ebd0562300.PNG)

When video controls are displayed, the entire caption does display, but it displays behind the video controls:

![img_0012](https://cloud.githubusercontent.com/assets/11525617/13301536/422e0b90-dafa-11e5-999d-439a75657c31.PNG)

This is happening in Chrome and Safari on an iPad Mini running iOS 8.0.

Incidentally, when you start a video playing on an iPad, and then during the course of the video you switch to fullscreen mode, when you first enter fullscreen mode, the first two or three captions you see will be positioned correctly, before it reverts to displaying too low and getting cut off.

robwalch:
@kylesessions can you share url with a player that duplicates this issue? Where are you loading 7.3.2 from? Do you have any CSS transforms on the player or video tag?

Thanks!

esboyd:
@kylesessions can you throw up a page using the intro video and JW 7.3.2? Thanks!

kylesessions:
@robwalch and @esboyd Thanks for the assistance. I'm using a self-hosted player. I downloaded 7.3.2 via the `License Keys & Downloads` link in the dashboard of my account. I do not have any CSS transforms on the player or video tag. I came across a suggestion on stackoverflow earlier about adding `webkit-transform: scale(1, 1)` to either the video element itself, or the container element, so I tried that, but it didn't seem to solve the problem.

Test page is here: https://s3.amazonaws.com/fgs-public/vidTest/vidTest3.html

The test page has 7.3.2, but I wasn't sure what you meant by the intro video, so I just used the video and captions from the front page of the site I'm working on.

robwalch:
[...]

PR #1090, opened 2016-03-02
Change caption position for audio inactive

pajong:
When audio mode, we always show the control bar.
When inactive, we change the bottom of the captions to be lower since we do not have control bar for non-audio.
This applied to audio player too, resulting the captions to be displayed below the control bar.

JW7-2098

### DIFF
--- a/src/css/flags/media-audio.less
+++ b/src/css/flags/media-audio.less
@@ -7,6 +7,9 @@
         .jw-controlbar {
             display: table;
         }
+        .jw-captions {
+            bottom: 3em;
+        }
     }
 }
 .jw-flag-media-audio {


robwalch:
Fixes #1067 in dev (7.4)

robwalch:
@pajong not seeing this work here http://player-pr-test-jenkins.longtailvideo.com/builds/946/archive/test/squash/bootstrap.html?primary=html5&edition=ads&feature=hls_basic

robwalch:
These changes do not override the rule which lowers the captions during inactive playback:
 
https://github.com/jwplayer/jwplayer/commit/752d25363f894bde32a9df12acb7ff4bbcff8a0e#diff-ed27457a8bbbc341d5a3b1818aeb534dR11

PR #1092, opened 2016-03-03
Change caption position for audio inactive

pajong:
When audio mode, we always show the control bar.
When inactive, we change the bottom of the captions to be lower since we do not have control bar for non-audio.
This applied to audio player too, resulting the captions to be displayed below the controlbar.

JW7-2098

### DIFF
--- a/src/css/flags/media-audio.less
+++ b/src/css/flags/media-audio.less
@@ -8,6 +8,12 @@
             display: table;
         }
     }
+    // This has higher specificity to overwrite caption position when user inactive in playing state
+    &.jw-flag-user-inactive &.jw-state-playing {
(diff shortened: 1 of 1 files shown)


src/css/flags/media-audio.less
```diff
@@ -8,6 +8,12 @@
             display: table;
         }
     }
+    // This has higher specificity to overwrite caption position when user inactive in playing state
+    &.jw-flag-user-inactive &.jw-state-playing {
```
pajong: This needs to be this specific. Without the jw-state-playing, the user-inactive caption has the same specificity, which sometimes overwrites (tested on local without the state - works fine, tested on pr builder test - caption moves down when inactive).

PR #1093, opened 2016-03-03
Removing a space that should not be there

pajong:
JW7-2098

### DIFF
--- a/src/css/flags/media-audio.less
+++ b/src/css/flags/media-audio.less
@@ -9,7 +9,7 @@
         }
     }
     // This has higher specificity to overwrite caption position when user inactive in playing state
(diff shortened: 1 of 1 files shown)


src/css/flags/media-audio.less
```diff
@@ -9,7 +9,7 @@
         }
     }
     // This has higher specificity to overwrite caption position when user inactive in playing state
-    &.jw-flag-user-inactive &.jw-state-playing {
+    &.jw-flag-user-inactive&.jw-state-playing {
```
robwalch: I think you can get rid of the ampersand too. Did you try?

``
&.jw-flag-user-inactive.jw-state-playing {
```

pajong:
@robwalch You are right, I removed the &.

PR #1097, opened 2016-03-04
Remove callInternal from public API

robwalch:
`callInternal` is undocumented and has been deprecated. It's purpose was to give jwplayer 6 normalized access to html5 and flash api methods. It is no longer used or needed so removing it will help reduce library file size.
 
JW7-2039


donato:
From 222.9 kb to 221.9 kb, nice!

PR #1102, opened 2016-03-08
Improvements to file extension parsing

RikHeijdens:

This PR changes the way file extensions are parsed, this change is required to make Akamai tokenized urls work on Android.

Fixes
AB-717

### DIFF
--- a/src/js/utils/strings.js
+++ b/src/js/utils/strings.js
@@ -59,7 +59,7 @@ define([
 
         path = path.substring(path.lastIndexOf('.')).split('?')[0].split('#')[0];
         if (path.lastIndexOf('.') > -1) {
-          return path.substr(path.lastIndexOf('.') + 1, path.length).toLowerCase();
+            return path.substr(path.lastIndexOf('.') + 1, path.length).toLowerCase();
         }
     };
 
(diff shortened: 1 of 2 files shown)


src/js/utils/strings.js
```diff
@@ -57,7 +57,7 @@ define([
             return azureFormat;
         }
 
-        path = path.substring(path.lastIndexOf('/') + 1, path.length).split('?')[0].split('#')[0];
```
RikHeijdens: This change passes all the unit tests, but are there any URL's that will break because of this change?

robwalch:
We should use regular expressions here. @RikHeijdens please add an example of an Akamai tokenized URL to https://github.com/jwplayer/jwplayer/blob/master/test/unit/strings-test.js#L16

RikHeijdens:
@robwalch I've added a test case see: https://github.com/jwplayer/jwplayer/commit/aed0d27409f429c0ef9a6e444e76d0adbd9dc1a4

src/js/utils/strings.js
```diff
@@ -57,7 +57,7 @@ define([
...
+        path = path.substring(path.lastIndexOf('.')).split('?')[0].split('#')[0];
```
robwalch: What if there's a '.' in the search or hash portion of the url? `lastIndexOf('.')` should not be performed before splitting off the search `?` and hash `#`.

RikHeijdens: Good point, I missed that one.

PR #1129, opened 2016-04-12
Ignore control cases when parsing id3

pajong:
0x00 and 0x03 are control characters and should not be parsed as strings.
JW7-2148

### DIFF
--- a/src/js/utils/id3Parser.js
+++ b/src/js/utils/id3Parser.js
@@ -24,6 +24,10 @@ define([
         i = startingIndex || 0;
         while (i < len) {
             c = array[i++];
+            // If the character is 3 (END_OF_TEXT) or 0 (NULL) then skip it
+            if (c === 0x00 || c === 0x03) {
+                continue;
+            }
             switch (c >> 4) {
                 case 0: case 1: case 2: case 3: case 4: case 5: case 6: case 7:
                 // 0xxxxxxx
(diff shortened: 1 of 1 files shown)


src/js/utils/id3Parser.js
```diff
@@ -24,6 +24,9 @@ define([
         i = startingIndex || 0;
         while (i < len) {
             c = array[i++];
+            if (c === 0x00 || c === 0x03) {
```
egreaves: A comment here would make it clear why we're skipping these characters:
0x00 = NULL
0x03 = EOL (end of line)

src/js/utils/id3Parser.js
```diff
@@ -139,29 +143,29 @@ define([
...
                         }
-                    } else{
+                    } else {
                         var encoding = array[startPos];
                         if (encoding === 1 || encoding === 2) {
-                            cue.value.data = id3Parser.utf16BigEndianArrayToStr(array, startPos+1);
+                            cue.value.data = id3Parser.utf16BigEndianArrayToStr(array, startPos + 1);
                         } else {
-                            cue.value.data = id3Parser.utf8ArrayToStr(array, startPos+1);
+                            cue.value.data = id3Parser.utf8ArrayToStr(array, startPos + 1);
```
robwalch: @pajong what happened to the update where you do `startPos++`?

PR #1133, opened 2016-04-14
Show browser captions in audio player with poster image

egreaves:
Ensures browser captions are displayed when the player is in audio-only mode. 
Fixes #
JW7-2362

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -859,17 +859,15 @@ define([
         function _onMediaTypeChange(model, val) {
             var isAudioFile = (val ==='audio');
             var provider = _model.getVideo();
-            var isNotFlash = (provider && provider.getName().name.indexOf('flash') !== 0);
-            var previewElem = _playerElement.getElementsByClassName('jw-preview')[0];
-            var mediaElem = _playerElement.getElementsByClassName('jw-media')[0];
+            var isFlash = (provider && provider.getName().name.indexOf('flash') === 0);
 
             utils.toggleClass(_playerElement, 'jw-flag-media-audio', isAudioFile);
 
-            if(isAudioFile && isNotFlash) {
-                // Put the preview element before media in order to display browser captions
-                _playerElement.insertBefore(previewElem, mediaElem);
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -858,7 +858,19 @@ define([
 
         function _onMediaTypeChange(model, val) {
             var isAudioFile = (val ==='audio');
+            var provider = _model.getVideo();
+            var isNotFlash = (provider && provider.getName().name.indexOf('flash') !== 0);
```
donato: don't have variables with "not" in their name

src/js/view/view.js
```diff
@@ -858,7 +858,19 @@ define([
...
+            var previewElem = _playerElement.getElementsByClassName('jw-preview')[0];
+            var mediaElem = _playerElement.getElementsByClassName('jw-media')[0];
+
             utils.toggleClass(_playerElement, 'jw-flag-media-audio', isAudioFile);
+
+            if(isAudioFile && isNotFlash) {
```
donato: missing a space here, did you run jshint? If so, can you fix the .jshintrc file to prevent this type of thing?

src/js/view/view.js
```diff
@@ -858,7 +858,19 @@ define([
 
         function _onMediaTypeChange(model, val) {
             var isAudioFile = (val ==='audio');
+            var provider = _model.getVideo();
+            var isNotFlash = (provider && provider.getName().name.indexOf('flash') !== 0);
+            var previewElem = _playerElement.getElementsByClassName('jw-preview')[0];
```
donato: We want to avoid DOM searches when possible. Can you think of a better way to do lines 863 and 864?

robwalch: The preview element is already available in view as `_preview.el`.

src/js/view/view.js
robwalch: This is already repeated twice so make it a function
```
function isFlash() {
  var provider = _model.getVideo();
  return (provider && provider.getName().name.indexOf('flash') === 0);
```

PR #1139, opened 2016-04-19
Prevent double encoding of html entities in playlist items

idvl:

Prevent html entities in playlist item titles from being double-encoded by handlebars.

### DIFF
--- a/src/js/view/components/playlist.js
+++ b/src/js/view/components/playlist.js
@@ -46,7 +46,7 @@ define([
                 return {
                     active : (idx === selectedIndex),
                     label : (idx+1)+ '.',
-                    title : item.title
+                    title : utils.createElement(item.title).textContent
                 };
             });
             return PlaylistTemplate(newList);


donato:
@idvl - can you give an example of what this will fix?

idvl:
Of course, totally forgot it.

Using a playlist with item titles containing HTML entities will display the title of the first item correctly in the player, because it is set with `.innerHTML`. 

The visual playlist however displays double encoded entities, because of the handlebars template using `{{this.title}}` instead of `{{{this.title}}}` (which is totally right).

The patch prevents double encoding through handlebars without breaking the current functionalities.

An example of a playlist containing HTML entities: 
```
...
playlist: [
    {
        title: "A &#34;title&#34; &#8211; with entities",
        sources: [{ file: "http://content.jwplatform.com/videos/xAU6qr2H-DZ7jSYgM.mp4" }]
    },
    {
        title: "Part 2 &#62; something&#8230;",
        sources: [{ file: "http://content.jwplatform.com/videos/xAU6qr2H-DZ7jSYgM.mp4" }]
    }
],
...
```

donato:
Good catch!

This reminds me that we should probably do some safety checking in title.js, the updateText method.

This is a good start 👍

PR #1140, opened 2016-04-19
Hide extra chromecast icon

donato:
This uses Google's presentation-api, which as of chrome 49
will prevent the browser from adding a casting icon to the
video tag.

Courtesy @mounirlamouri
JW7-2403

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -173,6 +173,10 @@ define([
         _videotag = _videotag || document.createElement('video');
         _videotag.className = 'jw-video jw-reset';
 
+        // prevent browser from showing second cast icon, using presentation-api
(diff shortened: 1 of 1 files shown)


donato:
Waiting on testing of player casting.

mounirlamouri:
If I understand correctly, this will add the attribute all the time. Ideally, we should add the attribute if the user of the API has enabled the Cast SDK. Otherwise, hiding the icon will prevent users to do basic video flinging.

avayvod:
Yes, I think JWPlayer needs to check if cast.appid is set before adding the attribute.

donato:
New PR here: https://github.com/jwplayer/jwplayer/pull/1141

PR #1141, opened 2016-04-20
Hide extra chromecast icon

donato:
This uses Google's Remote Playback API, which as of Chrome 49
will prevent the browser from adding a casting icon to the
video tag.

JW7-2403

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -173,6 +173,12 @@ define([
         _videotag = _videotag || document.createElement('video');
         _videotag.className = 'jw-video jw-reset';
 
+        // prevent browser from showing second cast icon
+        // https://w3c.github.io/remote-playback/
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -173,6 +173,12 @@ define([
         _videotag = _videotag || document.createElement('video');
         _videotag.className = 'jw-video jw-reset';
 
+        // prevent browser from showing second cast icon, using presentation-api
```
avayvod: Thanks for fixing this so fast!

"disableRemotePlayback" attribute is a part of Remote Playback API, not Presentation API: https://w3c.github.io/remote-playback/

Happy to chat more about these two APIs and how they relate to each other :)

avayvod:
Other than comment, the change looks good to me! Thanks for fixing this!

donato:
Thanks @avayvod, I've updated the comment and commit message.

PR #1142, opened 2016-04-20
Setup sideloaded tracks after preroll or midroll

egreaves:
Sideloaded tracks are removed from the videotag when playing an ad. This ensures that they're re-added when playback resumes after a preroll or midroll ad.
Fixes #
JW7-2418

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -551,7 +551,8 @@ define([
             if (_isSDK || !canRenderNatively) {
                 return;
             }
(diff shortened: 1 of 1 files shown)


dannyfinks:
@egreaves is there still a chance with these changes that after a postroll plays and a second playlist item begins with captions, that the captions would not display?

egreaves:
@dannyfinks in that scenario, the first condition (`tracks !== _itemTracks`) would be true and the sideloaded tracks would be added to the videotag.

PR #1156, opened 2016-04-28
Hide captions when displaying ads

egreaves:
The `jw-captions` class hides captions by default. We need to target `jw-captions-enabled` in order to hide captions during ad playback. 
Fixes #
JW7-2456

### DIFF
--- a/src/css/flags/ads.less
+++ b/src/css/flags/ads.less
@@ -1,7 +1,12 @@
 .jwplayer.jw-flag-ads {
 
     .jw-preview,
-    .jw-dock {
+    .jw-dock,
+    .jw-logo,
+    .jw-captions.jw-captions-enabled {
+        display: none;
+    }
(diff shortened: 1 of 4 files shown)


src/css/flags/ads.less
```diff
@@ -33,10 +33,14 @@
         }
     }
 
-    .jw-logo,
-    .jw-captions, video::-webkit-media-text-track-container {
```
robwalch: Just add `.jw-captions-enabled` to `.jw-captions` instead of making a new block:
```css
.jw-logo, .jw-captions.jw-captions-enabled, video::-webkit-media-text-track-container {
```
That's all the `&` operator in less does, but making a new block adds it as another rule.

robwalch: You could also merge these items with the preview and dock block above.

PR #1158, opened 2016-04-28
Check if 2 bytes exist before changing to utf16

pajong:
We got null characters when we have 1 byte but tried to convert into character.
JW7-2442


src/js/utils/id3Parser.js
```diff
@@ -58,7 +58,7 @@ define([
         out = '';
         len = array.length;
```
robwalch: `lastDoubleByte =  array.length - 1;`

PR #1160, opened 2016-04-28
Fix for live Flash streams

robwalch:
In 7.3 [changes to positionHandler](https://github.com/jwplayer/jwplayer/commit/51e58ebf07bdb1682a960bee7bc83c709e54b8a9) resulted in us no longer sending time events for Flash FLV streams that do not emit metadata events with a duration. These are special streams handled by the NetStream class in the same provider that handles mp4 and flv files. These streams do not provide any duration information and have return an impossibly large value for bytesTotal.

It used to be that if the duration was unknown Flash would return -1 and the player would see this as "Live". The first position handler with some media buffered would send a "bufferFull" event and playback would start.

This change fixes the regression and brings the position handler up to speed with our other providers. No time event will be sent until position or duration is known. If the NetStream position advances but duration is still unknown we assume it is a live stream. Live streams have a duration of Infinity.  

JW7-2451

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
+++ b/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
@@ -239,11 +239,20 @@ public class VideoMediaProvider extends MediaProvider {
 
     /** Interval for the position progress **/
     protected function positionHandler():void {
+        var pos:Number = _stream.time;
         var duration:Number = _item.duration;
-        if (duration <= 0) {
+        if (duration <= 0 && pos === 0) {
+            // don't send time or buffer event until duration or position is known
             return;
         }
-        var position:Number = Math.round(Math.min(_stream.time, duration) * 1000) / 1000;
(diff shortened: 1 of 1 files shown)


src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as
```diff
@@ -239,11 +239,19 @@ public class VideoMediaProvider extends MediaProvider {
...
             return;
         }
-        var position:Number = Math.round(Math.min(_stream.time, duration) * 1000) / 1000;
+        if (duration > 0) {
+            position = Math.min(position, duration);
+        } else if (position > 0) {
+            // Live Stream
+            duration = _item.duration = -1;
+        }
+        position = Math.round(position * 1000) / 1000;
```
pajong: What is this for?

robwalch: Makes position precision accurate to the millisecond.

PR #1161, opened 2016-04-29
Style native captions in Chrome/Safari

egreaves:
Apply player config captions styles to native captions in Chrome and Safari.
Fixes #
JW7-2068

### DIFF
--- a/src/js/view/captionsrenderer.js
+++ b/src/js/view/captionsrenderer.js
@@ -148,7 +148,7 @@ define([
         }
 
         /** Constructor for the renderer. **/
-        this.setup = function(options) {
+        this.setup = function(playerElement, options) {
             _captionsWindow = document.createElement('div');
             _textContainer = document.createElement('span');
             _captionsWindow.className = 'jw-captions-window jw-reset';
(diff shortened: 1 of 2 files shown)


src/js/view/captionsrenderer.js
```diff
@@ -192,7 +193,57 @@ define([
...
+                textStyleText += ' text-shadow: ' + textStyle.textShadow + ';';
+            }
+            textStyleText += '}';
+            backdropStyleText += '}';
+
+
+
+            var style = document.createElement('style');
+            style.textContent = windowStyleText + textStyleText + backdropStyleText;
+            playerElement.appendChild(style);
```
robwalch: We have a util for this "utils.css":

```js
cssUtils.css('#' + playerElement.id + ' .jw-video::-webkit-media-text-track-display', styleObject);
```

Don't construct the style text yourself. This util does that for you.

src/js/view/captionsrenderer.js
```diff
@@ -148,7 +148,7 @@ define([
         }
 
         /** Constructor for the renderer. **/
-        this.setup = function(options) {
+        this.setup = function(playerElement, options) {
```
robwalch: Avoid padding around elements like this. Just pass in the playerId needed.

src/js/view/captionsrenderer.js
robwalch: Here's an example of `utils.css()` in use:

https://github.com/jwplayer/jwplayer/blob/master/src/js/view/view.js#L242

PR #1163, opened 2016-04-29
Add json feed attributes to model

pajong:
We used to only get playlist from json feed.
We want to get all other attributes as well as playlist when we get a json feed.
JW7-2299

### DIFF
--- a/src/js/controller/setup-steps.js
+++ b/src/js/controller/setup-steps.js
@@ -126,7 +126,9 @@ define([
         if (_.isString(playlist)) {
             _playlistLoader = new PlaylistLoader();
             _playlistLoader.on(events.JWPLAYER_PLAYLIST_LOADED, function(data) {
-                _model.set('playlist', data.playlist);
+                for (var key in data) {
+                    _model.set(key, data[key]);
+                }
                 resolve();
             });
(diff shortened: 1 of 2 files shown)


src/js/playlist/loader.js
```diff
@@ -45,16 +46,15 @@ define([
                         pl = JSON.parse(loadedEvent.responseText);
                         // If the response is not a JSON array, try to read playlist of the response
                         if (!_.isArray(pl)) {
-                            pl = pl.playlist;
+                            jsonObj = pl;
```
robwalch:
```jw
if (_.isArray(pl)) {
 jsonObj = { playlist: pl };
} else if (_.isArray(pl.playlist)) {
jsonObj = pl;
} else {
throw null;
}
``

File: src/js/controller/setup-steps.js [hunk 2]
```diff
@@ -126,7 +126,9 @@ define([
         if (_.isString(playlist)) {
             _playlistLoader = new PlaylistLoader();
             _playlistLoader.on(events.JWPLAYER_PLAYLIST_LOADED, function(data) {
-                _model.set('playlist', data.playlist);
+                for (var key in data) {
+                    _model.set(key, data[key]);
+                }
```
robwalch: Too risky. As a work-around. Copy data.feedid is present. We don't want any other fields from this data.

PR #1164, opened 2016-04-29
Set cross origin when setting text tracks

pajong:
We were setting cors attribute too early, resulting the cors attribute to be there when redirect happens.
This caused the video src to have cors issue.
Setting cors when all data has been loaded, and when there are text tracks.
JW7-2460

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -198,6 +198,9 @@ define([
             }
             _setAudioTracks(_videotag.audioTracks);
             _setTextTracks(_videotag.textTracks);
+            if (_videotag.textTracks.length) {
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -198,6 +198,9 @@ define([
             }
             _setAudioTracks(_videotag.audioTracks);
             _setTextTracks(_videotag.textTracks);
+            if (_videotag.textTracks.length) {
```
robwalch: Put this in `_setSubtitlesTrack` right before setting `mode = "showing"`, if and only if the textTrack has a crossorigin src attribute (`utils.crossdomain(texttrack.src)`).

Clear the attribute in `_setVideotagSource`. And on stop?

PR #1166, opened 2016-04-29
Set cross origin when setting text tracks

pajong:
We were setting cors attribute too early, resulting the cors attribute to be there when redirect happens.
This caused the video src to have cors issue.
Setting cors when we actually start showing tracks.
JW7-2460

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -518,7 +518,8 @@ define([
             _currentAudioTrackIndex = -1;
             _currentTextTrackIndex = -1;
             _activeCuePosition = -1;
-            if(!_visualQuality.reason) {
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -1164,10 +1160,10 @@ define([
                    track.mode = 'disabled';
                 });
             }
-            if(index > 0 && index <= _textTracks.length) {
+            if (index > 0 && index <= _textTracks.length) {
                 _currentTextTrackIndex = index - 1;
+                _videotag.setAttribute('crossorigin', 'anonymous');
```
pajong: Tried to set it only when the tracks have cors issue, but track.src is undefined - textTracks does not have src in it.

PR #1168, opened 2016-04-29
Add json feed attributes to model

pajong:
We used to only get playlist from json feed.
We want to get feedid attribute as well as playlist when we get a json feed.
JW7-2299

### DIFF
--- a/src/js/playlist/loader.js
+++ b/src/js/playlist/loader.js
@@ -22,7 +22,7 @@ define([
             var status = utils.tryCatch(function() {
                 var childNodes = loadedEvent.responseXML ? loadedEvent.responseXML.childNodes : null;
                 var rss = '';
-                var pl;
+                var jsonObj;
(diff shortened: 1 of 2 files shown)


src/js/playlist/loader.js
```diff
@@ -35,26 +35,29 @@ define([
...
+                        var pl = JSON.parse(loadedEvent.responseText);
                         // If the response is not a JSON array, try to read playlist of the response
-                        if (!_.isArray(pl)) {
-                            pl = pl.playlist;
+                        if (_.isArray(pl)) {
+                            jsonObj = { playlist: pl };
+                        } else if (_.isArray(pl.playlist)) {
+                            jsonObj = pl;
+                        } else {
+                            _playlistError('Not a valid RSS/JSON feed');
```
robwalch: throw or use function with error text

robwalch:
test this please

PR #1169, opened 2016-04-29
Download fonts as font file instead of base64 data

jw-kaurand:
Fonts are no longer loaded as base64 strings.  Instead, basic JWPlayer are added as a dependency for view.js while @font-face definitions are required separately so we can load woff or ttf files after we know where to load them from.  Require is used so webpack creates a single package that loads the fonts when appropriate.  Adds a check for window.webpackJsonpjwplayer to see if we're running it in phantomjs because it causes errors despite it working like we want it to.

JW7-2448

### DIFF
--- a/src/css/imports/icons.less
+++ b/src/css/imports/icons.less
@@ -1,13 +1,5 @@
 @import "mixins";
 
-@font-face {
-    font-family: 'jw-icons';
-    src: url('../../../assets/fonts/jw-icons.woff') format('woff'),
-         url('../../../assets/fonts/jw-icons.ttf') format('truetype');
-    font-weight: normal;
-    font-style: normal;
-}
-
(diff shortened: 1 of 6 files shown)


src/css/imports/icons.less
```diff
@@ -1,13 +1,5 @@
 @import "mixins";
 
-@font-face {
-    font-family: 'jw-icons';
-    src: url('../../../assets/fonts/jw-icons.woff') format('woff'),
-         url('../../../assets/fonts/jw-icons.ttf') format('truetype');
-    font-weight: normal;
-    font-style: normal;
```
robwalch: revert this so we just require jwplayer.less

jw-kaurand:
Closed in favor of https://github.com/jwplayer/jwplayer/pull/1170

Issue #1128, opened 2016-04-12
Captions not showing by default on iPhone


kylesessions:
### Expected Behavior
When instantiating a player object with `jwplayer.playlist[n].tracks[n].default` set to `true`, the captions from `tracks[n]` should be displayed when the video is played.

### Actual Behavior

On iPhones, when the video is played, no captions are displayed until the user uses the iPhone's full-screen video UI to activate the desired captions track.

### Steps to reproduce

1. On an iPhone running iOS 9.x, visit www.omgyes.com/de/ (note: it is important to visit the German version, as captions are supposed to be turned off by default in English, but on in other site languages)
2. Tap the play icon in the top section of the page, beneath the text `... auf der Grundlage aktueller Forschung.` Sometimes the video won't load and the player will display the `!` triangle icon. In this case, simply reload the page and try again.
3. When the video is playing, observe no subtitles/captions. Then use the phone's UI to activate the `Deutsch` caption track and observe the captions appear.

### Environment

- iPhone 5/6
- iOS 9.x
- JW Player version 7.3.2, URL: www.omgyes.com/de/

robwalch:
@egreaves please investigate.

egreaves:
@kylesessions Thanks for pointing this out. This is caused by the captions tracks not being loaded before the player attempts to set the default index. We will be addressing this as part of the captions improvements slated for 7.5.

robwalch:
Let's get this in our next patch.

fdambrosio:
Hi,
I'm testing 7.5.0 beta 1 with audio captions. 
Now it works and the issue seems to be solved, but on iPhone player starts in fullscreen mode and caption text is cutted and aligned on the left.

test: http://www.fdambrosio.net/test/player7/audio-750.html

iPhone 5S
iOS 9
jwplayer 7.5.0 beta 1

egreaves:
Closing this issue since it was addressed in 7.4.1. We'll track @fdambrosio's comment in this more recent issue: to https://github.com/jwplayer/jwplayer/issues/1324

PR #1171, opened 2016-05-02
Honor default VTT track in player config

egreaves:
Addresses https://github.com/jwplayer/jwplayer/issues/1128.
Fixes #
JW7-2463

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -162,7 +162,6 @@ define([
             _textTracks = null,
             _audioTracks = null,
             _currentTextTrackIndex = -1,
-            _defaultTextTrackIndex = -1,
             _currentAudioTrackIndex = -1,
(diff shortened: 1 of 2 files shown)


src/js/providers/html5.js
```diff
@@ -1145,7 +1150,11 @@ define([
             }
             _addTracksListener(tracks, 'change', _textTrackChangeHandler);
             if (_textTracks && _textTracks.length) {
-                _this.trigger('subtitlesTracks', { tracks: _textTracks });
+                var subtitlesTracks =  { tracks: _textTracks };
+                if (_defaultTextTrackIndex > -1) {
```
robwalch: This if statement is not needed, since if `_defaultTextTrackIndex` is `-1` then  `defaultIndex` will be set to `0`. This is the default value in captions.js > `_selectDefaultIndex`.

PR #1178, opened 2016-05-03
Fix the cursor does not auto-hide under HTML5 fullscreen mode

### LINKED ISSUE #1174: Cursor won't disappear under fullscreen mode
### Expected Behavior

Under HTML5 fullscreen mode, the cursor disappears when the user is inactive.

### Actual Behavior

- The cursor remains.
- On some of my machines, the control bar even remains as well and sometimes blinks.
I am not sure whether it's caused by `Internet Download Manager` or not.
I revise some codes in `ui.js` to fix this but it looks really stupid.
```javascript
function moveHandler(evt){
    // IMD triggers mousemove without actual mouse moving under fullscreen mode
    // hence this stupid check is added
    var _startX_new = getCoord(evt, 'X');
    var _startY_new = getCoord(evt, 'Y');
    if (_startX !== _startX_new || _startY !== _startY_new) {
        _startX = _startX_new;
        _startY = _startY_new;
        if (_useMouseEvents || (_usePointerEvents && evt.pointerType !== 'touch')) {
            triggerEvent(events.touchEvents.MOVE, evt);
        }
    }
}
```

### Steps to reproduce

1. Play a video using HTML5.
2. Click on the fullscreen button on the control bar.
3. Do not move the mouse for a while and the cursor remains on the screen.

### Environment

- Windows 7 SP1 x64
- `Google Chrome 51.0.2704.29` and `IE11` are tested
- URL: https://www.jwplayer.com the video on the homepage

jfcherng:

Fix the cursor does not auto-hide under HTML5 fullscreen mode.

Fixes #1174

### DIFF
--- a/src/css/flags/fullscreen.less
+++ b/src/css/flags/fullscreen.less
@@ -10,7 +10,7 @@
         margin: 0;
         position: fixed;
 
-        &.jw-flag-user-inactive .jw-media {
+        &.jw-flag-user-inactive, &.jw-flag-user-inactive .jw-media {
             cursor: none;
             -webkit-cursor-visibility: auto-hide;
         }
--- a/src/css/flags/fullscreen.less
+++ b/src/css/flags/fullscreen.less
@@ -10,7 +10,7 @@
         margin: 0;
         position: fixed;
 
-        &.jw-flag-user-inactive {
+        &.jw-flag-user-inactive .jw-media {
             cursor: none;
             -webkit-cursor-visibility: auto-hide;
         }


robwalch:
Thanks @jfcherng. We'll get this into the following patch.

src/css/flags/fullscreen.less
```diff
@@ -10,7 +10,7 @@
         margin: 0;
         position: fixed;
 
-        &.jw-flag-user-inactive {
+        &.jw-flag-user-inactive .jw-media {
```
robwalch: Media may not fill the player container, so it may be safer to target both player and media containers:
```css
&.jw-flag-user-inactive, &.jw-flag-user-inactive .jw-media {
```

@jfcherng would this still work?

jfcherng: Yes, it works for sure.

PR #1184, opened 2016-05-10
Hiding cursor when the user is inactive

arup87:
–– Added cursor hide CSS to the user-inactive.less file
–– Removed cursor hide CSS from fullscreen.less
Fixes #
JW7-1601


robwalch:
@arup87 I still see the finger pointer in Chrome after the controlbar is hidden.

robwalch:
Oh nevermind. It's OK after a refresh.

PR #1187, opened 2016-05-11
Added custom property for the injected stylesheet

nithin-raj:
Added id as "jwplayer-styles" to injected style sheet which will provide the user on identifying the injected style easily.

### DIFF
--- a/src/js/utils/css.js
+++ b/src/js/utils/css.js
@@ -9,7 +9,7 @@ define([
         if (!_styleSheet) {
             _styleSheet = document.createElement('style');
             _styleSheet.type = 'text/css';
-            _styleSheet.id = 'jwplayer-styles';
+            _styleSheet.setAttribute("style-src","jwplayer-styles")
             document.getElementsByTagName('head')[0].appendChild(_styleSheet);
         }
         var cssText = '';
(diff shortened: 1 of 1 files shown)


robwalch:
The player can add more than one stylesheet to the page so this won't do. We'd want unique ids, but maybe add a custom attribute so that you could query all jw stylesheets in the DOM.

nithin-raj:
Agreed, added a custom attribute and removed the id property.

robwalch:
These changes will be superseded by the simple-style-loader. @johnBartos is working on consolidating utils.css with the style loader used to add jwplayer's view css. I like the idea of an attribute that 'stamps' the element as a jw element. This could be useful for us as well when embedding multiple players so that we can avoid injecting identical css for multiple players.

robwalch:
See #1201 and https://github.com/jwplayer/simple-style-loader/blob/master/addStyles.js#L87

Each style tag will have a 'data-jwplayer-id' attribute. The first will be a global styles element used by all players. The value for the attribute on that element will simply be the webpack module id of the loader module. Players that define their own styles will create a new element who's 'data-jwplayer-id' attribute value is set to the player id.

johnBartos:
I'm going to close this since the the new version (0.3.0) of simple-style-loader fixes this issue https://github.com/jwplayer/simple-style-loader. We use simple-style-loader to apply styles to the player. This feature will be available in 7.5.0 and the data tag looks like `data-jwplayer-id=...`

PR #1194, opened 2016-05-16
708 Captions positioning

egreaves:
Ensure browser rendered captions display above the control bar and within the viewport for all lines.

Fixes #
JW7-2123

### DIFF
--- a/src/css/flags/user-inactive.less
+++ b/src/css/flags/user-inactive.less
@@ -1,5 +1,3 @@
-@import "../imports/mixins";
-
 .jwplayer.jw-flag-user-inactive {
     &.jw-state-playing {
         .jw-controlbar,
@@ -20,9 +18,8 @@
             -webkit-cursor-visibility: auto-hide;
         }
         video::-webkit-media-text-track-container {
-            max-height: 100%;
+            max-height: none;
         }
-        .safari-only-captions-position(0, 0);
     }
 
     &.jw-state-buffering {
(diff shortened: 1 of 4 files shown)


src/css/flags/user-inactive.less
```diff
@@ -1,3 +1,6 @@
+@import "../imports/vars";
```
robwalch: Do you need to import vars here?

src/css/imports/captions.less
```diff
@@ -42,13 +43,16 @@
 
 .jwplayer {
     video::-webkit-media-text-track-container {
-        bottom: @controlbar-height * 0.75;
-
+        // With 16 lines of captions, 1em ~= 6.25% of height
+        max-height: calc(100% - 6.25 * (@controlbar-height));
```
robwalch: `calc`... fancy!

robwalch:
@egreaves Have a look at [Safari Technical Preview](https://developer.apple.com/safari/technology-preview/). Captions are not positioned correctly on the seconds playlist item when the controlbar is shown.

src/css/imports/mixins.less
```diff
@@ -58,6 +58,22 @@
...
 }
 
+.safari-only-captions-position(@margin, @translation) {
+
+    // this detects Safari 9+
+    // http://browserstrangeness.bitbucket.org/css_hacks.html#webkit
+
+    @supports ((overflow:-webkit-marquee) and (justify-content:inherit)) or (hanging-punctuation:allow-end) {
+        video::-webkit-media-text-track-container {
+            top: @margin;
```
robwalch: What is margin? Is there any harm in adding `top: 0` captions.less? It would be better to use rule that work everywhere.

robwalch:
@johnBartos this looks good to me. Have a look after it's built.

PR #1200, opened 2016-05-19
Fix controller method names

robwalch:
Incorrect renaming of playlist methods broke these calls. 

Fixes #
JW7-2096

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -219,7 +219,7 @@ define([
                     index: _model.get('item'),
                     item: _model.get('playlistItem')
                 });
-                
+
                 _this.trigger(events.JWPLAYER_CAPTIONS_LIST, {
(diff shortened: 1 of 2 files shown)


src/js/controller/controller.js
```diff
@@ -649,7 +649,6 @@ define([
             this.setCurrentAudioTrack = _setCurrentAudioTrack;
             this.getCurrentAudioTrack = _getCurrentAudioTrack;
             this.getAudioTracks = _getAudioTracks;
-            this._getCurrentQuality = _getCurrentQuality;
```
dannyfinks: @robwalch this one shouldn't have been removed

robwalch: Thanks for the catch! Fixed https://github.com/jwplayer/jwplayer/commit/34d0267b93513fe58b967bc173145b258df19678

PR #1202, opened 2016-05-19
Prevent repeated mediaType events

pajong:
Only fire mediaType events when there is a change in mediaType.
JW7-2446

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -73,6 +73,7 @@ define([
         };
 
         function _videoEventHandler(type, data) {
+            var evt = _.extend({}, data, {type: type});
             switch (type) {
                 case 'flashThrottle':
                     var throttled = (data.state !== 'resume');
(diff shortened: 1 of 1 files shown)


src/js/controller/model.js
```diff
@@ -93,8 +94,11 @@ define([
                     return;
 
                 case events.JWPLAYER_MEDIA_TYPE:
-                    this.mediaModel.set('mediaType', data.mediaType);
-                    break;
+                    if (this.mediaModel.get('mediaType') !== data.mediaType) {
+                        this.mediaModel.set('mediaType', data.mediaType);
+                        this.mediaController.trigger(type, evt);
```
jw-kaurand: Can we have this as a break instead and rely on the ```this.mediaController.trigger(type, evt);``` at the end of the handler instead?

pajong: No -  we do not want to trigger the event if the mediaType did not change.

PR #1203, opened 2016-05-19
Detect when fonts are loaded before displaying icons

jw-kaurand:
Detect when fonts are loaded before displaying icons by polling the width of a span.  The span starts out with a basic width and will have its size increase ones the icon font is applied and wider glyphs are used instead of the default glyphs.  This may be a more appropriate way to determine if the UI is ready in the future to make sure we don't need to do many repaints before we display the UI.

JW7-2481

### DIFF
--- a/src/css/imports/icons.less
+++ b/src/css/imports/icons.less
@@ -17,6 +17,12 @@
     .jw-icon;
 }
 
+.jw-font-load-detect {
+    position: absolute;
+    visibility: hidden;
+    display: inline-block;
+}
+
 /*** ICONS ***/
 .jw-icon-audio-tracks {
     &:before {
@@ -142,4 +148,3 @@
         content: "\e616";
     }
 }
-
--- a/src/css/imports/icons.less
+++ b/src/css/imports/icons.less
@@ -17,12 +17,6 @@
     .jw-icon;
 }
 
-.jw-font-load-detect {
-    position: absolute;
-    visibility: hidden;
-    display: inline-block;
-}
-
 /*** ICONS ***/
 .jw-icon-audio-tracks {
     &:before {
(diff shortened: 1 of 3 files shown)


robwalch:
test this please

src/css/imports/icons.less
```diff
@@ -17,6 +17,12 @@
     .jw-icon;
 }
 
+.jw-font-load-detect {
+    position: absolute;
+    visibility: hidden;
+    display: inline-block;
+}
```
robwalch: It looks like this class is used on a test element only. Why not use inline styles?

src/js/view/view.js
```diff
@@ -102,6 +103,41 @@ define([
             document.msExitFullscreen;
         _elementSupportsFullscreen = _requestFullscreen && _exitFullscreen;
 
+        function _applyStyles() {
+            var pollTarget = utils.createElement(
+                '<span class="jw-font-load-detect jw-reset">&#xe614&#xe614&#xe614</span>'
+            );
+            _playerElement.appendChild(pollTarget);
+            utils.css(pollTarget, {
```
robwalch: This adds a stylesheet to the DOM. `utils.style` adds styles to the element's styles attribute, inline.

src/js/view/view.js
```diff
@@ -102,6 +103,41 @@ define([
             document.msExitFullscreen;
         _elementSupportsFullscreen = _requestFullscreen && _exitFullscreen;
 
+        function _applyStyles() {
+            var pollTarget = utils.createElement(
+                '<span class="jw-font-load-detect jw-reset">&#xe614&#xe614&#xe614</span>'
```
robwalch: Removed 'jw-font-load-detect'

PR #1204, opened 2016-05-19
Guard to check if we have provider before using it for the rightclick

jw-kaurand:
Guard to check if we have provider before using it for the rightclick.


robwalch:
test this please

PR #1206, opened 2016-05-19
Fixed captions positions on mobile devices

egreaves:
Fixed captions position on mobile devices. Also adjusted the height of the captions container so it's ~0.5em above the control bar. This ensures that captions with line set to 'auto' display above the control bar.

Fixes #
JW7-2123


egreaves:
test this please

Issue #1098, opened 2016-03-04
VTT file containing "WEBVTT" as the first line throws "Invalid SRT file"


isiryder:
### Expected Behavior

A VTT file with only "WEBVTT" to the first line of the file, is a valid one.

So I think it shouldn't throw such error (and why SRT btw?).

I would expect it just to return, as if no file was added, without any error.

In the worst case return a significant error as "No captions in VTT file" or "Empty VTT files not allowed"

### Actual Behavior

Throws "Invalid SRT file" when passing as caption a VTT file that only contains "WEBVTT" as the first line.

NOTE: A VTT file that only contains "WEBVTT" as the first line should be valid. There is an online VTT  validator: https://quuz.org/webvtt/

### Steps to reproduce

1. Create VTT file with one line only containing "WEBVTT".
2. Using setup add that file as a track (I tested thumbnails and captions).
3. Check the console and see the error.

### Environment

- iMac
- OSX Yosemite
- JW Player version 7

robwalch:
Hi @isiryder,

Do you have a test page that reproduces this issue, or at least a url linking to the file in question? Also what type of media are you attempting to play in which browser?

Thanks in advance!

isiryder:
Hi @robwalch,

Unfortunately I don't have a visible test page right now. But I've uploaded the file in question: http://filebin.ca/2ZWuSubgCkFZ/empty.vtt

I am trying to play a simple mp4 video file using Chrome v48.0.2564.116 (64-bit).

Thanks!

johnBartos:
I'm trying to recreate this, but I'm not seeing any error thrown. Are you still having this issue @isiryder?

isiryder:
Yes, I am still having this issue. Testing with Chrome 50 and JWPlayer 7.4.2 (HTML version).

PR #1207, opened 2016-05-20
Don't throw if VTT file is valid but without captions

johnBartos:
- Don't throw if VTT is empty
- Don't add to captions menu if VTT is empty
- Download VTT before adding to list to prevent pre-emptive adding of empty track

See https://github.com/jwplayer/jwplayer/issues/1098

JW7-2440

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -170,7 +170,6 @@ define([
 
                 if (kind === 'captions' || kind === 'subtitles') {
                     if (track.file) {
-                        _addTrack(track);
                         _load(track);
                     } else if (track.data) {
                         _addTrack(track);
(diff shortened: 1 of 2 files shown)


src/js/controller/captions.js
```diff
@@ -239,6 +238,9 @@ define([
             } else {
                 status = utils.tryCatch(function() {
                     track.data = srt(xhr.responseText);
+                    if (track.data) {
```
egreaves: No need for a guard here. This is done in _addTrack

src/js/controller/captions.js
```diff
@@ -170,7 +170,6 @@ define([
 
                 if (kind === 'captions' || kind === 'subtitles') {
                     if (track.file) {
-                        _addTrack(track);
```
egreaves: Test this change with a `.vtt` file that contains captions.

johnBartos:
test this please

PR #1215, opened 2016-05-23
Feature/captions api

johnBartos:
- Expose captions getter and setter
- Clear styles before setting captions style without resetting player
- Set captions styles (including defaults) in model

Previously, changing captions styling required reloading the player. This PR exposes `getCaptions()` and `setCaptions` methods, the latter of which restyles captions on the fly.

Fixes #
JW7-2022


johnBartos:
test this pleas

src/js/view/captionsrenderer.js
```diff
@@ -193,6 +193,16 @@ define([
             this.populate(_model.get('captionsTrack'));
         };
 
+        this.clear = function () {
+            while(_display.firstChild) {
+                _display.removeChild(_display.firstChild);
+            }
```
robwalch: Use `utils.empty`. It's already required and used in this module. It's part of dom.js:
```
    dom.empty = function(element) {
        if (!element) {
            return;
        }
        while (element.childElementCount > 0) {
            element.removeChild(element.children[0]);
        }
    };
```

PR #1220, opened 2016-05-24
Bump loader version

johnBartos:
- Bump to next minor version

JW7-2472

### DIFF
--- a/package.json
+++ b/package.json
@@ -47,7 +47,7 @@
     "qunitjs": "^1.23.0",
     "raw-loader": "^0.5.1",
     "requirejs": "^2.2.0",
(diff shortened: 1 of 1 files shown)


package.json
```diff
@@ -47,7 +47,7 @@
     "qunitjs": "^1.23.0",
     "raw-loader": "^0.5.1",
     "requirejs": "^2.2.0",
-    "simple-style-loader": "~0.2.0",
+    "simple-style-loader": "~0.3.0",
```
robwalch: Why don't you just get rid of the tilde? We should maintain this version and keep it in sync with the last tested published version known to work.

PR #1224, opened 2016-05-31
Remove small black area for poster image

pajong:
When the stretch is set to uniform, and the black area is small enough,
we want to stretch the poster image to remove the black area.
JW7-2507

### DIFF
--- a/src/js/view/preview.js
+++ b/src/js/view/preview.js
@@ -4,7 +4,7 @@ define([
 ], function(_, utils) {
     var Preview = function(_model) {
         this.model = _model;
-        this.aspectRatioPromise = Promise.resolve(0);
+        this.loadImage = Promise.resolve(null);
 
         _model.on('change:playlistItem', onPlaylistItem, this);
         _model.on('change:mediaModel', onMediaModel, this);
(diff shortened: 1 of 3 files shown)


src/js/view/preview.js
```diff
@@ -50,6 +50,15 @@ define([
             var backgroundImage = '';
             if (_.isString(img)) {
                 backgroundImage = 'url("' + img + '")';
+
+                // Save the background image's width and height for resize check in view.js
+                var image = new Image();
+                image.src = img;
```
robwalch: Image loading is async.

src/js/view/view.js
```diff
@@ -698,6 +698,17 @@ define([
 
             // pass width, height from jwResize if present
             _resizeMedia(width, height);
+
+            if (_this.model.get('stretching') === 'uniform' && _preview) {
```
robwalch: ...`&& _preview.height && _preview.width` otherwise `imageAspectRatio` will be `NaN`.

src/js/view/view.js
```diff
@@ -698,6 +698,18 @@ define([
...
+                var playerAspectRatio = _playerElement.clientWidth / _playerElement.clientHeight;
+                // snap image to edges when the difference in aspect ratio is less than 9%
+                _preview.aspectRatioPromise.then(function(imageAspectRatio) {
+                    if (Math.abs(playerAspectRatio - imageAspectRatio) < 0.09) {
+                        utils.css.style(_preview.el, {
+                            backgroundSize: 'cover'
+                        });
```
robwalch: You need to add or remove the style. This only adds it. Look at the actions mixin again.

robwalch: https://github.com/JWPlayer/jwplayer-commercial/commit/040ce5fe6df4d8edb7d4a7b131d64b4ec479b2c6

src/js/view/preview.js
```diff
@@ -50,6 +51,17 @@ define([
...
```
robwalch: Set `src` after `onload`.

src/js/view/preview.js
```diff
@@ -50,6 +50,20 @@ define([
             var backgroundImage = '';
             if (_.isString(img)) {
                 backgroundImage = 'url("' + img + '")';
+
+                // Save the background image's width and height for resize check in view.js
+                if (this.image) {
+                    this.image.onload = null;
+                }
```
robwalch: This should be moved to top of setImage. We still need to do it if 'img' is null or not a string.

src/js/view/preview.js
```diff
@@ -46,16 +45,48 @@ define([
...
+            if (stretching === 'uniform') {
+                var playerAspectRatio = width / height;
+                // snap image to edges when the difference in aspect ratio is less than 9%
+                var image = this.image;
+                var backgroundSize = null;
+                if (image) {
+                    if (image.width === 0) {
+                        var _this = this;
+                        image.onload = function() {
+                            _this.resize(width, height, stretching);
```
robwalch: @pajong by moving the resize code into preview, we now retry on image load when the image is ready instead of using a Promise.

PR #1228, opened 2016-06-01
Bugfix/setcaptions spam

johnBartos:
- Wait until caption is loaded until trying to load
- Don't remove caption from list even if it is empty

JW7-2440

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -169,7 +169,12 @@ define([
                 kind = track.kind.toLowerCase();
 
                 if (kind === 'captions' || kind === 'subtitles') {
-                    _addTrack(track);
+                    if (track.file) {
+                        _addTrack(track);
+                        _load(track);
+                    } else if (track.data) {
+                        _addTrack(track);
+                    }
                 }
             }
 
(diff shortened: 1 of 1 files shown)


src/js/controller/captions.js
```diff
@@ -192,14 +188,23 @@ define([
             if (captionsMenuIndex !== 0) {
                 track = _tracks[captionsMenuIndex-1];
             }
+            if (track && track.shouldLoad) {
+                _load(track);
+                // Don't try to load again, even if there was an error
+                track.shouldLoad = false;
```
robwalch: How about `track.loaded`?

src/js/controller/captions.js
```diff
@@ -242,8 +246,8 @@ define([
                     track.data = srt(xhr.responseText);
                     if (track.data.length) {
                         _addTrack(track);
-                        _this.setCaptionsList(_captionsMenu());
-                        _selectDefaultIndex();
+                        // _this.setCaptionsList(_captionsMenu());
+                        // _selectDefaultIndex();
```
robwalch: What happened down here?

src/js/controller/captions.js
```diff
@@ -238,13 +242,7 @@ define([
                 });
             } else {
                 status = utils.tryCatch(function() {
```
johnBartos: do you think I could remove the trycatch? I took out the thrown error but im not sure if it may throw another unexpected one

robwalch: Look at the parser - I believe it throws when encountering error(s) - so you should try catch.

Also `utils.tryCatch` is a util that wraps execution in try catch blocks only in the release build. It's intended for event handling (noobs who write bad callbacks should not break the player). For code that handles valid exceptions use real `try {} catch(e) {}` blocks.

PR #1233, opened 2016-06-06
Volume and mute commands shouldn't be queued

robwalch:
Volume and mute can be set on the model when a provider is not available, so there is no need to queue these commands. Volume is set on providers when they are setup.

Cleaned up use of bind in controller, as well as `this` references and param names in model for set volume and mute.

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -295,28 +295,28 @@ define([
             _currentProvider = null;
         };
 
-        this.setVolume = function(vol) {
-            vol = Math.round(vol);
-            _this.set('volume', vol);
+        this.setVolume = function(volume) {
+            volume = Math.round(volume);
+            this.set('volume', volume);
             if (_provider) {
-                _provider.volume(vol);
+                _provider.volume(volume);
             }
-            var muted = (vol === 0);
-            if (muted !== _this.get('mute')) {
-                _this.setMute(muted);
+            var mute = (volume === 0);
+            if (mute !== this.get('mute')) {
+                this.setMute(mute);
             }
         };
 
-        this.setMute = function(state) {
-            if (!utils.exists(state)) {
-                state = !_this.get('mute');
+        this.setMute = function(mute) {
+            if (!utils.exists(mute)) {
+                mute = !this.get('mute');
             }
-            _this.set('mute', state);
+            this.set('mute', mute);
             if (_provider) {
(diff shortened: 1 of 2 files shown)


src/js/controller/model.js
```diff
@@ -295,28 +295,28 @@ define([
...
+            volume = Math.round(volume);
+            this.set('volume', volume);
             if (_provider) {
-                _provider.volume(vol);
+                _provider.volume(volume);
             }
-            var muted = (vol === 0);
-            if (muted !== _this.get('mute')) {
-                _this.setMute(muted);
+            var mute = (volume === 0);
```
johnBartos: If the volume cannot go below 0 it's better to write this as `var mute = !volume`

src/js/controller/model.js
```diff
@@ -295,28 +295,28 @@ define([
...
+            if (mute !== this.get('mute')) {
```
johnBartos: Is there a cost to setting mute to the state it's already in (i.e. setting mute to `true` if it's already `true`)? If not, then we can avoid the conditional and always set mute.

src/js/controller/model.js
```diff
@@ -295,28 +295,28 @@ define([
...
+            this.set('mute', mute);
             if (_provider) {
-                _provider.mute(state);
+                _provider.mute(mute);
             }
-            if (!state) {
-                var volume = Math.max(10, _this.get('volume'));
+            if (!mute) {
+                var volume = Math.max(10, this.get('volume'));
                 this.setVolume(volume);
```
johnBartos: Couldn't these two functions stack overflow in some cases? If mute is false and volume is 0 (for whatever reason)

src/js/controller/model.js
robwalch: Not really. We check for a change in `set` so that "change:" events are only sent when the value changes https://github.com/jwplayer/jwplayer/blob/master/src/js/utils/simplemodel.js

src/js/controller/model.js
robwalch: That works. I wouldn't say it's "better". The only change here was renaming `vol` to `volume`. Please comment on the changes or submit changes.

src/js/controller/model.js
robwalch: No. How could volume be 0 after `Math.max(10, _this.get('volume'));`???

PR #1234, opened 2016-06-07
Preview test pages for view states

robwalch:
This gets /manual/view.html and interactive.html working again. Once exception - I wasn't able to get the jw-icons webfont to load. We need a way to override the base path for css output by the require less loader for the woff file to load.


johnBartos:
alright ill continue from here

PR #1235, opened 2016-06-08
Call resize for preview when setImage is called

pajong:
When autostart is set to true, we only call setImage on complete event.
Since resize is not called yet, the black border is not removed.
Making playerAspectRatio a preview variable so that we can call resize on state change when setImage is called.
JW7-2507


pajong:
The black border issue was still there when autostart was set to true.

PR #1240, opened 2016-06-09
Feature/withcredentials

johnBartos:
- Add `withCredentials` boolean flag, settable in source, playlist, or provider (top level) config

Some fetch requests need to be sent with credentials. This item allows users to set whether or not their requests should respond to set-cookie headers and pass them along with requests. 

The lowest set `withCredentials` wins out, i.e. source overrides playlist which overrides top level. Also added a test.

JW7-2499 and JW7-2582


johnBartos:
test this please

PR #1241, opened 2016-06-09
Reuse the 608 captions track

egreaves:
Only add the 608 captions track to the menu if it contains cues.

Fixes #
JW7-2596

### DIFF
--- a/src/js/providers/tracks-mixin.js
+++ b/src/js/providers/tracks-mixin.js
@@ -83,6 +83,15 @@ define(['../utils/underscore',
                     _textTracksCache[i + track.kind] = track;
                 }
                 else if (track.kind === 'subtitles' || track.kind === 'captions') {
+                    // By setting the track mode to 'hidden', we can determine if the 608 track has cues
+                    var mode = track.mode;
+                    track.mode = 'hidden';
+                    if (!track.cues.length && this.getName().name === 'caterpillar' && track.label === 'Unknown CC') {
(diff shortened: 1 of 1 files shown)


src/js/providers/tracks-mixin.js
```diff
@@ -83,6 +83,15 @@ define(['../utils/underscore',
...
                 else if (track.kind === 'subtitles' || track.kind === 'captions') {
+                    // By setting the track mode to 'hidden', we can determine if the 608 track has cues
+                    var mode = track.mode;
+                    track.mode = 'hidden';
+                    if (!track.cues.length && this.getName().name === 'caterpillar' && track.label === 'Unknown CC') {
+                        // There's no method to remove tracks added via: video.addTextTrack in caterpillar.
+                        // This ensures the 608 captions track isn't added until it has cues
+                        continue;
+                    }
+                    track.mode = mode;
```
robwalch: How often does `setTextTracks` get called? There may be no cues in the initial segment(s) buffered and played, so what happens if we only find cues later on during playback? Will this check run again so that the track is added to the menu?

egreaves: @robwalch The streaming engine notifies the provider when cues are added. When the first cue is added, it gets called:
https://github.com/jwplayer/jwplayer-commercial/blob/master/src/js/providers/caterpillar.js#L114.

PR #1245, opened 2016-06-13
Properly set withCredentials on source

johnBartos:
- Order of setting withCredentials on source goes in order of 1. source 2. playlist 3. model
- Only fallback if the flag on the current level is undefined

In the previous ticket, `withCredentials` was set to true if any of the three booleans were `true` - false cases at lower levels (i.e. souce) were being overwritten by `true` values higher up (`playlist`, `model`). This change makes it such that fallback is performed only if the current level is undefined.

Fixes #
JW7-2582

### DIFF
--- a/src/js/playlist/playlist.js
+++ b/src/js/playlist/playlist.js
@@ -18,8 +18,13 @@ define([
 
         _.each(playlist, function(item) {
             item = _.extend({}, item);
-            item.allSources = _formatSources(item.sources, androidhls,
-                item.drm || configDrm, item.preload || preload, item.withCredentials || withCredentials);
+
+            item.allSources = _formatSources(item.sources,
+                androidhls,
+                item.drm || configDrm,
+                item.preload || preload,
+                _fallbackIfUndefined(item.withCredentials, withCredentials));
+
             item.sources = _filterSources(item.allSources, providers);
 
             if (!item.sources.length) {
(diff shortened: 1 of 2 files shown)


src/js/playlist/playlist.js
```diff
@@ -93,5 +99,9 @@ define([
         return null;
     }
 
+    function _fallbackIfUndefined(value, fallback) {
+        return typeof value !== 'undefined' ? value : fallback;
```
pajong:
```_.isUndefined''' is a function that checks if the value is undefined

File: src/js/playlist/playlist.js [hunk 2]
```diff
@@ -16,10 +16,16 @@ define([
     Playlist.filterPlaylist = function(playlist, providers, androidhls, configDrm, preload, feedid, withCredentials) {
         var list = [];
 
+
```
pajong: extra white space?

src/js/playlist/playlist.js
```diff
@@ -93,5 +98,9 @@ define([
         return null;
     }
 
+    function _fallbackIfUndefined(value, fallback) {
+        return typeof _.isUndefined(value) ? fallback : value;
```
pajong: Remove typeof

pajong: And rebase the commits

PR #1252, opened 2016-06-20
Fix fullscreen flash stretch issues

pajong:
In fullscreen mode exact fit, flash had a logic to change the stretch setting to uniform then resize.
Removing that logic so that stretching works as expected in fullscreen mode.
JW7-1890

### DIFF
--- a/src/flash/com/longtailvideo/jwplayer/view/View.as
+++ b/src/flash/com/longtailvideo/jwplayer/view/View.as
@@ -119,14 +119,7 @@ public class View extends Sprite {
 
     private function resizeMedia(width:Number, height:Number):void {
         if (_mediaLayer.numChildren > 0 && _model.media.display) {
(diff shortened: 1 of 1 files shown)


src/flash/com/longtailvideo/jwplayer/view/View.as
```diff
@@ -119,14 +119,7 @@ public class View extends Sprite {
 
     private function resizeMedia(width:Number, height:Number):void {
         if (_mediaLayer.numChildren > 0 && _model.media.display) {
-            var preserveAspect:Boolean = (_model.fullscreen && _model.stretching === Stretcher.EXACTFIT);
```
johnBartos: We don't care about preserving the aspect ratio anymore?

pajong: That's the exact thing that we are trying to fix in this ticket. We want to make it actually do exactfit when we set it to exactfit, instead of changing to uniform.

PR #1255, opened 2016-06-22
added some missing conf for first grunt launch

andy-b-84:

It didn't take me too much time to realise some soft wasn't installed, but these are my 2c of contribution

Fixes #
JW7-####

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,6 +57,8 @@ You also have the power to programatically set any configuration within the play
     # First time set up
     npm install -g grunt
     npm install
+    npm install webpack minimist
+    export AIR_HOME={AIRSDK_Compiler}
     
     # Build using
     grunt


README.md
```diff
@@ -57,6 +57,8 @@ You also have the power to programatically set any configuration within the play
     # First time set up
     npm install -g grunt
     npm install
+    npm install webpack minimist
+    export AIR_HOME={AIRSDK_Compiler}
```
robwalch: We can add webpack and minimist to the dev dependencies in package.json.

How does `export AIR_HOME={AIRSDK_Compiler}` help? Do you already have $AIRSDK_Compiler defined in your environment? If so maybe mxmlc is already in your path ad we can make changes to support that? In the future we plan to make Flash compilation optional and only warn users if mxmlc is not found / no swf compiled.

PR #1257, opened 2016-06-22
Try to play the next item in adpod on error

pajong:
When there is an error, try to play the next item in the ad pod.
JW7-2097

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -33,7 +33,8 @@ define([
             _options = {},
             _oldProvider,
             _oldpos,
-            _olditem;
+            _olditem,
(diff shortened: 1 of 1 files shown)


src/js/controller/instream-adapter.js
```diff
@@ -112,10 +113,31 @@ define([
...
+
         function _instreamForward(type, data) {
             if (type === 'complete') {
                 return;
             }
+            if (type === 'mediaError' || type === 'error') {
+                if (_array && _arrayIndex + 1 < _array.length) {
+                    _loadNextItem();
+                }
+            }
```
robwalch: We should not load the next item until after the error event is dispatched.

PR #1263, opened 2016-06-24
Fixes #1149: Remove bower dependencies

### LINKED ISSUE #1149: Replace bower components with npm modules
Our unit tests and karma configuration depend on bower. It would be nice to remove bower as a dependency and only use npm for dev dependencies.

hghazzi:
Removing bower dependency from jwplayer and replacing with npm modules


robwalch:
@hghazzi this is great! jwplayer-commercial needs the same.

PR #1266, opened 2016-06-27
Order qualities from high to low

pajong:
Handle qualities ordering in flash provider, so that we do not mess with the logic in adaptive switching.
This only changes the ordering in UI, and the quality changing works the same way by saving the original index.
JW7-2607

### DIFF
--- a/src/js/providers/flash.js
+++ b/src/js/providers/flash.js
@@ -5,8 +5,9 @@ define([
     'events/states',
     'utils/embedswf',
     'providers/default',
-    'utils/backbone.events'
-], function(utils, _, events, states, EmbedSwf, DefaultProvider, Events) {
+    'utils/backbone.events',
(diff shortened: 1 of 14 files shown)


robwalch:
@pajong conflicts

src/js/providers/flash.js
```diff
@@ -280,18 +313,13 @@ define([
 
                     // jwplayer 6 flash player events (forwarded from AS3 Player, Controller, Model)
                     _swf.on(events.JWPLAYER_MEDIA_LEVELS, function(e) {
```
robwalch: If the handlers are identical combine them:
```js
_swf.on([events.JWPLAYER_MEDIA_LEVELS, events.JWPLAYER_MEDIA_LEVEL_CHANGED], function(e) {
```

PR #1267, opened 2016-06-27
Style focused elements using hover colors

robwalch:
Remove the browser's focus ring on focused buttons and instead style elements the same as when they are being hovered over.

This is done so that the new aria label and tab indexing feature does not introduce new unexpected styling on websites using the player.

JW7-2628

=================
Other things to improve player behavior with tabbing (TODO / new tech-debt introduces by tabbing buttons):
1. When a button is tab focused in the control bar, the control bar should not hide
2. Change the behavior of space and enter keys when buttons are focused to click or toggle the focused button rather than always toggle playback

### DIFF
--- a/src/css/imports/mixins.less
+++ b/src/css/imports/mixins.less
@@ -88,12 +88,19 @@
         .jw-button-color {
             color: @inactive-color;
             .hover(@hover-color);
+            &:focus {
+                outline: none;
+                color: @hover-color;
+            }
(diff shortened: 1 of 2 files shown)


johnBartos:
If we want tabbing to be a feature we should style the ring, not remove it. It's difficult to tell what item the keyboard is focused on if the ring is absent.

PR #1272, opened 2016-06-28
Load providers on playlist not in setup

robwalch:
Always load providers when the playlist is set, or instream adapter wants to load an item, and the required provider(s) have not been loaded. 

JW7-2540

### DIFF
--- a/src/js/providers/providers-loaded.js
+++ b/src/js/providers/providers-loaded.js
@@ -1,12 +1,11 @@
 define([
-    'providers/html5',
-    'providers/flash'
-], function(html5, flash) {
+    //'providers/html5',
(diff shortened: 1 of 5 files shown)


src/js/providers/providers-loaded.js
```diff
@@ -1,12 +1,11 @@
 define([
-    'providers/html5',
-    'providers/flash'
-], function(html5, flash) {
+    //'providers/html5',
```
pajong: Do we still want to keep this if this file is not doing anything?

robwalch: I'll add a comment. Any providers that we want bundled in jwplayer.js should be required there.

PR #1291, opened 2016-07-05
Fix issues with captions not showing after midroll

pajong:
After midroll, we disable the text tracks when natively rendered.
We should have a check saying if the tracks inuse is false, we reload the tracks.
JW7-2652

### DIFF
--- a/src/js/providers/tracks-mixin.js
+++ b/src/js/providers/tracks-mixin.js
@@ -437,8 +437,14 @@ define(['../utils/underscore',
     }
 
     function _tracksAlreadySideloaded(tracks) {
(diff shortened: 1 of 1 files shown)


src/js/providers/tracks-mixin.js
```diff
@@ -437,8 +437,14 @@ define(['../utils/underscore',
     }
 
     function _tracksAlreadySideloaded(tracks) {
+        // Determine if the currently chosen track is removed and disabled
+        var textTracks = this._textTracks;
+        var disabled = false;
```
egreaves: Lets instead call this `inuse` since that's what we're trying to determine.

PR #1293, opened 2016-07-06
Track creation

egreaves:
TextTrack labels are read-only once created. We only want to reuse tracks where the labels are the same. This also addresses issues with duplicate embedded tracks in Safari and preventing the addition of cues without valid start and end times.

Fixes #
JW7-2652


egreaves:
Test this please

PR #1298, opened 2016-07-07
Listen to browser track changes

egreaves:
Listen to browser track changes. Ignore tracks that aren't in use.
Fixes #
JW7-2687

### DIFF
--- a/src/js/providers/tracks-mixin.js
+++ b/src/js/providers/tracks-mixin.js
@@ -96,13 +96,12 @@ define(['../utils/underscore',
             }
         }
 
-        if (this._textTracks) {
-            if (this._renderNatively) {
-                this.addTracksListener(this._textTracks, 'change', textTrackChangeHandler.bind(this));
-            }
-            if (this._textTracks.length) {
-                this.trigger('subtitlesTracks', {tracks: this._textTracks});
-            }
+        if (this._renderNatively) {
+            this.addTracksListener(this.video.textTracks, 'change', textTrackChangeHandler.bind(this));
+        }
+
+        if (this._textTracks.length) {
+            this.trigger('subtitlesTracks', {tracks: this._textTracks});
         }
     }
 
@@ -186,7 +185,7 @@ define(['../utils/underscore',
         if (cueData.useDTS) {
             // There may not be any 608 captions when the track is first created
             // Need to set the source so position is determined from metadata
-            if(!track.source) {
+            if (!track.source) {
                 track.source = cueData.source || 'mpegts';
             }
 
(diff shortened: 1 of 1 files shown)


src/js/providers/tracks-mixin.js
```diff
@@ -98,7 +98,7 @@ define(['../utils/underscore',
 
         if (this._textTracks) {
             if (this._renderNatively) {
-                this.addTracksListener(this._textTracks, 'change', textTrackChangeHandler.bind(this));
+                this.addTracksListener(this.video.textTracks, 'change', textTrackChangeHandler.bind(this));
```
pajong: This is somewhat confusing - we keep ```this._textTracks``` and check if that exists, and use ```this.video.textTracks``` to add listener to. It feels like we are unnecessarily keeping two copies of the same thing.

src/js/providers/tracks-mixin.js
```diff
@@ -300,20 +300,27 @@ define(['../utils/underscore',
...
-
-        if (!this._textTracks || this.video.textTracks.length > this._textTracks.length) {
+        var textTracks = this.video.textTracks;
+        var inUseTracks = _.filter(textTracks, function (track)  {
+            return track.inuse || !track._id;
+        });
+        if (!this._textTracks || inUseTracks.length > this._textTracks.length) {
             // If the video element has more tracks than we have internally..
-            this.setTextTracks.call(this, this.video.textTracks);
+            this.setTextTracks.call(this, textTracks);
```
pajong: Again, we use ```textTracks``` and ```this._textTracks``` - feels redundant.

src/js/providers/tracks-mixin.js
```diff
@@ -300,20 +300,27 @@ define(['../utils/underscore',
...
         var _selectedTextTrackIndex = -1, i = 0;
         for (i; i < this._textTracks.length; i++) {
             if (this._textTracks[i].mode === 'showing') {
                 _selectedTextTrackIndex = i;
                 break;
             }
         }
-        this.setSubtitlesTrack.call(this, _selectedTextTrackIndex + 1);
+        // Notifying the model when the index changes keeps the current index in sync in iOS Fullscreen mode
+        if(_selectedTextTrackIndex !== this._currentTextTrackIndex) {
```
pajong: Minor - space after ```if```

src/js/providers/tracks-mixin.js
```diff
@@ -300,20 +300,27 @@ define(['../utils/underscore',
...
+            console.log('new track index:',_selectedTextTrackIndex, this._currentTextTrackIndex);
```
pajong: We should remove the console log.

src/js/providers/tracks-mixin.js
hghazzi: I agree that we should eventually find a way to consolidate the video tag tracks and _textTracks to be one and the same, or group them in a way to get what we need when we need it.

PR #1299, opened 2016-07-11
Fix visual quality levels index for flash

pajong:
To sort quality from high to low, we only changed the UI from flash provider.
When we receive visualQuality from flash, the index of each level is not what our UI says.
Updating the visualQuality event so that we correctly give the index.
JW7-2719

### DIFF
--- a/src/js/providers/flash.js
+++ b/src/js/providers/flash.js
@@ -358,6 +358,11 @@ define([
                     }, this);
 
                     _swf.on('visualQuality', function(e) {
+                        // Get index from sorted levels from the level's index + 1 to take Auto into account
+                        var sortedIndex = (_qualityLevels.length) === 1 ? 0 :
+                            _getSortedIndex(_qualityLevels, e.level.index + 1) - 1;
+                        // Use extend so that the actual level's index is not modified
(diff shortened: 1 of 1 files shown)


src/js/providers/flash.js
```diff
@@ -358,6 +358,10 @@ define([
                     }, this);
 
                     _swf.on('visualQuality', function(e) {
+                        // Get index from sorted levels from the level's index + 1 to take Auto into account
+                        var sortedIndex = _getSortedIndex(_qualityLevels, e.level.index + 1) - 1;
```
hghazzi: Is Auto always an option in the Qualities List?

src/js/providers/flash.js
```diff
@@ -358,6 +358,11 @@ define([
                     }, this);
 
                     _swf.on('visualQuality', function(e) {
+                        // Get index from sorted levels from the level's index + 1 to take Auto into account
+                        var sortedIndex = (_qualityLevels.length) === 1 ? 0 :
+                            _getSortedIndex(_qualityLevels, e.level.index + 1) - 1;
```
hghazzi: @pajong So was that a valid concern?

pajong: Yup

PR #1300, opened 2016-07-12
Filter for supported track kinds

egreaves:

Metadata tracks are not included in the subtitlesTracks list. This ensures that we filter for only subtitles and captions tracks when comparing the counts of what's already included in the menu and track changes coming from the video tag.

Fixes #
JW7-2730

### DIFF
--- a/src/js/providers/tracks-mixin.js
+++ b/src/js/providers/tracks-mixin.js
@@ -301,7 +301,7 @@ define(['../utils/underscore',
     function textTrackChangeHandler() {
         var textTracks = this.video.textTracks;
         var inUseTracks = _.filter(textTracks, function (track)  {
-            return track.inuse || !track._id;
+            return (track.inuse || !track._id) && _kindSupported(track.kind);
         });
         if (!this._textTracks || inUseTracks.length > this._textTracks.length) {
             // If the video element has more tracks than we have internally..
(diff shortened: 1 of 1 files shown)


src/js/providers/tracks-mixin.js
```diff
@@ -385,6 +385,10 @@ define(['../utils/underscore',
         return providerName.indexOf('flash') === -1 && (utils.isChrome() || utils.isIOS() || utils.isSafari());
     }
 
+    function _kindSupported(kind) {
```
pajong: We should replace all of ```if (track.kind === 'subtitles' || track.kind === 'captions')``` with this function if we are creating this. Here is one place: https://github.com/jwplayer/jwplayer/blob/f8fef3d470d92509ac23dff16aa663b61ca1b8a9/src/js/providers/tracks-mixin.js#L69
Try to check for other places too.

hghazzi: @pajong Not sure all the kinds can be, but this is not exactly the same as `if (track.kind === 'subtitles' || track.kind === 'captions')`

pajong: @hghazzi This checks whether if ```subtitles``` or ```captions``` is in the string, and I think Evol created this function to check the same thing as ```if (track.kind === 'subtitles' || track.kind === 'captions')```.

PR #1305, opened 2016-07-14
Replace html breaks with line breaks

egreaves:
Multi-line captions need to be separated by line breaks, not html line break tags, so they can be rendered correctly.

Fixes #
JW7-2686

### DIFF
--- a/src/js/parsers/captions/srt.js
+++ b/src/js/parsers/captions/srt.js
@@ -50,7 +50,7 @@ define([
                 entry.begin = _seconds(line.substr(0, index));
                 entry.end   = _seconds(line.substr(index + 5));
                 // Remaining lines contain the text
-                entry.text = array.slice(idx + 1).join('<br/>');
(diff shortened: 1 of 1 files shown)


src/js/parsers/captions/srt.js
```diff
@@ -50,7 +50,7 @@ define([
                 entry.begin = _seconds(line.substr(0, index));
                 entry.end   = _seconds(line.substr(index + 5));
                 // Remaining lines contain the text
-                entry.text = array.slice(idx + 1).join('<br/>');
+                entry.text = array.slice(idx + 1).join('\r\n');
```
hghazzi: So there's absolutely no captioning that would expect html, correct?

egreaves: @hghazzi Not for line breaks. Some other [HTML tags](https://dvcs.w3.org/hg/text-tracks/raw-file/default/608toVTT/608toVTT.html#italics-and-underline) are honored, though.

PR #1306, opened 2016-07-14
Add listener to mediaController earlier

pajong:
Add listener as soon as model is set up, so that we do not lose some of the events.
JW7-2710

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -87,6 +87,7 @@ define([
             _setup.on(events.JWPLAYER_READY, _playerReady, this);
             _setup.on(events.JWPLAYER_SETUP_ERROR, this.setupError, this);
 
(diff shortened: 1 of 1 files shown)


src/js/controller/controller.js
```diff
@@ -87,6 +87,7 @@ define([
             _setup.on(events.JWPLAYER_READY, _playerReady, this);
             _setup.on(events.JWPLAYER_SETUP_ERROR, this.setupError, this);
 
+            _model.mediaController.on('all', _triggerAfterReady, this);
```
hghazzi: @egreaves Should we be concerned about moving this out of the `_playerReady` method and before all the other events for `mediaController` are setup?

hghazzi:
Confirmed that it won't be an issue, merging

PR #1307, opened 2016-07-14
Use requestAnimationFrame for responsive sizing

jw-kaurand:
Updates the container dimensions for the player using requestAnimationFrame so as to only update it when the DOM has updated.  This helps correct issues where the resize event can propagate incorrect container sizes when entering or exiting fullscreen.

JW7-2740

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -231,21 +231,23 @@ define([
                 containerWidth = Math.round(bounds.width),
                 containerHeight = Math.round(bounds.height);
 
-            if (containerWidth && containerHeight) {
-                if (containerWidth !== _lastWidth || containerHeight !== _lastHeight) {
-                    _lastWidth = containerWidth;
-                    _lastHeight = containerHeight;
-                    clearTimeout(_resizeMediaTimeout);
-                    _resizeMediaTimeout = setTimeout(_resizeMedia, 50);
-
-                    _model.set('containerWidth', containerWidth);
-                    _model.set('containerHeight', containerHeight);
-                    _this.trigger(events.JWPLAYER_RESIZE, {
-                        width: containerWidth,
-                        height: containerHeight
-                    });
-                }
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -250,7 +246,24 @@ define([
...
+        }
+
+        function _responsiveListener() {
+            if (!document.body.contains(_playerElement)) {
+                window.removeEventListener('resize', _responsiveListener);
+                if (_isMobile) {
+                    window.removeEventListener('orientationchange', _responsiveListener);
+                }
+            } else {
+                if (window.requestAnimationFrame) {
```
hghazzi: Looks like this is not supported in IE9, will that be an issue?

https://developer.mozilla.org/en-US/docs/Web/API/window/requestAnimationFrame#Browser_compatibility

src/js/view/view.js
```diff
@@ -250,7 +246,24 @@ define([
...
+                    cancelAnimationFrame(_resizeContainerRequestId);
+                    _resizeContainerRequestId = requestAnimationFrame(_setContainerDimensions);
```
hghazzi: Can we add `window.` to both of these function calls.

src/js/view/view.js
```diff
@@ -225,17 +226,12 @@ define([
...
                 containerHeight = Math.round(bounds.height);
 
-            if (!document.body.contains(_playerElement)) {
-                window.removeEventListener('resize', _responsiveListener);
-                if (_isMobile) {
-                    window.removeEventListener('orientationchange', _responsiveListener);
-                }
-            } else if (containerWidth && containerHeight) {
+            if (containerWidth && containerHeight) {
                 if (containerWidth !== _lastWidth || containerHeight !== _lastHeight) {
```
hghazzi: Can these two if statements be merged into one cleaner one?

johnBartos:
As a side note, this may fix the vpaid resizing bug

PR #1308, opened 2016-07-14
Make slider container height 1em for easy grabbability

johnBartos:
- Increase slider container height to 1em

Fixes #
JW7-2705

### DIFF
--- a/src/css/imports/slider.less
+++ b/src/css/imports/slider.less
@@ -1,7 +1,7 @@
 @import "vars";
 
 .jw-slider-container {
-    height: @slider-height;
+    height: 1em; // For easy grabbability - does not affect positioning of slider itself
     width: 100%;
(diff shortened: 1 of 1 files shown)


src/css/imports/slider.less
```diff
@@ -1,7 +1,7 @@
 @import "vars";
 
 .jw-slider-container {
-    height: @slider-height;
+    height: 1em; // For easy grabbability - does not affect positioning of slider itself
```
hghazzi: This looks like a change that could have unexpected side-effects.

johnBartos:
The 1em height is how it was before the skin changes

src/css/imports/slider.less
johnBartos: This is how it was before the skin changes

PR #1311, opened 2016-07-15
Hide controller during full slot ads on mobile

jw-kaurand:
Changed priority of CSS so that we can get the controlbar to hide on mobile during full slot ads.

JW7-2752

### DIFF
--- a/src/css/flags/ads.less
+++ b/src/css/flags/ads.less
@@ -60,7 +60,13 @@
     }
 }
 
-.jwplayer.jw-flag-ads-vpaid {
+.jwplayer.jw-flag-ads.jw-flag-touch {
+    .jw-controlbar {
+        display: table;
+    }
+}
+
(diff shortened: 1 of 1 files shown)


src/css/flags/ads.less
```diff
@@ -60,7 +60,13 @@
...
 }
 
-.jwplayer.jw-flag-ads-vpaid {
+.jwplayer.jw-flag-ads.jw-flag-touch {
+    .jw-controlbar {
+        display: table;
+    }
+}
+
+.jwplayer.jw-flag-ads-vpaid, .jwplayer.jw-flag-touch.jw-flag-ads-vpaid {
```
hghazzi: So is it just vpaid ads that we are concerned about?

jw-kaurand: Yeah, for the scope of this issue we're only concerned about full slot ads, which are VPAID.

PR #1315, opened 2016-07-15
Set dimensions on setup and resize

jw-kaurand:
Sets dimensions when we resize via the api and during startup. This fixes an issue where the viewable area analytics parameter was not being set correctly on the startup ping.

JW7-2740

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -442,7 +442,7 @@ define([
             // This setTimeout allows the player to actually get embedded into the player
             _api.on(events.JWPLAYER_READY, function() {
                 // Initialize values for containerWidth and containerHeight
-                _responsiveListener();
+                _setContainerDimensions();
 
                 _resize(_model.get('width'), _model.get('height'));
             });
@@ -802,7 +802,7 @@ define([
         this.resize = function(width, height) {
             var resetAspectMode = true;
             _resize(width, height, resetAspectMode);
-            _responsiveListener();
+            _setContainerDimensions();
         };
         this.resizeMedia = _resizeMedia;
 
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -802,7 +802,7 @@ define([
         this.resize = function(width, height) {
             var resetAspectMode = true;
             _resize(width, height, resetAspectMode);
-            _responsiveListener();
+            _setContainerDimensions();
```
pajong: ```_responsiveListener``` cancels the previous animationFrame, so it is required to be in resize.

jw-kaurand: Hmm, yes, though we want to execute _setContainerDimension immediately so we can update container sizes right away.  I think I can cancel any pending RAF calls in _setContainerDimensions for safety.

src/js/view/view.js
```diff
@@ -53,6 +53,8 @@ define([
             _rightClickMenu,
             _resizeMediaTimeout = -1,
             _resizeContainerRequestId = -1,
+            _resizeRAFFunc = window.requestAnimationFrame || function(rafFunc) { return window.setTimeout(rafFunc, 17); },
```
pajong: The name of the function is really hard to follow what that function is doing - can we rename this to something more intuitive or leave a comment on what resizeRAFFunc means? Same for ```_resizecCancelRAFFunc```.
Also, don't forget to rebase your commits :)

pajong: Also note that the build is failing because the line is too long.

jw-kaurand:
Closed in favor of https://github.com/jwplayer/jwplayer/pull/1323

PR #1326, opened 2016-07-19
Non-native captions rendering in audio mode

egreaves:
Ensure non-natively rendered captions are always above the control bar when the player is in audio mode.

Fixes #
JW7-2771


hghazzi:
@egreaves should this be merged in?

Issue #1325, opened 2016-07-19
Increase decimal scale of getPosition


maxstoller:
It seems `getPosition` returns a number number with 1 decimal place. Would it be possibly to modify this function to return a number rounded to two decimal places instead?

johnBartos:
@maxstoller We don't truncate the position, it's accurate to 6 places when applicable. I believe that 0's are automatically dropped though (so 14.000000 -> 14). What's your use case?

maxstoller:
Hi @johnBartos -- Thanks for the quick response. We're syncing playback of two videos. Here's a demo of getPosition returning 1 decimal places. Watch the JS console. https://jwplayer-demo.herokuapp.com/

johnBartos:
Ah, its because it's an RTMP stream. We're cutting it to the tenth there but it can easily be changed - do you want to contribute and submit a PR for it?

Here's the line to change: https://github.com/jwplayer/jwplayer/blob/master/src/flash/com/longtailvideo/jwplayer/media/RTMPMediaProvider.as#L428

And here's what it should look like:
https://github.com/jwplayer/jwplayer/blob/master/src/flash/com/longtailvideo/jwplayer/media/VideoMediaProvider.as#L255

maxstoller:
Done, thanks @johnBartos. I increased it to 5 decimal places, but I can cut it to 3 if needed.

PR #1327, opened 2016-07-19
increase scale of position number

maxstoller:
#1325 

Thanks!


johnBartos:
How about we just remove the rounding altogether? That's what we do in our HTML5 implementation

maxstoller:
Even better! Thanks @johnBartos. Is this going to make it to the commercial release? If not, does that mean we have to build it ourselves?

johnBartos:
Yeah it will make it in to 7.5, thanks for your contribution! I'll test this out and merge it in soon

maxstoller:
Great, thanks!

robwalch:
Check RTMP / API for impact on time and complete events.

PR #1333, opened 2016-07-22
Catch video.play() promise errors

robwalch:
Catches error logged in the console for **html5 provider**:

> Uncaught (in promise) DOMException: The play() request was interrupted by a call to pause().

More on `video.play()` promise:
https://developers.google.com/web/updates/2016/03/play-returns-promise?hl=en

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -446,6 +446,13 @@ define([
             return currentQuality;
         }
 
+        function _play() {
+            var promise = _videotag.play();
+            if (promise && promise.catch) {
+                promise.catch(utils.noop);
+            }
+        }
+
         function _completeLoad(startTime, duration) {
 
             _delayedSeek = 0;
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -446,6 +446,13 @@ define([
             return currentQuality;
         }
 
+        function _play() {
+            var promise = _videotag.play();
+            if (promise && promise.catch) {
```
hghazzi: This check is here because I guess not all browsers return a Promise?

robwalch: Since Chrome 50. Other browsers are on board or getting there. We could actually use this in iOS 10 to determine play() won't actually start playback because a user gesture is required.

We also need a similar catch in the commercial project for shaka and caterpillar.

robwalch:
We should move video-actions-mixing into this project. Use it in html5.js. And add catching of video.play() promise exceptions there so we get in shaka and caterpillar too.

dannyfinks:
@pajong ran some smoke tests - looks good to me as well

PR #1337, opened 2016-07-27
Captions no longer move when controls are disabled

hghazzi:
when the controls were disabled, the max width was still being changed
when the player was active.

JW7-2588

### DIFF
--- a/src/css/flags/controls-disabled.less
+++ b/src/css/flags/controls-disabled.less
@@ -10,4 +10,8 @@
     .jw-media {
         cursor: auto;
     }
+
+    video::-webkit-media-text-track-container {
+        max-height: none;
+    }
 }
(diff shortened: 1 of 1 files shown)


src/css/flags/controls-disabled.less
```diff
@@ -10,4 +10,8 @@
     .jw-media {
         cursor: auto;
     }
+
```
robwalch: Don't you also need to do this to `.jw-captions` for captions rendered in a div (Firefox)?

```css
.jw-captions {
    max-height: none;
}
```

hghazzi: You're right, fixing now. Thought I tested it in Firefox, but must have been Safari.

PR #1338, opened 2016-07-27
Fix embedded stylings blocking video

hghazzi:
Need to make the background transparent for caption elements to ensure
they are not overwritten by embedded stylings

JW7-2779

### DIFF
--- a/src/css/imports/captions.less
+++ b/src/css/imports/captions.less
@@ -50,6 +50,11 @@
     // Override vtt.js font-size so text resizes as the player size changes
     font-size: inherit;
     line-height: 1.3em;
+    background: rgba(0, 0, 0, 0);
+}
+
+.jw-text-track-container {
+    background: rgba(0, 0, 0, 0);
 }
 
 .jwplayer {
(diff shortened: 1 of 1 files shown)


src/css/imports/captions.less
```diff
@@ -50,6 +50,11 @@
     // Override vtt.js font-size so text resizes as the player size changes
     font-size: inherit;
     line-height: 1.3em;
+    background: rgba(0, 0, 0, 0);
```
robwalch: Use background-color: transparent;

hghazzi: gotcha

src/css/imports/captions.less
```diff
@@ -50,6 +50,11 @@
     // Override vtt.js font-size so text resizes as the player size changes
     font-size: inherit;
     line-height: 1.3em;
+    background: transparent;
+}
+
+.jw-text-track-container {
```
robwalch: I just noticed that this element does not have the .jw-reset class added to. @egreaves is there a reason for this? Every element in the player has this class to prevent page styles from affecting player elements.

PR #1339, opened 2016-07-27
[REVIEW] Parse frames correctly from time string

pajong:
We did not support time string with format hh:mm:ss:ff, where ff refers to frames.
Frames were treated as seconds, resulting captions with frames to have wrong start/end time.
Calculate the frames correctly when parsing.
JW7-2763

### DIFF
--- a/src/js/parsers/captions/dfxp.js
+++ b/src/js/parsers/captions/dfxp.js
@@ -10,6 +10,15 @@ define([
         validate(xmlDoc);
         var _captions = [];
         var paragraphs = xmlDoc.getElementsByTagName('p');
+        // Default frameRate is 30
+        var frameRate = 30;
+        var tt = xmlDoc.getElementsByTagName('tt');
+        if (tt && tt[0]) {
+            var parsedFrameRate = parseFloat(tt[0].getAttribute('ttp:frameRate'));
+            if (!isNaN(parsedFrameRate)) {
+                frameRate = parsedFrameRate;
+            }
+        }
         validate(paragraphs);
         if (!paragraphs.length) {
             paragraphs = xmlDoc.getElementsByTagName('tt:p');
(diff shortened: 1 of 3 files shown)


pajong:
We need an example TTML with frameRate included in it before this can be merged.

src/js/parsers/captions/dfxp.js
```diff
@@ -10,6 +10,8 @@ define([
         validate(xmlDoc);
         var _captions = [];
         var paragraphs = xmlDoc.getElementsByTagName('p');
+        var tt = xmlDoc.getElementsByTagName('tt');
+        var frameRate = (tt && tt[0].getAttribute) ? tt[0].getAttribute('ttp:frameRate') || 30 : 30;
```
robwalch: Simplify this. It's better to use an if statement than OR in a ternary statement.

src/js/parsers/captions/dfxp.js
```diff
@@ -10,6 +10,12 @@ define([
         validate(xmlDoc);
         var _captions = [];
         var paragraphs = xmlDoc.getElementsByTagName('p');
+        // Default frameRate is 30
+        var frameRate = 30;
+        var tt = xmlDoc.getElementsByTagName('tt');
+        if (tt && tt[0].getAttribute) {
+            frameRate = tt[0].getAttribute('ttp:frameRate') || frameRate;
```
robwalch: parseFloat and make sure !isNaN

src/js/utils/strings.js
```diff
@@ -82,25 +82,34 @@ define([
      * @param {String}    The input string. Supported are 00:03:00.1 / 03:00.1 / 180.1s / 3.2m / 3.2h
```
robwalch: update input string examples

robwalch:
@pajong tests are failing in FF

PR #1342, opened 2016-07-28
Side load captions once per item and don't clear after ads

robwalch:
Cleaned up new captions mixin.

Ensured sideloaded captions are only loaded once per playlist item and that loading is aborted if the item is skipped before loading completes.

Fixed leak that kept adding event listener which was never actually removed.

Removed code that cleared captions on provider destroy because it removed tracks on ad complete requiring captions to be loaded a again for the current playback item.

JW7-2826

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -460,7 +460,7 @@ define([
             if (sourceChanged || loadedSrc === 'none' || loadedSrc === 'started') {
                 _duration = duration;
                 _setVideotagSource(_levels[_currentQuality]);
-                _this.setupSideloadedTracks(this._itemTracks);
+                _this.setupSideloadedTracks(_this._itemTracks);
                 _videotag.load();
             } else {
                 // Load event is from the same video as before
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -144,7 +144,6 @@ define([
             _audioTracks = null,
             _currentAudioTrackIndex = -1,
             _activeCuePosition = -1,
-            _itemTracks = null,
             _visualQuality = { level: {} };
 
         // Find video tag, or create it if it doesn't exist.  View may not be built yet.
@@ -154,7 +153,6 @@ define([
         _videotag.className = 'jw-video jw-reset';
 
         this.isSDK = _isSDK;
-        this.itemTracks = _itemTracks;
         this.video = _videotag;
 
         // prevent browser from showing second cast icon
@@ -462,7 +460,7 @@ define([
             if (sourceChanged || loadedSrc === 'none' || loadedSrc === 'started') {
                 _duration = duration;
                 _setVideotagSource(_levels[_currentQuality]);
-                _this.setupSideloadedTracks(_itemTracks);
+                _this.setupSideloadedTracks(this._itemTracks);
                 _videotag.load();
             } else {
                 // Load event is from the same video as before
@@ -531,8 +529,6 @@ define([
 
                 dom.emptyElement(_videotag);
                 _currentQuality = -1;
-                _itemTracks = null;
-                _this.clearTracks();
                 // Don't call load in iE9/10 and check for load in PhantomJS
                 if (!_isMSIE && 'load' in _videotag) {
                     _videotag.load();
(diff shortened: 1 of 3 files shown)


src/js/providers/tracks-mixin.js
```diff
@@ -97,28 +98,38 @@ define(['../utils/underscore',
...
+        // Determine if the tracks are the same and the embedded + sideloaded count = # of tracks in the controlbar
+        var alreadyLoaded = itemTracks === this._itemTracks;
+        if (!alreadyLoaded) {
+            _cancelXhr(this._itemTracks);
+        }
+        this._itemTracks = itemTracks;
+
         this._renderNatively = _nativeRenderingSupported(this.getName().name);
-        if (this.isSDK || !tracks) {
+        if (!itemTracks || !itemTracks.length) {
```
hghazzi: `_.isEmpty(itemTracks)`. Also, `this.isSDK` is no longer needed?

src/js/providers/tracks-mixin.js
```diff
@@ -581,46 +591,42 @@ define(['../utils/underscore',
...
-                    } else {
-                        // make VTTCues from SRT track
-                        var cues = srt(xhr.responseText);
-                        var vttCues = _convertToVTTCues(cues);
-                        _addVTTCuesToTrack.call(this, track, vttCues);
-                    }
-                }.bind(this));
+                var responseText = xhr.responseText;
+
+                // TODO: parse SRT with using vttParser and deprecate srt module
```
hghazzi: Is there a tech debt ticket for this?

src/js/providers/tracks-mixin.js
```diff
@@ -97,28 +98,38 @@ define(['../utils/underscore',
...
+        this._itemTracks = itemTracks;
+
         this._renderNatively = _nativeRenderingSupported(this.getName().name);
-        if (this.isSDK || !tracks) {
+        if (!itemTracks || !itemTracks.length) {
             return;
         }
 
-        if (!_tracksAlreadySideloaded.call(this, tracks)) {
+        if (!alreadyLoaded) {
```
hghazzi: Were the checks in `_tracksAlreadySideloaded` unnecessary?

src/js/providers/tracks-mixin.js
```diff
@@ -351,6 +359,20 @@ define(['../utils/underscore',
...
+            if (xhr) {
+                xhr.onload = null;
+                xhr.onreadystatechange = null;
+                xhr.onerror = null;
+                if ('abort' in xhr) {
+                    xhr.abort();
+                }
+            }
+        });
+    }
```
hghazzi: There isn't an existing ajax utils methods that can take care of this?

src/js/providers/tracks-mixin.js
robwalch: They were the reason there were so many side effects and are no longer necessary since `this._itemTracks = null` was added to clearTracks.

src/js/providers/tracks-mixin.js
robwalch: The SDK check was a left over from an earlier version where that helped the player in the SDK. Currently this code doesn't work at all in SDK because they have their own provider so this check is not needed.

src/js/providers/tracks-mixin.js
robwalch: nope. we need one. This is it. It just needs to be exposed publicly in utils: https://github.com/jwplayer/jwplayer/search?utf8=%E2%9C%93&q=ajax+abort

src/js/providers/tracks-mixin.js
robwalch: No. Make one.

egreaves:
test this please

src/js/providers/tracks-mixin.js
hghazzi: https://github.com/jwplayer/jwplayer/issues/1343

src/js/providers/html5.js
```diff
@@ -462,7 +460,7 @@ define([
             if (sourceChanged || loadedSrc === 'none' || loadedSrc === 'started') {
                 _duration = duration;
                 _setVideotagSource(_levels[_currentQuality]);
-                _this.setupSideloadedTracks(_itemTracks);
+                _this.setupSideloadedTracks(this._itemTracks);
```
egreaves: use `_this` here

PR #1353, opened 2016-08-03
Fixed duration displaying NaN:NaN while loading

hghazzi:
When formatting time, we were not handling undefined and NaN correctly
as potential inputs.

JW7-2777

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -102,7 +102,7 @@ define([
 
                     /* falls through */
                 case events.JWPLAYER_MEDIA_META:
-                    var duration = data.duration;
+                    var duration = data.duration || 0;
                     if (_.isNumber(duration)) {
                         this.mediaModel.set('duration', duration);
                         this.set('duration', duration);
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -102,8 +102,8 @@ define([
 
                     /* falls through */
                 case events.JWPLAYER_MEDIA_META:
-                    var duration = data.duration || 0;
-                    if (_.isNumber(duration)) {
+                    var duration = data.duration;
+                    if (_.isNumber(duration) && !_.isNaN(duration)) {
                         this.mediaModel.set('duration', duration);
                         this.set('duration', duration);
                     }
--- a/src/js/polyfills/vtt.js
+++ b/src/js/polyfills/vtt.js
@@ -193,111 +193,6 @@
     }
   }
 
-  function parseCue(input, cue, regionList) {
-    // Remember the original input if we need to throw an error.
-    var oInput = input;
-    // 4.1 WebVTT timestamp
-    function consumeTimeStamp() {
-      var ts = parseTimeStamp(input);
-      if (ts === null) {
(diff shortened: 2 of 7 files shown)


hghazzi:
An oldie, but a goodie: https://www.destroyallsoftware.com/talks/wat

robwalch:
Try to find where this is coming from. Are `position` and `duration` undefined or NaN in the model by default?

hghazzi:
test this please.

hghazzi:
test this please

pajong:
@hghazzi This needs rebase, and the commit message should explain why we got NaN instead of a number.

src/js/controller/model.js
```diff
@@ -102,7 +102,7 @@ define([
 
                     /* falls through */
                 case events.JWPLAYER_MEDIA_META:
-                    var duration = data.duration;
+                    var duration = data.duration || 0;
```
pajong: It looks like we never tried to set the duration if ```data.duration``` is not a number. This changes the behavior, and probably fires an additional metadata event with a duration update. Can you explain which part passes ```NaN``` to the duration?

egreaves: @hghazzi - @pajong is right that this would now update the model to 0 in cases where the meta event has an invalid duration. The problem here was that `_.isNumber` evaluates to true for `NaN`.

hghazzi: @pajong @egreaves The provider changes I made below are where the duration is potentially being set to NaN in certain race conditions due to the browser. This is the common location where all providers pass through to set the duration, so I made this change to always ensure we have a default of 0 for invalid values. This is the same way we initialize the duration value to begin with, using `|| 0` as a fallback value.

pajong: @hghazzi It sounds like we do not need ``` _.isNumber``` check if we are making it default to 0 when duration is ```NaN```. Also, if we do it from the provider, why do we need to do the same thing here? We should probably just do it in here or in the providers, not duplicate the same behavior.

hghazzi: @pajong I think it is still valid to keep the `_.isNumber` since this event could be triggered from multiple providers and theres a chances that truthy non-numerical values could potentially be passed in. As for doing the check in the providers vs the model, it should definitely be in the model, but we know the bug occurred in HTML and Flash, so made the check there as well, to avoid the possibility that a NaN duration could affect other parts of the providers. Discussed this with @robwalch and I believe we agreed that this would be the correct way forward, without putting it in every provider.

egreaves: @hghazzi would it be safer, then, to remove `|| 0` and instead check the following before updating the duration in the model?
 `if (_.isNumber(duration) && !_.isNaN(duration)) {`

We shouldn't get `NaN` from the HTML5 or Flash providers anymore. But, if we do, we wont make unnecessary updates to the model that don't reflect the actual duration of the media being played.

hghazzi:
@pajong @egreaves @robwalch Updated the code so we don't trigger a duration change event if we receive a NaN from the model.

hghazzi:
test this please

PR #1354, opened 2016-08-04
Expose showControls function from views

pajong:
Allow instream adapter to have a function to show the control bar, so that we can show the control bar after a vpaid ad.
JW7-2762

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -1042,8 +1042,7 @@ define([
                 _instreamModel = null;
             }
             this.setAltText('');
-            utils.removeClass(_playerElement, 'jw-flag-ads');
-            utils.removeClass(_playerElement, 'jw-flag-ads-hide-controls');
+            utils.removeClass(_playerElement, ['jw-flag-ads', 'jw-flag-ads-hide-controls']);
             if (_model.getVideo) {
                 var provider = _model.getVideo();
                 provider.setContainer(_videoLayer);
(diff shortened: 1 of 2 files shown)


src/js/view/view.js
```diff
@@ -1027,6 +1027,10 @@ define([
             _controlbar.setAltText(text);
         };
 
+        this.showControls = function() {
+            utils.removeClass(_playerElement, 'jw-flag-ads-hide-controls');
+        };
```
robwalch: Now you can use `this.showControls()` in `this.destroyInstream` in place of `utils.removeClass(_playerElement, 'jw-flag-ads-hide-controls');`

or don't call utils.removeClass twice - it accepts arrays and space delimited strings:

`utils.removeClass(_playerElement, ['jw-flag-ads', 'jw-flag-ads-hide-controls']);`

src/js/view/view.js
```diff
@@ -1038,8 +1042,7 @@ define([
                 _instreamModel = null;
             }
             this.setAltText('');
-            utils.removeClass(_playerElement, 'jw-flag-ads');
-            utils.removeClass(_playerElement, 'jw-flag-ads-hide-controls');
+            utils.removeClass(_playerElement, ['jw-flag-ads', 'jw-flag-ads-hide-controls']);
```
robwalch: This is better because it only sets the element's `className` once.

PR #1355, opened 2016-08-04
Ensure that cues rendered with VTT.js are positioned correctly

egreaves:
Position cues within the video container when rendered with VTT.js. This ensures:

- Cue position is recalculated whenever the controlbar toggles or the video container's size changes
- Cues are placed in the next best line that fits within the container when 16 lines cannot be accommodated
- Cues with a negative line value are repositioned to fit within the container
- Cues always have a position value that's between 0-100%, factoring in the cueBox size

Also removed the `parseCue` function, which is used in `vttparser.js` and not needed in this file.

Fixes #
JW7-2861

### DIFF
--- a/src/js/polyfills/vtt.js
+++ b/src/js/polyfills/vtt.js
@@ -193,111 +193,6 @@
     }
   }
 
-  function parseCue(input, cue, regionList) {
-    // Remember the original input if we need to throw an error.
-    var oInput = input;
-    // 4.1 WebVTT timestamp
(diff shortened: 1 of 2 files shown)


robwalch:
Second test captions are cropped at the bottom by container at 28s
http://player-pr-test-jenkins.longtailvideo.com/builds/2565/archive/test/squash/bootstrap.html?primary=flash&edition=ads&feature=captions_config

PR #1362, opened 2016-08-09
Captions show after replaying and preroll

hghazzi:
Need to keep track on wether an ad is being played so that we can correctly persist the selected captions. This is done by saving the ad state in the model, and not just the overall state. Also needed to remove the tracks listener before we removed the cue so that it is not triggered unnecessarily.

JW7-2818

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -282,8 +282,12 @@ define([
                 loader.load(toLoad);
             }
 
+            function _getAdState() {
+                return _this._instreamAdapter && _this._instreamAdapter.getState();
+            }
+
             function _getState() {
-                var adState = _this._instreamAdapter && _this._instreamAdapter.getState();
+                var adState = _getAdState();
                 if (_.isString(adState)) {
                     return adState;
                 }
@@ -301,7 +305,7 @@ define([
                     return;
                 }
 
-                var adState = _this._instreamAdapter && _this._instreamAdapter.getState();
+                var adState = _getAdState();
                 if (_.isString(adState)) {
                     // this will resume the ad. _api.playAd would load a new ad
                     return _api.pauseAd(false);
@@ -377,7 +381,7 @@ define([
             function _pause() {
                 _actionOnAttach = null;
 
-                var adState = _this._instreamAdapter && _this._instreamAdapter.getState();
+                var adState = _getAdState();
                 if (_.isString(adState)) {
                     return _api.pauseAd(true);
                 }
@@ -641,6 +645,7 @@ define([
             this.getVisualQuality = _getVisualQuality;
             this.getConfig = _getConfig;
             this.getState = _getState;
+            this.getAdState = _getAdState;
 
             // Model passthroughs
             this.setVolume = _model.setVolume.bind(_model);
@@ -728,9 +733,8 @@ define([
             this.createInstream = function() {
                 this.instreamDestroy();
                 this._instreamAdapter = new InstreamAdapter(this, _model, _view);
-                // Persist the current captions track so the index
-                // is  set correctly after ad playback ends
-                _model.persistCaptionsTrack();
+                _model.set('adState', this.getAdState());
+
                 return this._instreamAdapter;
             };
 
@@ -738,12 +742,14 @@ define([
                 if (this._instreamAdapter) {
                     this._instreamAdapter.skipAd();
                 }
+                _model.set('adState', this.getAdState());
             };
 
             this.instreamDestroy = function() {
                 if (_this._instreamAdapter) {
                     _this._instreamAdapter.destroy();
                 }
+                _model.set('adState', this.getAdState());
             };
 
             _setup.start();
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -100,7 +100,7 @@ define([
                 case events.JWPLAYER_MEDIA_BUFFER:
                     this.set('buffer', data.bufferPercent);
 
-                    /* falls through */
+                /* falls through */
                 case events.JWPLAYER_MEDIA_META:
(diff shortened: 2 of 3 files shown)


src/js/controller/model.js
```diff
@@ -148,7 +148,11 @@ define([
                     this.setCurrentAudioTrack(data.currentTrack, data.tracks);
                     break;
                 case 'subtitlesTrackChanged':
-                    this.setVideoSubtitleTrack(data.currentTrack, data.tracks);
+                    var adState = this.get('adState');
+                    if (!_.isString(adState)) {
```
pajong: Please put a comment on why we need to do this.

src/js/controller/model.js
```diff
@@ -111,10 +111,10 @@ define([
 
                 case events.JWPLAYER_MEDIA_BUFFER_FULL:
                     // media controller
-                    if(this.mediaModel.get('playAttempt')) {
+                    if (this.mediaModel.get('playAttempt')) {
                         this.playVideo();
                     } else {
-                        this.mediaModel.on('change:playAttempt', function() {
+                        this.mediaModel.on('change:playAttempt', function () {
```
pajong: We don't put a space between ```function``` and ```()```

pajong:
@hghazzi Your commit message should explain what the problem was, and how the changes you made will solve the issue, so that the person reviewing the PR understands the issue and the solution.

hghazzi:
@pajong updated PR based on comments.

src/js/providers/tracks-mixin.js
```diff
@@ -295,6 +295,7 @@ define(['../utils/underscore',
         this._unknownCount = 0;
         this._activeCuePosition = null;
         if (this._renderNatively) {
+            this.removeTracksListener(this.video.textTracks, 'change', this.textTrackChangeHandler);
```
egreaves: I'd leave a comment here to make it clearer why we have to remove the listener before clearing the cues.

src/js/controller/controller.js
```diff
@@ -377,7 +381,7 @@ define([
             function _pause() {
                 _actionOnAttach = null;
 
-                var adState = _this._instreamAdapter && _this._instreamAdapter.getState();
+                var adState = _getAdState();
```
robwalch: I like this, but maybe it should always return a string?

src/js/controller/controller.js
```diff
@@ -728,22 +733,23 @@ define([
             this.createInstream = function() {
                 this.instreamDestroy();
                 this._instreamAdapter = new InstreamAdapter(this, _model, _view);
-                // Persist the current captions track so the index
-                // is  set correctly after ad playback ends
-                _model.persistCaptionsTrack();
+                _model.set('adState', this.getAdState());
```
robwalch: Instream has a reference to the model and should update the model with it's state when it's changes.

src/js/controller/controller.js
robwalch: nevermind. I see _instreamAdapter also returns false once destroyed.

src/js/controller/model.js
```diff
@@ -148,7 +148,13 @@ define([
...
                     break;
                 case 'subtitlesTrackChanged':
-                    this.setVideoSubtitleTrack(data.currentTrack, data.tracks);
+                    var adState = this.get('adState');
+
+                    // We only want to update the tracks ands persist the select track
+                    // when an ad is not playing
+                    if (!_.isString(adState)) {
+                        this.persistVideoSubtitleTrack(data.currentTrack, data.tracks);
+                    }
```
robwalch: Hmm... so the ads provider is emitting this event? There shouldn't be any captions to persist. I question why this is necessary.

src/js/controller/model.js
```diff
@@ -384,8 +390,8 @@ define([
 
         };
 
-        this.persistVideoSubtitleTrack = function(trackIndex) {
-            this.setVideoSubtitleTrack(trackIndex);
+        this.persistVideoSubtitleTrack = function(trackIndex, tracks) {
+            this.setVideoSubtitleTrack(trackIndex, tracks);
```
robwalch: Isn't setting the track one of the things that triggers "subtitlesTrackChanged"? Potential loop here.

src/js/controller/model.js
```diff
@@ -384,8 +390,8 @@ define([
...
             this.persistCaptionsTrack();
```
robwalch: Why does persistCaptionsTrack set the menu to "Off"? That should be the default on playlist item. It should not be set back to off based on tracks. it would be off already.

src/js/providers/tracks-mixin.js
```diff
@@ -295,6 +295,8 @@ define(['../utils/underscore',
         this._unknownCount = 0;
         this._activeCuePosition = null;
         if (this._renderNatively) {
+            // Removing listener first to ensure that removing cues does not trigger it unnecessarily
+            this.removeTracksListener(this.video.textTracks, 'change', this.textTrackChangeHandler);
```
egreaves: @robwalch we have to change the track mode to `hidden` to remove the cues, then set the mode to `disabled` after. This causes up to 2 unnecessary track change events if we don't remove the listener before.

egreaves:
This does not work for this [1st scenario](http://player-pr-test-jenkins.longtailvideo.com/builds/2646/archive/test/squash/bootstrap.html?primary=html5&edition=ads&feature=captions/captions_with_vast_linear) in iOS+iPhone.

hghazzi:
test this please

egreaves:
Closing this PR while the solution is being improved. The goal is to only listen for track changes when playing media that is not an ad. This will simplify the logic so that we can persist track changes that occur while in fullscreen on iOS.

PR #1364, opened 2016-08-09
Fix handling cues with missing end times

hghazzi:
When cues do not have end times, the start time of the following cue,
that does not have the same start time, should be used.

JW7-2854

### DIFF
--- a/src/js/view/captionsrenderer.js
+++ b/src/js/view/captionsrenderer.js
@@ -102,9 +102,26 @@ define([
         };
 
         this.getCurrentCues = function(allCues, pos) {
-            return _.filter(allCues, function (cue) {
-                return pos >= (cue.startTime) && (!cue.endTime || pos <= cue.endTime);
-            });
+            function getEndTime(cue, i) {
+                if (cue.endTime ||  _.last(allCues) === cue) {
+                    return cue.endTime;
+                }
+
+                if (cue.startTime < allCues[i+1].startTime) {
+                    return allCues[i+1].startTime;
+                }
+
+                return getEndTime(cue, i+1);
+            }
+
+            return _.reduce(allCues, function (cues, cue, i) {
+                var endTime = getEndTime(cue, i);
+
+                if (pos >= cue.startTime && (!endTime || pos <= endTime)) {
+                    cues.push(cue);
+                }
+                return cues;
+            }, []);
         };
 
         this.updateCurrentCues = function(cues) {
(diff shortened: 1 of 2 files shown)


src/js/view/captionsrenderer.js
```diff
@@ -102,9 +102,26 @@ define([
...
+            function getEndTime(cue, i) {
+                if (cue.endTime ||  _.last(allCues) === cue) {
+                    return cue.endTime;
+                }
+
+                if (cue.startTime < allCues[i+1].startTime) {
+                    return allCues[i+1].startTime;
+                }
+
+                return getEndTime(cue, i+1);
```
hghazzi: If using recursion here is a concern, I can make it iterative. It just seemed a lot cleaner this way.

test/unit/captionsrenderer-test.js
```diff
@@ -21,9 +21,45 @@ define([
...
+        var currentNumCues = [1, 1, 1, 1, 1, 1, 0, 1, 1, 1, 2, 1, 2, 4, 4, 1, 1, 1, 1, 0, 0, 0];
+
+        for (var i = 0; i < currentNumCues.length; i += 1) {
+            assert.equal(captionsRenderer.getCurrentCues(allCues, i).length, currentNumCues[i], 'Invalid number of cues at position ' + i);
+        }
+    });
+
+    test('should show the correct number of cues at any given position in time with some missing end times', function (assert) {
+        var allCues = [
+            new VTTCue(0, undefined, 'HG: Morning, Rob.'),
```
egreaves: @hghazzi using `undefined` throws a TypeError in Chrome and sets `endTime` to `NaN` in Safari. A more realistic test is just using 0 (or `null`, which also sets `endTime` to 0 in Chrome and Safari), as that was the case in the stream where the bug was first discovered. Also, our parser returns 0 for invalid input.

robwalch:
Fix the parsing issue in Flash https://github.com/jwplayer/jwplayer-commercial/blob/master/src/flash/com/longtailvideo/jwplayer/parsers/SRT.as#L55

I don't want to add work arounds here.

hghazzi:
@robwalch Do you think it might still be worth keeping this code as part of the fix to handle any potential edge case where end times are no available? Or should we maybe throw an exception since there should always be an end time?

robwalch:
Throwing an exception only in the debug player would be ok. Release builds should fail silently. There are examples of this in different parts of the player - See utils.tryCatch

I'd rather not hide the problem even for edge cases.

PR #1366, opened 2016-08-10
Update the controlbar for the new layout

egreaves:
This PR updates the controlbar with the following changes:
- Remove visual playlist and all supporting code
- Remove the previous button
- Place the next button after the duration
- Delete playlist container css rules from skins

Fixes #
JW7-2518


egreaves:
test this please

johnBartos:
Player size: smaller 👍

src/css/imports/icons.less
```diff
@@ -135,11 +135,4 @@
     &:before {
```
johnBartos: Can we get rid of the prev icon too?

src/js/view/controlbar.js
```diff
@@ -302,23 +280,15 @@ define([
         onPlaylist : function(model, playlist) {
             var display = (playlist.length > 1);
             this.elements.next.toggle(display);
-            this.elements.prev.toggle(display);
-            if(this.elements.playlist) {
-                this.elements.playlist.setup(playlist, model.get('item'));
-            }
         },
-        onPlaylistItem : function(model/*, item*/) {
+        onPlaylistItem : function(/*model, item*/) {
```
johnBartos: I think we should just remove these comments, or at least move them out of the signature

src/css/imports/icons.less
egreaves: Yep - good catch.

PR #1369, opened 2016-08-12
Support IMA HTML5 SDK over Flash Provider

robwalch:
In order to use the html5 IMA sdk over flash, our insteam-flash module needs to update playback state the same way that the insteam-html5 does.

In this code "provider" is actually the Google IMA or VAST plugin that passes a reference of itself to the instream adapter via `applyProviderListeners`. Listening to play/pause state events and updating the adModel keeps the player in sync with the state of the ad played by the plugin.

Other notes on changes:
- model listeners for instream-adapter where not being detached when insteam is destroyed (`_model.off(null, null, _instream)`)
- made state handler case statement fall-through per @egreaves suggestion

JW7-2640

### DIFF
--- a/src/js/controller/instream-flash.js
+++ b/src/js/controller/instream-flash.js
@@ -51,16 +51,7 @@ define([
                 }, this);
             }
 
-            this.swf.on('instream:state', function(evt) {
-                switch (evt.newstate) {
-                    case states.PLAYING:
-                        this._adModel.set('state', evt.newstate);
-                        break;
-                    case states.PAUSED:
-                        this._adModel.set('state', evt.newstate);
-                        break;
-                }
-            }, this)
+            this.swf.on('instream:state', this.stateHandler, this)
             .on('instream:time', function(evt) {
                 this._adModel.set('position', evt.position);
                 this._adModel.set('duration', evt.duration);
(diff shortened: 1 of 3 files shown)


robwalch:
test this please

src/js/controller/instream-flash.js
```diff
@@ -82,9 +73,24 @@ define([
...
+                provider.on(events.JWPLAYER_PLAYER_STATE, this.stateHandler, this);
             };
         },
 
+        stateHandler: function(evt) {
+            switch (evt.newstate) {
+            case states.PLAYING:
+                this._adModel.set('state', evt.newstate);
+                break;
+            case states.PAUSED:
```
egreaves: Any reason not to fall-through here since both states are handled the same?

robwalch: No. That's a good idea.

robwalch:
@dannyfinks Without these changes, https://github.com/jwplayer/jwplayer-ads-googima/pull/116 should not pass all GoogIMA Flash provider cucumber tests. Maybe have a look at http://player-pr-test-jenkins.longtailvideo.com/builds/2744/archive/test/ and the automation for GoogIMA.

dannyfinks:
@robwalch could be a lack of flash tests compared to html5...I will dig in to find some gaps

robwalch:
Just the buffering... as far as I can tell so far.

pajong:
@dannyfinks This is good to be merged if it looks good on your side.

dannyfinks:
@pajong this looks good to me as far as the changes here. I am just waiting on this to get merged to follow up more on the IMA Plugin changes.

PR #1371, opened 2016-08-15
Don't detach resize event handlers if the player isn't found on the page

johnBartos:
In Android Chrome, an `orientationchange` event will fire right as the page loads and before the player is attached to the DOM. This causes the `orientationchange` event handler to become detached (meaning it will never fire again). After some discussion we came to the conclusion that detaching the handlers isn't necessary because the `destroy` method detaches the handlers when the player is destroyed.

JW7-2910

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -263,6 +263,7 @@ define([
             if (document.body.contains(_playerElement)) {
                 _cancelDelayResize(_resizeContainerRequestId);
                 _resizeContainerRequestId = _delayResize(_setContainerDimensions);
+            }
         }
 
         // Set global colors, used by related plugin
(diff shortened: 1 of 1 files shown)


robwalch:
So when do they get detached? If you call setup on an existing player or remove one, the listener should be removed.

johnBartos:
@robwalch Before handlers are attached they're removed:

````
            window.removeEventListener('resize', _responsiveListener);
            window.addEventListener('resize', _responsiveListener, false);
            if (_isMobile) {
                window.removeEventListener('orientationchange', _responsiveListener);
                window.addEventListener('orientationchange', _responsiveListener, false);
            }
````

PR #1373, opened 2016-08-16
Persist captions after showing ads

hghazzi:
Whenever an ad plays, we need to not update the captions tracks, which occurs due to the listener being added to the providers. Added logic to avoid adding the listener to the provider when an ad starts playing.

JW7-2818


src/js/controller/controller.js
```diff
@@ -728,9 +732,6 @@ define([
             this.createInstream = function() {
                 this.instreamDestroy();
                 this._instreamAdapter = new InstreamAdapter(this, _model, _view);
-                // Persist the current captions track so the index
-                // is  set correctly after ad playback ends
-                _model.persistCaptionsTrack();
```
hghazzi: This is no longer needed with this fix.

src/js/providers/tracks-mixin.js
```diff
@@ -97,7 +97,8 @@ define(['../utils/underscore',
             }
         }
 
-        if (this._renderNatively) {
+        // If provider is being used to play an ad, don't add listener, so tracks won't change.
+        if (this._renderNatively && !this.instreamMode) {
```
egreaves: Checking if we're in `instreamMode` here means we may not end up removing an existing listener. It's important that we remove the listener at all times, but only add it when we're not in `instreamMode`.

src/js/controller/model.js
```diff
@@ -383,8 +388,8 @@ define([
 
         };
 
-        this.persistVideoSubtitleTrack = function(trackIndex) {
-            this.setVideoSubtitleTrack(trackIndex);
+        this.persistVideoSubtitleTrack = function(trackIndex, tracks) {
+            this.setVideoSubtitleTrack(trackIndex, tracks);
```
robwalch: Do we need to call `setVideoSubtitleTrack` here?

src/js/controller/model.js
```diff
@@ -219,6 +220,10 @@ define([
             _provider.volume(_this.get('volume'));
             _provider.mute(_this.get('mute'));
             _provider.on('all', _videoEventHandler, this);
+
+            if (this.get('instreamMode') === true) {
+                _provider.instreamMode = true;
```
robwalch: I wonder if we could pass this in to `changeVideoProvider` or `loadVideo`?

hghazzi: Took a look, and still seems like this would be the best place to contain both the setting and removal of the `instreamMode` flag

src/js/providers/tracks-mixin.js
```diff
@@ -102,7 +102,11 @@ define(['../utils/underscore',
...
             // Only bind and set this.textTrackChangeHandler once so that removeEventListener works
             this.textTrackChangeHandler = this.textTrackChangeHandler || textTrackChangeHandler.bind(this);
-            this.addTracksListener(this.video.textTracks, 'change', this.textTrackChangeHandler);
+
+            this.removeTracksListener(this.video.textTracks, 'change', this.textTrackChangeHandler);
+            if (!this.instreamMode) {
+                this.addTracksListener(this.video.textTracks, 'change', this.textTrackChangeHandler);
+            }
 
             if (utils.isEdge()) {
```
hghazzi: @egreaves I know this is something you just added to fix another bug, but does it need a `removeTracksListener`, and should it also check `this.instreamMode`?

egreaves: @hghazzi, yeah, it does. We should always remove existing listeners before adding one; there are side effects if we don't. Until we have to worry about displaying captions during ads, we don't need to listen to track changes while playing an ad. Also, one of my [PRs](https://github.com/jwplayer/jwplayer/pull/1374/files#diff-09efbcfa88512d3674c2d08e2b8a9496R275) that got merged yesterday made some updates to removing and adding listeners. You'll need to rebase to get those changes.

src/js/controller/model.js
egreaves: @robwalch @hghazzi yeah, we do. This behavior hasn't changed. When the cc menu selection changes, `setVideoSubtitleTrack` tells the provider what track should be selected. In the case of `TextTrack.change` events where the provider notifies the model, the provider returns early when `_provider.setSubtitlesTrack(trackIndex);` gets called in `setVideoSubtitleTrack `.

egreaves:
test this please

PR #1374, opened 2016-08-16
Show 608 captions in Edge

egreaves:

- Expand native rendering to Microsoft Edge. Since Edge has native HLS support, we can get 608/metadata TextTracks from the video element.
- Listen to the `addtrack` event in Edge since tracks aren't available when the `onloadeddata` event is raised. 

Fixes #
JW7-2494

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -177,5 +177,10 @@ define([
         return majorVersion;
     };
 
+    browser.nativeRenderingSupported = function (providerName) {
+        return providerName.indexOf('flash') === -1 &&
+        (browser.isChrome() || browser.isIOS() || browser.isSafari() || browser.isEdge());
+    };
+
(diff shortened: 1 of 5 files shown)


src/js/utils/browser.js
```diff
@@ -177,5 +177,10 @@ define([
         return majorVersion;
     };
 
+    browser.nativeRenderingSupported = function (providerName) {
+        return providerName.indexOf('flash') === -1 &&
+        (browser.isChrome() || browser.isIOS() || browser.isSafari() || browser.isEdge());
+    };
+
```
robwalch: This should not be a util. Ideally this would be set on the config level, and the default setting would be based on these rules.

robwalch: For not make it a module, not a util. Require where needed.

PR #1376, opened 2016-08-17
Convert VTTCue to TextTrackCue for rendering captions in Edge

egreaves:
Since Edge has support for HLS, the player now uses native rendering for 608 captions tracks. This means that cues from sideloaded tracks need to be of type `TextTrackCue` when  added to a TextTrack.

Fixes #
JW7-2678

### DIFF
--- a/src/js/providers/tracks-mixin.js
+++ b/src/js/providers/tracks-mixin.js
@@ -402,7 +402,7 @@ define(['../utils/underscore',
     //////////////////////
 
     function _addCueToTrack(track, vttCue) {
-        if (!utils.isEdge()) {
+        if (!utils.isEdge() || !window.TextTrackCue) {
             track.addCue(vttCue);
             return;
         }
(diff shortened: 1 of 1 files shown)


src/js/providers/tracks-mixin.js
```diff
@@ -401,6 +401,18 @@ define(['../utils/underscore',
     ////// PRIVATE METHODS
     //////////////////////
 
+    function _addCueToTrack(track, vttCue) {
+        if (!utils.isEdge()) {
```
hghazzi: Why did you change this from checking if VTTCue exists to checking if its edge.

egreaves: @hghazzi `TextTrackCue` also exists in browsers with the `VTTCue` interface, so that wasn't a reliable check. We only need to use `TextTrackCue` in Edge. In IE11, we polyfill `VTTCue` and render with `VTT.js`.

hghazzi: ok

robwalch: But you don't know for sure that it exists in all environments that return true for `isEdge` so it would be safer make sure we're not in Edge or don't have TextTrackCues before trying to use them.

`if (!utils.isEdge() || !window.TextTrackCue) {`

PR #1381, opened 2016-08-23
Replace html breaks with line breaks for proper VTTCue rendering of DFXP subtitles.

johanberonius:
Replace html breaks with line breaks for proper VTTCue rendering of DFXP subtitles.

This fixes that br-elements are ignored and multi line subtitles are displayed on a single line with no space in between.

".*?" in the RegExp is needed since the element may contain a xmlns-attribute.

See this change to the SRT-parser for reference: https://github.com/jwplayer/jwplayer/commit/d10b4faf738c0a2e4690805c2cdbc8adf0c7fae5

### DIFF
--- a/src/js/parsers/captions/dfxp.js
+++ b/src/js/parsers/captions/dfxp.js
@@ -30,7 +30,7 @@ define([
         for (var i = 0; i < paragraphs.length; i++) {
             var p = paragraphs[i];
             var rawText = (p.innerHTML || p.textContent || p.text || '');
-            var text = strings.trim(rawText).replace(/>\s+</g, '><').replace(/tts?:/g, '');
+            var text = strings.trim(rawText).replace(/>\s+</g, '><').replace(/tts?:/g, '').replace(/<br.*?\/>/g, '\r\n');
             if (text) {
                 var begin = p.getAttribute('begin');
                 var dur = p.getAttribute('dur');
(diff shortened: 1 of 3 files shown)


robwalch:
Thanks @johanberonius! We'll merge after verifying the change.

src/js/parsers/captions/dfxp.js
```diff
@@ -30,7 +30,7 @@ define([
         for (var i = 0; i < paragraphs.length; i++) {
             var p = paragraphs[i];
             var rawText = (p.innerHTML || p.textContent || p.text || '');
-            var text = strings.trim(rawText).replace(/>\s+</g, '><').replace(/tts?:/g, '');
+            var text = strings.trim(rawText).replace(/>\s+</g, '><').replace(/tts?:/g, '').replace(/<br.*?\/>/g, '\r\n');
```
hghazzi: @johanberonius Looks good. We have some unit tests for this already at `dxfp-test.js`, could you add some more to verify the cases you mentioned? Also, feel free to rename the test file to `dfxp-test.js`, just be sure to update it in `test/tests.js` 😸

johanberonius:
Added a test and renamed the file.

Also, discovered another bug where text content can be removed. 
See this comment: https://github.com/jwplayer/jwplayer/commit/1574b0d7bc4b9cdeecbf5a85273cea5a1da4054c#diff-6cc80f30e8fecb522d563e0fba5485eeR25

hghazzi:
Aweomse! Thanks, @johanberonius!

hghazzi:
Merging in your changes, @johanberonius could you create a separate issue regarding the bug you found?

PR #1383, opened 2016-08-23
Quick fix for mobile fullscreen icon

pajong:
After removing small UI, the icons were too big to fit on fullscreen android.
Making a quick fix before addressing this better with feed UI so that we do not display duration and time elapsed on mobile fullscreen small size.
JW7-2933

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -824,10 +824,10 @@ define([
             }
 
             _captionsRenderer.resize();
-            if (_isMobile) {
-                var hideTimeInControl = width && width < _minWidthForTimeDisplay;
-                utils.toggleClass(_playerElement, 'jw-flag-compact-player', hideTimeInControl);
-            }
+            
+            var hideTimeInControl = width && width < _minWidthForTimeDisplay;
+            utils.toggleClass(_playerElement, 'jw-flag-compact-player', hideTimeInControl);
+
         }
 
         this.resize = function(width, height) {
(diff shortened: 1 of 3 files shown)


src/js/view/controlbar.js
```diff
@@ -361,6 +364,16 @@ define([
...
             utils.toggleClass(this.elements.fullscreen.element(), 'jw-off', val);
         },
+        checkControlSize : function(containerWidth) {
+            this.containerWidth = containerWidth || this.containerWidth;
+            if (this.containerWidth && this.containerWidth < this._minWidthForTimeDisplay) {
+                this.elements.elapsed.style.display = 'none';
+                this.elements.duration.style.display = 'none';
+            } else {
+                this.elements.elapsed.style.display = '';
+                this.elements.duration.style.display = '';
```
robwalch: Isn't this already done elsewhere or was that code removed?

Can we please make a css flag for this "" instead of setting it inline? it looks like this is still in the player and we could continue using it by toggling the flag https://github.com/jwplayer/jwplayer/blob/2d59678c4cc70fb8c98b05dfcea24e8a95d0dab0/src/css/flags/compact-player.less

src/js/view/view.js
```diff
@@ -823,6 +824,10 @@ define([
             }
 
             _captionsRenderer.resize();
+            if (_isMobile) {
```
robwalch: Why does it matter if the environment is mobile? It's the same ui.

pajong: I had this check because we enlarged the icon size for mobile and that was causing it to cut off.
But I think you are right that we should still hide the time on small players.
Updated the PR.

PR #1385, opened 2016-08-24
Merge changes for caterpillar v2

robwalch:
Updates for better tracks and API support in HLS and HTML5 providers.

I made a few updates to HTML5 based on issues found while testing HLS in Safari:
- Fixed JS exception in visualquality handler fired after playback ended
- Fixed "bufferChange" event fired with duration value of 0


robwalch:
@egreaves @dannyfinks any reason we can't merge these changes into dev before the commercial cat-v2 work?

We want to make sure these changes don't cause regressions. If anything they should improve current behavior of providers.

dannyfinks:
@robwalch i'll verify in the pr builder if their has been any impact today.

egreaves:
@robwalch @dannyfinks The tracks specific changes allow caterpillar in FF to support non-native rendering of captions while preserving native behavior for ID3 metadata events. These changes shouldn't affect existing behavior in Chrome. Danny, I'm ok with merging this PR once you've verified there's been no impact.

PR #1388, opened 2016-08-29
DFXP captions parser: Namespace prefixes are removed from opening and closing tags but not from text content.

johanberonius:

Only remove XML namespace prefix from parsed captions text if it is part of an opening or closing tag.

Fixes #

That any text similar to namespace prefixes will be removed from regular text content.

### DIFF
--- a/src/js/parsers/captions/dfxp.js
+++ b/src/js/parsers/captions/dfxp.js
@@ -30,7 +30,7 @@ define([
         for (var i = 0; i < paragraphs.length; i++) {
             var p = paragraphs[i];
             var rawText = (p.innerHTML || p.textContent || p.text || '');
-            var text = strings.trim(rawText).replace(/>\s+</g, '><').replace(/tts?:/g, '').replace(/<br.*?\/>/g, '\r\n');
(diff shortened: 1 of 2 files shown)


src/js/parsers/captions/dfxp.js
```diff
@@ -30,7 +30,7 @@ define([
         for (var i = 0; i < paragraphs.length; i++) {
             var p = paragraphs[i];
             var rawText = (p.innerHTML || p.textContent || p.text || '');
-            var text = strings.trim(rawText).replace(/>\s+</g, '><').replace(/tts?:/g, '').replace(/<br.*?\/>/g, '\r\n');
+            var text = strings.trim(rawText).replace(/>\s+</g, '><').replace(/(<\/?)tts?:/g, '$1').replace(/<br.*?\/>/g, '\r\n');
```
johnBartos: Do you mind writing a comment for what this does?

johanberonius: See https://github.com/jwplayer/jwplayer/commit/1574b0d7bc4b9cdeecbf5a85273cea5a1da4054c#diff-6cc80f30e8fecb522d563e0fba5485eeR25

PR #1391, opened 2016-08-31
Next Up Display

egreaves:
Give users a visual indication of which item will be played next.

- Display a tooltip for the NextUp item when hovering over the `next` button.
- Keep the NextUp tooltip open (opened state) when playback position passes the `nextupoffset` (default: 10s before playback ends).
- Allow the user to close the NextUp tooltip when it is in the opened state.
- Play the NextUp item when the user clicks/taps on the tooltip and close the tooltip.
- Close the tooltip if the user seeks back to a position that is before the `nextupoffset`.
- In playlist mode, NextUp shows the next playlist item.
- In related mode, NextUp is only shown when autoplaying and the autoplay timer is set to 0. NextUp shows the same item that would be seen in the "Next Up" area if the overlay were open. 
- Do not show the NextUp tooltip when playing ads.

Fixes #
JW7-2511, JW7-2896

### DIFF
--- a/src/js/view/components/nextuptooltip.js
+++ b/src/js/view/components/nextuptooltip.js
@@ -1,38 +1,61 @@
 define([
     'utils/dom',
     'utils/ui',
-    'view/components/tooltip',
+    'utils/underscore',
     'utils/helpers',
     'templates/nextup.html'
-], function(dom, UI, Tooltip, utils, nextUpTemplate) {
-    var NextUpTooltip = Tooltip.extend({
-        'constructor' : function(_model, _api, nextEl, ariaText) {
-            this._model = _model;
-            this._api = _api;
-            this.sticky = false;
-            this.nextEl = nextEl;
-            this._nextUpText = ariaText || 'Next Up';
+], function(dom, UI, _, utils, nextUpTemplate) {
+    var NextUpTooltip = function(_model, _api, nextButton) {
+        this._model = _model;
+        this._api = _api;
+        this._nextButton = nextButton;
 
+        this.nextUpText = _model.get('localization').nextup;
+        this.state = 'tooltip';
+        this.relatedMode = false;
+
+        this.setup();
+    };
+
+    _.extend(NextUpTooltip.prototype, {
+        setup: function() {
             this.container = document.createElement('div');
             this.container.className = 'jw-nextup-container jw-reset';
+            var element = utils.createElement(nextUpTemplate());
+            this.addContent(element);
             this.hide();
-            this.openClass = 'jw-open';
 
-            this.onMediaModel(this._model, this._model.get('mediaModel'));
-            this._api.onPlaylistItem(this.onPlaylistItem.bind(this));
+            var relatedBlock = this._model.get('related');
+
+            // Next Up is always shown for playlist items
+            this.showNextUp = true;
+
+            // If there are related items, Next Up is shown when we're autoplaying and the timer is set to 0
+            if (relatedBlock) {
+                this.showNextUp = relatedBlock.oncomplete === 'autoplay' && relatedBlock.autoplaytimer === 0;
+            }
+
+            this.reset();
+            
+            // Events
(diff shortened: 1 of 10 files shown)


src/js/view/components/nextuptooltip.js
```diff
@@ -0,0 +1,238 @@
...
+        this.setup();
+    };
+
+    _.extend(NextUpTooltip.prototype, {
+        setup: function() {
+            this.container = document.createElement('div');
+            this.container.className = 'jw-nextup-container jw-reset';
+            var element = utils.createElement(nextUpTemplate());
+            this.addContent(element);
+            this.hide();
```
johnBartos: Is this necessary?

src/js/view/components/nextuptooltip.js
```diff
@@ -0,0 +1,238 @@
...
+        },
+        show: function() {
+            if (this.state === 'tooltip' || this.state === 'closed') {
+                dom.addClass(this.container, 'jw-nextup-container-visible');
+            }
+        },
+        hide: function(evt) {
+            // Hovering over the next button should not show/hide NextUp if it is in the opened state
+            if (this.state === 'opened' && evt.currentTarget === this._nextButton) {
```
johnBartos: how about a `hoverOut` method instead of checking for the target?

src/js/view/components/nextuptooltip.js
```diff
@@ -0,0 +1,238 @@
...
+        },
+        setNextUpItem: function(nextUpItem) {
+            if (!nextUpItem) {
+                this.reset();
+                return;
+            }
+
+            this.nextUpItem = nextUpItem;
+
+            this.closeButton = this.content.getElementsByClassName('jw-nextup-close')[0];
```
johnBartos: use `document.querySelector` instead of `getElementsByClassName()[0])`

src/js/view/components/nextuptooltip.js
```diff
@@ -0,0 +1,238 @@
...
+            this._model.on('change:duration', this.onDuration, this);
+
+            var playlist = this._model.get('playlist');
+            var nextUpIndex = (item.index + 1) % playlist.length;
+            var nextUpItem = playlist[nextUpIndex];
+            this.relatedMode = false;
+            this.setNextUpItem(nextUpItem);
+
+            this._related = this._related || this._api.getPlugin('related');
+            if (this._related && item.index === playlist.length - 1) {
```
johnBartos: Related only fires the `playlist` event when in related mode

src/js/view/components/nextuptooltip.js
```diff
@@ -0,0 +1,238 @@
...
+            // - in playlist mode but not playing an ad
+            // - autoplaying in related mode and autoplaytimer is set to 0
+            if (this.showNextUp && (val >= this.offset)) {
+                this.showTilEnd();
+            } else if (this.state === 'opened' || this.state === 'closed') {
+                this.state = 'tooltip';
+                this.hide();
+            }
+        },
+        element: function() {
```
johnBartos: is element necessary if it's actually just container?

src/js/view/components/nextuptooltip.js
```diff
@@ -0,0 +1,238 @@
...
+    var NextUpTooltip = function(_model, _api, nextButton) {
+        this._model = _model;
+        this._api = _api;
+        this._nextButton = nextButton;
+
+        this.nextUpText = _model.get('localization').nextup;
+        this.state = 'tooltip';
+        this.relatedMode = false;
+
+        this.setup();
```
johnBartos: can you call setup from the client code instead of the constructor?

src/js/view/components/nextuptooltip.js
egreaves: @johnBartos It's used in the view: 
`_controlsLayer.appendChild(_nextuptooltip.element());`

PR #1392, opened 2016-08-31
Check for null before checking the provider

pajong:
When providers.choose returns null, we need to check that before checking the chosen provider.
JW7-2841

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -248,7 +248,8 @@ define([
 
         // Give the option for a provider to be forced
         this.chooseProvider = function(source) {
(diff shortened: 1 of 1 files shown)


src/js/controller/model.js
```diff
@@ -248,7 +248,8 @@ define([
 
         // Give the option for a provider to be forced
         this.chooseProvider = function(source) {
-            return _providers.choose(source).provider;
```
robwalch: @pajong this only happens if there is something wrong with playlist filtering. If you see this happen, you need to fix the provider-supports filtering so that the item is filtered out.

PR #1398, opened 2016-09-04
Upgrade minimum required flash version for video acceleration support

robwalch:
Always use StageVideo. Setup is simplified with Flash 16 since StageVideo is always available.

http://blogs.adobe.com/flashplayer/2015/04/best-practices-for-high-performing-and-efficient-flash-video.html


robwalch:
Required FLEX/AIR SDK player lib is not available on Jenkins:
`>> /opt/flex/frameworks/flex-config.xml(56): Error: unable to open 'libs/player/16.1/playerglobal.swc'`

PR #1399, opened 2016-09-04
Make resuming a live stream in Safari behave like other providers

robwalch:
When resuming an HLS stream in JS and Flash playback we restart or seek to the end of the stream to keep the stream "Live" and in sync with the segments available in the live manifest.

Safari does not do this natively. Instead, it continues to buffer video while paused so that playback can be resumed from the point it was paused. It does this for a few minutes, but eventually it seeks ahead to keep up with the live stream.

Here, to make the behavior consistent, when video is resumed in Safari, if it's fallen more than 15 seconds behind the end, we'll seek ahead.

Users on iOS and Safari won't be able to take short breaks while watching live streams like they used to. Publishers who want their users to be able to pause for periods longer than 10 seconds and resume video from the same point should serve DVR streams with more than 120 seconds of video in their HLS manifests.

JW7-2297

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -654,7 +654,7 @@ define([
                     var seekableDuration = end - _getSeekableStart();
                     var isLiveNotDvr = seekableDuration < MIN_DVR_DURATION;
                     var behindLiveEdge = end - _videotag.currentTime;
-                    if (isLiveNotDvr && end && behindLiveEdge > 15) {
+                    if (isLiveNotDvr && end && (behindLiveEdge > 15 || behindLiveEdge < 0)) {
                         // resume playback at edge of live stream
                         _videotag.currentTime = Math.max(end - 10, end - seekableDuration);
                     }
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -636,12 +640,27 @@ define([
...
             clearTimeout(_playbackTimeout);
             _videotag.pause();
+            _beforeResumeHandler = function() {
+                var unpausing = _videotag.paused && _videotag.currentTime;
+                if (unpausing && _videotag.duration === Infinity) {
+                    var end = _getSeekableEnd();
+                    var seekableDuration = end - _getSeekableStart();
+                    var isLiveNotDvr = seekableDuration < MIN_DVR_DURATION;
+                    var behindLiveEdge = end - _videotag.currentTime;
+                    if (isLiveNotDvr && end && behindLiveEdge > 15) {
```
egreaves: Safari's seekable range may reset to  0 - n, where n < `_videotag.currentTime`, which would cause `behindLiveEdge` to be negative. This stalls the player until the seekable range catches up to the `_videotag.currentTime` value. We should also resume from the edge of the live stream if `behindLiveEdge < 0`.

robwalch: @egreaves Let's do it. Please add an additional commit to `feature/resume-html5-live` adds the `|| behindLiveEdge < 0` check.

PR #1402, opened 2016-09-08
Set defaults for Next Up, Playlist and Related titles

egreaves:
Set default localization values and allow customization of the Playlist, Related and Next Up titles.

Fixes #
JW7-2894

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -39,7 +39,9 @@ define([
             liveBroadcast: 'Live broadcast',
             loadingAd: 'Loading ad',
             rewind: 'Rewind 10s',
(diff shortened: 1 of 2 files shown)


src/js/api/config.js
```diff
@@ -39,7 +39,9 @@ define([
             liveBroadcast: 'Live broadcast',
             loadingAd: 'Loading ad',
             rewind: 'Rewind 10s',
-            nextup: 'Next Up'
+            nextuptext: 'Next Up',
```
johnBartos: 1) this should be camelCase like the rest of the props
2) Why include the text suffix? Nothing else does that - it's implicit

PR #1407, opened 2016-09-09
Add freewheel style and update pauseHandler

pajong:
Adding styles for freewheel plugin.
Updating pauseHandler for html5 provider to return out when pause is called in IDLE state.
JW7-2979


src/js/providers/html5.js
```diff
@@ -363,7 +363,7 @@ define([
 
         function _pauseHandler() {
```
robwalch: What about
```js
 if (!_attached) {
    return;
}
```
?

I think there's a reason we're only looking at 'complete'. If the video enters the paused state it's active.

src/css/states/idle.less
```diff
@@ -21,4 +21,9 @@
     .jw-title {
         display: block;
     }
+    .jw-media {
+        video {
+            pointer-events: none;
+        }
+    }
```
robwalch: If this is only a fix for Freewheel, it should go in ads.less 
```css
.jw-flag-ads-freewheel.jw-state-idle {
  .jw-media {
```

pajong:
test this please

PR #1410, opened 2016-09-11
Hide dock buttons when in controlbar only mode

egreaves:

Fixes #
JW7-3047


egreaves:
test this please

PR #1417, opened 2016-09-13
Do not polyfill VTTCue / Made it a module of VTTParser

robwalch:
Prevents exceptions in IE/Edge when Shaka attempts to use polyfilled VTTCue.
JW7-3075


src/js/providers/tracks-mixin.js
```diff
@@ -430,11 +430,11 @@ define(['../utils/underscore',
     //////////////////////
 
     function _addCueToTrack(track, vttCue) {
-        if (!utils.isEdge() || !window.TextTrackCue) {
+        if (!utils.isEdge() || !utils.isTrident() || !window.TextTrackCue) {
```
hghazzi: Is this the fix, and the rest of the changes are just moving the polyfill?

robwalch: No. The fix is to not add `VTTCue` to `window` because that makes Shaka use VTTCues when the browser does not support adding JavaScript object instances to TextTracks.

This just makes sure we don't make the same mistake in IE11.

PR #1418, opened 2016-09-13
Add minDvrWindow item-level config, create new model prop streamType

johnBartos:
- Add `minDvrWindow` on the item level with a default of 120
- New model prop: `minDvrWindow` (number) set on setting an item
- New model prop: `streamType` (`VOD`, `DVR`, or `LIVE`)
- `streamType` is set `onDuration` in the view
- Discontinue use of `utils.adaptiveType` except when determining `streamType` for the model
- Rename instances of `adaptiveType` to `streamType`
- Add tests 

Fixes #
JW7-2504 (1/2)

https://github.com/jwplayer/jwplayer-commercial/pull/2404 (2/2)

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -270,8 +270,8 @@ define([
             this.mediaModel = new MediaModel();
             this.set('mediaModel', this.mediaModel);
             this.set('position', item.starttime || 0);
-            this.set('duration', (item.duration && utils.seconds(item.duration)) || 0);
             this.set('minDvrWindow', item.minDvrWindow);
+            this.set('duration', (item.duration && utils.seconds(item.duration)) || 0);
             this.setProvider(item);
         };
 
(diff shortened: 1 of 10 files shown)


src/js/controller/model.js
```diff
@@ -271,7 +271,7 @@ define([
             this.set('mediaModel', this.mediaModel);
             this.set('position', item.starttime || 0);
             this.set('duration', (item.duration && utils.seconds(item.duration)) || 0);
-
+            this.set('minDvrWindow', item.minDvrWindow);
```
robwalch: Should this be set before duration, incase `item.duration != 0` and `change:duration` will determine `streamType`?

src/js/utils/parser.js
```diff
@@ -139,18 +139,21 @@ define([
 
     /**
      * Determine the adaptive type
+     * Duration can be positive or negative, but minDvrWindow should always be positive
      */
-    parser.adaptiveType = function(duration) {
+    parser.streamType = function(duration, _minDvrWindow) {
+        var minDvrWindow = _.isUndefined(_minDvrWindow) ? 120 : _minDvrWindow;
+        var streamType = 'VOD';
+
```
robwalch: We could return 'VOD' early if `duration >= 0`. The the only work is to check for 'DVR' conditions otherwise return 'Live'.

PR #1420, opened 2016-09-13
Sideload captions in the core player when not rendering natively

egreaves:

- Centralize sideloading of tracks into a module
- Remove dependency on window.VTTCue for better cross-provider track setup in IE/Edge
- Cleanup xhr requests before callback
- Sideload tracks in the core player when not rendering natively
- Sideload tracks in the mixin when rendering natively

Fixes #
JW7-2715


src/js/controller/captions.js
```diff
@@ -1,4 +1,7 @@
-define([], function() {
+define(['../utils/helpers',
+    '../utils/nativerenderingsupported',
+    '../controller/tracksLoader'
```
hghazzi: Can we give this files more consistent naming format? Seems we use `kebab-case` most frequently for filenames. Also, could we rename `nativerenderingsupported`? It's a util that does not explicitly identify it to be captions related and it's a little long. Maybe we could do `native-caption` and `tracks-loader` instead?

src/js/providers/tracks-mixin.js
```diff
@@ -428,6 +417,34 @@ define(['../utils/underscore',
...
+
+            var cue;
+            this._cuesByTrackId[track._id] = { cues: vttCues, loaded: true };
+
+            while((cue = vttCues.shift())) {
+                _addCueToTrack(textTrack, cue);
+            }
+        } else {
+            track.data = vttCues;
+        }
```
hghazzi: So if we don't render natively, we just set `track.data` to `vttcue`? If so this might be cleaner

```
if (!this._renderNatively) {
    track.data = vttCues;
    return;
}

var textTrack = this._tracksById[track._id];
.....

```

egreaves: @hghazzi this code will be removed because `track.data = vttCues;` is only done in `captions.js` now.

src/js/controller/tracks-loader.js
```diff
@@ -0,0 +1,110 @@
...
+    'parsers/parsers',
+    'parsers/captions/srt',
+    'parsers/captions/dfxp',
+    'parsers/captions/vttcue'
+], function(_, utils, parsers, srt, dfxp, VTTCue) {
+    var tracksLoader = {};
+
+    tracksLoader.loadFile = function(track, successHandler, errorHandler) {
+        track.xhr = utils.ajax(track.file, function(xhr) {
+            xhrSuccess(xhr, track, successHandler, errorHandler);
```
robwalch: This is not called in the right context:

`xhrSuccess.call(tracksLoader, xhr, track, successHandler, errorHandler);`

See squash/bootstrap.html?primary=html5&edition=ads&feature=api_set_captions scenario 2

src/js/controller/captions.js
```diff
@@ -53,14 +56,45 @@ define([], function() {
...
             // Clean up in case we're replaying
-            _itemHandler(_model,item);
+            _itemHandler(_model, item);
 
-            // listen for tracks coming from the provider
-            _model.mediaController.on('subtitlesTracks', _subtitlesTracksHandler, this);
+            _model.mediaController.off('subtitlesTracks', _subtitlesTracksHandler, this);
 
-            var captionsMenu = _captionsMenu();
-            this.setCaptionsList(captionsMenu);
```
robwalch: This should not have been removed, or moved into a branch of code where it is only set sometimes. The model's captionsList should always be an array, and it must be reset for each item.

test/squash/bootstrap.html?primary=html5&edition=ads&feature=captions_basic

`jwplayer().getCaptionsList()` returns `undefined`, because the captions list is not set for the first item when the item changes.

PR #1422, opened 2016-09-15
Bugfix/next up cached

johnBartos:
- Enable Next Up Tooltip on Related `cached` event

Fixes #
JW7-3095

### DIFF
--- a/src/js/view/components/nextuptooltip.js
+++ b/src/js/view/components/nextuptooltip.js
@@ -152,6 +152,7 @@ define([
             this._related = this._api.getPlugin('related');
             // Only switch to related mode if there is a related playlist
             this._related.on('playlist', this.onRelatedPlaylist.bind(this));
+            this._related.on('cached', this.onRelatedCached.bind(this));
         },
         onPlaylist: function(model, playlist) {
             this._playlist = playlist;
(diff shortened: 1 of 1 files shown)


src/js/view/components/nextuptooltip.js
```diff
@@ -202,22 +203,13 @@ define([
 
             this.offset = offset;
         },
+        onRelatedCached: function() {
+          this.enableRelatedMode();
```
egreaves: This makes the assumption that we have a playlist with at least one item.

src/js/view/components/nextuptooltip.js
```diff
@@ -152,6 +152,7 @@ define([
             this._related = this._api.getPlugin('related');
             // Only switch to related mode if there is a related playlist
             this._related.on('playlist', this.onRelatedPlaylist.bind(this));
+            this._related.on('cached', this.onRelatedCached.bind(this));
```
egreaves: If the cached event means the playlist hasn't changed, you can handle the events similarly:
`this._related.on('playlist cached', this.onRelatedPlaylist.bind(this));`

src/js/view/components/nextuptooltip.js
johnBartos: Neither cached nor playlist will fire when there are no related items, but I can add a check. However, the bug related to the Next button flashing is caused by playlists with a length of 1. Going to rework that in a future ticket.

src/js/view/components/nextuptooltip.js
johnBartos: The need to be on different handlers (`onRelatedPlaylist` vs `onRelatedCached`)

src/js/view/components/nextuptooltip.js
johnBartos: A check to see if there's actually a next up item is done so it's already safe.

PR #1423, opened 2016-09-15
Update styles for freewheel plugin

pajong:
Show control bars for freewheel ads.
Make freewheel span a click through, as this is an overlaying element on the video when freewheel overlay ad appears.
Make iframe inside the span a clickable, as this is the actual ad that needs to be clickable.

JW7-2985 JW7-2987


pajong:
@robwalch For nonlinear ads, they are always in an iframe.

PR #1424, opened 2016-09-16
DFXP captions parser should replace break elements with newlines even when `innerHTML` isn't supported.

johanberonius:

DFXP parser will replace `<br>`-elements with text nodes containing a newline before extracting text content from the paragraph.
Fixes that newlines are lost if `innerHTML` is unsupported. 

This fixes same issue as #1381 but for Edge.

### DIFF
--- a/src/js/parsers/captions/dfxp.js
+++ b/src/js/parsers/captions/dfxp.js
@@ -29,6 +29,13 @@ define([
 
         for (var i = 0; i < paragraphs.length; i++) {
             var p = paragraphs[i];
+
+            var breaks = p.getElementsByTagName('br');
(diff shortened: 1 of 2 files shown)


robwalch:
Where is this a problem outside of phantomjs? Fixes for phantomjs do not merit adding code to the player.

johanberonius:
Yes, in Edge. Subtitles were all on a single line. No no newlines were inserted since `<br>`-tags disappears when extracting text using `node.textContent`.
XML DOM nodes in Edge doesn't seem to have the property `innerHTML`.
I have run the test and a player with real subtitles in Edge and Firefox. 

All formatting in form of `<i>`, `<b>` or `<span>`-tags also disappears, that is still a problem.

PR #1427, opened 2016-09-19
Ads with shaka should never use flash provider

hghazzi:
When ads are played with shaka media, then it should use the html provider for the ads, even if the primary is set to flash.

JW7-3044

***Note:*** Must be merged with https://github.com/jwplayer/jwplayer-commercial/pull/2428

### DIFF
--- a/src/js/providers/providers.js
+++ b/src/js/providers/providers.js
@@ -6,10 +6,8 @@ define([
     ], function(Default, ProvidersSupported, ProvidersLoaded, _) {
 
     function Providers(config) {
-        this.providers = ProvidersSupported.slice();
         this.config = config || {};
-
-        this.reorderProviders();
+        this.providers = this.reorderProviders(this.config.primary);
     }
 
     Providers.loaders = {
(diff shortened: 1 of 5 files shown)


src/js/providers/providers.js
```diff
@@ -83,22 +83,34 @@ define([
...
-                this.providers.splice(html5Idx, 0, flashProvider);
+        getPrimary: function(primaryOverride) {
+            var primary = primaryOverride || this.config.primary;
+
+            if (primary === 'html5' || primary === 'flash') {
+                return primary;
             }
+
+            return 'html5';
+        },
```
robwalch: I don't think we need an html5 primary setting. Primary is either set to 'flash' or it is not.

It's better that we limit it to this so that deprecation becomes simpler and we don't complicate what this setting does.

hghazzi: @robwalch So this method is meant to identify which provider to move up to the front, hence returning `html5`. Maybe I should rename it to `getPrimaryProvider`?

hghazzi:
test this please.

PR #1429, opened 2016-09-19
Fixed deadzone in dock container

egreaves:
Fixed the dock container so the click event results in a play or pause when not hovering over a dock button.

Fixes #
JW7-2636, JW7-2869

### DIFF
--- a/src/css/imports/jwplayerelement.less
+++ b/src/css/imports/jwplayerelement.less
@@ -147,10 +147,10 @@
 }
 
 // These items can intercept pointer-events
-.jw-overlays > div,
+.jw-overlays .jw-plugin,
+.jw-dock .jw-dock-button,
 .jw-media,
 .jw-controlbar,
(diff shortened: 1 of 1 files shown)


src/css/imports/jwplayerelement.less
```diff
@@ -148,7 +148,12 @@
 
 // These items can intercept pointer-events
 .jw-overlays > div,
-.jw-media, .jw-controlbar, .jw-dock, .jw-logo, .jw-skip, .jw-display-icon-container,
+.jw-media,
+.jw-controlbar,
+.jw-dock > div,
```
robwalch: These types are queries can be very inefficient (from every div check if the parent has a .jw-dock parent). 

Let's make it the elements with the class we know we add directly to these containers:

```css
.jw-overlays .jw-plugin,
.jw-dock .jw-dock-button,
```

PR #1430, opened 2016-09-20
Don't dispatch Pause event while seeking

johnBartos:
- Don't dispatch Pause (or any other state change events) while seeking
- Call playVideo directly when seeking ends

JW7-2590


robwalch:
Going to reject the request since playback is paused while scrubbing and these changes raise issues with ad schedules and playback.

PR #1435, opened 2016-09-21
Set min width for time display to 640px 

johnBartos:
450px was way too low. 640px is a common player size and among the lowest widths before things start to look squished.

JW7-3120


robwalch:
Let's add the breakpoints we use in related to the player and then use those here. We'll want to multiplex breakpoints with `jw-flag-touch` since with the size controlbar 640 is the right size to hide times, but on desktop around 450 is best. And finally let's get rid of the compact mode class.

PR #1436, opened 2016-09-21
Load plugin once and use for all player instances

egreaves:
Previously, if there were multiple players on a page attempting to load the same plugin from different urls, the plugin would only be available in players that requested the plugin from the same url.

This change ensures that we load the plugin once and use it for all instances of the player on the page.

Fixes #
JW7-3100

### DIFF
--- a/src/js/plugins/loader.js
+++ b/src/js/plugins/loader.js
@@ -137,12 +137,27 @@ define([
                 }
 
                 var status = utils.tryCatch(function() {
-                    if (jsPlugin && pluginsConfig[pluginURL]) {
+                    if (jsPlugin) {
+                        var pluginConfig = pluginsConfig[pluginURL];
+
+                        if (!pluginConfig) {
+                            // Find the plugin's config by name if the plugin was already loaded from a different url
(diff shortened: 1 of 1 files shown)


src/js/plugins/loader.js
```diff
@@ -137,12 +137,27 @@ define([
...
+                            // Find the plugin's config by name if the plugin was already loaded from a different url
+                            pluginConfig = _.find(pluginsConfig, function(config, url) {
+                                return url.indexOf(pluginName) > 0;
+                            });
+
+                            if (!pluginConfig) {
+                                return;
+                            }
+
+                            utils.log('"' + pluginName + '" plugin already loaded from: ' + pluginURL);
```
robwalch: Let's not bother with all this. `url.indexOf(pluginName)` is not something to be counted on. 

Log and return. The player will continue to work as it has. The error will be logged and developers (and the player team) can use this info to fix their pages.

```js
if (!pluginConfig) {
    utils.log('JW Plugin already loaded', pluginName, pluginURL);
    return;
}
```

PR #1438, opened 2016-09-22
Dont pop up Next Up Tooltip during live playback

johnBartos:
- Detach 'change:position' handler when in Live mode.
- Reattach handler on a new Next Up item (will get detached again if it's live)

JW7-3105

### DIFF
--- a/src/js/view/components/nextuptooltip.js
+++ b/src/js/view/components/nextuptooltip.js
@@ -31,7 +31,7 @@ define([
             
             // Events
             this._model.on('change:mediaModel', this.onMediaModel, this);
-            this._model.on('change:position', this.onElapsed, this);
+            this._model.on('change:streamType', this.onStreamType, this);
             this._model.on('change:nextUp', this.onNextUp, this);
 
             this.closeButtonUI = new UI(this.closeButton, {'directSelect': true})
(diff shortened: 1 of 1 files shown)


src/js/view/components/nextuptooltip.js
```diff
@@ -157,6 +158,7 @@ define([
             var streamType = this._model.get('streamType');
             if (streamType === 'LIVE' || streamType === 'DVR') {
                 model.off('change:duration', this.onDuration, this);
+                model.off('change:position', this.onElapsed, this);
```
robwalch: Live/DVR/VOD state can change when duration changes. Maybe these listeners should be removed and added on 'change:streamType'?

src/js/view/components/nextuptooltip.js
```diff
@@ -144,9 +144,10 @@ define([
             }
 
             this._nextButton.toggle(true);
-            // Listen for duration changes to determine the offset
-            // for when next up should be shown
+            // Listen for duration changes to determine the offset from the end for when next up should be shown
             this._model.on('change:duration', this.onDuration, this);
+            // Listen for position changes so we can show the tooltip when the offset has been crossed
+            this._model.on('change:position', this.onElapsed, this);
```
robwalch: You remove the duration handler above, but not position.... you could be adding multiple position handlers.

johnBartos: Now detaching every time on`streamType` change and only adding when it's `VOD`

src/js/view/components/nextuptooltip.js
johnBartos: Done

src/js/view/components/nextuptooltip.js
```diff
@@ -50,15 +49,15 @@ define([
...
-            var style = {};
             this.nextUpImage = new Image();
             this.nextUpImage.onload = (function() {
                 this.nextUpImage.onload = null;
             }).bind(this);
             this.nextUpImage.src = url;
 
-            style.backgroundImage = 'url("' + url + '")';
-            return style;
+            return {
```
johnBartos: Cleanup

PR #1440, opened 2016-09-23
Next up display for skins

madikarizma:
* Add next up display styles for each skin
* Fix alignment issues per skin

JW7-3137

### DIFF
--- a/src/css/imports/nextup.less
+++ b/src/css/imports/nextup.less
@@ -5,12 +5,13 @@
   -webkit-font-smoothing: antialiased;
   -moz-font-smoothing: antialiased;
   background-color: transparent;
+  bottom: @nextup-position-bottom;
   cursor: pointer;
   max-width: 300px;
   min-width: 200px;
   opacity: 0;
   position: absolute;
-  right: 0.5em;
+  right: @nextup-position-right;
   transform: translateY(5px);
   transition: all 150ms ease;
   visibility: hidden;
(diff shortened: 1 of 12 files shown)


src/css/imports/nextup.less
```diff
@@ -5,15 +5,14 @@
   -webkit-font-smoothing: antialiased;
   -moz-font-smoothing: antialiased;
   background-color: transparent;
-  bottom: calc(@controlbar-height + 0.5em);
```
robwalch: So there's is no default bottom w/o a skin? You should put most seven skin settings here and only overwrite those needed in other skins. Skin authors shouldn't be required to include every rule.

src/css/skins/seven.less
```diff
@@ -54,4 +64,38 @@
...
+    background: @nextup-body-background;
+  }
+  .jw-nextup-thumbnail-visible + .jw-nextup-title::after {
+    background: linear-gradient(
+      -180deg,
+      fade(@nextup-title-overflow-background, 0%) 0%,
+      @nextup-title-overflow-background 100%
+    );
+  }
+  .jw-nextup-title {
```
johnBartos: Add these coloring styles to the `basic-skins-style` mixin in `mixins.less`

madikarizma: Something I need to do is get more familiar with how this code base works, but this and @robwalch's comment helped and the changes I made should be closer to what you had in mind.

johnBartos: Looks like you got it right, going to test it locally & merge

PR #1442, opened 2016-09-24
Timeout webpack requires

robwalch:
Timeout several potential points of failure in the setup process to provide accurate error messages when player setup cannot complete. The first polyfill or provider to timeout will trigger the setup error.

- Provider loading times out after 8 seconds with message 'Error: Could not load "`provider.name`" provider'
- Polyfill loading times out after 5 seconds with message 'Could not load polyfill: `polyfill.name`'

### DIFF
--- a/src/js/controller/setup-steps.js
+++ b/src/js/controller/setup-steps.js
@@ -81,7 +81,9 @@ define([
 
     function _loadPromisePolyfill(resolve) {
         if (!window.Promise) {
+            var timeout = setTimeout(_polyfillCouldNotLoad(resolve, 'Promise JS'), 8000);
             require.ensure(['polyfills/promise'], function (require) {
+                clearTimeout(timeout);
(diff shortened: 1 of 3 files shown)


src/js/controller/setup-steps.js
```diff
@@ -92,7 +94,9 @@ define([
 
     function _loadBase64Polyfill(resolve) {
         if (!window.btoa || !window.atob) {
+            var timeout = setTimeout(_polyfillCouldNotLoad(resolve, 'base64'), 8000);
```
johnBartos: Shouldn't `reject` be the first arg of `_polyfillCouldNotLoad` here?

johnBartos: Pass `reject` and use it for the timeout?

robwalch:
Closing until tests are fixed.

PR #1447, opened 2016-09-26
Show next up when item is valid and has a title or thumbnail

egreaves:
Only show the NextUp tooltip when:
- There is a valid item
- The item has at least a thumbnail or a title

Fixes #
JW7-3148


johnBartos:
test this please

PR #1448, opened 2016-09-26
Resolve next up display/control bar overlapping issue on touch players

madikarizma:
Fixed remaining bug where Next Up display overlapped control bar on mobile players.

JW7-3090

PR #1450, opened 2016-09-27
Don't adjust next up position at larger breakpoints

robwalch:
@madikarizma @johnBartos we should revisit touch + breakpoint flags. It might make sense to create higher level mobile breakpoints to simplify these rules and organize them in one place.

This change is a follow up to https://github.com/jwplayer/jwplayer/pull/1448

Fixes JW7-3090


johnBartos:
Maybe we should make a "larger-size" mixin, which encompasses breakpoints 3-6?

PR #1454, opened 2016-09-28
Fixed nextup bottom position for touch in core styles

robwalch:
Nextup was still being bumped up to far at smaller breakpoints with the seven skin and core styles.

JW7-3090


robwalch:
@madikarizma @johnBartos

PR #1458, opened 2016-09-30
Prevent spinning play wheel in iOS Safari

robwalch:
Throttling state change updates and toggling the dragging `.jw-flag-dragging` within state updates insures that css flags are set at once and not reset synchronously.

Fixes JW7-3233

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -657,29 +657,6 @@ define([
             _playerElement.onmouseup = handleMouseUp;
         }
 
-        function stopDragging(model) {
-            if (model.get('state') === states.PAUSED) {
-                model.once('change:state', stopDragging);
-                return;
-            }
-
-            // Handle the case where they begin seeking again before the last seek completes
-            if (model.get('scrubbing') === false) {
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -976,19 +953,27 @@ define([
 
         function _stateHandler(model, state) {
             _currentState = state;
-
+            // Throttle all state change UI updates except for play to prevent iOS 10 animation bug
```
johnBartos: Why don't we use _.throttle()?

http://underscorejs.org/#throttle

sidpatel13:
test this please

src/js/view/view.js
robwalch: Because we also want to cancel all timeouts when the view is destroyed.

johnBartos:
test this please

johnBartos:
Going to a new playlist item via the playlist interface doesn't show the buffering icon anymore

robwalch:
@johnBartos maybe if you throttle. Hopefully state changes are smoother now and if playback starts right away the buffering state doesn't show while playback is just starting.

PR #1461, opened 2016-09-30
Keep player textTracks in sync with browser textTracks

egreaves:

While there's no api method to [remove text tracks](https://lists.w3.org/Archives/Public/public-html/2015Feb/0017.html) programmatically added to the video tag, textTracks can be added/removed by Safari when playing HLS streams with ad breaks. This causes the player's internal tracklist to get out of sync with the browser. 

This PR addresses the following issues:
- Keep player and browser 608 and id3 tracks in sync
- Use the custom `name` property with the CC menu since `label` is read only and may not have a value in Safari
- Change track mode to `hidden` instead of `disabled` during ad playback so Safari can update `VTTCue.endTime`

Fixes #
JW7-2947

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -31,7 +31,6 @@ define(['utils/helpers',
             var tracks = e.tracks || [];
             for (var i = 0; i < tracks.length; i++) {
                 var track = tracks[i];
-                track.name = track.label || track.name || track.language;
                 _addTrack(track);
             }
             var captionsMenu = _captionsMenu();
@@ -104,6 +103,7 @@ define(['utils/helpers',
         function _addTrack(track) {
 
             track.data = track.data || [];
+            track.name = track.label || track.name || track.language;
 
             if (!track.name) {
                 track.name = 'Unknown CC';
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -31,7 +31,7 @@ define(['utils/helpers',
             var tracks = e.tracks || [];
             for (var i = 0; i < tracks.length; i++) {
                 var track = tracks[i];
-                track.label = track.label || track.name || track.language;
+                track.name = track.label || track.name || track.language;
                 _addTrack(track);
             }
             var captionsMenu = _captionsMenu();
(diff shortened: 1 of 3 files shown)


src/js/providers/tracks-mixin.js
```diff
@@ -546,6 +560,7 @@ define(['utils/underscore',
     }
 
     function _addTrackToList(track) {
+        // TODO: remove old track
```
johnBartos: todo: remove todo

johnBartos:
Tested the Captions suite in Chrome and it checked out

hghazzi:
test this please

hghazzi:
Tested on Safari desktop, came across some issues

robwalch:
@egreaves make sure you also test Firefox, and iOS. QA can help with Android Chrome and Edge/IE11.

egreaves:
@hghazzi what issues?

hghazzi:
Went through them with @egreaves and looks like they're all Safari issues

src/js/controller/captions.js
```diff
@@ -104,12 +103,13 @@ define(['utils/helpers',
...
+            track.name = track.label || track.name || track.language;
 
-            if (!track.label) {
-                track.label = 'Unknown CC';
+            if (!track.name) {
+                track.name = 'Unknown CC';
                 _unknownCount++;
                 if (_unknownCount > 1) {
-                    track.label += ' (' + _unknownCount + ')';
+                    track.name += ' (' + _unknownCount + ')';
```
hghazzi: This and `_createLabel` in `tracks-mixins.js` are pretty similar. Anything we can do to consolidate them?

src/js/providers/tracks-mixin.js
```diff
@@ -57,6 +65,11 @@ define(['utils/underscore',
                 if (!track._id) {
                     if (track.kind === 'captions' || track.kind === 'metadata') {
                         track._id = 'native' + track.kind;
+                        if (!track.label && track.kind === 'captions') {
+                            // track label is read only in Safari
+                            // 'captions' tracks without a label need a name in order for the cc menu to work
+                            track.name = _createLabel(track);
+                        }
```
hghazzi: Line 517 still tries to set label, is that valid?

`track.label = label`

PR #1467, opened 2016-10-05
Fixes for player rendering of captions

egreaves:

- Ensure all tracks have an `_id`
- Add both embedded and sideloaded tracks to the CC menu
- Position 708 captions correctly in FF
- Added unit tests

Fixes #
JW7-3239, JW7-3264


src/js/controller/captions.js
```diff
@@ -1,7 +1,9 @@
 define(['utils/helpers',
     'utils/render-captions-natively',
-    'controller/tracks-loader'
-], function(utils, renderCaptionsNatively, tracksLoader) {
+    'controller/tracks-loader',
+    'utils/track/create-id',
+    'utils/track/create-label'
+], function(utils, renderCaptionsNatively, tracksLoader, createTrackId, createTrackLabel) {
```
hghazzi: I think it now makes sense to put `renderCaptionsNatively` `createTrackId` `createTrackId` in one file called `captionUtils.js`. And to unit test them too. 😸

egreaves: Good suggestion.

PR #1472, opened 2016-10-07
Ensure track id is the same in all providers

egreaves:
- Ensure `track._id` is consistent across providers for sideloaded tracks
- Set `captionsIndex` before `captionsList` so the index is a part of the `onCaptionsList` event data

Fixes #
JW7-3290


sidpatel13:
This PR looks good. Checked the automated tests and also tested manually to confirm id attributes are populated with correct captions url.

PR #1473, opened 2016-10-07
Destroy advertising plugins on load

robwalch:
Advertising plugins need to be destroyed when load is called to stop ad playback before the playlist is changed.

This is tech-debt we should remove in 7.8/7.9 by calling `destroy` on playlist events in the plugins (destroy doesn't actually tear down the plugin itself, just all active advertising players and schedule managers running in the plugin). Instream destroy would also insure any ad playback mechanisms in the player are removed

```js
this.load = function (toLoad) {
    _controller.instreamDestroy();
    _controller.load(toLoad);
    return _this;
};
```

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -177,7 +177,7 @@ define([
         };
 
         this.load = function (toLoad) {
-            var plugin = this.getPlugin('vast') || this.getPlugin('googima');
+            var plugin = _this.getPlugin('vast') || _this.getPlugin('googima') || _this.getPlugin('freewheel');
             if (plugin) {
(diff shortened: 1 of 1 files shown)


robwalch:
adds-tech-debt label is to track closed issues that add or build on top of existing tech-debt.

robwalch:
@dannyfinks thought you merged this...

dannyfinks:
@robwalch I saw Jong was assigned as well so thought maybe you wanted him to review as well. My bad didn't notice the milestone included 7.7.1

robwalch:
We can leave it out and address the full tech-debt story #1476 as part of the FreeWheel release in 7.8.

PR #1482, opened 2016-10-10
Adding Airplay Support

hghazzi:
***Done***
- Available in Premium edition license or higher
- Airplay icon in the control bar when airplay config is enabled.
- Enabled native Airplay functionality
- Transition to Casting State
- Play, pause, seek, disconnect
- Hide the volume slider while AirPlay is active (currently behaves
like Chromecast)
- Airplay Icon active state
- Disabled Ad playback

JW7-1963


robwalch:
The PR builders printing cucumber results with TRIGGERED_BUILD_NUMBER - 1 should be updated to included fixes made in the jw7_commercial job.

src/js/view/controlbar.js
```diff
@@ -307,6 +312,15 @@ define([
         onCastActive : function(model, val) {
             utils.toggleClass(this.elements.cast.element(), 'jw-off', !val);
         },
+        onAirplayAvailable : function(model, val) {
+            this.elements.airplay.toggle(val);
+        },
+        onAirplayActive : function(model, val) {
+            utils.toggleClass(this.elements.airplay.element(), 'jw-off', !val);
+            if (this.elements.mute) {
+                this.elements.mute.toggle(!val);
```
hghazzi: Hey @robwalch, I would expect this to hide the volume button, but it doesn't seem to be working. Any ideas?

robwalch: It looks like it should.

robwalch:
test this please

src/js/view/view.js
```diff
@@ -381,6 +382,8 @@ define([
 
             _controlsLayer = _playerElement.getElementsByClassName('jw-controls')[0];
 
+            _controlbarLayer = _playerElement.getElementsByClassName('jw-controlbar')[0];
```
robwalch: Is this used anywhere? If not please remove.

src/js/view/controlbar.js
```diff
@@ -307,6 +312,12 @@ define([
         onCastActive : function(model, val) {
             utils.toggleClass(this.elements.cast.element(), 'jw-off', !val);
         },
+        onAirplayAvailable : function(model, val) {
+            this.elements.airplay.toggle(val);
+        },
+        onAirplayActive : function(model, val) {
+            utils.toggleClass(this.elements.airplay.element(), 'jw-off', !val);
+        },
```
robwalch: Can we move these handlers to commercial?

hghazzi: I would need to extend `controlbar.js` into commercial similarly to how we do it for `view.js`, which we don't have right now. Should I make that change, or leave it as tech debt?

robwalch: OK. We'll leave it as existing tech-debt since we do the same with the cast buttons.

alexprice91:
@robwalch do you have an example of how to enable this via config?

robwalch:
@alexprice91 it will be enabled by default in iOS and macOS Safari once released.

@hghazzi are there any configuration options available for AirPlay like disabling it or player edition requirements?

hghazzi:
Airplay should be enabled by default, like Rob said. It should be only
available for premium edition license or higher, and if you want to disable
Airplay, you add 'enabled': false to the 'cast' config that is currently
being used for chromecast.

alexprice91:
Oh, got it. Any update on release for Enterprise users?

robwalch:
@alexprice91 keep an eye on our [release notes](https://developer.jwplayer.com/jw-player/docs/developer-guide/release_notes/release_notes_7/) for the early beta release. It will not be in our upcoming 7.8 release later this month. We expect to follow up 7.8 with Airplay and Chromecast updates in December.

PR #1486, opened 2016-10-13
Flashplayer path in config should be relative to base path if unset.

johanberonius:

Flashplayer path in config should be relative to base path if unset.
Base path is already relative to script path if unset.

### Fixes

That flashplayer is not found if unset even if base path is set correctly in config. 
This happens if `getScriptPath('jwplayer.js')` does not find the path to `jwplayer.js`, eg. due to asset bundling.

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -75,7 +75,7 @@ define([
         __webpack_public_path__ = config.base;
         config.width  = _normalizeSize(config.width);
         config.height = _normalizeSize(config.height);
-        config.flashplayer = config.flashplayer || utils.getScriptPath('jwplayer.js') + 'jwplayer.flash.swf';
+        config.flashplayer = config.flashplayer || config.base + 'jwplayer.flash.swf';
 
         // Non ssl pages can only communicate with flash when it is loaded
         //   from a non ssl location
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -75,7 +75,7 @@ define([
         __webpack_public_path__ = config.base;
         config.width  = _normalizeSize(config.width);
         config.height = _normalizeSize(config.height);
-        config.flashplayer = config.flashplayer || config.base + 'jwplayer.flash.swf';
+        config.flashplayer = config.flashplayer || (utils.getScriptPath('jwplayer.js') || config.base) + 'jwplayer.flash.swf';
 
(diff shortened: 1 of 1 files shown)


src/js/api/config.js
```diff
@@ -75,7 +75,7 @@ define([
         __webpack_public_path__ = config.base;
         config.width  = _normalizeSize(config.width);
         config.height = _normalizeSize(config.height);
-        config.flashplayer = config.flashplayer || utils.getScriptPath('jwplayer.js') + 'jwplayer.flash.swf';
+        config.flashplayer = config.flashplayer || config.base + 'jwplayer.flash.swf';
```
robwalch: If bundling is the concern this should read:

`config.flashplayer = config.flashplayer || (utils.getScriptPath('jwplayer.js') || config.base) + 'jwplayer.flash.swf'; `

robwalch:
Is your main use case self-hosted? Are you setting base in your config?

src/js/api/config.js
johanberonius: > Is your main use case self-hosted? 

Yes.

> Are you setting base in your config?

Yes, `base` is set correctly. `flashplayer` also needs to be set or it will not be found even though it is in the default location. I would like to avoid that.

> If bundling is the concern this should read:...

Isn't that redundant since `config.base` is already set to  `utils.getScriptPath('jwplayer.js')`?
https://github.com/jwplayer/jwplayer/blob/master/src/js/api/config.js#L72
https://github.com/jwplayer/jwplayer/blob/master/src/js/utils/playerutils.js#L42

johanberonius: And shouldn't `config.base` take precedence? If it is set then that's where `flashplayer` should be located, right?

robwalch:
@johanberonius the choices we made here were to maintain compatibility with JW6 embeds. By default config.base is set to our CDN location and we do not want self hosted players where there is no base set in the config to reach out to the CDN to load a swf - that would cause security policy error in Flash where the host doesn't match the current domain.

If `getScriptPath('jwplayer.js')` returns an empty string we can fallback to base - per my suggestion, but we're not going to remove the script check. This is needed by existing setups.

johanberonius:
I understand. Fixed now, script path takes precedence with fallback to base path.
It now works works without having to set both `base` and `flashplayer`.

PR #1487, opened 2016-10-13
Feature/hlsjs/master

johnBartos:
- Don't reassign the prototype - it breaks ES6 class constructor functions


src/js/providers/providers.js
```diff
@@ -58,9 +58,11 @@ define([
...
         }
 
-        var F = function(){};
-        F.prototype = Default;
-        provider.prototype = new F();
+        if (name !== 'hlsjs') {
+            var F = function () {};
+            F.prototype = Default;
+            provider.prototype = new F();
+        }
```
robwalch: Can we remove this?

src/js/controller/tracks-helper.js
```diff
@@ -24,8 +24,8 @@ define([
             return { label: label, unknownCount: unknownCount };
         },
         renderNatively: function (providerName) {
-            return (providerName === 'html5' || providerName === 'shaka' || providerName === 'caterpillar') &&
-                (browser.isChrome() || browser.isIOS() || browser.isSafari() || browser.isEdge());
+            return (providerName === 'html5' || providerName === 'shaka' || providerName === 'caterpillar' || providerName === 'hlsjs') &&
+                (browser.isChrome() || browser.isIOS() || browser.isSafari() || browser.isEdge() || (providerName === 'hlsjs' && browser.isFF()));
```
robwalch: Let's add a tech-debt ticket to simplify this. Providers that can render natively should update the model to indicate this and all others will not. That was this method isn't used in every provider that uses mixins and captions.js and view.js.

src/js/providers/providers.js
robwalch: There are issues removing it so how about this?

`_.defaults(provider.prototype, Default);`

PR #1491, opened 2016-10-17
Expose the function to add skip button

pajong:
Expose the function to add skip button so that FreeWheel can use the function to setup a skip button.
JW7-3245

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -82,7 +82,7 @@ define([
 
             _instream.on('all', _instreamForward, this);
             _instream.on(events.JWPLAYER_MEDIA_TIME, _instreamTime, this);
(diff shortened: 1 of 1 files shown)


src/js/controller/instream-adapter.js
```diff
@@ -322,6 +331,7 @@ define([
 
         this.setText = function(text) {
             _view.setAltText(text ? text : '');
+            window.yo = this;
```
robwalch: dude

robwalch:
Vast is able to display a skip button without exposing this method. Do we not know at the time of loading an add if it can be skipped?

src/js/controller/instream-adapter.js
pajong: oops my bad

pajong:
We do not call loadItem with FreeWheel - there are things called slots that control the ad play back instead of loading items this way. We just use the adapter to start blocking and adding ad texts to control bar.

PR #1492, opened 2016-10-18
Fix cursor pointer for FW in flash provider

pajong:
JW7-3251

### DIFF
--- a/src/css/flags/ads.less
+++ b/src/css/flags/ads.less
@@ -55,6 +55,14 @@
         }
     }
 
+    .jw-overlays {
+        .jw-plugin {
+            video {
+                cursor: pointer;
+            }
+        }
+    }
+
     .jw-freewheel-nonlinear-container {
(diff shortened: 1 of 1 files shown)


src/css/flags/ads.less
```diff
@@ -55,6 +55,14 @@
         }
     }
 
+    .jw-overlays {
+        .jw-plugin {
+            video {
+                cursor: pointer;
+            }
+        }
+    }
```
hghazzi: This could simply be

```
.jw-overlays  .jw-plugin video {
    cursor: pointer;
}
```

robwalch: It will get compiled that way. Since we're using less we write our rules using a hierarchy. It make it easier to read and modify.

robwalch:
Make this specific to the freewheel plugin by adding it only to the div in the plugin.

Issue #1330, opened 2016-07-21
Safari | Audio HLS: poster image is not displayed


fdambrosio:
### Expected Behavior

When the player with HLS audio is ready or it's playing or it's paused, I expect that the poster image is displayed on Safari. 
### Actual Behavior

With Safari (audio player is loaded in html5 mode) the poster image isn't displayed in any case: ready, pause, playing. When player is loaded the poster is showed only 1 second. 
I think that there's some problems with skin too because I see gray bars at the top and bottom of the player, even above the control bar.
### Steps to reproduce
1. setup jwplayer HLS audio + image with Jwp 7.5.0 RC.1
2. open page with Safari
3. poster is not showed 
4. demo: http://www.fdambrosio.net/test/player7/safari.html
### Environment
- iMac
- MacOS 10.11.5 
- Safari 9.1.1
- Jwp 7.5.0 RC.1

demo: http://www.fdambrosio.net/test/player7/safari.html

johnBartos:
This is a known issue - on Chrome & FF, we make the video tag transparent to show the poster image behind it. Safari can't/won't make it transparent & renders as a black box instead.

robwalch:
...and the reason the video is on top of the poster is to display webvtt captions if present.

fdambrosio:
Ok, thanks for the answer, so we'll never have poster on Safari?

robwalch:
File a radar with Apple to allow transparency on video elements playing audio only sources. We'll do the same.

robwalch:
If you know you are playing audio and don't need captions you can display the video tag as hidden with CSS to show the poster. Is this something you need the player to do, or do need captions as well?

fdambrosio:
Yes, we need audio & poster image & captions that we show automatically by default

smfr:
I'm confused by this report. I see a poster image in Safari 10:
![jwplayer](https://cloud.githubusercontent.com/assets/20968/18977387/baf5091e-866e-11e6-859b-9d6ae304628b.png)

Is this not what it's supposed to look like?

You should also clarify that you're talking about the JWP poster, not <video poster="">.

smfr:
My tags were eaten. Not `<video poster="foo.png">`

robwalch:
The bug is no longer reproducible in this test case because the media is not accessible and has been hidden by the player. When an audio only stream is loaded in a video tag in Safari 10, the video tag is not transparent.

smfr:
Can I get a working test case?

[...]

PR #1497, opened 2016-10-20
Set poster image for video elements in html5-provider.

johanberonius:

Set the poster attribute on the video element in html5-provider to show the image until the first frame of video is displayed, or during entire playback for audio-only sources.
(I hope `init(item)` is the right place to do this.)
### Fixes

That a black background is displayed instead of the image for audio-only sources, eg. MP3 HLS-streams. See #1330

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -278,7 +278,7 @@ define([
             }
             _position = currentTime;
         }
-        
+
         function _getDuration() {
             var duration = _videotag.duration;
             var end = _getSeekableEnd();
@@ -291,7 +291,7 @@ define([
             }
             return duration;
         }
-        
+
         function _updateDuration(duration) {
             _duration = duration;
             if (_delayedSeek && duration && duration !== Infinity) {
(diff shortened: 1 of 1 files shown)


robwalch:
Hi @johanberonius @egreaves and I tried this and while it works for images hosted in the same domain or served with cors headers, we found that a poster served cross-origin without cors header breaks media playback.

johanberonius:
Ok. It worked well in our environment. We serve images from a subdomain different from the web- and streaming-server. And there are no CORS-header for images.
Only tested in Safari on macOS and iOS though.
I can set up a test-page to verify more browsers/platforms.

johanberonius:
A test page where the poster image is set on the video element when the player is ready is available here:
http://rawgit.com/sveriges-utbildningsradio/jwplayer-test/master/video-poster-image.html
Note that no CORS-headers are served for images.

egreaves:
Hi @johanberonius, 

Your approach definitely works well with your setup, where images are loaded from the same domain but a different subdomain than the video source, but this is not something we can implement in the provider for the reason @robwalch mentioned above. We would effectively be limiting our publishers' flexibility in loading assets from different domains, especially if they can't modify CORS rules on all the domains from which they serve content.

PR #1501, opened 2016-10-21
Allow muted autostart of ads on mobile

egreaves:
Publishers can now opt in to autostarting with ads if `autostart` is `true` and the advertising block has `autoplayadsmuted` set to `true`.

Fixes #
JW7-3159, JW7-3015

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -40,8 +40,7 @@ define([
                 buffer: 0
             });
 
-            // Videos autostart on mobile when the `muted` and `autoplay` attributes are set.
-            // Mute the player on a mobile device if autostart is set to true
+            // Videos can autoplay on mobile when the `muted` and `autoplay` attributes are set.
             if (this.autoStartOnMobile()) {
                 this.set('mute', true);
             }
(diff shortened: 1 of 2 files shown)


egreaves:
test this please

src/js/controller/model.js
```diff
@@ -407,7 +406,8 @@ define([
         };
 
         this.autoStartOnMobile = function() {
-            return this.get('autostart') && utils.isMobile() && !this.get('mobileSdk');
+            return this.get('autostart') && utils.isMobile() && !this.get('mobileSdk') &&
+                (!this.get('advertising') || this.get('advertising').autoplayadsmuted);
```
hghazzi: Did you sync up with @esboyd on the config name? Some thoughts: 

- If this is in the advertising block do we need to include the word `ads` in it? 
- If we say `autostart` for the media, we should use the same term for ads and not use `autoplay`. 
- If looking at just the config, there's nothing in the name that implies that mute on autostart is mobile only.
- unless I'm mistaken, I think we're moving towards camelCase for the config names. 

Maybe something like `muteOnMobileAutostart` or `muteOnMobilePlay`?

PR #1502, opened 2016-10-21
Wait until canplay on iOS to send the bufferFull event

egreaves:

When autoplaying on iOS, we were triggering the `bufferFull` event in `_completeLoad`, which is before there's enough data buffered to start playback. Doing so when the `canplay` event is raised guarantees the play promise will be fulfilled. 

For manual play attempts, the `bufferFull` event will be raised slightly later than before, but this does not affect playback and should not introduce any side effects. 

Fixes #
JW7-3362

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -346,7 +346,8 @@ define([
         }
 
         function _sendBufferFull() {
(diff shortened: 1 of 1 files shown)


egreaves:
test this please

PR #1506, opened 2016-10-25
Ignore pause events before playback starts

egreaves:

This ensures that we ignore pause events raised by the browser before playback starts when autoplaying on mobile.

Fixes #
JW7-3362


robwalch:
If the browser paused the video tag we need to forward that event. This change would prevent the player state from updating correctly if the user wants to pause the video before it starts.

Issue #1476, opened 2016-10-08
Remove advertising plugin workarounds


robwalch:
Advertising plugins need to be destroyed when load is called to stop ad playback before the playlist is changed. Currently this is being done in this project (which should not contain any ad related code).

https://github.com/jwplayer/jwplayer/blob/f8d48015eaddf9c265c49821e6772ddc05f81fda/src/js/api/api.js#L180

See https://github.com/jwplayer/jwplayer/pull/1473

This is tech-debt we should remove in 7.8/7.9 by calling `destroy` [on playlist - or a new playlistLoad event](https://github.com/jwplayer/jwplayer/pull/1509#issuecomment-256144716) in the plugins (destroy doesn't actually tear down the plugin itself, just all active advertising players and schedule managers running in the plugin). Instream destroy would also insure any ad playback mechanisms in the player are removed

``` js
this.load = function (toLoad) {
    _controller.instreamDestroy();
    _controller.load(toLoad);
    return _this;
};
```

robwalch:
See https://github.com/jwplayer/jwplayer/pull/1509#issuecomment-256144716

PR #1509, opened 2016-10-25
Destroy freeWheel plugin on new item load

pajong:
Add freewheel to plugin destroy list along with vast and googima.


robwalch:
We need to take another look at https://github.com/jwplayer/jwplayer/issues/1476. I know having the plugins destroy themselves on 'playlist' didn't do the trick. We should `trigger('playlistLoad', toLoad)` here and respond to that in ad plugins.

PR #1511, opened 2016-10-25
Remove max width of floating controlbars

madikarizma:
Floating control bars no longer have max-width set. Allows control bars to stretch to near the edge of player.

Fixes #
JW7-3247

### DIFF
--- a/src/css/imports/mixins.less
+++ b/src/css/imports/mixins.less
@@ -28,7 +28,6 @@
         display: inline-block;
         // width assignment is required for IE11 to center correctly
         width: 96%;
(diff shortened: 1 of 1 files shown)


robwalch:
This was a "feature" of JW6 where the width of the cbar was limited to 800px when going fullscreen because having it go across the screen is unnecessary.

robwalch:
JW7-3247 refers to controlbars that are not flush against the bottom of the player. At the smaller breakpoints that padding or margin should be removed. max-width is not the issue and only affects very large players which is not the concern.

PR #1513, opened 2016-10-25
Mobile Autostart Fixes

egreaves:

- Hide volume toggle in mobile SDKs
- Don't raise pause event before playback starts
- Remove autoplay attributes/properties if autoplay attempt fails

Fixes #
JW7-3392, JW7-3362, JW7-3373

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -152,6 +152,11 @@ define([
                     var visualQuality = _.extend({}, data);
                     this.mediaModel.set('visualQuality', visualQuality);
                     break;
+
+                case 'autoplayFailed':
+                    _provider.removeAutoplayAttributes();
+                    this.setMute(false);
+                    break;
             }
 
             this.mediaController.trigger(type, evt);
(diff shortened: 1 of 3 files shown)


src/js/providers/html5.js
```diff
@@ -456,7 +456,21 @@ define([
             if (promise && promise.catch) {
                 promise.catch(function(err) {
                     console.warn(err);
+                    // User gesture required to start playback
+                    if(err.name === 'NotAllowedError') {
```
pajong: JP lint: space after `if`

src/js/controller/model.js
```diff
@@ -152,6 +152,10 @@ define([
                     var visualQuality = _.extend({}, data);
                     this.mediaModel.set('visualQuality', visualQuality);
                     break;
+
+                case 'autoplayFailed':
+                    this.setMute(false);
```
egreaves: The player doesn't have the original config values once an update has been made to the mute state in the model. We always set the model's mute value to `true` on mobile when autoplaying, so this just undoes that change.

PR #1523, opened 2016-10-31
Move destroy on load to each plugin

pajong:
JW7-3298

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -177,11 +177,6 @@ define([
         };
 
         this.load = function (toLoad) {
-            var plugin = this.getPlugin('vast') || this.getPlugin('googima') || this.getPlugin('freewheel');
-            if (plugin) {
-                _controller._model.set('preInstreamState', 'instream-idle');
-                plugin.destroy();
-            }
             _controller.load(toLoad);
             return _this;
         };
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -255,11 +255,14 @@ define([
                 if (_model.get('state') === states.ERROR) {
                     _model.set('state', states.IDLE);
                 }
+                _model.set('preInstreamState', 'instream-idle');
+
                 _stop(true);
 
                 if (_model.get('autostart')) {
                     _model.once('itemReady', _play);
                 }
+                _this.trigger('destroyPlugin', {});
 
                 switch (typeof item) {
                     case 'string':


src/js/api/api.js
```diff
@@ -177,11 +177,8 @@ define([
         };
 
         this.load = function (toLoad) {
-            var plugin = this.getPlugin('vast') || this.getPlugin('googima') || this.getPlugin('freewheel');
-            if (plugin) {
-                _controller._model.set('preInstreamState', 'instream-idle');
-                plugin.destroy();
-            }
+            _controller._model.set('preInstreamState', 'instream-idle');
```
robwalch: Instream state should not be handled here. Remove and reset in controller when instream is instantiated.

src/js/api/api.js
```diff
@@ -177,11 +177,8 @@ define([
...
+            this.trigger(events.JWPLAYER_LOAD_PLAYLIST, {});
```
robwalch: Load may load a playlist or set the current item. See controller `_load`.
```js
                switch (typeof item) {
                    case 'string':
                        _loadPlaylist(item);
                        break;
                    case 'object':
                        var success = _setPlaylist(item);
                        if (success) {
                            _setItem(0);
                        }
                        break;
                    case 'number':
                        _setItem(item);
                        break;
                }
```
I don't think this event is correct where you've placed it. It would only make sense if item was a string, but we may still need to do something in the case that it's an object.

src/js/controller/controller.js
```diff
@@ -255,6 +255,8 @@ define([
                 if (_model.get('state') === states.ERROR) {
                     _model.set('state', states.IDLE);
                 }
+                _model.set('preInstreamState', 'instream-idle');
```
robwalch: Do we need this? I would like this to only be set in instream-adapter.

src/js/controller/controller.js
```diff
@@ -263,9 +265,11 @@ define([
 
                 switch (typeof item) {
                     case 'string':
+                        _this.trigger(events.JWPLAYER_LOAD_PLAYLIST, {});
                         _loadPlaylist(item);
                         break;
                     case 'object':
+                        _this.trigger(events.JWPLAYER_LOAD_PLAYLIST, {});
```
robwalch: Either we make a 'load' event that fires for all three cases, or we make a 'loadPlaylist" and a 'beforeSetPlaylist' event.

src/js/events/events.js
```diff
@@ -66,6 +66,7 @@ define([], function() {
         JWPLAYER_CONTROLS: 'controls',
         JWPLAYER_FULLSCREEN: 'fullscreen',
         JWPLAYER_RESIZE: 'resize',
+        JWPLAYER_LOAD_PLAYLIST: 'loadPlaylist',
```
robwalch: No more new sysmbols here please. Use strings. I would like to deprecate jwplayer.events. from the player. They are long, repetitive and make the player bigger.

src/js/controller/controller.js
pajong: Yes, we need this.
This is the commit which added this line: https://github.com/jwplayer/jwplayer/commit/0ccf9c8a9bebe0371ccf9a3341c3290256b8e877

pajong:
Squash test page created: https://github.com/jwplayer/jwplayer-commercial/pull/2631

src/js/controller/controller.js
robwalch: Where can we put it in instream-adapter so that all changes to 'preInstreamState' are done in insteam?

PR #1524, opened 2016-10-31
Show mute button for muted autoplay

egreaves:
Improvements to muted autoplay:
- Show a mute icon when autoplaying on mobile
- Don't show mute icon when the browser interrupts playback, pausing the player
- Decouple muted autoplay from the player's mute state (set by user interaction or config preference)
- Gracefully degrade when autoplay is not supported or fails, removing the mute button and setting the player back to the original `mute` state

Fixes #
JW7-2955

### DIFF
--- a/src/css/flags/ads.less
+++ b/src/css/flags/ads.less
@@ -4,7 +4,8 @@
     .jw-dock,
     .jw-logo,
     .jw-captions.jw-captions-enabled,
-    .jw-nextup-container {
+    .jw-nextup-container,
+    .jw-autostart-mute {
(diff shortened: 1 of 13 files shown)


egreaves:
test this please

dannyfinks:
@egreaves I see that i'm unable to single tap on the player while the mute icon is displayed to display the control-bar. I am not sure if we want that experience. Also if you were to double tap and go to fullscreen and than return the player than no longer shows the mute icon but the control-bar is not displayed while the player is paused.

PR #1533, opened 2016-11-08
Remove autoplay muted icon when mute state changes

pajong:
Whenever there is a change to mute state, we should stop showing autostart muted.
JW7-2955

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -405,9 +405,17 @@ define([
             this.set('nextUp', nextUp);
         };
 
+        function _autoStartSupportedIOS() {
+            if (!utils.isIOS()) {
+                return false;
+            }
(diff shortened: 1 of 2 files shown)


src/js/controller/model.js
```diff
@@ -405,9 +405,17 @@ define([
             this.set('nextUp', nextUp);
         };
 
+        function _autoStartSupportedIOS() {
+            if (!utils.isIOS()) {
+                return false;
+            }
+            // Autostart only supported in iOS 10 or higher - check if the version is 9 or less
+            return !(utils.isIOS(6) || utils.isIOS(7) || utils.isIOS(8) || utils.isIOS(9));
```
johnBartos: Can you just return utils.isIOS(10) instead?

pajong: We need to change that if new iOS comes out.

PR #1539, opened 2016-11-10
New Chromecast Support (Do Not Merge)

hghazzi:
***Will add details***

JW7-1964


src/js/controller/model.js
```diff
@@ -398,6 +398,10 @@ define([
 
         this.setNextUp = function (nextUp) {
             this.set('nextUp', nextUp);
+
+            if (_provider && _provider.next) {
+                _provider.next(nextUp);
+            }
```
robwalch: I'd like us to avoid adding a new provider method.

PR #1542, opened 2016-11-11
Set autostartMuted to false when setMute is called

pajong:
When setMute is called with mute false, set autostartMuted to false.
Listen to the change in autostartMuted property to handle unmute with ads properly.
JW7-3469

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -652,6 +652,7 @@ define([
                 utils.addClass(_playerElement, 'jw-flag-autostart');
                 _model.set('autostartMuted', true);
                 _model.on('change:autostartFailed', _autoplayUnmute);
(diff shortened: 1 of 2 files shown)


src/js/view/view.js
```diff
@@ -829,15 +830,11 @@ define([
...
                 mute = false;
             }
-
             _model.off('change:autostartFailed', _autoplayUnmute);
             _model.off('change:mute', _autoplayUnmute);
+            _model.off('change:autostartMuted', _autoplayUnmute);
             _model.set('autostartFailed', undefined);
             _model.set('autostartMuted', undefined);
-
-            if (_instreamModel) {
```
pajong: This was redundant, because api.setMute calls instream mute function.

PR #1544, opened 2016-11-11
Merge MAUI branch to master

pajong:


src/js/view/components/nextuptooltip.js
```diff
@@ -74,10 +75,12 @@ define([
                 return;
             }
             dom.addClass(this.container, 'jw-nextup-container-visible');
+            dom.addClass(this._playerElement, 'jw-flag-nextup');
         },
         hide: function() {
             dom.removeClass(this.container, 'jw-nextup-container-visible');
             dom.removeClass(this.container, 'jw-nextup-sticky');
+            dom.removeClass(this._playerElement, 'jw-flag-nextup');
```
robwalch: I don't think this was a good call. When muted, only show the mute button. Nextup is not more important.

PR #1553, opened 2016-11-15
Replace "middle" with "center"

mattcoz:
Replaces instances of "middle" with "center" to adhere to the valid VTT values.

Fixes #
JW7-1474

The enum value of "middle" is not valid, the correct value is "center". This applies to line, position, and align.

https://w3c.github.io/webvtt/#the-vttcue-interface

### DIFF
--- a/src/js/parsers/captions/vttparser.js
+++ b/src/js/parsers/captions/vttparser.js
@@ -188,7 +188,7 @@ define(['parsers/captions/vttcue'], function(VTTCue) {
             // Apply default values for any missing fields.
             cue.region = settings.get('region', null);
             cue.vertical = settings.get('vertical', '');
-            cue.line = settings.get('line', -1);
+            cue.line = settings.get('line', 'auto');
             cue.lineAlign = settings.get('lineAlign', 'start');
             cue.snapToLines = settings.get('snapToLines', true);
             cue.size = settings.get('size', 100);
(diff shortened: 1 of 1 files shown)


johnBartos:
I thought the issue was that `cue.position` only accepts a double or `'auto'`

johnBartos:
Here's the real fix: `cue.position = settings.get("position", "auto");`

See: https://github.com/mozilla/vtt.js/issues/354

Fix your PR and I'll merge it!

mattcoz:
@johnBartos Yes, and undefined is not a double or 'auto'.

mattcoz:
I agree that your fix is better for fixing the original issue, as it doesn't make sense to try to convert cue.align to cue.position. The same should probably be applied to cue.positionAlign. But, "middle" is still not a valid value, it should be "center". Also, the valid values for positionAlign are "line-left", "center", "line-right", and "auto".

mattcoz:
Ok, it seems that Apple and the W3C might disagree on whether it should be "middle" or "center". The captions don't show on the iPhone I have for testing when "center" is used. If I change it back to "middle" they show, but they aren't aligned properly. It seems that they don't understand the "auto" keyword for position.

mattcoz:
Ok, I'll create a new pull request with fixes for iOS.

johnBartos:
@mattcoz It would be great if you could PR this to vtt.js too

PR #1554, opened 2016-11-16
Workaround for iOS autoplay HLS not completing

pajong:
Sometimes after midroll on iOS autoplay, the duration is not updated properly if playback is an HLS stream, resulting the stream to never go to a complete state.
Making a workaround for this specific case so that the stream ends.
JW7-3477

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -387,6 +387,12 @@ define([
                 return;
             }
 
+            // Workaround for iOS not completing after midroll with HLS streams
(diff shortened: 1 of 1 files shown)


robwalch:
@pajong in this situation we should try to provide Apple with a test case to reproduce the issue. Setup a video tag, with src = the m3u8. Then on canseek set the currentTime to the point where the midroll ends or after it. Let's see if the end event fires then. If we can reproduce it not firing with just a video tag and setting currentTime to seek, let's send that to Apple for a fix.

PR #1556, opened 2016-11-16
Expose function to set autostartMuted to false

pajong:
This function is needed for ad platforms to listen to volume change in vpaid 2 ads and clear autostartMuted icon.
JW7-3503

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -177,6 +177,10 @@ define([
             }
         };
 
+        this.clearAutostartMute = function() {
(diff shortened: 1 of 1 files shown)


src/js/controller/instream-adapter.js
```diff
@@ -177,6 +177,10 @@ define([
             }
         };
 
+        this.clearAutostartMute = function() {
+            _model.set('autostartMuted', false);
+        };
```
robwalch: Shouldn't setting the player's mute state to false do this (`player.setMute(false)`)?

PR #1558, opened 2016-11-17
Ignore primary:flash when Flash is blocked

robwalch:
When `primary: 'flash'` is set, and `utils.flashVersion()` is not 0, embed a swf during setup that calls back to confirm the Flash plugin is not blocked. If it doesn't callback in 1.5 seconds, set primary to `undefined` and continue setup.

The swf being embedded is a new swf called "flash.loader.swf" and is an instance of `FlashHealthCheck.as`. It is < 1k in size. It's not actually loading anything - I just couldn't think of a name that described it's role that didn't sound silly or suspicious.

When embedded it fills a temporary container that is styled to match the player's dimensions. This is so that any rules applied to the size of an object tag are applied (ex: a browser may block a 5x5 swf but not a 640x360 one).

JW7-3414

### DIFF
--- a/src/js/controller/setup-steps.js
+++ b/src/js/controller/setup-steps.js
@@ -2,11 +2,12 @@ define([
     'plugins/plugins',
     'playlist/loader',
     'utils/scriptloader',
+    'utils/embedswf',
     'utils/constants',
     'utils/underscore',
     'utils/helpers',
     'events/events'
-], function(plugins, PlaylistLoader, ScriptLoader, Constants, _, utils, events) {
+], function(plugins, PlaylistLoader, ScriptLoader, EmbedSwf, Constants, _, utils, events) {
 
     var _pluginLoader,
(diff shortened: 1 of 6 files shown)


src/js/controller/setup-steps.js
```diff
@@ -139,6 +144,50 @@ define([
...
 
+    function _checkFlash(resolve, _model, _api) {
+        var primaryFlash = _model.get('primary') === 'flash';
+        var flashVersion = utils.flashVersion();
+        if (primaryFlash && flashVersion) {
+            var flashHealthCheckId = '' + _model.get('id') + '-' + Math.random().toString(16).substr(2);
+            var originalContainer = _api.getContainer();
+            var parentElement = originalContainer.parentElement;
+            if (!parentElement) {
+                // couldn't perform test because setup container has not parent
```
hghazzi: "has not parent"

PR #1561, opened 2016-11-17
Use middle instead of center since it leads to errors in ios/chrome

johnBartos:
Chrome: `vttparser.js:195 The provided value 'center' is not a valid enum value of type AlignSetting.`

Reverting back to middle still fixes the firefox issues but doesn't throw in Chrome. Firefox 50 seems to say to use `center` but `middle` still works, apparently. https://bugzilla.mozilla.org/show_bug.cgi?id=1276130

@mattcoz


mattcoz:
Apple says "middle", the W3C says "center". Apple also disagrees on the usage of "auto" for "line" and "position", only accepting double values. I'm thinking that checking for the default values beforehand will help us detect which values we should be using. I'm working on a patch for that now.

PR #1563, opened 2016-11-17
Use VTTCue to determine whether to use center/middle and double values instead of "auto"

mattcoz:
Different browsers disagree on the default values for align, line, and position.

Firefox: "center", "auto", "auto"
Chrome: "middle", "auto", "auto"
Safari: "middle", -1, 50

Safari does not support "auto" at all for line and position.

My fix creates an empty VTTCue, checks the default values, and changes the parser behavior to match.

### DIFF
--- a/src/js/parsers/captions/vttparser.js
+++ b/src/js/parsers/captions/vttparser.js
@@ -126,7 +126,7 @@ define(['parsers/captions/vttcue'], function(VTTCue) {
     }
     
     var defaults = new VTTCue(0,0,0);
-    var center = defaults.align == 'middle' ? 'middle' : 'center';
+    var center = defaults.align === 'middle' ? 'middle' : 'center';
 
     function parseCue(input, cue, regionList) {
         // Remember the original input if we need to throw an error.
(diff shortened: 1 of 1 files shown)


johnBartos:
Thanks for the PR! Unless we're running into problems specifically caused by this I want to hold off on merging this into 7.8.

mattcoz:
The problem is with Safari not understanding "auto" for "position". This causes the captions to be improperly aligned. My code takes it a step further by following the W3C's "auto" logic to set the "position" value based on the "align" value.

https://w3c.github.io/webvtt/#webvtt-cue-position

src/js/parsers/captions/vttparser.js
```diff
@@ -124,6 +124,9 @@ define(['parsers/captions/vttcue'], function(VTTCue) {
             callback(k, v);
         }
     }
+    
+    var defaults = new VTTCue(0,0,0);
+    var center = defaults.align === 'middle' ? 'middle' : 'center';
```
egreaves: Hi @mattcoz,

Can you add a comment here so it's clear why we need to determine the default align value?
 
Bugs have been filed with [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=617989) and [Safari](https://bugs.webkit.org/show_bug.cgi?id=158478), so eventually this check won't be necessary. 

The WebVTT spec was updated here https://github.com/w3c/webvtt/pull/244 and, so far, FF is the only browser that matches the spec.

src/js/parsers/captions/vttparser.js
```diff
@@ -166,34 +169,42 @@ define(['parsers/captions/vttcue'], function(VTTCue) {
...
                         break;
                 }
             }, /:/, /\s/);
 
             // Apply default values for any missing fields.
             cue.region = settings.get('region', null);
             cue.vertical = settings.get('vertical', '');
-            cue.line = settings.get('line', 'auto');
+            var line = settings.get('line', 'auto');
+            if (line === 'auto' && defaults.line === -1) {
```
egreaves: A comment here would be helpful so we know this conditional applies to Safari.

src/js/parsers/captions/vttparser.js
```diff
@@ -166,34 +169,42 @@ define(['parsers/captions/vttcue'], function(VTTCue) {
...
+            }
+            cue.line = line;
             cue.lineAlign = settings.get('lineAlign', 'start');
             cue.snapToLines = settings.get('snapToLines', true);
             cue.size = settings.get('size', 100);
-            cue.align = settings.get('align', 'middle');
-            cue.position = settings.get('position', 'auto');
+            cue.align = settings.get('align', center);
+            var position = settings.get('position', 'auto');
+            if (position === 'auto' && defaults.position === 50) {
```
egreaves: Here as well.

PR #1566, opened 2016-11-22
Workaround for Android 'inert-visual-viewport' behavior

robwalch:
This is limited to horizontal sliders on Android where pageXOffset is not 0 - assuming this means the viewport is offset from zooming and not scrolling the page. It's not intended for all viewport settings where horizontal scrolling without zooming the viewport can occur. Most pages only scroll vertically at 100% (not zoomed), and only scroll horizontally when zoomed in so this should help compatibility with 'inert-visual-viewport' behavior.

JW7-1078

### DIFF
--- a/src/js/view/components/slider.js
+++ b/src/js/view/components/slider.js
@@ -5,6 +5,18 @@ define([
     'utils/helpers'
 ], function(Extendable, UI, SliderTemplate, utils) {
 
+    var getRailBounds = function(elementRail) {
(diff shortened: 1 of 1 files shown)


jw-kaurand:
This works pretty well for dragging.  Can we expand this solution to also work when tapping on the time slider?

robwalch:
Thx @jw-kaurand. Fixed it.

PR #1575, opened 2016-11-29
Clean up console output while running unit tests

hghazzi:
When running unit tests with karma, we were outputting log messages and 404 errors. Mocked out the logging and fixed the errors so that they do not appear in the console when running karma or grunt.


src/js/utils/backbone.events.js
```diff
@@ -185,9 +185,10 @@ define([
         var ev, i = -1, l = events.length;
         while (++i < l) {
             try {
-                (ev = events[i]).callback.apply(ev.context || context, args);
+                ev = events[i];
+                ev.callback.apply(ev.context || context, args);
             } catch(e) {
-                console.log('Error in "' + name + '" event handler:', e);
+                console.log('Error in "', name, '" event handler:', e);
```
robwalch: The string concatenation was intentional. We don't want spaces around the event name like this:
`" name "`.

test/unit/ajax-test.js
```diff
@@ -135,7 +135,7 @@ define([
         var done = assert.async();
 
         utils.ajax({
-            url: '//playertest.longtailvideo.com/vast/preroll.xml',
+            url: require.toUrl('./data/playlist.xml'),
```
robwalch: Make sure this works when running tests in the browser (open test/qunit/) and in karma.

PR #1576, opened 2016-11-30
Feature/combined maui controls ui update

madikarizma:
* Introduces fixed time slider on small player sizes
* Make fixed time slider a configuration option to always use fixed time slider controls vs. default
* On small player sizes... play, rewind, and next icons move to new display icons container above control bar
* Fixed next up display position for skins with inset control bar with max width
* Several updates that improve how other elements in the player react and appear with fixed time slider
* Updates to skins to share consistent fixed time slider appearance while keeping skin colors in tact
* Ensure fixed time slider CSS is still easy enough to override/customize

Fixes #
JW7-3262
JW7-3326


johnBartos:
I think we need to adjust the captions safe region when in small mode

screen shot 2016-11-30 at 13 58 05

johnBartos:
Long lists are also a slight problem:

screen shot 2016-11-30 at 14 00 11

johnBartos:
I think the icon here is technically centered but visually it looks a bit too high

screen shot 2016-11-30 at 14 01 11

johnBartos:
Another thing I'm seeing is that at small sizes (e.g. 325px), all the controlbar buttons are visible for a second or two and then disappear. Intentional?

dannyfinks:
Should related and sharing dock button not show up and small sizes?

madikarizma:
@dannyfinks I believe I was told to hide this, but I will double-check.

src/css/imports/display.less
```diff
@@ -44,45 +44,134 @@
...
+  .jw-breakpoint-0.jw-state-paused &,
+  .jw-breakpoint-0.jw-state-buffering &,
+  .jw-breakpoint-1.jw-state-playing &,
+  .jw-breakpoint-1.jw-state-paused &,
+  .jw-breakpoint-1.jw-state-buffering & {
+    padding-top: 14px;
+  }
+}
+
+@keyframes jw-display-icon-container-slide-up {
```
robwalch: Animations should be added to skins, not core player styles. This applies the animation to all skins, but makes it difficult to override or customize in customer skins.

Animations would be better organized as styles/mixins included in all skins.

egreaves:
- [ ] Duration shows above time slider below breakpoint 1 or when `timeSliderFixed = true`
- [ ] Captions are displayed on or below the time slider at breakpoints 3 and below
- [ ] The time slider doesn't display correctly when the controlbar is showing during ad playback
- [ ] Pause button in the display container does not work on touch devices
- [ ] The play/pause, rewind and next buttons are showing in the controlbar area at breakpoints where those buttons are shown in the display container

@madikarizma we can address the captions display in a separate ticket since it's outside of the scope of this PR.

egreaves:
Not sure if we want to address these now or as separate bug tickets.

- [ ] At/below ~355px in breakpoint-1, the dock icons' touch target overlaps with the display icons.
- [ ] Logo does not move up when controls are hidden
- [ ] Can’t navigate to HD/CC/Volume button tooltips at breakpoint 5 or less
- [ ] At breakpoint 6+, "Next Up" isn't positioned above the next button
- [ ] Audio player (controlbar only) does not display correctly when `"timeSliderAbove": true` is set in the config

PR #1578, opened 2016-12-01
Resume ad playback when the play display icon is tapped on mobile

egreaves:
Allow the user to resume playback by tapping the `play` display icon when an ad is paused.

Fixes #
JW7-3563


robwalch:
The model's state should not be buffering while an ad is playing. This is likely an issue with instream on mobile. This change could allow tampering with the playback state by tapping while video is loading.

egreaves:
@robwalch I considered that and was trying to track down what caused the model to be in a buffering state in the first place. I'll continue along that path.

PR #1582, opened 2016-12-05
Allow toggling controls on mobile when player is paused

egreaves:

On a mobile device when the player is paused, tapping on the view toggles the controls instead of resuming playback. Playback resumes when the user taps the display icon or the play button in the control bar.

Fixes #
JW7-3248

### DIFF
--- a/src/css/flags/touch.less
+++ b/src/css/flags/touch.less
@@ -44,4 +44,10 @@
             display: none;
         }
     }
+    &.jw-state-paused.jw-flag-controls-hidden {
+        .jw-controls {
+            display : none;
+        }
+    }
+
 }
--- a/src/css/flags/touch.less
+++ b/src/css/flags/touch.less
@@ -1,4 +1,5 @@
 @import "../imports/icons";
+@import "controls-disabled";
 
 .jw-flag-touch {
     &.jw-breakpoint-7,
--- a/src/css/flags/touch.less
+++ b/src/css/flags/touch.less
@@ -1,5 +1,4 @@
 @import "../imports/icons";
-@import "controls-disabled";
 
 .jw-flag-touch {
     &.jw-breakpoint-7,
@@ -45,9 +44,4 @@
             display: none;
         }
     }
-    
-    &.jw-state-paused.jw-flag-controls-hidden {
-        .jw-flag-controls-disabled;
-    }
-
 }
--- a/src/css/flags/touch.less
+++ b/src/css/flags/touch.less
@@ -44,10 +44,9 @@
             display: none;
         }
     }
+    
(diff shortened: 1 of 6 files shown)


src/js/view/view.js
```diff
@@ -928,6 +934,18 @@ define([
             _controlsTimeout = setTimeout(_userInactive, _timeoutDuration);
         }
 
+        function _toggleControls() {
```
pajong: ```_model.set('controls', true | false)```

egreaves: @pajong this would cause the player to override the `"controls": false` setting in the config, which we don't want to do.

pajong:
```
utils.toggleClass(_playerElement, 'jw-flag-controls-hidden', _controlsHidden);
_captionsRenderer.renderCues(true);
_controlsHidden = !_controlsHidden;
```

pajong:
@robwalch Can you please also take a look at this PR about introducing a new flag?

src/css/flags/touch.less
```diff
@@ -44,4 +45,9 @@
             display: none;
         }
     }
+    
+    &.jw-state-paused.jw-flag-controls-hidden {
+        .jw-flag-controls-disabled;
```
robwalch: Use the same rules as jw-flag-user-inactive (playing).

src/js/view/view.js
```diff
@@ -54,6 +54,7 @@ define([
             _captionsRenderer,
             _audioMode,
             _showing = false,
+            _controlsHidden = false,
```
robwalch: Instead of adding this property, use `utils.hasClass`

src/js/view/view.js
```diff
@@ -487,20 +488,25 @@ define([
             }
         }
 
-        function _touchHandler() {
-            if( (_model.get('state') === states.IDLE ||
+        function _touchHandler(displayIcon) {
+            if ((_model.get('state') === states.IDLE ||
```
robwalch: Can we reduce the number of calls to `_model.get('state')`? Too many.

src/js/view/view.js
```diff
@@ -928,6 +934,12 @@ define([
             _controlsTimeout = setTimeout(_userInactive, _timeoutDuration);
         }
 
+        function _toggleControls() {
+            _controlsHidden = !_controlsHidden;
+            utils.toggleClass(_playerElement, 'jw-flag-controls-hidden', _controlsHidden);
```
robwalch: Can we rename this jw-flag-controls-hidden?

.jwplayer.jw-flag-controls-hidden

egreaves:
Test this please

PR #1583, opened 2016-12-06
use host for equality test

mrbar42:
cases where a port is part of the hostname will fail the condition.

this changes the comparison to both be host based

### DIFF
--- a/src/js/providers/flash.js
+++ b/src/js/providers/flash.js
@@ -17,7 +17,7 @@ define([
         var a = document.createElement('a');
         a.href = config.flashplayer;
 
-        var sameHost = (a.hostname === window.location.host);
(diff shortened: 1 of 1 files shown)


johnBartos:
Looks reasonable, thanks for the PR. Going to see if there's a reason it's `hostname` now (it was `host` once before)

https://github.com/jwplayer/jwplayer/commit/84331cbe7557463bf416930787310c6c356fe1ac

robwalch:
What we're trying to do here is guess if Chrome will throttle the swf. Ports should be compared on the page and swf, so this is fix is correct. Chrome is removing the throttling functionality, replacing it with HTML5 by default in the coming months and through 2017. The flash throttled check will be removed once it's no longer needed in Chrome.

PR #1588, opened 2016-12-08
Chromecast Support (Dont Merge, creating PR build)

hghazzi:


hghazzi:
test this please

src/js/view/controlbar.js
```diff
@@ -90,7 +126,7 @@ define([
                 mute: muteButton,
                 volume: volumeSlider,
                 volumetooltip: volumeTooltip,
-                cast: button('jw-icon-cast jw-off', this._api.castToggle, this._localization.cast),
+                cast: !utils.isSafari()? castButton(this._localization.cast) : button('jw-icon-airplay jw-off', this._api.castToggle, this._localization.airplay),
```
robwalch: I'm not sure `isSafari` is the right check here. There are a few browsers that shouldn't have any cast element. What is your take on this?

robwalch: @hghazzi hello

hghazzi: How did we handle the fullscreen element in Safari iOS before we could do inline?

If we just want to make sure it never shows the casting element, we can flip the logic to only create the chromecast element for non-iOS Chrome, otherwise it will create an airplay element that will always be hidden for non-Safari browsers. 

If we don't want to generate the element in unsupported browsers, I can rework the logic to do so.

PR #1591, opened 2016-12-09
Add missing “interaction” play reason metadata

robwalch:
I found many calls in the view that would result in a play with no play reason data. This is not the reason for all unknown plays, but should improve reporting on "interaction" plays.

JW7-3344

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -182,11 +182,11 @@ define([
         };
 
         this.play = function (state, meta) {
-            if (!_.isBoolean(state)) {
+            if (_.isObject(state) && state.reason) {
                 meta = state;
             }
             if (!meta) {
-                meta = {reason: 'external'};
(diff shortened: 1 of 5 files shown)


robwalch:
@pajong make sure you test this. I just added code where I saw it was missing or needed improvement.

@dannyfinks this shows there are a few gaps in the interaction play reason testing. If the controlbar is shown, or the seek is called with keyboard, this will kick off playback.

dannyfinks:
Thanks for digging in. I'll update the tests to include these interactions.

robwalch:
test this please

pajong:
test this please

dannyfinks:
@robwalch when I focus on the player and start playback via keyboard shortcut's 1-9 when the player is idle the play reason is returned as UI Interaction. Perhaps it should be 4?

PR #1594, opened 2016-12-12
Add state listener for vpaid ad state

pajong:
JW7-2617


robwalch:
This is supposed to be done via the "instream provider" set with `instream.applyProviderListeners(provider)`.

Freewheel and Googima do it. You would just need to add it to vpaid playback for VAST.

PR #1595, opened 2016-12-13
adds accessibility support to nextUp feature

### LINKED ISSUE #1587: jw-nextup-close button lacks aria tags
### Expected Behavior

I would like to see the jw-nextup-close button in jwplayer pass an accessibility testing framework as beautifully as its other buttons.

### Actual Behavior

Nested within the `jw-nextup-container` class is a button with the class of `jw-nextup-close`. It seems like this is something that cannot currently be disabled. However, this button lacks aria-label, aria-hidden status, or role (whatever is most applicable in this case).

### Steps to reproduce

Look at any embedded jwplayer through an accessibility framework. [Here is an example sent through the WAVE accessibility report tool](http://wave.webaim.org/report#/https://support.jwplayer.com/customer/portal/articles/1406723#fndtn-method-1). You may have to select "No styles" to the left of the report to see all of the jwplayer parts because the button-in-question is a hidden button.

I'll take a screencap as a shortcut: example of jw-nextup-close button

You can see that it does not have what the other buttons have.

### Environment

Using Mac OSX, Chrome (but I think this is an issue regardless of these factors). JW Player 7.8.2 was tested -- link is above!

### Extra

Thanks so much for otherwise being very compliant and very accessible! Apologies if this is already a known-issue; did a search but didn't see anything related.

ablwr:

Adds aria-label support for usually-hidden 'close' button in Next Up feature.

Fixes #1587

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -40,6 +40,7 @@ define([
             loadingAd: 'Loading ad',
             rewind: 'Rewind 10s',
             nextUp: 'Next Up',
+            nextUpClose: 'Next Up Close',
             related: 'Related'
         },
         renderCaptionsNatively: false
--- a/src/js/view/components/nextuptooltip.js
+++ b/src/js/view/components/nextuptooltip.js
@@ -11,6 +11,7 @@ define([
         this._nextButton = nextButton;
         this._playerElement = playerElement;
         this.nextUpText = _model.get('localization').nextUp;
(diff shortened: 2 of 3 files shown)


egreaves:
Thanks @ablwr! Looks good.

ablwr:
🎉 Thank YOU! for the guidance. 🎉

PR #1600, opened 2016-12-14
Fix dock and display icons colliding on smaller part of breakpoint 1

madikarizma:
Acceptable that the only remaining scenario is when there are more than 2 dock icons and the player size is below ~270px. We can't really get around this without a design change that should be saved for JW8.

Fixes #
JW7-3636

### DIFF
--- a/src/css/imports/display.less
+++ b/src/css/imports/display.less
@@ -54,7 +54,7 @@ display icons
 
   display: table;
   height: 100%;
-  padding-bottom: 66px;
+  padding: @mobile-touch-target 0 66px;
   position: relative;
   width: 100%;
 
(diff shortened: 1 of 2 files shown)


src/css/imports/display.less
```diff
@@ -91,7 +91,7 @@ display icons
     &::before {
       font-size: @mobile-touch-target * 0.5;
       position: relative;
-      top: 1px;
+      // top: 1px;
```
egreaves: Remove this since we no longer need it

src/css/imports/display.less
```diff
@@ -91,7 +91,6 @@ display icons
     &::before {
       font-size: @mobile-touch-target * 0.5;
       position: relative;
-      // top: 1px;
```
madikarizma: @egreaves

PR #1604, opened 2016-12-16
Pass in hlsjsdefault param to source

pajong:
This is used to filter sources.
JW7-3614

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -504,7 +504,8 @@ define([
             function _setPlaylist(p) {
                 var playlist = Playlist(p);
                 playlist = Playlist.filterPlaylist(playlist, _model.getProviders(), _model.get('androidhls'),
-                    _model.get('drm'), _model.get('preload'), _model.get('feedid'), _model.get('withCredentials'));
+                    _model.get('drm'), _model.get('preload'), _model.get('feedid'), _model.get('withCredentials'),
+                    _model.get('hlsjsdefault'));
 
                 _model.set('playlist', playlist);
 
--- a/src/js/playlist/playlist.js
+++ b/src/js/playlist/playlist.js
@@ -13,7 +13,8 @@ define([
     };
 
     /** Go through the playlist and choose a single playable type to play; remove sources of a different type **/
-    Playlist.filterPlaylist = function(playlist, providers, androidhls, configDrm, preload, feedid, withCredentials) {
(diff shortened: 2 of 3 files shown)


src/js/controller/controller.js
```diff
@@ -504,7 +504,8 @@ define([
             function _setPlaylist(p) {
                 var playlist = Playlist(p);
                 playlist = Playlist.filterPlaylist(playlist, _model.getProviders(), _model.get('androidhls'),
-                    _model.get('drm'), _model.get('preload'), _model.get('feedid'), _model.get('withCredentials'));
+                    _model.get('drm'), _model.get('preload'), _model.get('feedid'), _model.get('withCredentials'),
+                    _model.get('hlsjsdefault'));
```
robwalch: Let's refactor `filterPlaylist` so that it takes `model` as the third param.

src/js/playlist/playlist.js
```diff
@@ -13,7 +13,8 @@ define([
     };
 
     /** Go through the playlist and choose a single playable type to play; remove sources of a different type **/
-    Playlist.filterPlaylist = function(playlist, providers, androidhls, configDrm, preload, feedid, withCredentials) {
+    Playlist.filterPlaylist = function(playlist, providers, androidhls, configDrm,
+                                       preload, feedid, withCredentials, hlsjsdefault) {
```
robwalch: Replace `androidhls, configDrm, preload, feedid, withCredentials, hlsjsdefault` with `model`.

`_formatSources` should take `item, model`

johnBartos:
This should be done in commercial the same way we used the `hlshtml` option. `hlsjs` shouldn't appear anywhere in OS and putting this option per-source is a PITA for an experimental flag

src/js/providers/providers.js
```diff
@@ -90,16 +90,21 @@ define([
                 var html5Idx = _.indexOf(providers, _.findWhere(providers, {name: 'html5'}));
                 providers.splice(html5Idx, 0, flashProvider);
             }
+            if (hlsjsdefault) {
+                var hlsIndex =  _.indexOf(providers, _.findWhere(providers, {name: 'hlsjs'}));
+                var hlsjsProvider = providers.splice(hlsIndex, 1)[0];
+                providers.unshift(hlsjsProvider);
```
robwalch: We can just insert hlsjs before html5 in commercial. The supports check won't let ungated Browsers use it.

PR #1616, opened 2016-12-20
Refactor playlist filter, and add hlsjsdefault

pajong:
Refactor to input model instead of a long list of model attributes.
Added hlsjsdefault to the sources.
Updated tests accordingly.
JW7-3614


pajong:
please test this

PR #1617, opened 2016-12-20
Improve display container for different player sizes, states, and flags.

madikarizma:
* Refactored some code
* Heavily commented conditional selectors (otherwise can be confusing)
* Fixed JW7-3689 where display icon position on taller videos would be at bottom (fix is not possible if aspect ratio in config is set to make video taller and it is wide—unless we add another player flag)

Fixes #
JW7-3689

### DIFF
--- a/src/css/flags/time-slider-above.less
+++ b/src/css/flags/time-slider-above.less
@@ -17,7 +17,7 @@
 
     // then give dock some padding
     .jw-dock {
-      padding: 0 2px;
+      padding: 0 1%;
     }
 
   }
@@ -33,7 +33,7 @@
 
     // then give dock buttons some space between each other
     .jw-dock-button {
-      margin: 4px 2px;
+      margin: 2% 1%;
     }
 
   }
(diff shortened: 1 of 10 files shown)


src/css/imports/dock.less
```diff
@@ -6,14 +6,6 @@
     display: block;
     opacity: 1;
 
-    .jw-breakpoint-0 & {
```
egreaves: Good move removing time slider specific rules

src/css/flags/time-slider-above.less
```diff
@@ -2,28 +2,127 @@
...
-
-      .jw-controls {
-        background: none;
-      }
-
+    /* do not show the controls background linear gradient when video is idle
+    state (except when cast available and we show controls on idle state) or
+    playing state when user inactive (the gradient in other scenarios blocks
+    video from clashing with the video colors, making it easy to see/use) */
+    &.jw-state-idle:not(.jw-flag-cast-available) .jw-controls,
```
egreaves: I prefer nesting `.jw-controls` here

src/css/flags/time-slider-above.less
```diff
@@ -35,55 +134,50 @@
...
-        height: @mobile-touch-target * 1.5;
-        padding: @slider-fixed-height (@mobile-touch-target / 4) 0;
-      }
-
+    broadcast and the override sets it to the height needed for anything that
+    isn't a live broadcast. the delay the player has before it detects a live
+    broadcast and adds causes a flicker. Handling control bar in this way fixes
+    that experience in time slider above mode. */
+    &:not(.jw-flag-ads) .jw-controlbar,
+    &:not(.jw-flag-live) .jw-controlbar {
```
egreaves: nest `.jw-controlbar`

src/css/flags/time-slider-above.less
madikarizma: @egreaves Same. I'm unsure why I did this here and for .jw-controlbar.

madikarizma:
@egreaves Updated

PR #1623, opened 2016-12-21
Fix next up container overlapping with time slider

madikarizma:
At some breakpoints, next up container would overlap with time slider/controls when that breakpoint added padding to the control bar area. Padding has been removed to create consistency throughout all breakpoints. Next up code under time slider above flag has been moved to it's appropriate place.

Fixes #
JW7-3677


egreaves:
@madikarizma Next Up is much higher above the time slider at breakpoint 6+ than at lower breakpoints 
screen shot 2016-12-21 at 9 54 17 pm

PR #1638, opened 2016-12-27
Ad playback changes for new timeslider flag

madikarizma:
* Remove breakpoints.less now that all breakpoint code has been moved to appropriate places
* Fix ads/live flag UI for time slider above flag
* Remove icon resizing at bigger breakpoints as it was decide this should be a later feature (was removed for last week's MAUI merge)

Fixes #
JW7-3157


dannyfinks:
@madikarizma @egreaves one of the issues I noticed that wasn't addressed in this PR is that when the control-bar fades during ad playback the skip button is moved down from the MAUI changes. Its position should be fixed irregardless of the control-bar visibility.

dannyfinks:
@madikarizma take a look at an IMA ad in fullscreen mode, the control-bar is no longer transparent

PR #1640, opened 2016-12-27
Use showNextUp param to toggle display of next up.

pajong:
JW7-3599

### DIFF
--- a/src/js/view/components/nextuptooltip.js
+++ b/src/js/view/components/nextuptooltip.js
@@ -139,8 +139,7 @@ define([
                 this.showNextUp = false;
                 return;
             }
-
-            this.showNextUp = true;
+            this.showNextUp = nextUp.showNextUp;
             this._nextButton.toggle(true);
             this.setNextUpItem(nextUp);
         },


src/js/view/components/nextuptooltip.js
```diff
@@ -140,7 +140,7 @@ define([
                 return;
             }
 
-            this.showNextUp = true;
+            this.showNextUp = nextUp.showNextUp;
```
egreaves: `nextUp.showNextUp` is `undefined` for playlists. 

Test with: `feature=playlist/playlist_next_up_display`.

src/js/view/components/nextuptooltip.js
```diff
@@ -139,8 +139,8 @@ define([
                 this.showNextUp = false;
                 return;
             }
-
-            this.showNextUp = true;
+            var showNextUp = nextUp.showNextUp;
+            this.showNextUp = (showNextUp === undefined) ? true : showNextUp;
```
egreaves: This isn't intuitive. I think it's better to  set `showNextUp` for playlist mode in the other PR.

PR #1644, opened 2016-12-28
Adjust captions container max height for each breakpoint

egreaves:
- Keep captions just above the control bar for each breakpoint. 
- At breakpoint 0, use the space above the controlbar buttons

Fixes #
JW7-3412


robwalch:
> I'll handle the updates when @madikarizma 's PR for controlbar timeSliderAbove coloration is merged.

@jw-kaurand which PR is that? Please include a link or PR #.

PR #1647, opened 2016-12-28
Fix state when mobile autostart fails

hghazzi:
If you disconnect form a chromecast while its a mobile device, the
state will be incorrect.

JW7-3739

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -61,14 +61,12 @@ define([
                     this.set('flashThrottle', throttled);
                     this.set('flashBlocked', throttled);
                     break;
-
                 case 'flashBlocked':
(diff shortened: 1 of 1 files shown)


src/js/controller/model.js
```diff
@@ -147,14 +138,15 @@ define([
...
                     break;
-
                 case 'visualQuality':
                     var visualQuality = _.extend({}, data);
                     this.mediaModel.set('visualQuality', visualQuality);
                     break;
-
                 case 'autoplayFailed':
                     this.set('autostartFailed', true);
+                    if (this.mediaModel.get('state') === states.PLAYING) {
```
egreaves: Let's create a local var for `this.mediaModel` since it's used several times.

PR #1650, opened 2016-12-29
Hide display icons and the controls gradient during a VPAID ad

jw-kaurand:
Hide display icons and the controls gradient during a VPAID ad so that do not receive play/pause buttons at small sizes and we do not get the jw-controls background gradient at small sizes.

JW7-3706


egreaves:
test this please

PR #1652, opened 2016-12-29
Bump of specificity of googima controlbar

pajong:
This had less specificity than inactive, so controlbar was being hidden when inactive.
JW7-3740

### DIFF
--- a/src/css/flags/ads.less
+++ b/src/css/flags/ads.less
@@ -78,7 +78,7 @@
 
 }
 
-.jwplayer.jw-flag-ads-googleima {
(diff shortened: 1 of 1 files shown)


src/css/flags/ads.less
```diff
@@ -78,7 +78,7 @@
 
 }
 
-.jwplayer.jw-flag-ads-googleima {
+.jwplayer.jw-flag-ads.jw-flag-ads-googleima {
```
egreaves: This results in controls being shown all the time for googima ads instead of when `vpaidcontrols === true`.

PR #1656, opened 2016-12-30
Improve time slider above background gradient. #3753

madikarizma:
* Remove gradient from controls element
* Apply gradient to controlbar element
* Fixes captions, ads, and any other displays appearing above the time slider in time slider above move from being covered by gradient background

Fixes #
JW7-3753

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -307,23 +307,47 @@ define([
         this.handleColorOverrides = function() {
             var id = _model.get('id');
 
-            function addStyle(elements, attr, value) {
+            function getRgba(color, opacity) {
+                var canvas, context, data;
+
+                canvas = document.createElement('canvas');
+                canvas.height = 1;
+                canvas.width = 1;
+
+                context = canvas.getContext('2d');
+                context.fillStyle = color;
+                context.fillRect(0, 0, 1, 1);
+
+                data = context.getImageData(0, 0, 1, 1).data;
+
+                return 'rgba(' + data[0] + ', ' + data[1] + ', ' + data[2] + ', ' + opacity + ')';
+            }
+
+            function addStyle(elements, attr, value, bundle) {
                 if (!value) {
                     return;
                 }
 
-                elements = utils.prefix(elements, '#' + id + ' ');
+                /* if bundle is true, bundle the first selector string to the
+                player element versus defaulting to setting as a child of. i.e.
+                #player1.jw-breakpoint-0 vs. #player1 .jw-breakpoint-0 */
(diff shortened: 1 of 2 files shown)


egreaves:
@madikarizma  `0.75` opacity is a bit harsh on a white video frame: 
screen shot 2016-12-30 at 1 09 17 am

It looks a little better at `0.5` but not ideal:
screen shot 2016-12-30 at 1 11 47 am

I think we need to increase the height of the gradient to be closer to this:

screen shot 2016-12-30 at 1 20 48 am

It'll make for a softer transition from 0.x > 0 opacity while not obscuring the video.

madikarizma:
@egreaves Below addressed in latest commits

madikarizma:
@egreaves Latest commit also addresses the following:
* Enable bundling first selector to player element via addStyle function vs only allowing selectors to be applied to element as child selectors. i.e. `skin = {
      "active": "red",
      "inactive": "yellow",
      "background": "blue"
    }`
* Add small player styles/overrides for background color on time slider
and control bar when colors are defined in config

egreaves:
@madikarizma I like the change for the background color override. I still think it looks a bit harsh on a white frame and the time slider is barely visible. 

screen shot 2016-12-31 at 9 59 07 am

Also, merge the latest from master into your PR to get the skins color fixes.

src/js/view/view.js
```diff
@@ -307,12 +307,31 @@ define([
...
+
+                canvas = document.createElement('canvas');
+                canvas.height = 1;
+                canvas.width = 1;
+
+                context = canvas.getContext('2d');
+                context.fillStyle = color;
+                context.fillRect(0, 0, 1, 1);
+
+                data = context.getImageData(0, 0, 1, 1).data;
```
robwalch: This can be done with math is color is always a hex string.

If we're uncertain of the format and must use this method we should reuse the same canvas/context.

src/js/view/view.js
```diff
@@ -367,6 +391,20 @@ define([
...
+                '.jw-flag-time-slider-above .jw-background-color.jw-controlbar',
+            ], 'background', backgroundColorGradient, true);
+
+            addStyle([
+                // time slider gradient background color for small player size only
+                '.jw-breakpoint-0 .jw-background-color.jw-slider-time',
+                '.jw-breakpoint-1 .jw-background-color.jw-slider-time',
+                '.jw-flag-time-slider-above .jw-background-color.jw-slider-time',
+            ], 'background', 'none', true);
+
```
robwalch: What's going on here?

madikarizma:
Closed in favor of https://github.com/jwplayer/jwplayer/pull/1673

PR #1659, opened 2016-12-30
Position Freewheel nonlinear above control bar with timeSliderAbove

jw-kaurand:
Position Freewheel nonlinear above control bar with timeSliderAbove.  This must be done because Freewheel does not position its nonlinear in its plugin container.

JW7-3768

### DIFF
--- a/src/css/flags/time-slider-above.less
+++ b/src/css/flags/time-slider-above.less
@@ -367,7 +367,8 @@
     }
 
     /* Move non-linear ad display above the time slider */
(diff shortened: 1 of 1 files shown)


robwalch:
All freewheel styles should be defined in freewheel.

https://github.com/jwplayer/jwplayer-ads-freewheel/blob/master/src/css/styles.less

This is how it's going to have to be for all ads plugins so either use a less specific plugin contain class here or include a fixed size in the plugin.

PR #1662, opened 2017-01-03
Pass fullscreen event through to providers

johnBartos:


robwalch:
Revert this in master. `provider.setFullscreen` is only to be used with html5 when the player DOM element cannot go fullscreen and the video tag is the only element that can render in fullscreen (iOS).

To update a provider with new player with height and stretching option use `provider.resize`.

@johnBartos @egreaves

PR #1668, opened 2017-01-03
[REVIEW] Move attach/detach out of providers

pajong:


src/js/controller/instream-html5.js
```diff
@@ -92,6 +92,7 @@ define([
             if (_currentProvider) {
                 _currentProvider.detachMedia();
                 _currentProvider.off();
+                _currentProvider.off(null, null, _this);
```
robwalch: `_currentProvider.off()` should have removed all event listeners. Why do that and `_currentProvider.off(null, null, _this)`?

src/js/controller/model.js
```diff
@@ -233,6 +233,16 @@ define([
             this.set('renderCaptionsNatively', _provider.renderNatively);
         };
 
+        this.detachMedia = function() {
+            _provider.off('all', _videoEventHandler, this);
+            return _provider.detachMedia();
+        };
+
+        this.attachMedia = function() {
```
robwalch: You probably want to add `_provider.off('all', _videoEventHandler, this);` here to make sure never end up attaching the same listeners more than once.

src/js/providers/flash.js
```diff
@@ -222,10 +220,8 @@ define([
                     }
                 },
                 detachMedia: function() {
```
robwalch: Can we remove `attachMedia` and `detachMedia` completely? The `_beforecompleted` logic should be handled in the model.

src/js/providers/html5.js
```diff
@@ -745,22 +703,15 @@ define([
...
                 _this.trigger(events.JWPLAYER_MEDIA_BEFORECOMPLETE);
-                // This event may trigger the detaching of the player
-                //  In that case, playback isn't complete until the player is re-attached
-                if (!_attached) {
-                    return;
-                }
 
-                _playbackComplete();
+                //TODO: not sure how to handle this yet
+                //_playbackComplete();
```
robwalch: Handle it in the model when it receives `JWPLAYER_MEDIA_BEFORECOMPLETE`.

src/js/controller/instream-adapter.js
```diff
@@ -83,18 +83,21 @@ define([
             _instream.on('all', _instreamForward, this);
             _instream.on(events.JWPLAYER_MEDIA_TIME, _instreamTime, this);
             _instream.on(events.JWPLAYER_MEDIA_COMPLETE, _instreamItemNext, this);
+            _instream.on(events.JWPLAYER_MEDIA_BEFORECOMPLETE, function() {
```
robwalch: We can only trigger 'complete' if there is no postroll, or after a postroll is complete.

src/js/providers/flash.js
```diff
@@ -341,19 +323,8 @@ define([
                         this.trigger(e.type);
                     }, this);
 
-                    _swf.on(events.JWPLAYER_MEDIA_BEFORECOMPLETE, function(e){
-                        _beforecompleted = true;
+                    _swf.on(events.JWPLAYER_MEDIA_BEFORECOMPLETE, function(e) {
```
robwalch: What if providers just fire 'complete' and the model turns it into 'beforeComplete'? Then you don't have to handle 'beforeComplete' in instream.

src/js/providers/html5.js
```diff
@@ -703,6 +659,10 @@ define([
             }
         };
 
+        this.playbackComplete = function() {
+            clearTimeout(_playbackTimeout);
+        };
```
robwalch: Instead of adding `provider.playbackComplete`, just clear this on ended. We don't need it in the providers.

PR #1669, opened 2017-01-04
Resume flash streams after cast disconnection

hghazzi:
The flag that handles the queue commands was not being updated
correctly, causing the flash provider to not be able to resume playing
the current item.

JW7-3755


hghazzi:
test this please

PR #1674, opened 2017-01-04
Enable small player color override styles

madikarizma:
* Lazy creation of the canvas color context
* Update color override comments
* Fix bg gradient override being set. When background color is not set in config, background gradient override was still being set. Fixed.
* Update color overrides
* Enable bundling first selector to player element via addStyle
function vs only allowing child selectors of player element
* Add small player styles/overrides for background color on time slider
and control bar

JW7-3732

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -79,6 +79,9 @@ define([
             // Used to differentiate tab focus events from click events
             _focusFromClick = false,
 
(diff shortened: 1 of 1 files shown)


robwalch:
@madikarizma @jw-kaurand Adding a custom bg color to a couple of views in test/manual/view.html  would help verify this one.

PR #1676, opened 2017-01-04
Update cast icon

hghazzi:
Updated css of cast icon to look better on both android and desktop

JW7-3721

### DIFF
--- a/src/css/imports/controlbar.less
+++ b/src/css/imports/controlbar.less
@@ -94,6 +94,9 @@
   display: none;
   margin: 0;
   padding: 0;
+  width: 32px;
+  position: relative;
+  top: 1px;
 
   button {
     --connected-color: @active-color;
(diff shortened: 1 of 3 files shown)


src/css/imports/controlbar.less
```diff
@@ -94,6 +94,9 @@
   display: none;
   margin: 0;
   padding: 0;
+  width: 32px;
```
robwalch: Can we use `em`s? 32px when not zoomed (16px fontsize) is 2em.

robwalch: Use Command + / Command - (Menu > View > Zoom In/Out) to compare the icon to it's neighbors.

hghazzi:
@robwalch updated

robwalch:
Thanks!

PR #1677, opened 2017-01-05
Show control bar on vpaid if vpaid-controls is set

pajong:
JW7-3740


dannyfinks:
@pajong is only IMA impacted by this?

pajong:
@dannyfinks Yes, this only impacts googima.

PR #1681, opened 2017-01-05
Improve time slider above gradient on white video frames

madikarizma:
Fixes #
JW7-3732


madikarizma:
@msiddique I'm not sure how much better we can do here without blowing out white icons too much.

PR #1685, opened 2017-01-06
Do not fire ad complete on skipped

pajong:
Ad skip and ad complete both call instreamItemNext to see if there is any other ads in the ad pod.
Do not fire adComplete event in instreamItemNext, since skipping an ad can also fire adComplete.
Instead, make instreamItemComplete function that fires adComplete, and let that function call instreamItemNext to check for next adpod item.
JW7-3777

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -82,7 +82,7 @@ define([
 
             _instream.on('all', _instreamForward, this);
             _instream.on(events.JWPLAYER_MEDIA_TIME, _instreamTime, this);
-            _instream.on(events.JWPLAYER_MEDIA_COMPLETE, _instreamItemNext, this);
+            _instream.on(events.JWPLAYER_MEDIA_COMPLETE, _instreamItemComplete, this);
             _instream.init();
 
             // Make sure the original player's provider stops broadcasting events (pseudo-lock...)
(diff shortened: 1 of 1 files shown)


src/js/controller/instream-adapter.js
```diff
@@ -164,8 +164,10 @@ define([
             }
 
             if (_array && _arrayIndex + 1 < _array.length) {
-                // fire complete event
-                this.trigger(events.JWPLAYER_MEDIA_COMPLETE, data);
+                // fire complete event if not ad skipped
+                if (e.type !== events.JWPLAYER_AD_SKIPPED) {
+                    this.trigger(events.JWPLAYER_MEDIA_COMPLETE, data);
+                }
```
robwalch: @pajong this shouldn't be necessary. This points to a problem with the ad plugins and instream that you missed:

`_instreamItemNext` is fired on mediaComplete. Skipping an add should not trigger mediaComplete.
```
_instream.on(events.JWPLAYER_MEDIA_COMPLETE, _instreamItemNext, this);
```

Did it at some point? Because it looks like you already tried to work around this with `setupSkipButton` `customNext`:
```js
        this.setupSkipButton = function(skipoffset, options, customNext) {
            _model.set('skipButton', false);
            if (customNext) {
                _instreamItemNext = customNext;
            }
            _instream._adModel.set('skipMessage', options.skipMessage);
            _instream._adModel.set('skipText', options.skipText);
            _instream._adModel.set('skipOffset', skipoffset);
            _model.set('skipButton', true);
        };
```

pajong: ```_instreamItemNext``` is also called when ad is skipped.
The ```customNext``` was created for freewheel, since freewheel handles ad pod by itself.
Updated the PR so that ```_instreamItemNext``` does not fire complete event. Made a separate function that is triggered when instream item completes.

PR #1686, opened 2017-01-06
Use Object.define to workaround Chrome 56 object property bug

### LINKED ISSUE #1684: Player fails to load flv files on Chrome 56
### Expected Behavior

Player should play flvs on all browsers

### Actual Behavior

Chrome 56 a console error occurs and the player never loads

### Steps to reproduce

1. https://erikdonohoo.github.io/jwplayer-flash-player-issue/ . (view the error)
2. Code is here https://github.com/erikdonohoo/jwplayer-flash-player-issue

### Environment

- Mac OS X 10.12.2 (also 10.12.1)
- Chrome 56 (current beta, soon to be stable in a week)

This is a major issue that will soon impact all users trying to play flash videos on chrome.  We are happily paying customers right now, but since we use a flash video recorder in the browser, we require the ability to play flvs.

robwalch:
In Chrome 56 setting properties on object elements fails. The player depends on methods it adds to the object element for communication between Flash and JavaScript.

These changes workaround the issue and ensure we only setup with primary flash when swf object properties are being set.

We've opened an issue with Chrome to resolve the issue for existing players:
https://bugs.chromium.org/p/chromium/issues/detail?id=678986

JW7-3794
Fixes #1684

### DIFF
--- a/src/js/controller/setup-steps.js
+++ b/src/js/controller/setup-steps.js
@@ -166,7 +166,6 @@ define([
                 height: height.toString().indexOf('%') > 0 ? height : (height + 'px')
             });
             var swf = EmbedSwf.embed(flashHealthCheckSwf, testContainer, flashHealthCheckId, null);
-            parentElement.replaceChild(testContainer, originalContainer);
             var done = function() {
                 if (embedTimeout === -1) {
                     return;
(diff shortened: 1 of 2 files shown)


robwalch:
#flashpocalypse

PR #1689, opened 2017-01-06
Remove Unnecessary :not

jw-kaurand:
Removes unnecessary :not that was interfering with styles of the sharing plugin

JW7-3795

### DIFF
--- a/src/css/flags/time-slider-above.less
+++ b/src/css/flags/time-slider-above.less
@@ -358,7 +358,7 @@
     }
 
     /* Move non-linear ad display above the time slider */
-    .jw-plugin:not(.jw-plugin-related) {
+    .jw-plugin {
       bottom: (@mobile-touch-target * 1.5);
     }
 


pajong:
@jw-kaurand That is a line that @hghazzi  added. @hghazzi Can you make sure that removing this will not break something else?

src/css/flags/time-slider-above.less
```diff
@@ -358,7 +358,7 @@
     }
 
     /* Move non-linear ad display above the time slider */
-    .jw-plugin:not(.jw-plugin-related) {
+    .jw-plugin {
```
hghazzi: Don't think I made this, change, might have been a change that @madikarizma made, or a bad merge. @jw-kaurand this css should not really do anything to related because the overlay should would cover up the time slide, correct?

jw-kaurand: It shouldn't affect the related plugin, however if we start enumerating all the plugins that should not be affected here then we get a style which will never be able to be overridden.  This ends up making it very hard to set the bottom back to 0, like when we're in the inactive state and there's a Google IMA nonlinear ad.  Sharing and Related are handling setting their bottom values inside the own CSS so we don't want to handle it here.

PR #1691, opened 2017-01-09
Cleaned up grunt and console output

hghazzi:
- Removed references to a module in grunt that is no longer used
- Mocked out server requests that were outputting errors in the console
while building


test/unit/ajax-test.js
```diff
@@ -221,6 +224,8 @@ define([
                 done();
             }
         });
+        server.respond();
```
robwalch: This will create the same response in every browser. The purpose of this test is to verify xhr error handling across platforms.

hghazzi:
test this please

robwalch:
So that mocked response matches the native results across all browsers? Are you sure? Have you tested it?

hghazzi:
I didn't test it myself, but everything is passing in jenkins, though we don't run unit tests for IE in browserstack. I see the we have `grunt build:remote` that is supposed to run IE tests, but that's not part of the default build. Is there anywhere we do run unit tests against IE?

https://jenkins.jwplayer.com/view/JW7/job/jw7-commercial-unit-tests/6795/testReport/

According to the docs, it should work for all browsers and platforms except for IE by default. To support IE, we need to load `sinon-ie-1.17.6.js`, which we currently don't do. 

http://sinonjs.org/docs/#server

robwalch:
Add `grunt karma:browserstack_ie11`

hghazzi:
updated jenkins to run unit tests in browserstack ie11, tests passed

https://jenkins.jwplayer.com/view/JW7/job/jw7-commercial-unit-tests/6830/testReport/

test/unit/ajax-test.js
robwalch: We're not just testing various response handlers, we're testing integration with XHR. This replaces XHR with fake XHR. That limits the code coverage we get because it polyfills the browsers the tests run in.

robwalch:
@hghazzi please fix this: (open console) http://player-pr-test-jenkins.longtailvideo.com/builds/5079/archive/node_modules/jwplayer/test/

Maybe it* just needs to be added to the archive list.

hghazzi:
@robwalch no more console error

http://player-pr-test-jenkins.longtailvideo.com/builds/5268/archive/node_modules/jwplayer/test/

PR #1693, opened 2017-01-09
More tweaks to MAUI and Casting

hghazzi:
- Icon no longer changes between different player sizes
- Airplay icon has active color when active

JW7-3798


src/css/flags/time-slider-above.less
```diff
@@ -187,8 +187,16 @@ of the file to be defined separately.
...
 
     .jw-group {
-      .jw-icon-cast button {
-        width: 37px;
+      .jw-icon-cast {
+        width: auto;
+
+        button {
+          font-size: inherit;
+          width: 44px;
```
jw-kaurand: It'd more consistent if we define width and height relative to @mobile-touch-target so that everything's pulling 44px from that variable.

madikarizma: +1

madikarizma: @hghazzi Just took care of this

PR #1694, opened 2017-01-09
Trigger firstFrame before dom manipulation

johnBartos:
Get a more accurate time to first frame by triggering before DOM manipulation.

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -337,13 +337,15 @@ define([
 
         function _playingHandler() {
             _this.setState(states.PLAYING);
+            // For accuracy, trigger firstFrame before manipulating the DOM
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -337,13 +337,15 @@ define([
 
         function _playingHandler() {
             _this.setState(states.PLAYING);
+            // For accuracy, trigger firstFrame before manipulating the DOM
+            _this.trigger(events.JWPLAYER_PROVIDER_FIRST_FRAME, {});
+
```
robwalch: I don't think we're loosing more than a millisecond adding attributes to the video tag. We try to trigger player events at the end of function (after any mutation) to avoid any errors caused by event handler changes.

johnBartos: Fair enough, it's very quick. Closing this out.

PR #1698, opened 2017-01-09
Fix dark icon colors for five/vapor under time slider above flag

madikarizma:
Also, fix bug preventing inset control bar overrides under time slider above flag

JW7-3805

### DIFF
--- a/src/css/skins/five.less
+++ b/src/css/skins/five.less
@@ -22,8 +22,6 @@
 
     @option-active-bg-color: rgba(255, 255, 255, 0.6);
 
-
-
     @rail-height: .2em;
     @thumb-size: 1em;
     @cue-height: .4em;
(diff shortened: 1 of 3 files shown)


src/css/skins/five.less
```diff
@@ -133,4 +131,24 @@
         border-radius: @five-border-radius;
     }
 
+    // override control bar element colors that don't look good in time slider above mode
+    &.jw-flag-time-slider-above,
+    &.jw-breakpoint-0,
+    &.jw-breakpoint-1 {
+      .jw-controlbar {
+        .jw-text {
```
egreaves: Couldn't we just set the `@inactive-color` and `@hover-color` vars for `.jw-flag-time-slider-above` etc for both skins?

madikarizma: @egreaves The goal here was to only override those colors for control bar elements and when in time slider above flag/small player size. We don't want to override the colors anywhere else. However, there is a cleaner way I could do this.. thinking about it.

robwalch: What's wrong with custom styling in a skin file?

egreaves: @robwalch nothing's wrong with it. I was asking if we could update the color var, which should have been a comment, not a review, but after talking with @madikarizma I understand why what I was thinking isn't possible. I'm ok with merging this.

PR #1702, opened 2017-01-10
Fix display icon alignment inconsistency

madikarizma:
* Buffering state would cause a small jump up in location of display icon
* Buffering icon not centered on player viewport when under user inactive flag because padding of element was not consistent with other states

JW7-3742


madikarizma:
@jw-kaurand Good catch. Last commit should have fixed it.

PR #1720, opened 2017-01-17
Truncate ad text in control bar

egreaves:

Ensure ad text doesn't break the control bar by pushing buttons on the right up. This truncates the text with an ellipses and keeps the control bar's height fixed.
Fixes #
JW7-3829

### DIFF
--- a/src/css/flags/ads.less
+++ b/src/css/flags/ads.less
@@ -47,18 +47,6 @@
 
         .jw-text-alt {
             display: inline-block;
-            height: @controlbar-height;
-            line-height: @controlbar-height;
-            margin: auto -0.5em;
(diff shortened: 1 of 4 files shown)


robwalch:
You haven't handled "Live" controlbar.

Can the same be acheived without changing the positioning of the center groups? That's a very major layout change.

robwalch:
@madikarizma can you take a look at this? Evol's changes look a bit drastic for keeping the controlbar groups (center + right) on a single line.

egreaves:
@robwalch I see the "Live" issue. Without changing the center group positioning, lengthy text would either push the right group outside the player or up above the center group. I'll work with @madikarizma to improve the solution.

PR #1733, opened 2017-01-18
Add audio mode check to prevent undefined

madikarizma:
Audio mode flag would be applied despite value being undefined, which is because _model.get('audioMode') hasn't been set prior to calling _setTimesliderFlags to determine player flags. Added call to check audio mode and give it boolean value.

Bug created by result of 'undefined' value:
screen shot 2017-01-18 at 5 54 06 pm


egreaves:
This PR addresses the audio flag not being set correctly: https://github.com/jwplayer/jwplayer/pull/1732

PR #1736, opened 2017-01-19
Fix control bar showing during playback despite user inactive

madikarizma:
Inset control bars do not hide during playback/buffering when user inactive flag takes over.

@robwalch @egreaves This may be the one scenario where there's no choice but to have a big selector, as the following is pretty specific and overrides the flags for inset control bar skins:

&:not(.jw-flag-time-slider-above):not(.jw-flag-audio-player):not(.jw-flag-small-player)

JW7-3860

### DIFF
--- a/src/css/imports/controlbar.less
+++ b/src/css/imports/controlbar.less
@@ -2,7 +2,7 @@
 @import "vars";
 
 .jw-controlbar {
-  display: table;
+  display: inline-block; // for correct display in IE browsers
   position: absolute;
   bottom: 0;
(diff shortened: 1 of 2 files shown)


robwalch:
The reason inset control bars show is because of the inset display setting:
```css
display: inline-block; // for correct display in IE browsers
```

Can we just make that the default and remove it from here?

madikarizma:
@robwalch Appears to hold up with this approach. It's possible IE might not like it, but we'll see in testing. Other browsers are ok.

PR #1744, opened 2017-01-20
Prevent the display icon from being tappable in IMA ads on mobile

jw-kaurand:
Prevent the display icon from being tappable in IMA ads on mobile so that we do not receive two taps on the UI, which is interpretted as a quick pause and unpause action.

JW7-3888

### DIFF
--- a/src/css/flags/ads.less
+++ b/src/css/flags/ads.less
@@ -67,6 +67,13 @@
         .jw-controlbar {
             font-size: 1em;
         }
+
+        .jw-display-icon-display {
+            &,
+            .jw-icon-display {
+                pointer-events: none;
+            }
+        }
     }
 
     &.jw-skin-seven .jw-controlbar {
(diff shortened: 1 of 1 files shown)


src/css/flags/ads.less
```diff
@@ -67,6 +67,13 @@
         .jw-controlbar {
             font-size: 1em;
         }
+
+        .jw-display-icon-display {
+            &,
```
hghazzi: Not very familiar with this usage of `&`. Is this the same as 

```
.jw-display-icon-display {
  pointer-events: none;
  .jw-icon-display {
    pointer-events: none;
  }
}
```

egreaves: It's the same as:
```
.jw-display-icon-display, 
.jw-display-icon-display .jw-icon-display {
      pointer-events: none;
}
```

PR #1755, opened 2017-01-24
Base Style Fixes for MAUI

robwalch:
- Set basic styles in player, not in seven skin so that custom skins without color and background-color styles are not broken.
- Don't apply any user-inactive rules when playing media-audio
- Set display-icon-container width and height to "auto" to prevent custom skins from turning them into ovals
- Float display-icon-container(s) left so that the 3 center button always show horizontally even when custom skins change the display mode to block or table
- Make controlbar display table by default and only inline-block for inset controlbar skins

JW7-3901

### DIFF
--- a/src/css/imports/controlbar.less
+++ b/src/css/imports/controlbar.less
@@ -2,7 +2,7 @@
 @import "vars";
 
 .jw-controlbar {
-  display: inline-block; // for correct display in IE browsers
+  display: table;
   position: absolute;
   bottom: 0;
   height: @controlbar-height;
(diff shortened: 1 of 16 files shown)


src/css/imports/controlbar.less
```diff
@@ -2,7 +2,7 @@
 @import "vars";
 
 .jw-controlbar {
-  display: inline-block; // for correct display in IE browsers
+  display: table;
```
egreaves: Since you changed this back to `table`, you can also remove the rule specific to time-slider-above:

```
.jw-flag-time-slider-above:not(.jw-flag-ads-googleima) {

  .jw-controlbar {
    display: table;
  }
```

PR #1763, opened 2017-01-25
Use the original colors for time sliders in timeSliderAbove mode

jw-kaurand:
Remove some styles that set the background on the jw-progress to none.  It uses the colors defined in the skin instead.
Removed instances of some variables which were only ever set and not ever used.

JW7-3927

### DIFF
--- a/src/css/flags/time-slider-above.less
+++ b/src/css/flags/time-slider-above.less
@@ -282,7 +282,6 @@
       .jw-knob {
         border: none;
         box-shadow: none;
-        border-radius: 0;
       }
 
       .jw-knob {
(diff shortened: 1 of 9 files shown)


src/css/flags/time-slider-above.less
```diff
@@ -286,9 +280,13 @@
       .jw-progress,
       .jw-buffer,
       .jw-knob {
-        background: none;
         border: none;
         box-shadow: none;
+        border-radius: 0;
```
robwalch: Why 'undo' this? Shouldn't radius only be set in skin mixins and seven styles only?

PR #1766, opened 2017-01-26
Fix for rewind icon for audio only players

hghazzi:
JW7-3912

### DIFF
--- a/src/css/imports/mixins.less
+++ b/src/css/imports/mixins.less
@@ -366,6 +366,7 @@
 
               &.jw-icon-rewind {
                   line-height: @controlbar-height;
+                  margin: -1px 0;
               }
           }
         }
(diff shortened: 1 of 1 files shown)


src/css/imports/mixins.less
```diff
@@ -363,6 +363,10 @@
           .jw-text-countdown {
               height: auto;
               line-height: normal;
+
+              &.jw-icon-rewind {
```
robwalch: This looks too specific. Is there a `.jw-icon-rewind` class definition higher up you can add this too (like where the :before rule is defined for the icon)?

hghazzi: there are no other references for .jw-icon-rewind that are applicable.

robwalch: icons.less

PR #1769, opened 2017-01-27
Fixes for timeSliderAbove=false

egreaves:
- Fix inset control bar at small breakpoints for custom skins
- Show display icons on view even when `timeSliderAbove = false`

Fixes #
JW7-3943

### DIFF
--- a/src/css/imports/mixins.less
+++ b/src/css/imports/mixins.less
@@ -23,12 +23,13 @@
 
 .inset-controlbar() {
 
+  &:not(.jw-flag-small-player) {
     .jw-controlbar {
-        display: inline-block;
+      display: inline-block;
     }
+  }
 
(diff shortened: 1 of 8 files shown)


src/css/imports/mixins.less
```diff
@@ -396,20 +400,6 @@
...
                 }
             }
-
-            // When we're in an error or buffering state...
-            &.jw-error,
-            &.jw-state-error,
-            &.jw-state-buffering {
-                // Show the normal state for the icons so we do not imply its clickable
-                .jw-display-icon-container {
-                    background-color: @display-bkgd-color;
```
robwalch: @jw-kaurand this was making the bg color of the buffering icon gray in skins because it's more specific than the skin color rules. When we don't want something to happen, don't override, omit.

PR #1774, opened 2017-01-27
Ensure that background-image is removed for Safari

jw-kaurand:
Ensure that background-image is removed for Safari in order to allow overrides to clear out any gradients that could affect display

JW7-3908

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -391,7 +391,7 @@ define([
                 '.jw-option.jw-active-option',
                 // slider fill color
                 '.jw-progress'
-            ], 'background', activeColor);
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -391,7 +391,7 @@ define([
                 '.jw-option.jw-active-option',
                 // slider fill color
                 '.jw-progress'
-            ], 'background', activeColor);
+            ], 'background', 'none ' + activeColor);
```
pajong: What does this 'none' in front of active color do?

egreaves: @pajong in Safari, it's necessary to clear the background image so that it does not affect custom colors or gradients.

PR #1780, opened 2017-01-31
Implement qualityLabels

johnBartos:
RFC on the quality-labels utils. Not sure how clear/clean the code is - I developed it with TDD in mind.

https://github.com/jwplayer/jwplayer-commercial/pull/2920 (2/2) 
JW7-2266

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -126,6 +126,8 @@ define([
             config.playlist = [ obj ];
         }
 
+        config.qualityLabels = config.qualityLabels || config.hlslabels;
+
         return config;
     };
 
--- a/src/js/providers/flash.js
+++ b/src/js/providers/flash.js
@@ -48,57 +48,6 @@ define([
             }
         };
 
-        var _customLabels = _getCustomLabels();
-
-        function _getNearestCustomLabel(sourceKBps) {
-            // get indexed value
-            var label = _customLabels[sourceKBps];
-            if (!label) {
-                //find nearest
-                var lastDiff = Infinity;
-                var i = _customLabels.bitrates.length;
-                while (i--) {
-                    var diff = Math.abs(_customLabels.bitrates[i] - sourceKBps);
-                    if (diff > lastDiff) {
-                        break;
-                    }
-                    lastDiff = diff;
-                }
-                label = _customLabels.labels[_customLabels.bitrates[i + 1]];
-                // index
-                _customLabels[sourceKBps] = label;
-            }
-            return label;
-        }
-
-        /** Indexed Custom Labels **/
-        function _getCustomLabels() {
-            var hlsLabels =_playerConfig.hlslabels;
-            if(!hlsLabels) {
-                return null;
-            }
-            var labels = {};
-            var bitrates = [];
-            for (var bitrate in hlsLabels) {
-                var key = parseFloat(bitrate);
-                if (!isNaN(key)) {
-                    var rateKBps = Math.round(key);
-                    labels[rateKBps] = hlsLabels[bitrate];
-                    bitrates.push(rateKBps);
-                }
-            }
-            if (bitrates.length === 0) {
-                return null;
-            }
-            bitrates.sort(function(a, b) {
-                return a - b;
-            });
-            return {
-                labels: labels,
-                bitrates: bitrates
-            };
-        }
-
         function checkFlashBlocked() {
             _flashBlockedTimeout = setTimeout(function() {
                 Events.trigger.call(_this, 'flashBlocked');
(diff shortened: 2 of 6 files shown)


test/unit/quality-labels.js
```diff
@@ -0,0 +1,99 @@
...
+       assert.equal(actual, expected);
+    });
+
+    test('retuns null when no custom labels are given', function (assert) {
+        var expected = null;
+        var actual = qualityLabels.getCustomLabel(null, 4000000);
+        assert.equal(actual, expected);
+    });
+
+    test('retuns null when no custom labels are empty', function (assert) {
```
egreaves: remove "no" from the test description

test/unit/quality-labels.js
```diff
@@ -0,0 +1,99 @@
...
+        var expected = null;
+        var actual = qualityLabels.getCustomLabel([], 4000000);
+        assert.equal(actual, expected);
+    });
+
+    test('retuns null when bandwidth is undefined', function (assert) {
+        var expected = null;
+        var actual = qualityLabels.getCustomLabel([], undefined);
```
egreaves: Though this passes, it's misleading because your bandwidths array is also empty, which could indicate a false positive. Consider expanding this to test `bandwidth === undefined` with both an empty array and an array containing at least 1 bandwidth. Alternatively, you could just test it with a non-empty bandwidths array since the test above covers the case where `bandwidths.length === 0`.

src/js/api/config.js
```diff
@@ -126,6 +126,8 @@ define([
             config.playlist = [ obj ];
         }
 
+        config.qualityLabels = config.qualityLabels || config.hlslabels;
```
hghazzi: We can verify with @esboyd, but I believe all config names you be entirely lowercase `qualitylabels`.

johnBartos: `hlslabels` isn't camelcase

test/unit/quality-labels-test.js
```diff
@@ -0,0 +1,99 @@
...
+    var test = QUnit.test.bind(QUnit);
+
+    var customLabels  = {
+        1000: 'low',
+        2000: 'medium',
+        3000: 'high',
+    };
+
+    var bandwidths = _.keys(customLabels);
+
```
hghazzi: Can you add `Qunit.module` before each method thats being tested? Its a useful separator, just like how we would use `describe` in mocha.

test/unit/quality-labels-test.js
```diff
@@ -0,0 +1,99 @@
...
+        var actual = qualityLabels.createLabel(360, undefined, false);
+        assert.equal(actual, expected);
+    });
+
+    test('creates a label using bandwidth when height is not available', function (assert) {
+        var expected = '1337 kbps';
+        var actual = qualityLabels.createLabel(undefined, 1337000,  false);
+        assert.equal(actual, expected);
+    })
+
```
hghazzi: Add a test with args `undefined, 1337000, true` to verify that parentheses are not added.

test/unit/quality-labels-test.js
```diff
@@ -0,0 +1,99 @@
...
+        var expected = null;
+        var actual = qualityLabels.getCustomLabel(customLabels, undefined);
+        assert.equal(actual, expected);
+    });
+
+    test('retuns null when no bandwidths are given', function (assert) {
+        var expected = null;
+        var actual = qualityLabels.findClosestBandwidth([], 4000);
+        assert.equal(actual, expected);
+    });
```
hghazzi: Same

test/unit/quality-labels-test.js
```diff
@@ -0,0 +1,99 @@
...
+        var expected = null;
+        var actual = qualityLabels.getCustomLabel(null, 4000000);
+        assert.equal(actual, expected);
+    });
+
+    test('retuns null when custom labels are empty', function (assert) {
+        var expected = null;
+        var actual = qualityLabels.getCustomLabel([], 4000000);
+        assert.equal(actual, expected);
+    });
```
hghazzi: Edge case, but add a test where you pass in a non array value, since that would pass a guard that is only looking for a falsey value.

src/js/providers/flash.js
```diff
@@ -125,11 +74,8 @@ define([
...
                 // Set original index
                 level.index = i;
-                // Translate sources into quality levels, assigning custom levels if present
-                if (_customLabels && level.bitrate) {
-                    // get label with nearest rate match
-                    var sourceKbps = Math.round(level.bitrate / 1000);
-                    level.label = _getNearestCustomLabel(sourceKbps);
+                if (level.label !== 'Auto') {
+                    level.label = utils.generateLabel(level, _playerConfig.qualityLabels);
                 }
```
hghazzi: Do we need to handle the case where two labels could end up having the same label if they are both closest to the same bandwidth?

src/js/api/config.js
johnBartos: wontfix, see `minDvrWindow` and `timeSliderAbove`

test/unit/quality-labels-test.js
johnBartos: qualityLabels should actually be {} in this case

src/js/providers/flash.js
johnBartos: They will both have the same label

test/unit/quality-labels-test.js
johnBartos: done

test/unit/quality-labels-test.js
johnBartos: done

PR #1781, opened 2017-01-31
Let Edge handle cleanup for non-sideloaded tracks in HLS streams

egreaves:
We don't need to and shouldn't attempt to remove cues from tracks added by Edge during native HLS playback.

Fixes #
JW7-3926


egreaves:
test this please

PR #1785, opened 2017-02-01
Fit jw-icon-rewind:before to the size of the icon

jw-kaurand:
Fit jw-icon-rewind:before to the size of the icon for correct safari highlights and display

Sets the 10 second rewind button's before pseudo element to fill its associated element.  This makes it so Safari will allow the entire icon can highlight correctly because Safari will only set the color on the part of the glyph that exists in that container.  This also fixes the issue where part of the icon is cut off in the audio only mode because it does not have a large enough line-height set.

JW7-3912 JW7-3664

### DIFF
--- a/src/css/imports/mixins.less
+++ b/src/css/imports/mixins.less
@@ -306,81 +306,77 @@
     }
 
     .custom-skin-styles() {
-        .jw-display-icon-container {
-            border-radius: @ui-corner-round;
+      .jw-display-icon-container {
+        border-radius: @ui-corner-round;
+      }
+
+      .jw-dock-button {
+        border-radius: @ui-corner-round;
+      }
+
+      .jw-controlbar .jw-controlbar-center-group .jw-text-alt {
+        top: 50%;
+        transform: translateY(-50%);
+        margin: auto;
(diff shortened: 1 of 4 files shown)


jw-kaurand:
test this please

jw-kaurand:
test this please

egreaves:
@jw-kaurand This works for desktop Safari, but does not work for Safari in iOS.

jw-kaurand:
It looks like it works in some situations on iOS but not in others.  I'll have to take a look at what's going on in some of the failure cases.

src/css/imports/mixins.less
```diff
@@ -317,78 +314,75 @@
...
-              line-height: normal;
-          }
-        }
+                .jw-icon-inline,
+                .jw-icon-tooltip,
+                .jw-text-elapsed,
+                .jw-text-duration,
+                .jw-text-countdown {
+                    height: auto;
+                    line-height: normal;
```
egreaves: @jw-kaurand this causes the rewind button to be cropped in iOS Safari. It needs to be an explicit line-height.

PR #1786, opened 2017-02-01
Move Seven skin IE 11 style to the seven.less file

jw-kaurand:
Moves a seven skin IE 11 style to the seven.less file instead of existing in the ads.less file.  This prevents alignment in all other skins in IE 11.

JW7-3969

### DIFF
--- a/src/css/skins/seven.less
+++ b/src/css/skins/seven.less
@@ -48,4 +48,8 @@
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
   }
+
+  &.jw-ie .jw-text-alt {
+    margin: 0.5em 0;
+  }
 }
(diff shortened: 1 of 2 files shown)


jw-kaurand:
test this please

src/css/skins/seven.less
```diff
@@ -48,4 +48,8 @@
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
   }
+
+  &.jw-ie .jw-text-alt {
+    margin: 0.5em 0;
```
egreaves: The seven skin gets this from the control bar: https://github.com/jwplayer/jwplayer/blob/master/src/css/imports/controlbar.less#L40.

The goal is to implement a comprehensive solution for centering text in IE11 for all the other skins. Removing the margin from `ads.less` still doesn't get us there.

pajong:
@jw-kaurand Merged by accident. I reverted the change.

PR #1792, opened 2017-02-03
Add conditionals to color overrides

jw-kaurand:
Conditionally applies colorization values if they are set.  This makes it so that "background: none" will not be applied by the shorthand property if we do not have an active, inactive or background color defined.  Previously, no value would exist when setting background color and would then be ignored, but with the default addition of "none" we need to make sure these are only applied conditionally.
Removes the if(!value) check as it is no longer used.

JW7-3908

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -367,12 +367,7 @@ define([
             // look good.
             var activeColor = _model.get('skinColorActive'),
                 inactiveColor = _model.get('skinColorInactive'),
-                backgroundColor = _model.get('skinColorBackground'),
-                backgroundColorGradient = backgroundColor ? 'transparent linear-gradient(180deg, ' +
-                    getRgba(backgroundColor, 0) + ' 0%, ' +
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -381,58 +377,64 @@ define([
...
+                // Apply background color
+                addStyle([
+                    // general background color
+                    '.jw-background-color'
+                ], 'background', 'none ' + backgroundColor);
+
+                addStyle([
+                    // for small player, set the control bar gradient to the config background color
+                    '.jw-flag-time-slider-above .jw-background-color.jw-controlbar'
+                ], 'background', 'none ' + backgroundColorGradient, true);
```
egreaves: We only need to add this when `_model.get('timeSliderAbove') !== false`

jw-kaurand: If `_model.get('timeSliderAbove') === false` then we won't ever apply the `jw-flag-time-slider-above` right?  I think we can just let it be a dead style instead of adding to the size with the `if` statement if so.

PR #1793, opened 2017-02-03
Expose toKbps method in utils

hghazzi:
to be shared between hls and shaka providers

JW7-3411


src/js/utils/quality-labels.js
```diff
@@ -74,5 +74,6 @@ define([
         createLabel: createLabel,
         getCustomLabel: getCustomLabel,
         findClosestBandwidth: findClosestBandwidth,
+        toKbps: toKbps
```
johnBartos: Needs a comma here

hghazzi:
updated

PR #1794, opened 2017-02-03
Use less fallbacks when creating track._id

egreaves:

Reduce the number of properties used to create `track._id`. This makes it predictable to determine tracks added by the browser during HLS playback in Edge.

Fixes #
JW7-3926


egreaves:
test this please

PR #1795, opened 2017-02-06
Feature/remove attached

pajong:


pajong:
@robwalch @johnBartos Made a new PR against v7.10 instead of master

PR #1808, opened 2017-02-09
Convert ISO language codes to label for Dash audio tracks

hghazzi:
If the language attribute from the tracks in the manifest are ISO 639-1
or ISO 639-2 language codes, convert to appropriate label name

JW7-3951

https://github.com/jwplayer/jwplayer-commercial/pull/2951

### DIFF
--- a/src/js/utils/language.js
+++ b/src/js/utils/language.js
@@ -16,12 +16,8 @@ define([
         'ru': 'Russian',
         'es': 'Spanish'
     };
-
-    /**
-     * A map of 3-letter language codes (ISO 639-2) to 2-letter language codes (ISO 639-1)
-     * Commented out until needed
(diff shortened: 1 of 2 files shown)


src/js/utils/language.js
```diff
@@ -0,0 +1,51 @@
...
+        'ja': 'Japanese',
+        'pt': 'Portuguese',
+        'ru': 'Russian',
+        'es': 'Spanish'
+    };
+
+    /**
+     * A map of 3-letter language codes (ISO 639-2) to 2-letter language codes (ISO 639-1)
+     */
+    var threeCharMap = {
```
robwalch: I'd like to hold back on this map until we find the need to support ISO 639-2 codes.

PR #1809, opened 2017-02-09
Prevent time slider tooltip from going outside the bounds of the player

egreaves:

Fixes #
JW7-3875

### DIFF
--- a/src/js/view/components/timeslider.js
+++ b/src/js/view/components/timeslider.js
@@ -15,6 +15,7 @@ define([
             this.text.className = 'jw-text jw-reset';
             this.img  = document.createElement('div');
             this.img.className = 'jw-reset';
+            this.resetWidth();
 
             var wrapper = document.createElement('div');
             wrapper.className = 'jw-time-tip jw-background-color jw-reset';
@@ -32,6 +33,16 @@ define([
 
         update : function(txt) {
             this.text.innerHTML = txt;
+        },
+        width : function () {
+            if (!this.containerWidth && this.container) {
+                this.containerWidth = utils.bounds(this.container).width;
+            }
+
+            return this.containerWidth;
+        },
+        resetWidth : function () {
+            this.containerWidth = 0;
         }
     });
 
@@ -170,6 +181,7 @@ define([
                 return;
             }
 
+            var playerWidth = this._model.get('containerWidth');
             var _railBounds = utils.bounds(this.elementRail);
             var position = (evt.pageX ? (evt.pageX - _railBounds.left) : evt.x);
             position = utils.between(position, 0, _railBounds.width);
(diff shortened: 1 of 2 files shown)


src/js/view/components/timeslider.js
```diff
@@ -197,7 +209,18 @@ define([
...
+            var widthPct = _railBounds.width / 100;
+            var tolerance = playerWidth - _railBounds.width;
+            var timeTipPct = 0;
+            if (timeTipWidth > tolerance) {
+                // timeTip may go outside the bounds of the player. determine the % of tolerance needed
+                timeTipPct = (timeTipWidth - tolerance) / (2 * 100 * widthPct);
+            }
+            var safePct = Math.min(1 - timeTipPct, Math.max(timeTipPct, pct));
+
+            this.timeTip.el.style.left = (safePct * 100) + '%';
```
robwalch: This should be applied with `utils.css.style` to take advantage of it's check that prevents redundant writes.

Also `toFixed(1)` the percent to limit how frequently the value can change. x1000 precision should be good no?

src/js/view/components/timeslider.js
```diff
@@ -193,11 +205,23 @@ define([
...
+            var timeTip = this.timeTip;
+
+            timeTip.update(timetipText);
             this.showThumbnail(time);
 
-            utils.addClass(this.timeTip.el, 'jw-open');
-            this.timeTip.el.style.left = (pct*100) + '%';
+            utils.addClass(timeTip.el, 'jw-open');
+
+            var timeTipWidth = timeTip.width();
```
jw-kaurand: It looks like this isn't working correctly for the individual thumbnails test on http://player-pr-test-jenkins.longtailvideo.com/builds/5277/archive/test/squash/bootstrap.html?primary=html5&edition=ads&feature=cuepoints_thumbnails&timeSliderAbove=true

Does this need to be using the jw-overlay that the tooltip contains to determine its size?  On that page, the tooltip width is much smaller than expected while its child overlay element is at the correct size.

jw-kaurand:
I think we also need to update the width of the tooltip in showTimeTooltip around line 210 so that we can make sure we handle the case where a user transitions from a time to a cuepoint text in the controlbar.  Use the following setup to see what I mean:

{
    "primary": "flash",
"timeSliderAbove": true,
    "file": "//content.bitsontherun.com/videos/q1fx20VZ-52qL9xLP.mp4",
    "image": "//content.bitsontherun.com/thumbs/3XnJSIm4-480.jpg",
    "analytics": {
        "debug": true
    },
    "advertising": {
        "client": "../bin-debug/googima.js",
        "schedule": {
            "preroll": {
                "type": "nonlinear",
                "offset": "pre",
                "tag": "//playertest.longtailvideo.com/adtags/ima-sample-banner-ad-15s.xml"
            },
            "midroll": {
                "type": "nonlinear",
                "offset": "10",
                "tag": "//playertest.longtailvideo.com/adtags/nonlinear.xml"
            }
        }
    }
}

screen shot 2017-02-10 at 1 20 14 pm

PR #1811, opened 2017-02-09
Create util to check if levels are redundant 

johnBartos:
Break function previously used just in just hlsjs into a util so shaka can use it too. Maybe this should go into a new `level-utils` file?

JW7-2476

(1/2) https://github.com/jwplayer/jwplayer-commercial/pull/2954

### DIFF
--- a/src/js/utils/quality-labels.js
+++ b/src/js/utils/quality-labels.js
@@ -69,11 +69,25 @@ define([
         return Math.floor(bandwidth / 1000);
     }
 
+    // Use an empty object as the context and populate it like a hash map
+    function hasRedundantLevels(levels) {
+        if (!_.isArray(levels)) {
+            return false;
+        }
+        return _.some(levels, function (level) {
+            var key = level.height || level.bandwidth;
+            var foundDuplicate = this[key];
+            this[key] = 1;
+            return foundDuplicate;
+        }, {});
+    }
+
     return {
         generateLabel: generateLabel,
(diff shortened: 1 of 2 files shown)


src/js/utils/quality-labels.js
```diff
@@ -69,11 +69,25 @@ define([
...
+        if (!_.isArray(levels)) {
+            return false;
+        }
+        return _.some(levels, function (level) {
+            var key = level.height || level.bandwidth;
+            var foundDuplicate = this[key];
+            this[key] = 1;
+            return foundDuplicate;
+        }, {});
+    }
```
hghazzi: A question for @esboyd, should this handle the case where two quality levels match the same label when using  `qualityLabels`?

hghazzi: Also this looks like it could be used for audio tracks as well.

https://github.com/jwplayer/jwplayer-commercial/blob/v7.10/src/js/providers/shaka.js#L418

src/js/utils/quality-labels.js
```diff
@@ -69,11 +69,25 @@ define([
         return Math.floor(bandwidth / 1000);
     }
 
+    // Use an empty object as the context and populate it like a hash map
```
robwalch: This comment applies to `return _.some...`, not to `hasRedundantLevels`. Please put the comment over the scope whose context you are referring to.

PR #1815, opened 2017-02-10
Consolidate states files into states.less

egreaves:

- Consolidate all states files into a single file
- Group similar behavior across states

Fixes #
JW7-3971


src/css/states/states.less
```diff
@@ -0,0 +1,189 @@
+@import "../imports/vars";
+@import "../imports/icons";
+
+// idle
+
+.jw-state-idle {
+  .jw-icon-display {
+    .jw-icon-play;
```
jw-kaurand: It could be useful to consolidate the location where we set the icon to use.  E.G. if we did this like `.jw-state-idle, .jw-state-paused { .jw-icon-display { .jw-icon-play; }}` it'd be easier to track down and manage these.

PR #1823, opened 2017-02-13
Add thumbstrip functionality to touch and PointerEvent devices

jw-kaurand:
Adds thumbstrip functionality to touch and PointerEvent devices by activating listeners for touch and mouse events instead of just mouse events.  In order to support touch platforms which will not activate the mouseover handler that will set the this.activeCue property, we have to use javascript to check what the closest cue is in the hover/drag handler.  We will set a cue that is within 5 pixels away to be the active cue so that we can show the text for the cue if necessary.

JW7-3167

### DIFF
--- a/src/js/view/components/timeslider.js
+++ b/src/js/view/components/timeslider.js
@@ -2,11 +2,12 @@ define([
     'utils/underscore',
     'utils/helpers',
     'utils/constants',
+    'utils/ui',
     'view/components/slider',
     'view/components/tooltip',
     'view/components/chapters.mixin',
     'view/components/thumbnails.mixin'
-], function(_, utils, Constants, Slider, Tooltip, ChaptersMixin, ThumbnailsMixin) {
+], function(_, utils, Constants, UI, Slider, Tooltip, ChaptersMixin, ThumbnailsMixin) {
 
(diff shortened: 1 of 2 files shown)


src/js/view/components/timeslider.js
```diff
@@ -207,6 +211,18 @@ define([
             }
 
             var timetipText;
+
+            // With touch events, we never will get the hover events on the cues that cause cues to be active.
+            // Therefore use the info we about the scroll position to detect if there is a nearby cue to be active.
+            if (evt.sourceEventType === 'touch') {
```
egreaves: Would using the UI util to add the appropriate event listeners in the chapters mixin work in lieu of doing this here?
https://github.com/jwplayer/jwplayer/blob/master/src/js/view/components/chapters.mixin.js#L62

egreaves:
@jw-kaurand make this PR against the `v7.10` branch

src/js/view/components/timeslider.js
jw-kaurand: I don't believe we can do something like that.  The problem is that a touch event always targets the original element that it first targeted.  There isn't really a listener that can be put on those that will recieve the touchmove events that result from the drag.

PR #1825, opened 2017-02-14
Add thumbstrip functionality to touch and PointerEvent devices

jw-kaurand:
Adds thumbstrip functionality to touch and PointerEvent devices by activating listeners for touch and mouse events instead of just mouse events. In order to support touch platforms which will not activate the mouseover handler that will set the this.activeCue property, we have to use javascript to check what the closest cue is in the hover/drag handler. We will set a cue that is within 5 pixels away to be the active cue so that we can show the text for the cue if necessary.

Since the last time this was PR'd, I moved the instanceof checks for determining the original pointerType to a util so that we don't run the instanceof checks whenever we normalize an event, and so we can give out the source event for those who might be interested. We check the pointerType from the evt.sourceEvent instead.

JW7-3167

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -33,7 +33,9 @@ define([
     }
 
     function normalizeUIEvent(type, srcEvent, target) {
-        var source;
+        var source,
+            sourceType;
+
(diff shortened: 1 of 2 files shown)


jw-kaurand:
test this please

PR #1832, opened 2017-02-16
Don’t setup/show nextUp tooltip if nextUpDisplay = false

egreaves:

Do not setup the next button when `nextUpDisplay: false`. The nextUp tooltip should not show on hover or after crossing the nextUpOffset threshold.

Fixes #
JW7-3600

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -695,8 +695,10 @@ define([
             _nextuptooltip = new NextUpToolTip(_model, _api, _controlbar.elements.next, _playerElement);
             _nextuptooltip.setup();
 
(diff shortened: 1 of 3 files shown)


src/js/view/view.js
```diff
@@ -695,8 +695,10 @@ define([
             _nextuptooltip = new NextUpToolTip(_model, _api, _controlbar.elements.next, _playerElement);
```
robwalch: Why even instantiate when `! nextUpDisplay`?

robwalch:
Requires refactor to separate next button in controlbar from nextup tooltip

PR #1835, opened 2017-02-18
Don't show nextUp when nextUpDisplay: false

egreaves:

- Setup the next button in the control bar
- Don't instantiate `nextuptooltip` when `nextUpDisplay: false`
- Use the following model properties to control nextUp:
    - `nextUpEnabled` (bool) - Enables nextUp behavior when there is a next item containing a title or thumbnail
    - `nextUpVisible` (bool) - show/hide nextUp tooltip based on the next button's `hover` state
    - `nextUpSticky` (bool) - Initialized to `null` for each playlistItem and can be set to `true`. When `true`, the nextUp tooltip includes the close button and is visible until playback of the current item ends or if the user closes the tooltip. When set to `false`, nextUp only shows on hover and `nextUpSticky` cannot be set to `true`. 

Fixes #
JW7-3600

### DIFF
--- a/src/js/view/components/nextuptooltip.js
+++ b/src/js/view/components/nextuptooltip.js
@@ -5,10 +5,9 @@ define([
     'utils/helpers',
     'templates/nextup.html'
 ], function(dom, UI, _, utils, nextUpTemplate) {
-    var NextUpTooltip = function(_model, _api, nextButton, playerElement) {
+    var NextUpTooltip = function(_model, _api, playerElement) {
         this._model = _model;
         this._api = _api;
-        this._nextButton = nextButton;
         this._playerElement = playerElement;
         this.nextUpText = _model.get('localization').nextUp;
         this.nextUpClose = _model.get('localization').nextUpClose;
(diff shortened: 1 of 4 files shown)


src/js/view/controlbar.js
```diff
@@ -422,6 +438,9 @@ define([
                 this.elements.duration.innerHTML = 'Live';
                 this.elements.durationLeft.innerHTML = 'Live';
             }
+        },
+        onNextUp: function(model, nextUp) {
+            this.elements.next.toggle(!!nextUp);
```
johnBartos: Could `elements.next` or `elements.next.toggle` be `undefined` here?

egreaves: @johnBartos next is always defined when the controlbar is setup

src/js/view/components/nextuptooltip.js
```diff
@@ -26,35 +25,34 @@ define([
...
-            this.onMediaModel(this._model, this._model.get('mediaModel'));
+            this.onMediaModel(model, model.get('mediaModel'));
 
             // Close button
             new UI(this.closeButton, {'directSelect': true})
-                .on('click tap', this.hide, this);
+                .on('click tap', function() {
+                    model.set('nextUpSticky', false);
+                    model.set('nextUpVisible', false);
+                    this.toggle(model, false);
```
johnBartos: If the change handler for `nextUpSticky` & `nextUpVisible` is `toggle`, do you need to call toggle after setting either of these to false?

egreaves: Good catch - that's not necessary.

egreaves:
test this please

PR #1849, opened 2017-02-23
Account for activeTab with viewability

johnBartos:
JW7-3538

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -188,7 +188,7 @@ define([
                 });
             });
 
-            _model.on('change:viewable', _onVisibilityChange);
+            _model.on('change:visibility', _onVisibilityChange);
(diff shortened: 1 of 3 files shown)


src/js/controller/controller.js
```diff
@@ -188,14 +188,8 @@ define([
...
 
-            _model.on('change:visibility', function(model, visibility) {
-                var viewable = visibility >= 0.5;
-                if (viewable && _model.get('playOnViewable')) {
-                    _autoStart();
-                } else if (!viewable && utils.isMobile()) {
-                    _this.pause({ reason: 'autostart' });
-                }
-            });
+            _model.on('change:viewable', _onVisibilityChange);
```
egreaves: I don't see `viewable` being set anywhere. should this be `visibility`?

johnBartos: good catch

PR #1858, opened 2017-02-27
Implement end of stream detection in the HTML5 provider 

johnBartos:
JW7-3269

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -9,8 +9,8 @@ define([
     'providers/default',
     'utils/backbone.events',
     'providers/tracks-mixin',
-    'utils/browser'
-], function(getIsAndroidHLS, cssUtils, utils, dom, _, events, states, DefaultProvider, Events, Tracks) {
+    'utils/time-ranges',
+], function(getIsAndroidHLS, cssUtils, utils, dom, _, events, states, DefaultProvider, Events, Tracks, timeRangesUtil) {
 
     var clearTimeout = window.clearTimeout,
         STALL_DELAY = 256,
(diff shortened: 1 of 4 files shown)


src/js/providers/html5.js
```diff
@@ -998,6 +1015,48 @@ define([
...
+
+            if (live && _lastEndOfBuffer === endOfBuffer) {
+                if (!_staleStreamTimeout) {
+                    _staleStreamTimeout = setTimeout(function () {
+                        _stale = true;
+                        checkStreamEnded();
+                    }, _staleStreamDuration);
+                }
+            } else {
+                clearTimeout(_staleStreamTimeout);
```
robwalch: This should probably be called anywhere we call `clearTimeout(_playbackTimeout);`

src/js/providers/html5.js
```diff
@@ -998,6 +1015,48 @@ define([
...
+                });
+                return true;
+            }
+
+            return false;
+        }
+
+        function atEndOfStream() {
+            if (_videotag.duration !== Infinity) {
+                return;
```
robwalch: return false;

src/js/providers/html5.js
```diff
@@ -998,6 +1015,48 @@ define([
...
+                _this.trigger(events.JWPLAYER_MEDIA_ERROR, {
+                    message: 'The live stream is either down or has ended'
+                });
+                return true;
+            }
+
+            return false;
+        }
+
+        function atEndOfStream() {
```
robwalch: How about `atEdgeOfLiveStream`?

src/js/providers/html5.js
johnBartos: done

PR #1859, opened 2017-02-28
Improve setup and ready for viewability

robwalch:
- Load the IntersectionObserver polyfill on setup
- Add the view element to the DOM during setup
- View starts with "jw-state-setup" class replaced after ready > requestAnimationFrame
- Fire ready on view resize instead of calling resize on player ready
- Add 'defer' setup step to ensure ready is fired asynchronously after setup is called (removing `defer(_playerReadyNotify)` from controller)
- Remove listeners from view on view.destroy
- Captions container height is 100% or max-height - Don't set height with JS
- New "spec" reporter for karma test runner prints tests run [:man_technologist:](https://jenkins.jwplayer.com/job/jw7-opensource-unit-tests/1504/console)

This prepares the player for doing more view setup in the setup steps so that view-ability can be determined on ready.

Ready is not fired if player container is not in the DOM.


karma.conf.js
```diff
@@ -4,7 +4,9 @@ module.exports = function( config ) {
     var isJenkins = !!process.env.JENKINS_HOME;
     var serverPort = process.env.KARMA_PORT || 9876;
     var testReporters = [
-        'progress',
+        'spec',
+        // 'progress',
+        // 'dots',
```
hghazzi: let's remove the comments

src/js/controller/controller.js
```diff
@@ -202,12 +202,12 @@ define([
                 _setup = null;
 
                 _view.on('all', _triggerAfterReady, _this);
+                _view.once(events.JWPLAYER_RESIZE, _playerReadyNotify);
```
johnBartos: Can you leave a comment as to why we wait for a resize event?

test/config.js
```diff
@@ -30,16 +30,7 @@
...
-    if (window.wallaby) {
-        window.wallaby.delayStart();
-        base = '../../';
-        for (var file in window.wallaby.tests) {
-            if (Object.prototype.hasOwnProperty.call(window.wallaby.tests, file)) {
-                deps.push(window.wallaby.tests[file]);
-            }
-        }
-        callback = window.wallaby.start;
-    } else if (window.__karma__) {
```
robwalch: https://wallabyjs.com/

karma.conf.js
robwalch: I will. progress is pretty useless unless you care about the number of tests before the runner is done which we should not. "dots" is actually really useful when debugging because it outputs console logs to stdio.

src/js/view/view.js
```diff
@@ -1064,11 +1054,6 @@ define([
             _stateHandler(model, model.get('state'));
         }
 
-        function _resizeAspectModeCaptions() {
```
egreaves: When not rendering natively, we need this in order to position captions correctly for aspect mode players. We can get the container height off the model instead, though.

src/js/view/view.js
```diff
@@ -840,10 +834,6 @@ define([
                 _resizeMediaTimeout = setTimeout(_resizeMedia, 250);
             }
 
-            if (_model.get('aspectratio')) {
```
egreaves: We should also check  `!_model.get('renderCaptionsNatively')` so we don't resize the captions container unnecessarily.

hghazzi:
@robwalch Do we still need `wallaby.js`?

robwalch:
@hghazzi no. We don't use wallaby.js

hghazzi:
OK, let's remove it from the root directory.

robwalch:
Done

PR #1860, opened 2017-02-28
Dock icons switching positions

hghazzi:
- Hide share dock button using css
- Only add dock button if it doesn’t exist in array

JW7-3863


hghazzi:
test this please

PR #1861, opened 2017-02-28
ESLint underscore.js

hghazzi:
I just happened to come across a duplicate decelration of `_.any` in `underscore.js`, so decided to lint the file. I only temporarily removed it from the ignore list to lint, because I did not want to change any of logic (such as using `===` instead of `==`), and only made the stylistic changes (mainly adding curly braces around `if` statements).

### DIFF
--- a/src/js/utils/underscore.js
+++ b/src/js/utils/underscore.js
@@ -5,6 +5,12 @@
 
 // https://github.com/jashkenas/underscore/blob/1f4bf626f23a99f7a676f5076dc1b1475554c8f7/underscore.js
 
+/* eslint no-eq-null: 0 */
+/* eslint eqeqeq: 0 */
+/* eslint no-void: 0 */
+/* eslint guard-for-in: 0 */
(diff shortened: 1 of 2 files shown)


src/js/utils/underscore.js
```diff
@@ -358,19 +417,29 @@ define([], function() {
...
+        if (nativeBind && func.bind === nativeBind) {
+            return nativeBind.apply(func, slice.call(arguments, 1));
+        }
+        if (!_.isFunction(func)) {
+            throw new TypeError;
+        }
         args = slice.call(arguments, 2);
-        return bound = function() {
-            if (!(this instanceof bound)) return func.apply(context, args.concat(slice.call(arguments)));
+        bound = function() {
```
johnBartos: I think `return bound = function()` is fine. Does the linter complain?

hghazzi: Yeah, [no-return-assign](http://eslint.org/docs/rules/no-return-assign)

PR #1862, opened 2017-02-28
Bugfix/seeking drag bounces back

karimJWP:
prevents firing an outdated position and causing the timeslider to jump back and forth

Fixes #
JW7-3479


karimJWP:
@reviewer: After dragging ends, the first time position is outdated.  My solution prevents the first time update from firing. Any recomendations on a cleaner approach?

PR #1863, opened 2017-03-01
Add support for html5 audio tag using useAudioTag config property

jimjenkins5:

On iOS (especially iPhone) media played in a video tag is always played fullscreen. This makes JW Player useless if you want to play audio while keeping the main page on the screen (i.e. for read along audio).  Also, using a video tag to play audio only content is questionable practice considering the intent of the video tag (evident by the name).

This PR adds a simple configuration property to which uses an audio tag instead of a video tag for the player.

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -130,14 +130,15 @@ define([
 
         // Find video tag, or create it if it doesn't exist.  View may not be built yet.
         var element = document.getElementById(_playerId);
-        var _videotag = (element) ? element.querySelector('video') : undefined;
+        var tagType = _playerConfig.useAudioTag ? 'audio' : 'video';
+        var _videotag = (element) ? element.querySelector(tagType) : undefined;
 
         function _setAttribute(name, value) {
             _videotag.setAttribute(name, value || '');
         }
 
         if (!_videotag) {
-            _videotag = document.createElement('video');
+            _videotag = document.createElement(tagType);
 
             if (_isMobile) {
                 _setAttribute('jw-gesture-required');


jthomerson:
+1 This is a very important feature to support, as indicated by many StackOverflow questions.

robwalch:
Hi @jimjenkins5,

Thanks for the contribution. We support the use a video tag if present in the target element used to setup the player:

```html
<div id="jwp"><video src=""></video></div>
<script>
jwplayer('jwp').setup({});
</script>
```

What do you think about us setting "useAudioTag" when the setup received an audio tag?

Ultimately we want to improve setup to do progressive enhancement of media tags. This is a step in that direction. Depending on the media you provide to the player however, the current provider will need to make a choice about what type of element to use.

robwalch:
I think what you have here is good. Let me see what other feedback the team has.

I assume `mediaType` event will always report 'audio' with an audio tag if `tag.videoHeight === 0`.
  
There might also be a use case where you want to set `"useAudioTag"` on the item level for mixed playlists. I don't think the value vs level of effort would be worth it for us to support.

As far as the progressive enhancement goes. I don't think you need to do anything. It would be more of a follow up on our end.

src/js/providers/html5.js
```diff
@@ -130,14 +130,15 @@ define([
 
         // Find video tag, or create it if it doesn't exist.  View may not be built yet.
         var element = document.getElementById(_playerId);
-        var _videotag = (element) ? element.querySelector('video') : undefined;
+        var tagType = _playerConfig.useAudioTag ? 'audio' : 'video';
+        var _videotag = (element) ? element.querySelector(tagType) : undefined;
```
robwalch: If we are going to query for an existing tag, we should query for video or audio. So we only need to look at the config when creating a media element, and at that point create an video or audio element.

johnBartos:
@jimjenkins5 Thanks for the work, happy to carry this the last mile to get it into 7.11.

robwalch:
This can be done once #1997 is released by including an audio tag in the element used to setup the player:

```js
<div id="player"><audio/></div>
<script>jwplayer('player').setup({ file: 'audio.m4a' })</script>
```

PR #1868, opened 2017-03-02
Fix controlbar only idle state on ready

robwalch:
New setup process set's the player container's state flag to ".jw-state-setup" until the player is ready. For control-bar only the state was not getting updated to idle, so the controls remained not-displayed. This ensures the state flag always gets updated right before the ready event.

JW7-4064


egreaves:
test this please

PR #1869, opened 2017-03-02
Bugfix/optimize vttparser

johnBartos:
Doesn't fix that IOS 9 issue we had but it makes VTT parsing faster/less blocking.


dannyfinks:
@johnBartos we are not getting tracks for this sideloaded example - http://player-pr-test-jenkins.longtailvideo.com/builds/5723/archive/test/harness.html?edition=ads&config=captions_vtt

johnBartos:
test this please

src/js/parsers/captions/vttparser.js
```diff
@@ -409,50 +379,59 @@ define(['parsers/captions/vttcue'], function(VTTCue) {
...
                 }
-            } catch (e) {
 
-                // If we are currently parsing a cue, report what we have.
-                if (self.state === 'CUETEXT' && self.cue && self.oncue) {
-                    self.oncue(self.cue);
+                currentCueBatch = 0;
+                if (self.buffer) {
+                    try {
+                        requestFrame(processBuffer);
```
robwalch: Does this need to be in a try catch block? The only code executed is `requestAnimationFrame` or `setTimeout`.

Any error parsing will fire asynchronously and no get caught.

PR #1871, opened 2017-03-02
Set jw-text-alt height in mixins for IE11

jw-kaurand:
Setting height to auto for the absolutely positioned elements causes height to be 0 in IE11.  Setting a default line height and height in the default skins and specifically sets the values later on for individual skins.

JW7-4034

### DIFF
--- a/src/css/imports/controlbar.less
+++ b/src/css/imports/controlbar.less
@@ -35,7 +35,7 @@
     top: -1px;
     bottom: 0;
     width: 100%;
-    height: auto;
+    height: @controlbar-height;
     line-height: @controlbar-height;
     margin: 0.5em 0;
     padding-right: 0.5em;
--- a/src/css/imports/mixins.less
+++ b/src/css/imports/mixins.less
@@ -345,6 +345,7 @@
           }
         }
 
+        .jw-text-alt,
         .jw-icon-inline,
         .jw-icon-tooltip,
         .jw-text-elapsed,


src/css/imports/mixins.less
```diff
@@ -345,6 +345,7 @@
           }
         }
 
+        .jw-text-alt,
```
robwalch: This issue impacts ima?

src/css/imports/controlbar.less
```diff
@@ -35,7 +35,7 @@
     top: -1px;
     bottom: 0;
     width: 100%;
-    height: auto;
+    height: @controlbar-height;
```
robwalch: Why was it set to auto? Why not fix it with `.jw-ie {}`?

src/css/imports/mixins.less
jw-kaurand: I thought it had at first but it actually looks okay there.  Will remove it and shift it downwards into an `.jw-ie` block.

src/css/imports/controlbar.less
jw-kaurand: It was set to auto because it lays out fine in other browsers, but in IE some other parts like ads use `display: table;` which will make it lay out correctly in there.  It seems like a better solution would be to have an `.jw-ie` defined in control-bar that does this work across ads and live as they'll both end up using the same set of code.

PR #1872, opened 2017-03-02
Chromeless UI Bugs

hghazzi:
- Error description displays for small players, padding adjusted so it
does not overlap with error title. Item description still stays hidden.
Affects with and without controls.

- Found another bug: Error display icon container now centered in
player for small players with controls.

- Poster image no longer show in controlbar only mode.

- Next Up container now hidden when controls are disabled.

JW7-4059

### DIFF
--- a/src/css/flags/controls-disabled.less
+++ b/src/css/flags/controls-disabled.less
@@ -1,6 +1,7 @@
 .jw-flag-controls-disabled {
-    .jw-controls {
+    .jw-controls, .jw-nextup-container {
         visibility: hidden;
+        opacity: 0;
     }
 
     .jw-logo {
--- a/src/css/flags/controls-disabled.less
+++ b/src/css/flags/controls-disabled.less
@@ -1,7 +1,6 @@
 .jw-flag-controls-disabled {
     .jw-controls, .jw-nextup-container {
         visibility: hidden;
-        opacity: 0;
     }
 
     .jw-logo {
--- a/src/css/flags/time-slider-above.less
+++ b/src/css/flags/time-slider-above.less
@@ -7,12 +7,6 @@
     display: table;
   }
 
-  &.jw-state-idle:not(.jw-flag-cast-available) {
-    .jw-display {
-      padding: 0;
-    }
-  }
-
   /* ==================================================
   display
   */
@@ -25,6 +19,12 @@
         padding-top: @mobile-touch-target;
         padding-bottom: @mobile-touch-target * 1.5;
       }
+  }
+
+  &.jw-state-idle:not(.jw-flag-cast-available), &.jw-state-error {
+    .jw-display {
+      padding: 0;
+    }
   }
 
   /* ==================================================
--- a/src/css/imports/title.less
+++ b/src/css/imports/title.less
@@ -47,5 +47,6 @@
 
   .jw-title-secondary {
     display: none;
+    margin-top: 5px;
   }
 }
--- a/src/css/imports/title.less
+++ b/src/css/imports/title.less
@@ -31,7 +31,7 @@
     margin-top: -.5em;
 }
 
-.jw-flag-small-player:not(.jw-flag-audio-player) {
+.jw-flag-small-player {
 
   .jw-title {
     background: linear-gradient(180deg, fade(mix(black, white, 80%), 75%), fade(mix(black, white, 80%), 0%));
--- a/src/css/imports/title.less
+++ b/src/css/imports/title.less
@@ -31,7 +31,7 @@
     margin-top: -.5em;
 }
 
-.jw-flag-small-player {
+.jw-flag-small-player:not(.jw-flag-audio-player) {
 
   .jw-title {
     background: linear-gradient(180deg, fade(mix(black, white, 80%), 75%), fade(mix(black, white, 80%), 0%));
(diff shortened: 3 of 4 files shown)


src/css/flags/controls-disabled.less
```diff
@@ -1,6 +1,7 @@
 .jw-flag-controls-disabled {
-    .jw-controls {
+    .jw-controls, .jw-nextup-container {
         visibility: hidden;
+        opacity: 0;
```
hghazzi: added to override opacity for `jw-nextup-container`

src/css/flags/time-slider-above.less
```diff
@@ -27,6 +21,12 @@
       }
   }
 
+  &.jw-state-idle:not(.jw-flag-cast-available), &.jw-state-error {
+    .jw-display {
+      padding: 0;
+    }
+  }
+
```
hghazzi: Moved to increase specificity for the existing case and for error case. Ensures that the error display icon stays centered

src/css/imports/title.less
```diff
@@ -48,7 +47,5 @@
 
   .jw-title-secondary {
     display: none;
-    margin-top: 5px;
```
hghazzi: no longer needed since padding is adjusted in `states.less`

src/css/states/states.less
```diff
@@ -97,8 +97,14 @@
...
 body .jwplayer.jw-state-error {
-  .jw-title .jw-title-primary {
-    white-space: normal;
+  .jw-title {
+    .jw-title-primary {
+      white-space: normal;
+    }
+    .jw-title-secondary {
+      display: block;
+      padding-top: 0.75em;
```
hghazzi: displays description during error state. `padding-top: 0.75em;` resets padding for `.jw-title-secondary` that was `padding: 0.75em 0;`

src/css/states/states.less
```diff
@@ -163,10 +169,10 @@ body .jwplayer.jw-state-error {
...
-.jw-state-idle,
+.jw-state-idle:not(.jw-flag-audio-player),
 .jwplayer.jw-state-complete:not(.jw-flag-casting):not(.jw-flag-audio-player),
-body .jw-error,
-body .jwplayer.jw-state-error {
+body .jw-error:not(.jw-flag-audio-player),
+body .jwplayer.jw-state-error:not(.jw-flag-audio-player) {
   .jw-preview {
     display: block;
   }
```
hghazzi: Probably a better (and obvious) way of doing this, but couldn't figure it out. Only show preview for non controlbar only players

src/css/flags/controls-disabled.less
robwalch: opacity is irrelevant when visibility is hidden. Why override it?

src/css/imports/title.less
robwalch: Setting it here set's it once and prevents it from changing when the state changes which is a performance improvement.

robwalch: This also removes the padding above item descriptions in the idle state. you're now only setting it in the error state.

src/css/states/states.less
robwalch: Yeah, don't change these rules. Add a new more specific rule for jw-flag-audio-player that set's jw-preview display to none.

src/css/flags/time-slider-above.less
robwalch: 👍

src/css/flags/controls-disabled.less
robwalch: Make sure only visibility: hidden is set for the problem cases:

".jw-flag-controls-disabled" makes the controls container hidden
but this is overriden by visible changes in 
".jw-nextup-container-visible" and ".jw-nextup-sticky .jw-nextup-close"

src/css/imports/title.less
robwalch: Undo and apply margin in error state / audio-player

src/css/states/states.less
```diff
@@ -172,6 +179,34 @@ body .jwplayer.jw-state-error {
...
 }
 
+.jw-state-idle,
+.jwplayer.jw-state-complete:not(.jw-flag-casting),
+body .jwplayer.jw-state-error {
+  &.jw-flag-audio-player {
```
robwalch: There is only one state (error) that needs fixing. idle and complete are fine and should not be given new 2. 

> Poster shwon in controlbar-only mode
`body .jwplayer.jw-state-error .jw-preview { display: block; }`

src/css/states/states.less
```diff
@@ -172,6 +179,34 @@ body .jwplayer.jw-state-error {
...
+    }
+
+    .jw-title-secondary {
+      width: auto;
+      display: inline-block;
+      padding-left: 0;
+    }
+  }
+}
+
```
robwalch: state-error is the issue, not jw-error

src/css/imports/title.less
```diff
@@ -32,7 +31,7 @@
     margin-top: -.5em;
 }
 
-.jw-flag-small-player {
+.jw-flag-small-player:not(.jw-flag-audio-player) {
```
robwalch: Please remove the not. It makes this rule more specific which can introduce regressions in other areas

PR #1873, opened 2017-03-03
Improve captions font scaling

egreaves:

- Scale captions text as the browser would. 
- Adjust the scale if a fontSize is provided on setup.
- Hide browser captions in idle, complete and error states

Fixes #
JW7-3967, JW7-2723

### DIFF
--- a/src/js/view/captionsrenderer.js
+++ b/src/js/view/captionsrenderer.js
@@ -26,6 +26,7 @@ define([
         var _captionsWindow;
         var _textContainer;
         var _VTTRenderer;
+        var _captionsFontScale;
 
         _display = document.createElement('div');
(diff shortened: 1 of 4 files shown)


src/js/view/captionsrenderer.js
```diff
@@ -183,6 +185,22 @@ define([
...
+
+            if (fontSize) {
+                if (height) {
+                    // Set scale based on custom font size and player height
+                    scale = fontSize / height;
+                } else {
+                    _model.once('change:containerHeight', _setFontScale);
+                }
+            }
+            _model.set('captionsFontScale', scale);
```
robwalch: Why use the model to store a setting only used in one module?

PR #1876, opened 2017-03-03
Fix Related/Playlist Tooltip

hghazzi:
Needed to do an additional check on the button objects since they can
have the same id but different tooltip text

JW7-3863


johnBartos:
Should we just not allow buttons with the same id? That seems like the most sane solution.

hghazzi:
@johnBartos id here is 'share' and 'related', to identify the doc button.

PR #1888, opened 2017-03-07
Hide close

hghazzi:
JW7-4059


src/css/flags/controls-disabled.less
```diff
@@ -1,5 +1,6 @@
 .jw-flag-controls-disabled {
-    .jw-controls, .jw-nextup-container {
+    .jw-controls, .jw-nextup-container,
```
egreaves: We should just hide the nextup container with `display: none`. There's no need to use `visibility: hidden` because none of the child elements of the nextup container need to be visible when controls are disabled.

hghazzi:
test this please

PR #1895, opened 2017-03-10
Ensure hexadecimal pairs are created for values less than 16

jw-kaurand:
This fixes an issue where arrayToInt will return incorrect hexadecimal values if parsing the number will result in one character instead of two.  This can end up omitting significant ‘0’ values and generate incorrect INT values.

JW7-4121


jw-kaurand:
Will need to wait on West posting the stream again to test this fix.  However does appear to work with the archived data that we have.

jw-kaurand:
Tested this using the live version of the stream that was presenting the issue and it's working correctly.

PR #1899, opened 2017-03-14
QoE Enhancements

robwalch:
Use a running timer that updates when the `currentTime` getter is called and on using setInterval. The interval is set to run 20 times a second, adding the delta to currentTime on each run with a maximum delta of 1000 milliseconds. Any period greater than a second that passes while the device or browser is inactive will not be added to the running clock.

Also added a running tally of active state in `qoe.dump()` to make the `qoe()` API method useful for developers who want to get an active running tally of all states for the current playlist item:

```js
// (Setup a player and begin playback)

// Before these changes you would only get a total of previous states:
console.log(jwplayer().qoe().item.sums);
{
    idle: 145,
    loading: 406
}

 // Now the current state is tallied with each call
console.log(jwplayer().qoe().item.sums);
{
    idle: 145,
    loading: 406,
    playing: 6989
}
```

States that the player has not entered are still not logged. The complete state would look like:
```js
{
    idle: 150, // idle time before attempting playback (includes some setup time with autostart)
    loading: 586, // time spent buffering
    playing: 3134, // time spent playing
    paused: 568, // time spent paused
    complete: 3094 // idle time after playback (replay/recommendations state)
}
```

JW7-4056

### DIFF
--- a/src/js/utils/clock.js
+++ b/src/js/utils/clock.js
@@ -1,39 +1,41 @@
 define([], function() {
 
-    var AudioClockPolyfill = function() {
-        var currentTime = 0;
-        var time = new Date().getTime();
+    var performance = window.performance;
+
+    var supportsPerformance = !!(performance && performance.now);
+
+    var getTime = function() {
+        if (supportsPerformance) {
+            return performance.now();
+        }
+        return new Date().getTime();
+    };
+
+    var Clock = function() {
+        var started = getTime();
+        var updated = started;
 
         var updateClock = function() {
-            var delta = new Date().getTime() - time;
+            var delta = getTime() - updated;
             if (delta > 250) {
                 delta = 250;
             } else if (delta < 0) {
                 delta = 0;
             }
-            time += delta;
-            currentTime += delta;
+            updated += delta;
         };
         setInterval(updateClock, 50);
 
         Object.defineProperty(this, 'currentTime', {
             get: function() {
                 updateClock();
-                return currentTime / 1000;
+                return updated - started;
             }
         });
     };
 
-    var Clock = function() {
-        if ('AudioContext' in window) {
-            this.timer = new AudioContext();
-        } else {
-            this.timer = new AudioClockPolyfill();
-        }
-    };
-
     Clock.prototype.now = function() {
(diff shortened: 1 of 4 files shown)


src/js/utils/clock.js
```diff
@@ -0,0 +1,40 @@
+define([], function() {
+
+    var AudioClockPolyfill = function() {
+        var currentTime = 0;
+        var time = new Date().getTime();
+
+        var updateClock = function() {
+            var delta = new Date().getTime() - time;
+            if (delta > 250) {
```
egreaves: Is 250 just a tolerance value?

robwalch: Yes. Updated the description. AudioContext is only missing in older IE browsers. For those, anytime over a quarter of a second where JavaScript could not be executed will not be added to the running clock. The getting handles precision < 50ms by calculating the delta since start or the last call and adding it to currentTime.

egreaves: Thanks for clarifying.

robwalch:
Safari only supports prefixed webkitAudioContext and starts suspended - won't resume in iOS without a gesture. On Android Chrome, the clock runs even when the device is locked or the tab is in the background - it could be that was the case only because  I was debugging but we can't count on the audio clock to run or stop running across browsers according to application or device activity.

robwalch:
I'm interested in scenarios where observing active tab and/or the audio clock's delta on each update could inform whether or not the device/browser/media/network activity was suspended. Right now it doesn't look like that's very reliable as each browser handles WebAudio, MediaElement and JavaScript execution in different ways.

To use AudioContext on mobile we'll also need to do:

```js
var AudioContext = window.AudioContext || window.webkitAudioContext;
var audioCtx = new AudioContext();
// since state and resume may or may not be implemented try to resume and later make sure curretTime is advancing
audioCtx.resume && audioCtx.resume();
var currentTime = audioCtx.currentTime;
var started = currentTime > 0;
var delta = currentTime - lastTime;
if (started && delta > 0) {
  return delta; // we can use this
} else {
  // it's not updating - maybe JavaScript is running but the context is suspended
}
lastTime = currentTime;
```

PR #1900, opened 2017-03-14
Fix captions font scaling, line height and background

egreaves:

- Adjust captions text size correctly based on `fontSize` and the browser's default scaling coefficient.
- Update captions text background when at least one of the backgroundColor or backgroundOpacity values is set.
- Set line height to `normal` for browser-rendered captions to prevent the browser from rendering `VTTCues` with too much vertical spacing. There's a slight cue overlap in some instances, but I'd say this is a better tradeoff than having a full line gap between cues at full screen player sizes when there are multiple cues.

Fixes #
JW7-4133

### DIFF
--- a/src/css/imports/captions.less
+++ b/src/css/imports/captions.less
@@ -47,11 +47,11 @@
 .jw-text-track-display {
     // Override vtt.js font-size so text resizes as the player size changes
     font-size: inherit;
-    line-height: 1.5em;
+    line-height: 1.5;
 }
 
 .jw-text-track-cue {
-    background-color: rgba(0, 0, 0, 0.498039);
+    background-color: rgba(0, 0, 0, 0.5);
     color: #fff;
     padding: .1em .3em;
 }
@@ -65,7 +65,7 @@
         }
         &::-webkit-media-text-track-container {
              max-height: ~"calc(100% - 40px)";
-             line-height: 1.3em;
+             line-height: normal;
          }
         &::-webkit-media-text-track-display {
             // Ensure captions aren't cropped when the pseudo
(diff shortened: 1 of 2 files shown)


src/css/imports/captions.less
```diff
@@ -47,11 +47,11 @@
 .jw-text-track-display {
     // Override vtt.js font-size so text resizes as the player size changes
     font-size: inherit;
-    line-height: 1.5em;
+    line-height: 1.5;
```
johnBartos: Did you mean to remove `em` here?

src/js/view/captionsrenderer.js
```diff
@@ -10,9 +10,11 @@ define([
 
     var _defaults = {
         back: true,
-        backgroundOpacity: 100,
+        backgroundOpacity: 50,
         edgeStyle: null,
+        fontSize: 14,
```
johnBartos: 14 px?

src/css/imports/captions.less
egreaves: That was intentional. The browser handles line-height better with unitless numbers. ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/line-height#Prefer_unitless_numbers_for_line-height_values))

johnBartos: The more you know ✨

src/js/view/captionsrenderer.js
egreaves: `fontSize` is expected to be a number ([ref](https://support.jwplayer.com/customer/portal/articles/1482067-styling-captions-for-fcc-compliance)).

PR #1905, opened 2017-03-14
Controls module

johnBartos:
Getting this PR up now for an early review

JW7-4073

### DIFF
--- a/.babelrc
+++ b/.babelrc
@@ -1,0 +1,5 @@
+{
+    "presets": [
+        "es2015-loose"
+    ]
+}
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -15,9 +15,11 @@ define([
     'events/states',
     'events/events',
     'view/error',
-    'controller/events-middleware'
+    'controller/events-middleware',
+    'view/controls/loader'
 ], function(Config, InstreamAdapter, _, Setup, Captions, Model, Storage,
-            Playlist, PlaylistLoader, utils, View, Events, changeStateEvent, states, events, error, eventsMiddleware) {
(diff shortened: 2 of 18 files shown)


.babelrc
```diff
@@ -0,0 +1,5 @@
+{
+    "presets": [
+        "es2015-loose"
+    ]
+}
```
robwalch: I've heard you can put this in package.json...

src/js/controller/controller.js
```diff
@@ -153,6 +155,21 @@ define([
...
+                    ControlsLoader.load()
+                        .then(function (Controls) {
+                            _view.addControls(new Controls(document, _this.currentContainer));
+                        })
+                        .catch(function (reason) {
+                            _this.triggerError({
+                                message: 'Controls failed to load',
+                                reason: reason
+                            });
+                        });
```
robwalch: Good idea since this will translate well to using System.import with webpack2.

src/js/view/view.js
```diff
@@ -20,7 +20,7 @@ define([
     'templates/player.html',
     'view/breakpoint',
     'view/components/button',
-    'view/display-container',
+    'view/display-container'
```
robwalch: I thought you were a fan of the trailing comma

johnBartos: (shrug)

johnBartos:
So I think the beef here is that the timeslider et. al are being updated only on `timeupdate` events, incurring a small visual jitter when re-enabling controls (could be up to 250ms). The `live` text during live mode could be up to a few seconds while we wait for another frag to load/parse.

robwalch:
test this please

PR #1907, opened 2017-03-14
setCaptions sets and reset styles correctly

hghazzi:
- Resetting: clear styles before resetting default styles
- Setting: For Safari, set style in one block instead of multiple times. Convert maps to strings when `!important` is used.

JW7-4086


src/js/utils/css.js
```diff
@@ -58,6 +60,29 @@ define([
         }
     };
```
hghazzi: Wanted to avoid making any significant changes here, but could probably add logic to to convert object to string if any field has `!important` in it.

src/js/utils/css.js
```diff
@@ -17,6 +17,8 @@ define([
                 return;
             }
             var el = document.createElement('div');
+
+            // TODO: Add warning if any styles have !important, since they will be stripped out here.
```
hghazzi: Wanted to avoid making any significant changes here, but could probably add logic to to convert object to string if any field has !important in it.

robwalch:
@hghazzi this is all wrong. Find something else to work on.

robwalch:
@hghazzi most of what you wrote is done in one line by using the css utils properly.

https://github.com/jwplayer/jwplayer/blob/master/test/unit/css-test.js#L34

What we probably need to do here is remove the `if (!style)` check or use an empty string to remove previously set rules. The former is an optimization so that we don't add empty rules so I rather not get rid of it.

https://github.com/jwplayer/jwplayer/blob/master/src/js/utils/css.js#L16

PR #1908, opened 2017-03-14
Get activeTab when page visibility API is webkit prefixed

robwalch:
Fixes viewable support for older versions of iOS, Android and other browsers still using the webkit prefixed page visibility API.

http://caniuse.com/#feat=pagevisibility (view all)

JW7-4153

### DIFF
--- a/src/js/utils/activeTab.js
+++ b/src/js/utils/activeTab.js
@@ -1,0 +1,16 @@
+define([], function() {
+    if ('hidden' in document) {
+        return function () {
+            return !document.hidden;
+        };
+    }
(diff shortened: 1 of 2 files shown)


robwalch:
test this please

src/js/utils/activeTab.js
```diff
@@ -0,0 +1,16 @@
+define([], function() {
+    if ('hidden' in document) {
```
johnBartos: Does this need to be a function? AMD can import plain vars

robwalch:
test this please

PR #1911, opened 2017-03-15
Hide Safari shadow dom media controls with css

egreaves:

Safari sets an opacity value for its media controls which obscure the player upon exiting full screen on iOS devices. This hides it with CSS.

Fixes #
JW7-4148


egreaves:
test this please

PR #1913, opened 2017-03-15
Checks for tracks updates in Safari

karimJWP:
Listens for changes in textTracks after onloadeddata

Occasionally the cc button will be absent when playing a stream containing in manifest vtt on safari

Fixes #
JW7-4050


egreaves:
test this please

PR #1914, opened 2017-03-16
Enable ES6 in source and tests

johnBartos:
ES6 everything!


johnBartos:
OS build and tests pass, for full success it must be built alongside https://github.com/jwplayer/jwplayer-commercial/pull/3169

PR #1916, opened 2017-03-17
Fix race condition where duration is set before streamType updates…

karimJWP:
… to vod - JW7-2115

Listens for changes in streamType to update duration, preventing a race condition

Fixes #
JW7-2115

### DIFF
--- a/src/js/view/controlbar.js
+++ b/src/js/view/controlbar.js
@@ -233,7 +233,6 @@ define([
             this._model.on('change:captionsList', this.onCaptionsList, this);
             this._model.on('change:captionsIndex', this.onCaptionsIndex, this);
             this._model.on('change:streamType', this.onStreamTypeChange, this);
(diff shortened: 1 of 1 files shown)


karimJWP:
Note to reviewer: must be merged with [this PR](https://github.com/jwplayer/jwplayer-commercial/pull/3148)

src/js/view/controlbar.js
```diff
@@ -233,6 +233,7 @@ define([
             this._model.on('change:captionsList', this.onCaptionsList, this);
             this._model.on('change:captionsIndex', this.onCaptionsIndex, this);
             this._model.on('change:streamType', this.onStreamTypeChange, this);
```
robwalch: Since there is already a listener, call `onDuration` from inside `onStreamTypeChange`. Just make sure to pass duration correctly.

karimJWP:
@robwalch changes addressed

PR #1917, opened 2017-03-18
Clean up provider-supported imports

robwalch:


dannyfinks:
@robwalch looks like VAST Ads error out here. Following example autostarts with preroll, but we get an ad error that their is no compatible creatives. http://player:pl4y3r@player-pr-test-jenkins.longtailvideo.com/builds/6043/archive/test/harness.html?edition=ads&config=vast-autostart-preroll.json&setup=1

PR #1920, opened 2017-03-20
Fix api loads during an ad causing incorrect player state

jw-kaurand:
Fixes the issue where the player would not end up in the idle state after using api.load during an ad.  It does this by ensuring that the provider is reattached by triggering destroyPlugin before the player is stopped.

JW7-4172

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -424,8 +424,8 @@ define([
                 }
                 _model.set('preInstreamState', 'instream-idle');
 
-                _stop(true);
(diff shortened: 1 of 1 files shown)


sidpatel13:
Change looks good in the PR build:
http://player-pr-test-jenkins.longtailvideo.com/builds/6046/archive/test/squash/bootstrap.html?primary=html5&edition=ads&feature=api_load_playlist_ads&debug=1
Running our automated tests against the branch to see if there are any failures

PR #1924, opened 2017-03-21
Split controls styles from chromeless player styles

jw-kaurand:
plits out the basic jwplayer layout files from control specific styles. Each set of styles are contained in separate folders.  Controls styles are only included when model.controls is set to true.


robwalch:
@jw-kaurand are these changes still a work in progress?

src/css/skins-test.less
```diff
@@ -1,25 +1,31 @@
-@import "imports/vars";
```
robwalch: Can we remove this file? You left duplicate imports of "jwplayerelement" down there...

src/css/controls.less
```diff
@@ -0,0 +1,45 @@
+@import "shared-imports/mixins";
+
+// TO REMOVE
+@import "controls/imports/jwplayerelement";
```
robwalch: > TO REMOVE?
Yes please. "jwplayerelement" makes no sense - most of the classes styles are on controls.

src/css/controls.less
```diff
@@ -0,0 +1,45 @@
...
+@import "controls/states";
+
+// flags
+@import "controls/flags/time-slider-above";
+@import "controls/flags/cast";
+@import "controls/flags/cast-available";
+@import "controls/flags/live";
+// Duplicate styles between controls-hidden/disabled
+@import "controls/flags/controls-hidden";    // To remove later
+@import "controls/flags/controls-disabled";
```
robwalch: These should no longer be needed. What do we need to think about keeping, replacing or testing after removal?

src/css/controls/imports/autostartmute.less
```diff
@@ -0,0 +1,23 @@
+@import "../../shared-imports/vars";
+@import "../../shared-imports/icons";
```
robwalch: Why would icons be shares? What part of the icon imports apply to anything outside of controls icons?

jw-kaurand: There's one situation when it's used and that is on the error screen to display the error icon in the center of the player.  I was going to try removing that dependency by making the `jw-display-container` not display in controls: false

robwalch: What happens in 7.10 with errors & controls false?

We're going to need to use the error player state when controls is false, because there is no display container without controls.

jw-kaurand:
This has an associated PR in commercial at https://github.com/jwplayer/jwplayer-commercial/pull/3182.  This is in a stable state now and should be ready for review

robwalch:
@jw-kaurand these changes are no longer in sync with v7.11 and cannot be merged.

jw-kaurand:
Closed in favor of the rebased version of this found here: https://github.com/jwplayer/jwplayer/pull/1950

PR #1927, opened 2017-03-22
Implement ES6 in source and tests

johnBartos:


.babelrc
```diff
@@ -1,5 +1,5 @@
 {
     "presets": [
-        "es2015-loose"
+        "es2015"
```
robwalch: What's wrong with loose? Changing this makes all scripts larger by requiring unused to code in transpiled class constructors.

Issue #1136, opened 2016-04-18
Container should not always require an id attribute


coulson84:
### Expected Behavior

If I pass the jwplayer() factory a reference to a node instead of an id string everything should work no problem.
### Actual Behavior

If no id is on the node reference passed to the jwplayer() creator then VAST ads fail to display because the initial video element that is inserted in to the DOM cannot be found
### Steps to reproduce

``` javascript
let videoPlayer = document.createElement('div');
document.body.appendChild(videoPlayer);

jwplayer(videoPlayer).setup({
            file: 'http://content.jwplatform.com/videos/Wf8BfcSt-kNspJqnJ.mp4',
            image: 'http://content.jwplatform.com/thumbs/Wf8BfcSt-640.jpg',
            width: 580,
            height: 370,
            autostart: false,
            advertising: {
                client: 'vast',
                schedule: {
                    'myAds': {
                        'offset': 'pre',
                        'tag': `http://www.adotube.com/php/services/player/OMLService.php?avpid=oRYYzvQ&platform_version=vast20&ad_type=linear&groupbypass=1&HTTP_REFERER=http://www.longtailvideo.com&video_identifier=longtailvideo.com,test`
                    }
                }
            }
        });
```
### Environment
- Browser: Chrome 49.0.2623
- OS: Ubuntu 15.04
- JW Player version: 7.3.6

---

This template is provided as a guide to filing bugs and feature requests. Please replace 
all text _in italics_ with details describing your issue. When submitting a feature request, 
"Steps to reproduce" can be substituted with a user story or list of acceptance criteria.

robwalch:
You can work around this issue by adding an id to the element:

```
let videoPlayer = document.createElement('div');
videoPlayer.id = 'video-player-1';
document.body.appendChild(videoPlayer);
```

coulson84:
This is already how I worked around the issue. I just think it is pretty pointless being able to pass a reference to a node instead of an id string if the node you pass still has to have an id on anyway.

donato:
Agreed that it sucks, but it's an age-old decision. The issue is communication between Flash and the player, as seen here https://github.com/jwplayer/jwplayer/blob/2d59678c4cc70fb8c98b05dfcea24e8a95d0dab0/src/flash/com/longtailvideo/jwplayer/player/SwfEventRouter.as#L67

Would it better suit your needs if we just generated a random id when none is given?

ithcy:
[...]

PR #1929, opened 2017-03-24
Support player element out-of-DOM during setup

robwalch:
This fixes setup for pages that remove the player element during setup. It also adds support for setting up the player with an element rather than an id and adding the player's container on or after the ready event.

Example 1 - element is removed during setup and added on or after ready:
```html
<body>
<div id="player-container">
    <div id="player"></div>
</div>
<script type="text/javascript">
    var container = document.body.querySelector('#player-container');
    jwplayer("player").setup({
        file: '//video.mp4'
    }).on('ready', function() {
        document.body.appendChild(container);
    });
    document.body.removeChild(container);
</script>
</body>
```

Example 2 - player is setup with element not in DOM, element is added afterwards
```html
<body>
<div id="player-container"></div>
<script type="text/javascript">
    var element = document.createElement('div');
    jwplayer(element).setup({
        file: '//video.mp4'
    }).on('ready', function() {
        document.body.appendChild(this.getContainer());
    });
</script>
</body>
```

Fixes JW7-4200 and #1136

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -239,6 +239,7 @@ define([
         }
 
         function _setContainerDimensions() {
+            var inDOM = document.body.contains(_playerElement);
             var bounds = _bounds(_playerElement);
             var containerWidth = Math.round(bounds.width);
             var containerHeight = Math.round(bounds.height);
@@ -247,6 +248,10 @@ define([
 
             // If the container is the same size as before, return early
             if (containerWidth === _lastWidth && containerHeight === _lastHeight) {
+                // Listen for player to be added to DOM
+                if (!_lastWidth || !_lastHeight) {
+                    _responsiveListener();
+                }
                 return;
             }
             // If we have bad values for either dimension, return early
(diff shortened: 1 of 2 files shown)


src/js/view/view.js
```diff
@@ -255,30 +260,41 @@ define([
                 if (!_lastWidth || !_lastHeight) {
                     _responsiveListener();
                 }
+                _model.set('inDom', inDOM);
+                // Fire resize 0,0 if the player element is not in the DOM
+                // This allows setup to complete even if element was removed from DOM
+                if (!inDOM) {
+                    _resized(containerWidth, containerHeight);
```
johnBartos: Does `Math.round(bounds.width)` always return `0` on each browser?

robwalch: Yes. `utils.bounds` returns 0 for all dimensions when the body does not contain the element parameter. If the element is in the browser `getBoundingClientRect` will force a repaint and dimensions will be returned, except for maybe IE which will return 0 until after a reflow / repaint.

PR #1930, opened 2017-03-24
Css transitions

karimJWP:
Transitions.less class is imported into the seven skin to create fade effects

Adds transitions to Seven Skin

Fixes #
JW7-861

### DIFF
--- a/src/css/imports/transitions.less
+++ b/src/css/imports/transitions.less
@@ -1,45 +1,28 @@
-.jw-transitions {
-    //.jw-controls {
-    //    opacity: 0;
-    //    -webkit-transition: opacity 10s;
-    //    -moz-transition: opacity 10s;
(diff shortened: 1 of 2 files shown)


src/css/imports/transitions.less
```diff
@@ -0,0 +1,55 @@
+.jwplayer.jw-skin-seven.jw-state-playing:not(.jw-flag-ads):not(.jw-flag-overlay-open-related) {
+    &.jw-flag-user-inactive {
+        .jw-overlays,
+        .jw-controls {
+            opacity: 0;
+            -webkit-transition: opacity 1s;
```
johnBartos: Autoprefixer should do this; `transition` should be fine

egreaves:
@karimJWP this PR should be created against the `v7.11` branch

PR #1932, opened 2017-03-24
Set Captions options that aren't defaults

egreaves:

Only apply style blocks when captions styles are different than the defaults. Default captions are styled in CSS, with the exception of the font size.

Fixes #
JW7-4086

### DIFF
--- a/src/js/utils/css.js
+++ b/src/js/utils/css.js
@@ -7,21 +7,18 @@ define([
         playerId = playerId || 'all-players';
         var cssText = '';
         if (typeof styles === 'object') {
-            var style;
-            for (style in styles) {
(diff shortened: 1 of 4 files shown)


src/js/utils/css.js
```diff
@@ -123,11 +120,34 @@ define([
...
+    }
+
     return {
         css: _css,
         style: _style,
         clearCss: styleLoader.clear,
         transform: transform,
-        hexToRgba: hexToRgba
+        hexToRgba: hexToRgba,
+        toStyleString: toStyleString
```
robwalch: Let's remove `toStyleString` asap. I don't want us adding additional util methods for something that the `css()` util should do.

PR #1935, opened 2017-03-27
Implement stylelint linter ruleset for LESS 

jw-kaurand:
Implements a set of linting rules for styling JWPlayer.  This is added as a build step and therefore LESS files must pass the linter before the player can be compiled.  All associated style updates that are required to pass the linter's rules have been applied.  Some additional rules will be added in the future as only rules which do no force a significant refactor have been applied.

JW7-4099


jw-kaurand:
Additional rules to be applied in the future can be found here: https://github.com/jwplayer/jwplayer/issues/1936

Feel free to add some suggestions to this.

johnBartos:
Let's try to add an NPM script to run this linter

jw-kaurand:
Closed in favor of the following PR https://github.com/jwplayer/jwplayer/pull/1952

PR #1938, opened 2017-03-27
Ungate Edge from flash throttle check

robwalch:
See if Chrome flash throttle detection works in Edge https://blogs.windows.com/msedgedev/2016/04/07/putting-users-in-control-of-flash/#d4BR0hI8mvmC8KMx.97

JW7-4118

### DIFF
--- a/src/js/providers/flash.js
+++ b/src/js/providers/flash.js
@@ -19,7 +19,8 @@ define([
 
         var sameHost = (a.host === window.location.host);
 
-        return utils.isChrome() && !sameHost;
+        // JW7-4118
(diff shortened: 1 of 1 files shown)


robwalch:
Edge does not dispatch throttled events like Chrome does, so the solution for Chrome does not work in Edge. The swf player is fully functional - it just doesn't render any video or audio. This is not how throttling with Chrome works where the plugin's audio buffer blocks playback and throttled events give us a hint that even though ActionScript can run and execute JavaScript, playback won't start without user interaction.

PR #1940, opened 2017-03-27
Feature/css transitions

karimJWP:
Know issue: closing related does not fade away

Fixes #
JW7-861


src/css/imports/transitions.less
```diff
@@ -0,0 +1,55 @@
...
+            opacity: 1;
+            transition: opacity 1s;
+        }
+    }
+}
+
+.jwplayer.jw-skin-seven.jw-flag-user-inactive.jw-state-playing:not(.jw-flag-media-audio):not(.jw-flag-audio-player):not(.jw-flag-ads-vpaid-controls):not(.jw-flag-casting):not(.jw-flag-ads) {
+    .jw-dock,
+    .jw-controlbar {
+        display: block;
```
robwalch: Transitions should not change the display of elements. If we need controls to be hidden without changing display that should be done in existing CSS rules to support transitions.

You can transition opacity from 0 to 1 and visibility from 'hidden' to 'visible' (and vice versa) to achieve this.

src/css/imports/transitions.less
```diff
@@ -0,0 +1,55 @@
...
+        }
+        &:hover {
+            .jw-knob {
+                width: @thumb-size;
+            }
+        }
+    }
+    .jw-overlays {
+         opacity: 0;
+         transition: opacity 1s;
```
robwalch: Because you've declared that elements with the `.jw-overlays` class will transition opacity here, you do not need to add the `transition` style again below for more specific `.jw-overlays` rules.

src/css/imports/transitions.less
```diff
@@ -0,0 +1,55 @@
+.jwplayer.jw-skin-seven {
```
robwalch: Start this file by defining which classes will make element transition which properties:

```css
.jw-controlbar,
.jw-controls-right,
.jw-display,
.jw-overlays {
    transition: opacity visibility 1s;
}
````

Then you don't have to add `transition` anywhere else, just change the property that will be transitioned. There doesn't need to be any specificity here because `transition` isn't likely to be overridden.

**Comment by karimJWP** [comment]
@robwalch comments addressed

File: src/css/flags/user-inactive.less [hunk 4]
```diff
@@ -11,13 +11,20 @@
           cursor: none;
           -webkit-cursor-visibility: auto-hide;
       }
+      .jw-dock {
+        display: block;
```
robwalch: Why are you adding `display: block` here? Is it set to none somewhere else? If so that's where you remove none and replace it wit these visibility/opacity 0.

src/css/imports/controlbar.less
```diff
@@ -153,9 +153,11 @@
 .jw-flag-autostart,
 .jw-flag-user-inactive.jw-state-playing,
 .jw-flag-user-inactive.jw-state-buffering {
-  &:not(.jw-flag-media-audio):not(.jw-flag-audio-player):not(.jw-flag-ads-vpaid-controls):not(.jw-flag-casting) {
+  &:not(.jw-flag-media-audio):not(.jw-flag-audio-player):not(.jw-flag-ads-vpaid-controls):not(.jw-flag-casting):not(.jw-flag-ads) {
```
robwalch: You can't add `:not(.jw-flag-ads)`. That will increase specificity which could introduce regressions. There already is going to be a rule for ads that you need to modify.

src/css/imports/transitions.less
```diff
@@ -0,0 +1,7 @@
+.jw-controls-right,
+.jw-controlbar,
+.jw-display,
+.jw-overlay,
+.jw-dock {
```
robwalch: jw-dock is inside jw-controls-right, so there is no reason to do the same transition on both. Just fade in jw-controls-right and everything inside fades too.

You could even make it simpler and just fade in jw-controls most of these are in jw-controls.

src/css/skins/seven.less
```diff
@@ -48,4 +47,15 @@
     -webkit-font-smoothing: antialiased;
     -moz-osx-font-smoothing: grayscale;
   }
+
+  .jw-slider-container {
+      .jw-knob {
+          width: 0;
```
robwalch: What about the height? What's the point of this? Why not just display: none?

src/css/flags/user-inactive.less
karimJWP: Thanks for catching that, i didn't realize the '.jwplayer.jw-flag-controls-hidden;' above was triggering controls-hidden.less

src/css/imports/transitions.less
karimJWP: @robwalch it seems that jw-controls is only altered for .jw-flag-user-inactive when time-slider is above (time-slider-above.less), and in controls-hidden.less under .jw-flag-controls-hidden, but that code never runs. Should we add styles for jw-controls when .jw-flag-user-inactive? without the styles, the transitions never kick-in. 
All other comments are addressed, thanks for the awesome feedback 👍  * 💯

PR #1943, opened 2017-03-28
Set/Clear requestAnimationFrame & setTimeout correctly

hghazzi:
This bug was causing us to clear the timeout id using the
requestAnimationFrame id, which affected casting

JW7-4207

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -1039,7 +1039,7 @@ define([
 
             _currentState = model.get('state');
             // Throttle all state change UI updates except for play to prevent iOS 10 animation bug
-            clearTimeout(_previewDisplayStateTimeout);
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -1039,7 +1039,7 @@ define([
 
             _currentState = model.get('state');
             // Throttle all state change UI updates except for play to prevent iOS 10 animation bug
-            clearTimeout(_previewDisplayStateTimeout);
+            _cancelDelayResize(_previewDisplayStateTimeout);
```
robwalch: @hghazzi you missed the one in `destroy`.

hghazzi: thanks, will create a new PR

PR #1944, opened 2017-03-28
When setting container, add video tag before child

pajong:
This is to avoid video tag from coming in front of a child that should display on top of the video.
Specifically, this makes a vpaid nonlinear ad to hide behind the video tag.
ADS-5


src/js/providers/html5.js
```diff
@@ -744,7 +744,11 @@ define([
 
         this.setContainer = function(containerElement) {
             _container = containerElement;
-            containerElement.appendChild(_videotag);
+            if (containerElement.firstChild) {
```
johnBartos: Passing `null` to the first arg of `insertBefore` is allowed; this can be simplified to `containerElement.insertBefore(containerElement.firstChild, _videotTag)`

robwalch: @pajong there should be a commercial PR that does the same for HLS and DASH providers.

@johnBartos Please verify passing `null` or `undefined` as the first arg to `insertBefore` is OK in IE9.

src/js/providers/html5.js
```diff
@@ -744,7 +744,7 @@ define([
 
         this.setContainer = function(containerElement) {
             _container = containerElement;
-            containerElement.appendChild(_videotag);
+            containerElement.insertBefore(containerElement.firstChild, _videotag);
```
robwalch: only `insertBefore` if there is a `containerElement.firstChild`, otherwise append. Also you got it backwards - video first then firstChild - BTW are you testing this @pajong.

pajong:
Messed up the ordering. Verified that ```insertBefore(element, null)``` works in IE9.

PR #1949, opened 2017-04-03
Feature/css transitions v7.11

karimJWP:
Fixes #
JW7-861

### DIFF
--- a/src/css/imports/transitions.less
+++ b/src/css/imports/transitions.less
@@ -9,10 +9,6 @@
             }
         }
     }
-    .jw-overlays {
-         opacity: 0;
-         transition: opacity 1s;
-    }
     &.jw-flag-overlay-open-related {
          .jw-overlays {
              opacity: 1;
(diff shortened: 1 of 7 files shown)


src/css/imports/transitions.less
```diff
@@ -0,0 +1,7 @@
+.jw-controls-right,
+.jw-controlbar,
+.jw-display,
+.jw-overlay,
+.jw-dock {
```
karimJWP: @robwalch copying your comment from previous branch 
"robwalch 5 days ago Member
jw-dock is inside jw-controls-right, so there is no reason to do the same transition on both. Just fade in jw-controls-right and everything inside fades too.

You could even make it simpler and just fade in jw-controls most of these are in jw-controls.
       @karimJWP
karimJWP 4 days ago Member
@robwalch it seems that jw-controls is only altered for .jw-flag-user-inactive when time-slider is above (time-slider-above.less), and in controls-hidden.less under .jw-flag-controls-hidden, but that code never runs. Should we add styles for jw-controls when .jw-flag-user-inactive? without the styles, the transitions never kick-in.
All other comments are addressed, thanks for the awesome feedback 👍 * 💯

robwalch:
@karimJWP please work with @jw-kaurand to update these changes to work with the latest CSS files.

PR #1950, opened 2017-04-03
Split controls styles from chromeless player styles

jw-kaurand:
Splits out the basic jwplayer layout files from control specific styles. Each set of styles are contained in separate folders. Controls styles are only included when model.controls is set to true.

JW7-4074


src/js/view/controls/controls.js
```diff
@@ -44,6 +44,12 @@ define([
         return { reason: 'interaction' };
     };
 
+    // Include the separate chunk that contains the controls styles.  Check webpackJsonjwplayer so we don't
+    // run this in phantomjs because it breaks despite working in browser and including files like we want to.
+    if (window.webpackJsonpjwplayer) {
```
robwalch: We don't need these if statements anymore since unit tests are webpacked.

PR #1957, opened 2017-04-04
Adds browsers prefixes via PostCSS Autoprefixer

jw-kaurand:
This adds PostCSS and its autoprefixer plugin to both the web pack and grunt build processes.  The prefixes added are based on the browsers present in the browserslist file.  Some nonstandard styles are retained for features specific to the platform.

JW7-4100


src/css/controls/imports/nextup.less
```diff
@@ -2,8 +2,9 @@
 @import "icons";
 
 .jw-nextup-container {
-    -webkit-font-smoothing: antialiased;
-    -moz-font-smoothing: antialiased;
+    font-smoothing: antialiased;
+    /* OSX Safari specific style for improving legibility */
+    -moz-osx-font-smoothing: grayscale;
```
robwalch: Isn't `-moz` Firefox specific?

PR #1958, opened 2017-04-05
Pass feed data to "playlist" and "playlistItem" events

karimJWP:
Fixes #
JW7-4127

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -124,6 +124,7 @@ define([
 
             config.playlist = [ obj ];
         } else if (_.isArray(configPlaylist.playlist)) {
+            // The "playlist" in the config is actually a feed that contains a playlist
             config.feedData = configPlaylist;
             config.playlist = configPlaylist.playlist;
         }
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -107,7 +107,8 @@ define([
             delete config.aspectratio;
         }
 
-        if (!config.playlist) {
+        var configPlaylist = config.playlist;
+        if (!configPlaylist) {
             // This is a legacy fallback, assuming a playlist item has been flattened into the config
             var obj = _.pick(config, [
                 'title',
@@ -122,6 +123,9 @@ define([
             ]);
 
             config.playlist = [ obj ];
+        } else if (_.isArray(configPlaylist.playlist)) {
+            config.feedData = configPlaylist;
+            config.playlist = configPlaylist.playlist;
         }
 
         config.qualityLabels = config.qualityLabels || config.hlslabels;
--- a/src/js/playlist/playlist.js
+++ b/src/js/playlist/playlist.js
@@ -13,10 +13,12 @@ define([
     };
 
     /** Go through the playlist and choose a single playable type to play; remove sources of a different type **/
-    Playlist.filterPlaylist = function(playlist, model) {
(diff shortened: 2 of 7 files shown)


src/js/playlist/playlist.js
```diff
@@ -18,6 +18,8 @@ define([
         var providers = model.getProviders();
         var preload = model.get('preload');
         var feedid = model.get('feedid');
+        var feedInstanceId = model.get('feed_instance_id');
+        var pinSetId = model.get('pin_set_id');
```
robwalch: This solution works on setup, but after that, if another playlist is loaded, these model properties would be invalid and should not be added to each item.

@karimJWP @dannyfinks

src/js/playlist/playlist.js
```diff
@@ -43,6 +45,14 @@ define([
                 item.feedid = item.feedid || feedid;
             }
 
+            if (item.feedInstanceId || feedInstanceId) {
+                item.feedInstanceId = item.feedInstanceId || feedInstanceId;
+            }
+
+            if (item.pinSetId || pinSetId) {
+                item.pinSetId = item.pinSetId || pinSetId;
+            }
```
robwalch: get rid of this `pinSetId` copying - because not all items in the playlist are pinned. This should already be in the item.

karimJWP:
@robwalch changes addressed

robwalch:
I've added feed data at the item level and in playlist events so that data only available at the top level of feeds can now be accessed in playlist events and playlist items.

Setup also accepts a `playlist` block that is a complete feed object with a playlist array.

```js
// setup example
jwplayer().setup({
   playlist: {
    title: 'Feed Title',
    pubdate: '2017-04-06',
    playlist: [
      {
        file: 'video.mp4'
      }
    ]
  }
});

// playlist events now contain feedData
jwplayer().on('playlist', event => console.log(event));

> {
  type: 'playlist',
  playlist: [
      {
        file: 'video.mp4'
        sources: ...
        feedData: ...
      }
  ],
  feedData: {
    title: 'Feed Title',
    pubdate: '2017-04-06',
  }
}

// playlist items now contain feedData
jwplayer().on('playlist', event => console.log(event));

> {
  type: 'playlistItem',
  index: 0,
  item: [
      {
        file: 'video.mp4'
        sources: ...
        feedData: {
          title: 'Feed Title',
          pubdate: '2017-04-06',
        }
      }
  ]
}
```

src/js/api/config.js
```diff
@@ -122,6 +123,9 @@ define([
             ]);
 
             config.playlist = [ obj ];
+        } else if (_.isArray(configPlaylist.playlist)) {
+            config.feedData = configPlaylist;
+            config.playlist = configPlaylist.playlist;
```
karimJWP: @robwalch  if config.feedData is equal to configPlaylist, should configPlaylist be renamed? It seems confusing that config.playlist would be configPlaylist.playlist

robwalch: When you inline a feed, it looks like this:
```js
jwplayer().setup({
   playlist: {
    playlist: [
      {
        file: 'video.mp4'
      }
    ],
    title: 'Feed Title',
    pubdate: '2017-04-06'
  }
});
```
So the "playlist" in the config is actually a feed that contains a playlist.

karimJWP: 👍

PR #1961, opened 2017-04-06
Disable knob animation on touch devices 

karimJWP:
disables knob animation on touch devices 

Fixes #
JW7-4242

### DIFF
--- a/src/css/skins/seven.less
+++ b/src/css/skins/seven.less
@@ -51,14 +51,16 @@
         -moz-osx-font-smoothing: grayscale;
     }
 
-    .jw-slider-container {
-        .jw-knob {
-            display: none;
-        }
-
-        &:hover {
+    &:not(.jw-flag-touch) {
+        .jw-slider-container {
             .jw-knob {
-                display: block;
+                display: none;
+            }
+
+            &:hover {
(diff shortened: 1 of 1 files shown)


src/css/skins/seven.less
```diff
@@ -50,14 +50,16 @@
         -moz-osx-font-smoothing: grayscale;
     }
 
-    .jw-slider-container {
-        .jw-knob {
-            display: none;
-        }
-
-        &:hover {
+    &:not(.jw-flag-touch) {
```
robwalch: This increases specificity, which should be avoided.

If the goal is to override existing ruled on touch devices, add a new `.jw-flag-touch` that overrides the existing rule.

karimJWP: addressed

robwalch:
@karimJWP please rebase to address the build error. It looks like the commercial build is still pulling down the old template files and that's why the build fails.

karimJWP:
@robwalch PR rebased

egreaves:
@karimJWP the knob always shows on desktop in `timeSliderAbove` mode: 
screen shot 2017-04-10 at 10 41 34 pm

PR #1968, opened 2017-04-10
Fix errors in dock template

johnBartos:
- Watch out for hanging `<`
- Add default parameters to avoid `undefined` in the markup

JW7-4245

### DIFF
--- a/src/js/view/controls/templates/dock.js
+++ b/src/js/view/controls/templates/dock.js
@@ -11,13 +11,12 @@ export default (dockButtons = []) => {
     );
 };
 
-
-const dockButton = (buttonClass, buttonId, image, tooltipText) => {
-    const style = image ? `style='background-image: url("${image}")` : '';
+const dockButton = (buttonClass = '', buttonId = '', image, tooltipText) => {
+    const style = image ? `style="background-image: url(${image})"` : '';
     const tooltipHtml = tooltipText ? tooltip(tooltipText) : '';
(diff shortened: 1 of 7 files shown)


src/js/view/controls/templates/dock.js
```diff
@@ -16,8 +16,8 @@ const dockButton = (buttonClass, buttonId, image, tooltipText) => {
     const style = image ? `style='background-image: url("${image}")` : '';
```
robwalch: Missing closing single-quote on `style='... url()`

src/js/view/controls/templates/dock.js
```diff
@@ -16,8 +16,8 @@ const dockButton = (buttonClass, buttonId, image, tooltipText) => {
...
     const tooltipHtml = tooltipText ? tooltip(tooltipText) : '';
     return (
-        `<div class="jw-dock-button jw-background-color jw-reset ${buttonClass || ''}" button="${buttonId}">` +
-            `<div class="jw-icon jw-dock-image jw-reset" ${style}</div>` +
+        `<div class="jw-dock-button jw-background-color jw-reset ${buttonClass || ''}" button=${buttonId}">` +
+            `<div class="jw-icon jw-dock-image jw-reset ${style}"></div>` +
```
robwalch: This is incorrect. `style=` is a new attribute - not part of the class attribute.

PR #1971, opened 2017-04-10
Bugfix/font smoothing

robwalch:
Remove "font-smoothing" and add private "-webkit-font-smoothing" style for icons and text. Remove some duplicate style definitions - mainly get rid of jw-icon font style duplication by moving it out of mixins.

JW7-4241


robwalch:
FYI @jw-kaurand

PR #1974, opened 2017-04-11
Hide logo when player is inactive on chromeless player with logo.hide = true

robwalch:
Move user active/inactive css logic back into the player view to support logo hiding functionality that is enabled in custom logo config options:

```
    "logo": {
        "hide": true, // huzza!
        "file": "logo.png"
    }
```

I'm not a fan of these changes, and seriously think, as part of chromeless this feature should be deprecated; The feature should still be there so that the logo hides with controls, but when controls are disabled, logo hiding should also be disabled.

### DIFF
--- a/src/js/view/utils/user-activity.js
+++ b/src/js/view/utils/user-activity.js
@@ -1,0 +1,53 @@
+define([
+    'utils/backbone.events',
+    'utils/helpers',
+    'utils/underscore',
+], function (Events, utils, _) {
+
+    const ACTIVE_TIMEOUT = utils.isMobile() ? 4000 : 2000;
+
+    return class UserActivity {
+        constructor() {
+            _.extend(this, Events);
+
+            // Alphabetic order
+            // Any property on the prototype should be initialized here first
+            this.activeTimeout = -1;
+            this.showing = false;
+            this.activeListeners = {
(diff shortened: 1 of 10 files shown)


src/js/view/utils/user-activity.js
```diff
@@ -0,0 +1,53 @@
+define([
```
robwalch: export default class

src/js/view/view.js
```diff
@@ -12,12 +12,13 @@ define([
...
     'view/utils/flag-no-focus',
     'view/utils/clickhandler',
+    'view/utils/user-activity',
     'view/captionsrenderer',
     'view/logo',
     'view/preview',
     'view/title',
 ], function(events, states, Events, utils, _, activeTab, raf, requestFullscreenHelper, setBreakpoint, flagNoFocus,
-            ClickHandler, CaptionsRenderer, Logo, Preview, Title) {
+            ClickHandler, UserActivity, CaptionsRenderer, Logo, Preview, Title) {
```
robwalch: import UserActivity

robwalch:
Closing. Hiding the logo in chromeless will be deprecated. I will make a new PR that includes some of the CSS and template improvements.

PR #1978, opened 2017-04-11
Fullscreen fix when escaping or ad click

hghazzi:
Missing handler for document fullscreen events

JW7-4235


hghazzi:
@robwalch I know you mentioned something else to watch out for, do you remember was it is?

PR #1985, opened 2017-04-13
Only add controls if no error class on player element

hghazzi:
JW7-4236


src/js/controller/controller.js
```diff
@@ -206,9 +206,11 @@ define([
                 if (enable) {
                     ControlsLoader.load()
                         .then(function (Controls) {
-                            if (!_view.isSetup) {
+                            var hasError = _this.currentContainer.classList.contains('jw-error');
```
johnBartos: `const`

PR #1986, opened 2017-04-13
Raise captions on pause

karimJWP:
When the state changes to PAUSED, call resize to adjust captions for control bar

Fixes #
JW7-3952

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -714,6 +714,7 @@ define([
                     if (_model.get('controls')) {
                         _captionsRenderer.resize();
                     }
+                    break;
                 default:
                     break;
             }
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -107,7 +107,6 @@ define([
             }
 
             _resizeMedia(containerWidth, containerHeight);
-
             _captionsRenderer.resize();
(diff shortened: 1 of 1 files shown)


egreaves:
test this please

src/js/view/view.js
```diff
@@ -711,6 +710,10 @@ define([
                 case states.PLAYING:
                     _resizeMedia();
                     break;
+                case states.PAUSED:
+                    if (_model.get('controls')) {
+                        _captionsRenderer.resize();
+                    }
```
hghazzi: @karimJWP Build is failing because a `break;` needs to be added to this case. The linter does not allow fallthroughs.

https://jenkins.jwplayer.com/job/jw7-commercial/6550/console

karimJWP: damn, good catch! Thanks!

robwalch: You only need to call `_captionsRenderer.renderCues(true);` This does less work and is what is used in other places to reposition text above the cb.

karimJWP:
retest this please

hghazzi:
@karimJWP can you rebase your branch against master? This that should fix the build

PR #1989, opened 2017-04-14
Check null for applyProviderListeners

pajong:
Allow for `applyProviderListeners` t be called from ads plugins without a provider to reset the current ads provider event listeners (html5 only). Return early when `applyProviderListeners` is called without a provider.

ADS-218

### DIFF
--- a/src/js/controller/instream-html5.js
+++ b/src/js/controller/instream-html5.js
@@ -60,6 +60,10 @@ define([
             // check provider after item change
             _checkProvider(provider);
 
+            if (!provider) {
+                return;
+            }
+
             // Match the main player's controls state
             provider.off(events.JWPLAYER_ERROR);
             provider.on(events.JWPLAYER_ERROR, function(data) {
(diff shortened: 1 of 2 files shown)


src/js/controller/instream-html5.js
```diff
@@ -60,6 +60,10 @@ define([
             // check provider after item change
             _checkProvider(provider);
 
+            if (!provider) {
```
MaxMikheyenko: should this be above `_checkProvider(provider);`?

robwalch:
Yes you are returning when provider is null, but the real change here is that 
`_checkProvider(null)` does something... explain what that something is and how it fixes ADS-218.

src/js/controller/instream-html5.js
robwalch: no. he wants to reset listeners.

robwalch:
@MaxMikheyenko @egreaves If this was supposed to be merged into 7.10.x, you should have made a PR or coordinated the cherry pick.

PR #1994, opened 2017-04-18
Improve performance of player view(s)

robwalch:
Revisited changes in #1980, and implemented performance optimizations for player setup, resize and view-ability checks.

JW7-4265

### DIFF
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -94,16 +94,17 @@ define([
                 const playDisplayIcon = new PlayDisplayIcon(model);
                 const nextDisplayIcon = new NextDisplayIcon(model, api);
 
-                // Toggle playback on mobile
-                if (touchMode) {
-                    playDisplayIcon.on('tap', () => {
-                        this.trigger(events.JWPLAYER_DISPLAY_CLICK);
-                        this.userActive(1000);
-                        api.play(reasonInteraction());
-                    });
-                } else {
-                    // On desktop allow media element to capture all play/pause toggle clicks
+                playDisplayIcon.on('click tap', () => {
+                    this.trigger(events.JWPLAYER_DISPLAY_CLICK);
+                    this.userActive(1000);
+                    api.play(reasonInteraction());
+                });
+
+                if (utils.isChrome() && !touchMode) {
+                    // On Chrome desktop allow media element to capture all play/pause toggle clicks
+                    // This allows swfs to capture clicks on start preventing flash-throttling
                     playDisplayIcon.el.style.pointerEvents = 'none';
+                    playDisplayIcon.icon.style.pointerEvents = 'none';
                 }
 
                 displayContainer.addButton(rewindDisplayIcon);
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -44,7 +44,7 @@ define([
         return { reason: 'interaction' };
     };
 
-    require('css/controls.less');
+    let stylesInjected = false;
 
     return class Controls {
         constructor(context, playerContainer) {
@@ -73,6 +73,10 @@ define([
                 mouseout: () => this.userActive()
             };
             this.dimensions = {};
+            if (!stylesInjected) {
+                stylesInjected = true;
+                require('css/controls.less');
+            }
         }
 
         enable(api, model) {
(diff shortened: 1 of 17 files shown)


src/js/view/controls/controls.js
```diff
@@ -73,6 +73,10 @@ define([
                 mouseout: () => this.userActive()
             };
             this.dimensions = {};
+            if (!stylesInjected) {
+                stylesInjected = true;
+                require('css/controls.less');
```
johnBartos: This doesn't need to be ensured?

robwalch: No. We're not splitting out the css. Just timing when the module is imported.

src/js/view/view.js
```diff
@@ -716,6 +724,17 @@ define([
                 default:
                     break;
             }
```
hghazzi: Seems to me that we can merge the case statement above and the one below into one.

src/js/view/view.js
```diff
@@ -328,11 +325,27 @@ define([
...
+            this.updateBounds();
+            updateVisibility();
+
+            _model.change('state', _stateHandler);
+            _model.on('change:fullscreen', _fullscreen);
+            _model.on('change:activeTab', updateVisibility);
+            _model.on('change:fullscreen', updateVisibility);
+            _model.on('change:intersectionRatio', updateVisibility);
+            _model.change('visibility', function observedVisibilityChange(model, visibility) {
+                if (visibility && !_lastVisibility) {
```
johnBartos: Use last model prop from change event

src/js/view/view.js
```diff
@@ -328,11 +325,27 @@ define([
...
-            _setContainerDimensions();
+            this.updateBounds();
+            updateVisibility();
+
+            _model.change('state', _stateHandler);
+            _model.on('change:fullscreen', _fullscreen);
+            _model.on('change:activeTab', updateVisibility);
+            _model.on('change:fullscreen', updateVisibility);
+            _model.on('change:intersectionRatio', updateVisibility);
+            _model.change('visibility', function observedVisibilityChange(model, visibility) {
```
johnBartos: arrow function too

src/js/view/view.js
robwalch: No we cannot.

robwalch: @hghazzi I didn't think it was worth it (to keep it short requires fall-throughs and inner if statements. Can you provide an example? Maybe just the first one is two if statements.

robwalch:
```js
            if (state === states.COMPLETE) {
                _api.setFullscreen(false);
            }
            
            switch (state) {
                case states.IDLE:
                case states.ERROR:
                case states.COMPLETE:
                    _captionsRenderer.hide();
                    break;
                default:
                    _captionsRenderer.show();
                    if (state === states.PAUSED &&_controls && !_controls.showing) {
                        _captionsRenderer.renderCues(true);
                    }
                    break;
            }
```

src/js/view/view.js
robwalch: Will do

src/js/view/view.js
hghazzi: How about this?

```
            if (state === states.COMPLETE) {
                _api.setFullscreen(false);
            }
            
            switch (state) {
                case states.IDLE:
                case states.ERROR:
                case states.COMPLETE:
                    _captionsRenderer.hide();
                    break;
                case states.PAUSED:
                    _captionsRenderer.show();
                    if (_controls && !_controls.showing) {
                        _captionsRenderer.renderCues(true);
                    }
                 break;
                default:
                    _captionsRenderer.show();
                    break;
            }
```

I like breaking out the fullscreen check because then the switch case is just related to the captions logics. And this way, we may be dupe call to `_captionsRenderer.show();`, which is clearer and avoids doing a state check in `default` which doesn't feel right.

robwalch:
@dannyfinks @msiddique please rerun this PR build once tests are updated [🍹](http://jenkins.jwplayer.com/job/jw7-pr-builder-opensource/1775/)

robwalch:
@dannyfinks click on display-icon is reenabled for all browsers except Chrome desktop which needs to pass clicks through to the media element to prevent flash-throttling.

PR #1999, opened 2017-04-20
Move aria label util into module

hghazzi:
To better break up the UI logic from the rest of the player, move aria labels into its own module. 

JW7-4282

### DIFF
--- a/src/js/controller/aria.js
+++ b/src/js/controller/aria.js
@@ -1,0 +1,19 @@
+define([
+
+], function() {
+
+    const addLabel = function(element, ariaLabel) {
+        if (!element || !ariaLabel) {
+            return;
+        }
+
+        element.setAttribute('aria-label', ariaLabel);
+        element.setAttribute('role', 'button');
+        element.setAttribute('tabindex', '0');
+    };
+
+    return {
+        addLabel: addLabel
+    };
+});
+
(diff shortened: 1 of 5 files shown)


johnBartos:
Use the PR template

src/js/controller/aria.js
```diff
@@ -0,0 +1,19 @@
...
+
+        element.setAttribute('aria-label', ariaLabel);
+        element.setAttribute('role', 'button');
+        element.setAttribute('tabindex', '0');
+    };
+
+    return {
+        addLabel: addLabel
+    };
+});
```
robwalch: Put this in the utils directory. It's not a controller / has no business logic.

src/js/view/controls/components/tooltip.js
```diff
@@ -1,8 +1,9 @@
 define([
     'utils/backbone.events',
+    'controller/aria',
```
robwalch: 'utils/aria'

src/js/utils/aria.js
```diff
@@ -0,0 +1,18 @@
...
+        }
+
+        element.setAttribute('aria-label', ariaLabel);
+        element.setAttribute('role', 'button');
+        element.setAttribute('tabindex', '0');
+    };
+
+    return {
+        addLabel: addLabel
+    };
```
robwalch: Just return the function. Waste of an object and `. addLabel` can't be minified.

PR #2000, opened 2017-04-20
Player API/View fixes

robwalch:

Address issues introduced by #1994 which led to JW7-4265 being rejected.

Fixes issues blocking release of 7.11.0-alpha.1:
- `player.resize()` not updating height
- `readyEvent.viewable` missing
- `player.setup(elementNotInDOM)` no longer firing 'ready' / completing setup

Some of the regressions came from firing "resize" from within the `updateBounds` call. This has now been split out so there are three calls to fully update a player view: `updateBounds`, `updateStyles` and finally `checkResized` which will dispatch a "resize" event when player dimensions have changed. It's important that the view fire a "resize" event for setup to complete, and it's important for performance that these methods be split up.

Order of visibility/viewable calls in the controller have been updated to ensure that "viewable" is added to the ready event.

#### Addresses Issue(s):

JW7-4265

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -251,8 +251,6 @@ define([
 
                 _view.on('all', _triggerAfterReady, _this);
 
-                _model.change('visibility', _onVisibilityChange);
-
                 var related = _api.getPlugin('related');
                 if (related) {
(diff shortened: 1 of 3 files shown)


dannyfinks:
@robwalch fixes look good to me

PR #2002, opened 2017-04-21
Lazy instantiate intersection observer and set container classes in view

robwalch:

1. Makes sure that the IntersectionObserver is instantiated when needed after the polyfill has been loaded. Moving the instantiation to a modules top level scope meant we tried creating it before the polyfill was available.
2. Move code that adds flag classes to the player view so that for players with fixed dimensions, classes can be added before the view is added to the DOM / before controls are loaded. This avoids some layout invalidation when controls are loaded, and allows controlbar-only audio players to render correctly on embed.

Fixes UI update issues in browser that rely on the IntersectionObserver polyfill, as well as improves the view DOM state at setup.

If the intersection observer is not instantiated an error will be thrown. This is intentional. We need view-ability to work for the player's UI and analytics to function correctly.

#### Addresses Issue(s):

JW7-4280 JW7-4281 JW7-4265

### DIFF
--- a/src/js/view/utils/audio-mode.js
+++ b/src/js/view/utils/audio-mode.js
@@ -1,0 +1,20 @@
+export const CONTOLBAR_ONLY_HEIGHT = 44;
+
+export const isAudioMode = function(model) {
+    const playerHeight = model.get('height');
+    if (model.get('aspectratio')) {
+        return false;
+    }
+    if (typeof playerHeight === 'string' && playerHeight.indexOf('%') > -1) {
+        return false;
+    }
+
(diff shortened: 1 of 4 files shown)


src/js/view/utils/audio-mode.js
```diff
@@ -0,0 +1,20 @@
+export const CONTOLBAR_ONLY_HEIGHT = 44;
```
egreaves: `CONTROLBAR`

dannyfinks:
@robwalch ran some tests in IE11 and Firefox and everything passed.

PR #2007, opened 2017-04-24
Create standard error module

johnBartos:
Creates an Error module which standardizes errors through numerical code values

Errors throughout the player are hand-coded, inconsistent, and hard to program against. Using a standardized numerical system will simplify our code base and ensure errors are uniform across plugins

This module is currently unused in this project but used in Commercial - reviews here are limited to feedback specifically about the structure and approach to this module. Error codes and categories can be decided & changed in a later PR.

https://github.com/jwplayer/jwplayer-commercial/pull/3348

#### Addresses Issue(s):

JW7-2606

### DIFF
--- a/src/js/utils/error.js
+++ b/src/js/utils/error.js
@@ -41,14 +41,12 @@ export const networkErrorData = (url, statusCode, responseText) => {
 const createMessage = (category, code, severity, data) => {
     let message = '';
     switch (category) {
-        case errorCategory.NETWORK: {
+        case errorCategory.NETWORK:
             message = createNetworkErrorMessage(code, data);
             break;
-        }
-        default: {
+        default:
             message = 'Unknown Error';
             break;
-        }
     }
 
     return message;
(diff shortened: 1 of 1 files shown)


johnBartos:
🙄 browserstack timeout

src/js/utils/error.js
```diff
@@ -0,0 +1,91 @@
...
+    // Crossdomain error
+    CROSSDOMAIN_ERROR: 4002,
+    // Accessing HTTP over HTTPS
+    BLOCKED_MIXED_CONTENT: 4003
+};
+
+export const errorSeverity = {
+    RECOVERABLE: 1,
+    FATAL: 2
+};
```
robwalch: Do these need to be maps? Can we export each const so they all become variables that can be minified?

Ex: `export const CATEGORY_PLAYER = 1;`

johnBartos: I'd rather have a file `error-category.js` and inside `export const PLAYER = 1` such that you can `import * as ErrorCategory from error-category`. Need to balance developer ease with minification.

johnBartos: Update: It seems that doing it your way does not allow minification either.

johnBartos:
Closed since this is now its own repo

PR #2012, opened 2017-04-25
API setConfig

hghazzi:
For supported attributes, use `setConfig` to set multiple attributes at
once.

JW7-4150

### DIFF
--- a/src/js/api/set-config.js
+++ b/src/js/api/set-config.js
@@ -1,0 +1,62 @@
+const _ = require('utils/underscore');
+
+const supportedFields = [
+    'repeat',
+    'volume',
+    'mute',
+    'autostart'
+];
+let model;
+
+function setVolume(mute, volume) {
+    const muteSet = !_.isUndefined(mute);
+    const volumeSet = !_.isUndefined(volume);
+
+    if (volumeSet) {
+        model.setVolume(volume);
+        if (!muteSet) {
+            model.setMute(volume === 0);
+        }
+    }
+
+    if (muteSet) {
+        model.setMute(mute);
+    }
+}
+
+function setAutoStart(controller, autoStart) {
+    model.setAutoStart(autoStart);
+
+    if (model.get('state') === 'idle' && autoStart === true) {
+        controller._play({ reason: 'autostart' });
+    }
+}
+
+export default (controller, newConfig = {}) => {
+    model = controller._model;
+
+    if (!_.size(newConfig)) {
+        return;
+    }
+
+    let newValue;
+    for (let field of supportedFields) {
+        newValue = newConfig[field];
+
+        if (_.isUndefined(newValue)) {
+            continue;
+        }
+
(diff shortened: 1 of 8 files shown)


src/js/api/set-config.js
```diff
@@ -0,0 +1,71 @@
+define([
```
johnBartos: Use es6 export. To import `_`, do `const _ = require('utils/underscore');`

src/js/api/set-config.js
```diff
@@ -0,0 +1,71 @@
...
+            model.set('autostartMuted', true);
+        }
+        model.set('playOnViewable', autoStartOnMobile || autoStart === 'viewable');
+
+        if (model.get('state') === 'idle' && autoStart === true) {
+            controller._play({ reason: 'autostart' });
+        }
+    }
+
+    return (controller, newConfig = {}) => {
```
johnBartos: I believe it would be more testable to pass in `controller` and `model` instead of just `controller`

src/js/api/set-config.js
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    function setAutostart(controller, autoStart) {
+        model.set('autostart', autoStart);
+
+        const autoStartOnMobile = model.autoStartOnMobile();
+        if (autoStartOnMobile) {
+            model.set('autostartMuted', true);
+        }
+        model.set('playOnViewable', autoStartOnMobile || autoStart === 'viewable');
```
johnBartos: Since this is lifted directly from https://github.com/jwplayer/jwplayer/blob/master/src/js/controller/setup-steps.js#L294 I think it would be better to break both into a single method

hghazzi:
@johnBartos Addressed changes, chose to keep accessing model from controller since it is a property that is instantiated in its constructor and should always be there.

src/js/api/set-config.js
```diff
@@ -0,0 +1,62 @@
...
+let model;
+
+function setVolume(mute, volume) {
+    const muteSet = !_.isUndefined(mute);
+    const volumeSet = !_.isUndefined(volume);
+
+    if (volumeSet) {
+        model.setVolume(volume);
+        if (!muteSet) {
+            model.setMute(false);
```
egreaves: this should be `true` if `volume === 0`

hghazzi:
test this please

hghazzi:
test this please

src/js/controller/controller.js
```diff
@@ -789,6 +791,10 @@ define([
             this.setCaptions = _view.setCaptions;
             this.next = _nextUp;
 
+            this.setConfig = function(newConfig) {
```
johnBartos: More concisely expressed as `this.setConfig = (newConfig) => setConfig(_this, newConfig);`

hghazzi: 👍

PR #2013, opened 2017-04-25
Add title to video tag

karimJWP:
Adds the item's title to the video tag.

This allows the video's title to appear on iOS device's lock-screens.

Suggesting potential edge case scenarios would be greatly appreciated. Explored playlist with ads where some items don't have titles. 

Enables title to be displayed in the iOS lock screen, instead of video's URL

JW7-3410

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -400,8 +400,10 @@ define([
             _model.on('change:fullscreen', updateVisibility);
             _model.on('change:intersectionRatio', updateVisibility);
             _model.on('change:visibility', redraw);
+            _model.on('itemReady', itemReady);
 
             updateVisibility();
+            itemReady();
 
             // Always draw first player for icons to load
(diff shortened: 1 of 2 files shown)


src/js/view/view.js
```diff
@@ -409,6 +411,12 @@ define([
             this.checkResized();
         };
 
+        function itemReady() {
+            var title = _model.get('playlistItem').title || '';
```
johnBartos: `const`

src/js/view/view.js
```diff
@@ -409,6 +411,12 @@ define([
...
+            var videotag = _videoLayer.querySelector('video, audio');
```
johnBartos: What happens if the player is not viewable?

karimJWP: @johnBartos title is still set. Checked when player does not appear on page and when item is loaded when player is in background tab

karimJWP:
@johnBartos all comments addressed

PR #2018, opened 2017-04-26
Set outline-color on dock buttons to remove focus ring

robwalch:

Adds the `. jw-button-color` class on dock button icons. This class is added to all icon buttons for custom foreground colors and resetting some default browser styles we do not want.

`outline-color` needs to be set on elements with `tabindex` so that focus rings are not drawn when clicking on buttons. The `. jw-button-color` class does this making the dock button icons more consistent with other UI buttons.

No

no

#### Addresses Issue(s):

None


hghazzi:
Do we want to add `jw-button-color` to https://github.com/jwplayer/jwplayer/blob/master/src/js/utils/aria.js#L11?

PR #2026, opened 2017-04-28
Add option to not pause when instream init

pajong:
Add a flag to not pause the video tag on instream init.

Googima iOS pauses the ad when the video tag is paused on ad init,
because the same video tag is used to play the ad and the content.

The same code was there but reverted after a fix, but a change from IMA SDK may have caused the problem to occur again.
The commit that reverted the change: https://github.com/jwplayer/jwplayer/commit/786dce1831810f24557831855eea04bfd86ca881

https://github.com/jwplayer/jwplayer-ads-googima/pull/211

#### Addresses Issue(s):

ADS-187

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -73,7 +73,7 @@ define([
 
         this.type = 'instream';
 
-        this.init = function() {
+        this.init = function(sharedVideoTag) {
 
             // Keep track of the original player state
             _oldProvider = _model.getVideo();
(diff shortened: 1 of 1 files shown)


egreaves:
@pajong please add the appropriate milestone to this PR

src/js/controller/instream-adapter.js
```diff
@@ -73,7 +73,7 @@ define([
 
         this.type = 'instream';
 
-        this.init = function() {
+        this.init = function(sharedVideoTag) {
```
johnBartos: Can we do this without a flag parameter?

johnBartos: @pajong

pajong: No we cannot

robwalch:
@pajong https://github.com/jwplayer/jwplayer-ads-googima/pull/211 was closed but this PR was not. Is there a new IMA PR?

pajong:
@robwalch https://github.com/jwplayer/jwplayer-ads-googima/blob/master/src/PlaylistItemManager.js#L202
That was already merged by Max

johnBartos:
@pajong Is this still 7.11?

PR #2030, opened 2017-05-01
Split breakpoint get/set logic into separate functions & dispatch a breakpoint event separate of updating the class

johnBartos:
1) Split the getting & setting of breakpoints into separate functions
2) Dispatches a breakpoint event when the player is resized
3) Sets the breakpoint class only when updating container styles

Plugins previously relied on the assumption that when a resize event is received, a breakpoint will set on the player className. This assumption is not true anymore with viewability.

I'd like @robwalch to check that my breakpoints are being dispatched & set in the correct handlers. 

https://github.com/jwplayer/jwplayer-plugin-related/pull/139

#### Addresses Issue(s):

JW7-4311

### DIFF
--- a/src/js/view/utils/breakpoint.js
+++ b/src/js/view/utils/breakpoint.js
@@ -1,4 +1,4 @@
-const utils = require('utils/helpers');
+const domUtils = require('utils/dom');
 
 export function getBreakpoint(width) {
     let breakpoint = 0;
@@ -24,5 +24,5 @@ export function getBreakpoint(width) {
 
 export function setBreakpoint(playerElement, breakpointNumber) {
     const breakpointClass = 'jw-breakpoint-' + breakpointNumber;
-    utils.replaceClass(playerElement, /jw-breakpoint-\d+/, breakpointClass);
+    domUtils.replaceClass(playerElement, /jw-breakpoint-\d+/, breakpointClass);
 }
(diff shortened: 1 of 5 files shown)


src/js/view/utils/breakpoint.js
```diff
@@ -1,31 +1,28 @@
-define([
-    'utils/helpers'
-], function (utils) {
-    return function setBreakpoint(playerElement, playerWidth, playerHeight) {
-        var width = playerWidth;
-        var height = playerHeight;
+const utils = require('utils/helpers');
```
robwalch: `require('utils/dom');` dom utils is more specific than helpers which is really only meant to bundle utils into publicly available `jwplayer.utils` and to reduce imports in large modules.

src/js/view/view.js
```diff
@@ -150,6 +150,7 @@ define([
                     width: containerWidth,
                     height: containerHeight
                 });
+                _this.trigger(events.JWPLAYER_BREAKPOINT,{ breakpoint: getBreakpoint(containerWidth) });
```
robwalch: Odd that the linter doesn't complain about this - no space after comma, no wrapping object literal.

johnBartos: Yeah that's weird, I'll see if I can get a rule for that

johnBartos:
Should we only fire breakpoint if the breakpoint has changed?

PR #2042, opened 2017-05-04
Use DOM utils to modify CSS classes

robwalch:

Uses our DOM utils to add and remove classes from cations renderer elements after setup.

Setting `element.className` to it's own value still marks the element for redraw. Our utils check this before setting className to avoid unnecessary repaints or layout invalidation.

@egreaves shouldn't we remove ".jw-captions-window-active" when there are no captions?


egreaves:
@robwalch yes, we should.

PR #2048, opened 2017-05-08
Don't attempt to set video tag title when using flash

karimJWP:
It avoids setting the video tag title when using the flash provider

The 'Add title attribute to <video> tag' PR https://github.com/jwplayer/jwplayer/pull/2013 broke the flash provider

Would it be best to check for flash before adding the event listener ? (on itemReady)

n/a

#### Addresses Issue(s):
Enables title to be displayed in the iOS lock screen, instead of video's URL (without breaking flash)
JW7-3410

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -416,6 +416,7 @@ define([
         };
 
         function itemReady() {
+            const provider = _model.getVideo();
             if (provider && provider.getName().name.indexOf('flash') === 0) {
                 return;
             }
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -416,6 +416,9 @@ define([
         };
 
         function itemReady() {
+            if (provider && provider.getName().name.indexOf('flash') === 0) {
+                return;
+            }
             const title = _model.get('playlistItem').title || '';
             var videotag = _videoLayer.querySelector('video, audio');
             videotag.setAttribute('title', title);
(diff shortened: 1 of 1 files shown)


karimJWP:
@msiddique can we run the flash tests before even reviewing ?

johnBartos:
The `getName` function just returns `flash`:[

````
    FlashProvider.getName = function() {
        return { name: 'flash' };
    };
````

You can simplify your check

**Comment by karimJWP** [comment]
@johnBartos you mean instead of checking for the indexOf('flash') ? I thought about that too but i noticed there are other methods in view.js that check for flash using the indexOf; I assume there might be (or might have been) a reason for that ; if you confirm it's safe i'll remove it

**Comment by johnBartos** [comment]
I don't see what could go wrong.

File: src/js/view/view.js [hunk 1]
```diff
@@ -415,8 +415,12 @@ define([
             this.checkResized();
         };
 
-        function itemReady() {
-            const title = _model.get('playlistItem').title || '';
+        function itemReady(item) {
+            const provider = _model.getVideo();
+            if (provider && provider.getName().name.indexOf('flash') === 0) {
```
egreaves: @karimJWP to @johnBartos' point, you can just compare `provider.getName().name === 'flash'` here. I'll merge once that's done.

karimJWP:
@egreaves @johnBartos  comments addressed

egreaves:
@johnBartos for HLS playback in flash, `getProvider()` returns `flash_adaptive`, so an equality check would fail.

PR #2050, opened 2017-05-08
Fix VTT captiosn disappearing in Shaka on replay

karimJWP:
Checks for tracks before obtaining the length

Prevents Commercial PR https://github.com/jwplayer/jwplayer-commercial/pull/3434 from failing unit tests

n/a

https://github.com/jwplayer/jwplayer-commercial/pull/3434

#### Addresses Issue(s):
dash vtt disappear on replay
JW7-4298

### DIFF
--- a/src/js/providers/tracks-mixin.js
+++ b/src/js/providers/tracks-mixin.js
@@ -503,7 +503,7 @@ define(['utils/underscore',
     }
 
     function _removeCues(renderNatively, tracks) {
-        if (tracks.length) {
+        if (tracks && tracks.length) {
             _.each(tracks, function(track) {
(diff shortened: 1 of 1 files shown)


johnBartos:
@karimJWP Does this actually do anything? `_.each` does nothing if the array is empty; if this is to prevent non-array objects from passing through, use a more explicit function like `_.isArray`

karimJWP:
@johnBartos it's to prevent unit test crash where the method is called with null tracks. It checks for an instantiated object before calling its length. I've seen it done almost everywhere where tracks.length is called

johnBartos:
@karimJWP Oops, misread the diff. LGTM

PR #2052, opened 2017-05-09
Prevents setting video tag title in incompatible Providers, and handles double encoded titles

karimJWP:
Checks that the provider is not youtube or chromecast when setting title, and handles double encoded titles.

Player will crash if an attempt to set a video tag title is done on providers that do not support it, and the iOS lock screen will display double encoded symbols if the title is not converted.

is this the best way to decode double encoded strings? I mimicked what was done in title.js

n/a

#### Addresses Issue(s):

JW7-3410

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -425,7 +425,8 @@ define([
             }
             const title = item.title || '';
             var videotag = _videoLayer.querySelector('video, audio');
-            videotag.setAttribute('title', title);
+            videotag.innerHTML = title;
+            videotag.setAttribute('title', videotag.innerHTML);
         }
 
         function redraw(model, visibility, lastVisibility) {
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -417,15 +417,15 @@ define([
 
         function itemReady(item) {
             const provider = _model.getVideo();
+            var videotag = _videoLayer.querySelector('video, audio');
             // Youtube, chromecast and flash providers do no support video tags
-            if (provider && /^(youtube|chromecast|flash)/.test(provider.getName().name)) {
+            if (!videotag) {
                 return;
             }
-            var videotag = _videoLayer.querySelector('video, audio');
             const dummyDiv = document.createElement('DIV');
             // Writing a string to innerHTML completely decodes multiple-encoded strings
             dummyDiv.innerHTML = item.title || '';
-            videotag.setAttribute('title', dummyDiv.innerHTML);
+            videotag.setAttribute('title', dummyDiv.textContent);
         }
 
         function redraw(model, visibility, lastVisibility) {
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -416,7 +416,6 @@ define([
         };
 
         function itemReady(item) {
-            const provider = _model.getVideo();
             var videotag = _videoLayer.querySelector('video, audio');
             // Youtube, chromecast and flash providers do no support video tags
             if (!videotag) {
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -417,13 +417,15 @@ define([
 
         function itemReady(item) {
             const provider = _model.getVideo();
+            // Youtube, chromecast and flash providers do no support video tags
             if (provider && /^(youtube|chromecast|flash)/.test(provider.getName().name)) {
                 return;
             }
             const title = item.title || '';
             var videotag = _videoLayer.querySelector('video, audio');
-            videotag.innerHTML = title;
-            videotag.setAttribute('title', videotag.innerHTML);
+            const dummyDiv = document.createElement('DIV');
+            dummyDiv.innerHTML = title;
+            videotag.setAttribute('title', dummyDiv.innerHTML);
         }
 
         function redraw(model, visibility, lastVisibility) {
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -423,7 +423,7 @@ define([
             }
             var videotag = _videoLayer.querySelector('video, audio');
             const dummyDiv = document.createElement('DIV');
-            // The item's title is set to the div's innerHTML to decrypt strings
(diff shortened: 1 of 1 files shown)


johnBartos:
Does setting the innerHTML of a video tag have side-effects on any browser?

egreaves:
@johnBartos @karimJWP though I can't think of a scenario where the video tag has child elements that this code would essentially replace, we should use a dummy div so as not to change the video's `innerHTML`.

src/js/view/view.js
```diff
@@ -417,12 +417,16 @@ define([
 
         function itemReady(item) {
             const provider = _model.getVideo();
-            if (provider && provider.getName().name.indexOf('flash') === 0) {
+            if (provider &&
```
egreaves: `/^(youtube|chromecast|flash)/.test(provider.getName().name)`

johnBartos: While we're at it, drop a comment explaining why these providers are excluded, or put this logic in a function (with a good name, of course)

src/js/view/view.js
```diff
@@ -417,12 +417,16 @@ define([
...
+                (provider.getName().name === 'youtube' ||
+                 provider.getName().name === 'chromecast' ||
+                 provider.getName().name.indexOf('flash') === 0 )) {
                 return;
             }
             const title = item.title || '';
             var videotag = _videoLayer.querySelector('video, audio');
-            videotag.setAttribute('title', title);
+            videotag.innerHTML = title;
```
egreaves: @johnBartos @karimJWP though I can't think of a scenario where the video tag has child elements that this code would essentially replace, we should use a dummy div so as not to change the video's `innerHTML`.

src/js/view/view.js
```diff
@@ -417,12 +417,15 @@ define([
...
             const provider = _model.getVideo();
-            if (provider && provider.getName().name.indexOf('flash') === 0) {
+            // Youtube, chromecast and flash providers do no support video tags
+            if (provider && /^(youtube|chromecast|flash)/.test(provider.getName().name)) {
                 return;
             }
             const title = item.title || '';
             var videotag = _videoLayer.querySelector('video, audio');
-            videotag.setAttribute('title', title);
+            const dummyDiv = document.createElement('DIV');
```
johnBartos: Can you leave a comment explaining why you need to `innerHTML` this?

karimJWP:
@egreaves @johnBartos  comments addressed

johnBartos:
https://connect.microsoft.com/IE/feedback/details/743819

Writing to `innerHTML` is different per browser; IE isn't completely spec compliant. However, I believe this bug has no cosmetic effect.

src/js/view/view.js
```diff
@@ -417,12 +417,15 @@ define([
 
         function itemReady(item) {
             const provider = _model.getVideo();
-            if (provider && provider.getName().name.indexOf('flash') === 0) {
+            // Youtube, chromecast and flash providers do no support video tags
+            if (provider && /^(youtube|chromecast|flash)/.test(provider.getName().name)) {
```
robwalch: You know, we could just return if the querySelector comes back with nothing `if (!videotag) { return }`

robwalch: That would be better

src/js/view/view.js
```diff
@@ -417,12 +417,15 @@ define([
...
                 return;
             }
-            const title = item.title || '';
             var videotag = _videoLayer.querySelector('video, audio');
-            videotag.setAttribute('title', title);
+            const dummyDiv = document.createElement('DIV');
```
robwalch: It sounds stingy, but do we really want to create a new element to decode titles for every item that plays? This element could be a static element used by all players for all string decoding.

robwalch: Consider reading `textContent` rather than `innerHTML `.

src/js/view/view.js
```diff
@@ -417,12 +417,15 @@ define([
...
+            // Writing a string to innerHTML completely decodes multiple-encoded strings
+            dummyDiv.innerHTML = item.title || '';
+            videotag.setAttribute('title', dummyDiv.innerHTML);
```
robwalch: Consider reading `textContent` rather than `innerHTML `.

src/js/view/view.js
karimJWP: that's a cool idea @robwalch , should i tackle it now or add it as a tech debt item ?

robwalch: Up to you.

src/js/view/view.js
egreaves: @robwalch @karimJWP `textContent` does not decode html encoded strings. Also, setting `innerHTML` then reading `innerText` ensures <, >, ", ' and ` get decoded correctly.

robwalch: Right. I'm suggesting that you should write innerHTML (which you are) and then read textContent for the value you set on the the video tag title. I'm not sure about innerText support in IE9.

src/js/view/view.js
karimJWP: @robwalch I'll exclude it for now, and bring it up for Today's tech debt meeting. There's probably more dummy divs across the player, it's worth seeing how feasible it is to change those to all use the same static element used by all players, but not sure how risky this is

src/js/view/view.js
egreaves: FYI @karimJWP - both `innerText` and `textContent` are supported in IE9+, but reading `textContent` is way more performant.

karimJWP:
@robwalch @egreaves all comments addressed

src/js/view/view.js
```diff
@@ -417,12 +417,15 @@ define([
 
         function itemReady(item) {
             const provider = _model.getVideo();
```
egreaves: Since we're checking for the presence of a videotag with `_videoLayer`, we no longer need the provider in this function.

karimJWP:
test this please

karimJWP:
@egreaves tests pass and all comments were addressed

PR #2055, opened 2017-05-10
Playback Rate API and Per Provider Playback Rate Management

jw-kaurand:
This PR adds API properties to manage setting the playback rate in the model and for providers.

This is needed because different providers either handle playback rate change differently or do not support them.  The API changes allows a developer to modify the playback rate, but also prevents the playback rate from being increase for providers which should not be allowed to do so.  When setting the playback rate in a provider, we also receive a value returned from the provider which is the rate that it will agree to play back at, which simplifies managing this value in the model.

Switching between providers, instream providers, and is the and between playlist items are the points that need the most attention.

There are other changes in the ads and commercial  projects related to this one.

#### Addresses Issue(s):
JW7-4319


src/js/api/callbacks-deprecate.js
```diff
@@ -11,6 +11,7 @@ define([
             onFullscreen: events.JWPLAYER_FULLSCREEN,
             onMeta: events.JWPLAYER_MEDIA_META,
             onMute: events.JWPLAYER_MEDIA_MUTE,
+            onPlaybackRateChange: events.JWPLAYER_PLAYBACK_RATE_CHANGED,
```
robwalch: This is for backwards support only. Do not add new methods to the list of ones we intend to deprecate.

src/js/controller/model.js
```diff
@@ -381,6 +388,19 @@ define([
             }
         };
 
+
+        this.setPlaybackRate = function(playbackRate) {
+            if (!_provider || !_.isNumber(playbackRate) || this.get('streamType') === 'LIVE' || !_attached) {
```
robwalch: We shouldn't need a provider to set the rate on the model. The player's rate should be set and providers should be updated when needed.

src/js/events/events.js
```diff
@@ -70,6 +70,7 @@ define([], function() {
         JWPLAYER_PLAYLIST_LOADED: 'playlist',
         JWPLAYER_AUDIO_TRACKS: 'audioTracks',
         JWPLAYER_AUDIO_TRACK_CHANGED: 'audioTrackChanged',
+        JWPLAYER_PLAYBACK_RATE_CHANGED: 'playbackRateChanged',
```
robwalch: I would like us to start replacing these with ES6 constants and state the intent to deprecate `jwplayer.events`.

test/data/api-methods.js
```diff
@@ -111,6 +113,7 @@ define({
     onMute: null,
     onPause: null,
     onPlay: null,
+    onPlaybackRateChange: null,
```
robwalch: remove

src/js/controller/model.js
```diff
@@ -256,6 +260,9 @@ define([
                 this.playbackComplete();
             }
 
+            // Restore the playback rate to the provider in case it changed while detached and we reused a video tag.
+            this.setPlaybackRate(this.get('playbackRate'));
```
robwalch: I agree that `model.playbackRate` should reflect the current playback rate of the player, but when the API sets playbackRate, we should set `defaultPlaybackRate` as the desired playbackRate, and then allow the providers and media controller logic in the model do the rest.

See https://developer.mozilla.org/en-US/Apps/Fundamentals/Audio_and_video_delivery/WebAudio_playbackRate_explained

test/data/api-methods.js
jw-kaurand: Removing

src/js/controller/model.js
jw-kaurand: Yeah, we can attempt to set it on the provider if we have a provider, otherwise just set it up our intended playback rate value in the model if not.

src/js/controller/model.js
jw-kaurand: I would like to maintain the `defaultPlaybackRate` property on the video tag, however that property ends up getting cleared out in a number of circumstances.  For example, DASH really does a number on the video tag's properties since it will reset a lot of them between playlist items. Because of that the provider is maintaining the playback rate property to reset it later.  This sort of logic could be done in provider on attach and detach and this can just query the provider for the playback rate that's expected to be used.

jw-kaurand:
The current state of this PR with Squash tests can be seen here: http://player-pr-test-jenkins.longtailvideo.com/builds/7199/archive/test/squash/  Playback Rate API tests can be found on that page.  For ease of targeting the right player with the "setPlaybackRate()" API call it may be easier to open individual tests in the harness.

src/js/controller/model.js
```diff
@@ -381,6 +388,19 @@ define([
             }
         };
 
+
+        this.setPlaybackRate = function(playbackRate) {
+            if (!_.isNumber(playbackRate) || this.get('streamType') === 'LIVE' || !_attached) {
+                return false;
```
robwalch: Why return a boolean?

src/js/providers/html5.js
```diff
@@ -103,6 +103,7 @@ define([
         var _lastEndOfBuffer = null;
         var _stale = false;
         var _edgeOfLiveStream = false;
+        var _playbackRate = 1;
```
robwalch: Remove

src/js/controller/model.js
```diff
@@ -381,6 +388,19 @@ define([
...
+            }
+
+            // Clamp the rate between 0.25x and 4x speed
+            var clampedRate = Math.max(0.25, Math.min(playbackRate, 4));
+
+            // Providers which support changes in playback rate will return the rate that we changed to
+            this.set('playbackRate', _provider ? _provider.setPlaybackRate(clampedRate) : clampedRate);
```
robwalch: Set "defaultPlaybackRate". Only set "playbackRate" based on default setting when a provider is available or when reattaching the provider.

src/js/controller/model.js
```diff
@@ -256,7 +260,10 @@ define([
                 this.playbackComplete();
             }
 
-            return _provider.attachMedia();
+            _provider.attachMedia();
+
+            // Restore the playback rate to the provider in case it changed while detached and we reused a video tag.
+            this.set('playbackRate', _provider.getPlaybackRate());
```
robwalch: Use "defaultPlaybackRate".

src/js/controller/model.js
```diff
@@ -229,6 +230,9 @@ define([
             // Mute the video if autostarting on mobile. Otherwise, honor the model's mute value
             _provider.mute(this.autoStartOnMobile() || _this.get('mute'));
 
+            // Set the playback rate to be the value that the provider supports and will play at
+            this.set('playbackRate', _provider.setPlaybackRate(_this.get('playbackRate')));
```
robwalch: Use "defaultPlaybackRate".

jw-kaurand:
The current build of this with the commercial version of it can be found here: http://player-pr-test-jenkins.longtailvideo.com/builds/7226/archive/test/harness.html?edition=ads

src/js/controller/model.js
```diff
@@ -229,6 +232,9 @@ define([
             // Mute the video if autostarting on mobile. Otherwise, honor the model's mute value
             _provider.mute(this.autoStartOnMobile() || _this.get('mute'));
 
+            // Set the playback rate to be the value that the provider supports and will play at
+            this.set('playbackRate', _provider.setPlaybackRate(_this.get('defaultPlaybackRate')));
```
egreaves: I don't like the pattern of returning a value from a setter. It's an extra line of code, but with SRP in mind:
```
_provider.setPlaybackRate(_this.get('defaultPlaybackRate'));
this.set('playbackRate', _provider.getPlaybackRate());
```

src/js/controller/model.js
```diff
@@ -381,6 +390,22 @@ define([
...
+                return;
+            }
+
+            // Clamp the rate between 0.25x and 4x speed
+            var clampedRate = Math.max(0.25, Math.min(playbackRate, 4));
+
+            this.set('defaultPlaybackRate', clampedRate);
+            // Providers which support changes in playback rate will return the rate that we changed to
+            if (_provider) {
+                this.set('playbackRate', _provider.setPlaybackRate(clampedRate));
```
egreaves: Same comment as above.

src/js/providers/default.js
```diff
@@ -43,6 +43,13 @@ define([
         getCurrentAudioTrack: noop,
         setCurrentAudioTrack: noop,
 
+        setPlaybackRate: function() {
+            return 1;
```
egreaves: The getter should be used for determining the `playbackRate`.

src/js/controller/model.js
```diff
@@ -229,6 +232,10 @@ define([
             // Mute the video if autostarting on mobile. Otherwise, honor the model's mute value
             _provider.mute(this.autoStartOnMobile() || _this.get('mute'));
 
+            // Set the playback rate to be the value that the provider supports and will play at
+            _provider.setPlaybackRate(_this.get('defaultPlaybackRate'));
```
egreaves: Wouldn't this accomplish the same thing as these 2 lines?
`this.setPlaybackRate(this.get('defaultPlaybackRate'));`

jw-kaurand: Yes, though it would rerun some additional, though cheap, logic.

egreaves:
@robwalch good catch. 
@jw-kaurand this should be good to go once that's addressed. Also, I took the liberty of renaming `speed` to `playbackRate` and updated the ticket.

src/js/events/events.js
robwalch: Let's follow the HTML5 spec, and call this event 'ratechange'.

jw-kaurand:
Updated the HTML5 to send out playbackRateChanged events when the video tag dispatches ratechanged events.  The model now listens for these events and updates the `playbackRate` property when it receives the event.  This keeps the model updated to the current playback rate.  The current version of this PR build with the Commercial version of the playback-rate-api-rebased branch can be found here -
 http://player-pr-test-jenkins.longtailvideo.com/builds/7251/archive/test/squash/

fdambrosio:
Hi, I don't know if this is a related problema, but I've an issue using playbackRate on Chrome 58 & Android 5.1.1. 
You can test this page: http://web-0.radioradicale.it/prova/player7/speed.html 

On Android, when I click the button 'speed 1.5X' the playbackRate value change (see console) but the media doesn't speed but it continue to play with 1.0 rate. 

Android playrate bug is fixed for Chrome > 52

src/js/controller/model.js
```diff
@@ -223,12 +228,20 @@ define([
                 this.set('flashBlocked', false);
             }
 
+            // If there is no current provider, then the video tag might have a preset playback rate...
+            if (!_provider) {
+                this.set('defaultPlaybackRate', _currentProvider.getPlaybackRate());
+            }
```
robwalch: Why does not having `_provider` set mean we should set 'defaultPlaybackRate'?

'defaultPlaybackRate' should be controlled by the user. 'playbackRate' should reflect the provider's rate.

src/js/controller/model.js
```diff
@@ -381,6 +397,23 @@ define([
...
+
+            // Clamp the rate between 0.25x and 4x speed
+            var clampedRate = Math.max(0.25, Math.min(playbackRate, 4));
+
+            this.set('defaultPlaybackRate', clampedRate);
+            // Providers which support changes in playback rate will return the rate that we changed to
+            if (_provider) {
+                _provider.setPlaybackRate(clampedRate);
+                this.set('playbackRate', _provider.getPlaybackRate());
```
robwalch: Let's wait for "ratechange". This update is unnecessary.

package.json
```diff
@@ -35,6 +35,7 @@
     "import-loader": "1.0.1",
     "intersection-observer": "0.2.1",
     "jquery": "3.2.1",
+    "jwplayer-errors": "2.0.0",
```
robwalch: Why are we adding this package in this PR?

jw-kaurand: Removed.  I guess I added --save-dev accidentally.

src/js/controller/model.js
jw-kaurand: Changed so that we set `playbackRate` on the model based on the tag's rate.  `defaultPlaybackRate` is specifically based on user selection/API selection.

jw-kaurand:
@robwalch @egreaves Updated OS the Commercial PRs.  The result should be able to be found in build 7284.

src/js/controller/model.js
```diff
@@ -111,6 +113,7 @@ define([
                             this.playVideo();
                         }, this);
                     }
+                    _provider.setPlaybackRate(this.get('defaultPlaybackRate'));
```
robwalch: You should use `model.setPlaybackRate` in case the streamType has already changed to "LIVE".

src/js/controller/model.js
```diff
@@ -223,12 +228,26 @@ define([
...
 
             // Mute the video if autostarting on mobile. Otherwise, honor the model's mute value
             _provider.mute(this.autoStartOnMobile() || _this.get('mute'));
 
+            if (!presetPlaybackRate) {
+                // Attempt setting the playback rate to be the user selected value
+                this.setPlaybackRate(this.get('defaultPlaybackRate'));
+            } else {
+                // Set to playback rate that was in use before setting up the player.  defaultPlaybackRate is user value
+                this.set('playbackRate', presetPlaybackRate);
```
robwalch: Why do we need this feature? Why not ignore playbackRate before setup.

It seems more likely to prevent a call to `api.setPlaybackRate` from working than anything else.

src/js/controller/model.js
```diff
@@ -381,6 +392,33 @@ define([
...
+            }
+            this.set('streamType', streamType);
+        };
+
+        this.setPlaybackRate = function(playbackRate) {
+            if (!_attached || !_.isNumber(playbackRate)) {
+                return;
+            }
+
+            var clampedRate;
```
egreaves: You could initialize `playbackRate` then check if it's not a live stream.

PR #2058, opened 2017-05-12
Slice language string to a length of 2 when mapping 2-char ISO codes

johnBartos:
Limits ISO language code to a length of 2

Some language codes (in my case, from a DASH stream) come instead as two hyphenated strings, with the first being the actual code:

`pt-br` (Brazilian Portuguese)

All we care about are the first two chars.

Should we map twice, maybe, when the string is hyphenated? 

https://github.com/jwplayer/jwplayer-commercial/pull/3466

#### Addresses Issue(s):
JW7-4313

### DIFF
--- a/src/js/utils/language.js
+++ b/src/js/utils/language.js
@@ -14,11 +14,15 @@ define([
         ja: 'Japanese',
         pt: 'Portuguese',
         ru: 'Russian',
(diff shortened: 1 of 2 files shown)


hghazzi:
want to write some unti tests? :)

PR #2063, opened 2017-05-15
Improve Play Reason Tracking

robwalch:

1. Prevent last play reason from being repeated
2. Set playReason to "external" for API calls that don’t set a reason

Play reasons are not tracked correctly for playlist navigation with related, and any case where a specific reason for seek, next, prev and setting the playlist item is not specified.

It's important that the previous 'playReason' removed when a new play call with no reason is called. This was not happening before:
  
```js
function _play(meta) {
  if (meta) {
    _model.set('playReason', meta.reason);
  }
```

Now we will always update playReason - `undefined` is better than the previous reason which would be incorrect, because we'll track those events as "unknown".

```js
function _play(meta = {}) {
  _model.set('playReason', meta.reason);
```

Not yet. These changes should help support JW7-4270

#### Addresses Issue(s):

JW7-2537

### DIFF
--- a/src/js/api/api-actions.js
+++ b/src/js/api/api-actions.js
@@ -8,17 +8,17 @@ define([
         //   Ultimately they should be moved into this file
         var passthroughs = [
             // 'setup',
+            // 'remove',
             // 'load',
             // 'play',
             // 'pause',
-            // 'remove',
+            // 'playlistNext',
+            // 'playlistPrev',
+            // 'playlistItem',
+            // 'seek',
 
-            'seek',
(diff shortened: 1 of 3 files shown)


donato:
Wooo!

Btw, @robwalch - what if they set a meta object, but it doesn't have a reason attribute? Will ES6 do a merge? Is that even a use-case?

robwalch:
@donato, It won't. It's OK for the reason to be undefined, but it would not be good if a user could trigger an exception by passing in a value this in not an object.

Issue #2061, opened 2017-05-14
Skin CSS files in 7.11 syntax error while building


astierler:
When upgrading to 7.11.0, my webpack build process broke while including stormtrooper.css
(I downloaded the 7.11.0 zip package from my JW Dashboard).

Also noticed this version has sourceMaps at bottom of CSS while previous did not - so not sure if related to a build process gone awry?

      Invalid CSS after "...ax-height:calc(": expected expression (e.g. 1px, bold), was "~'100%' - ~'38px')}"
      in resources/assets/jwplayer/7.11.0/skins/stormtrooper.css (line 1, column 1686)

This is the current syntax (in 7.11.0)

    max-height: calc(~'100%' - ~'38px')

Previous version (7.10.x and before) compile fine

    max-height: calc(100% - 38px)

Not a huge issue as I modified my local CSS file, but just posting this here in case others run into similar issue.

### Environment

- JW Player 7.11.0
- Webpack 2.5.1 (but probably any CSS linter may throw an error)

astierler:
Updating on this.. realizing that pulling in the skin is done remotely anyways (or via `base`) but leaving here in case this syntax error actually should be fixed

jw-kaurand:
@astierler I'll take a look into the issue.  Thanks for the feedback!

jw-kaurand:
@astierler Turns out the issue was related to an upgrade in our build process that was made for v7.11.0.  The fix for this should be merged now, and as a precaution to prevent such issues in the future I have  created a tech debt ticket that will create a process to lint the CSS outputted from our build process to check for these sorts of errors in the future (https://github.com/jwplayer/jwplayer/issues/2067).  This change is currently scheduled to be included in the v7.12.0 release.

Thanks again for your feedback!

PR #2066, opened 2017-05-17
Remove string escapes that causes errors in CSS

jw-kaurand:
This PR removes unnecessary escaping of calc properties in LESS that were causing errors in the CSS.  This was an oversight while switching from RECESS to LESS.

This is necessary because it causes the intended CSS to not be applied and creates errors in the outputted stylesheets.

#### Addresses Issue(s):
https://github.com/jwplayer/jwplayer/issues/2061

### DIFF
--- a/src/css/shared-imports/mixins.less
+++ b/src/css/shared-imports/mixins.less
@@ -26,13 +26,13 @@
         }
 
         .jw-captions {
-            max-height: ~"calc(~'100%' - ~'46px')";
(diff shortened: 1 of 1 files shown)


jw-kaurand:
In order to ensure the quality of the CSS we output.  We should also lint the outputted CSS files as well: https://github.com/jwplayer/jwplayer/issues/2067

jw-kaurand:
retest this please

PR #2068, opened 2017-05-17
Chromecast captions

karimJWP:
Displays CC button while casting and forwards setCaptions method call to chrome cast provider.

To enable toggling captions when casting and to allow dynamic caption styling.

The implementation for forwarding caption styling requests to the chrome cast provider via the model feels hacky

https://github.com/jwplayer/jwplayer-commercial/pull/3497/files

#### Addresses Issue(s):
Allow captions to be displayed with Chromecast
JW7-3381

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -785,9 +785,7 @@ define([
             this.setCaptions = function(options) {
                 const provider = this.getProvider();
                 if (provider.name === 'chromecast') {
-//                    _model.styleCaptions(options);
-                    const castProvider = _video();
-                    castProvider.styleCaptions(options);
+                    _model.styleCaptions(options);
                 }
                 _view.setCaptions(options);
             }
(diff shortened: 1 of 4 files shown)


src/js/controller/controller.js
```diff
@@ -782,7 +782,13 @@ define([
             this.resize = _view.resize;
             this.getSafeRegion = _view.getSafeRegion;
             this.setCues = _view.addCues;
-            this.setCaptions = _view.setCaptions;
+            this.setCaptions = function(stylingOptions) {
+                const provider = this.getProvider();
+                if (provider.name === 'chromecast') {
+                    _model.styleCaptions(stylingOptions);
```
egreaves: The view passes the options to `captionsrenderer.js`, which extends the defaults with the custom options and stores the modified options in the model. (See: https://github.com/jwplayer/jwplayer/blob/master/src/js/view/captionsrenderer.js#L173).

Let's discuss improving this.

PR #2069, opened 2017-05-17
Add "nextShown" and "nextClick" events to the player for next button and tooltip

robwalch:

Fire a "nextShown" event when the next up tooltip is shown. "hover" and "time" reasons are added to calls that toggle on the tooltip.

Fire a "nextClick" event when the next button or next tooltip is clicked.

These events contain data needed to evaluate impressions and clicks of upcoming playlist items and recommendations.

Model `setNextUp` method has been removed because it's not needed. And, a destroy method has been added to the nextup tooltip to clear model listeners registered to it's instance - we should do this for all control components that add listeners to the model OR reuse the same controls instance each time "controls" are toggled on and off.

This PR requires additional data in the `nextUp` object added in https://github.com/jwplayer/jwplayer-plugin-related/pull/147

#### Addresses Issue(s):

JW7-4347 JW7-4357

### DIFF
--- a/src/js/view/controls/nextuptooltip.js
+++ b/src/js/view/controls/nextuptooltip.js
@@ -3,11 +3,14 @@ import nextUpTemplate from 'view/controls/templates/nextup';
 define([
     'utils/dom',
     'utils/ui',
+    'utils/underscore',
+    'utils/backbone.events',
     'utils/helpers',
-], function(dom, UI, utils) {
+], function(dom, UI, _, Events, utils) {
 
     return class NextUpTooltip {
         constructor(_model, _api, playerElement) {
+            _.extend(this, Events);
             this._model = _model;
             this._api = _api;
             this._playerElement = playerElement;
@@ -15,6 +18,7 @@ define([
             this.nextUpClose = _model.get('localization').nextUpClose;
             this.state = 'tooltip';
             this.enabled = false;
+            this.shown = false;
             this.reset();
         }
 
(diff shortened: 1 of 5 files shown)


src/js/view/controls/nextuptooltip.js
```diff
@@ -154,7 +171,7 @@ define([
             const showUntilEnd = val >= this.offset;
             if (showUntilEnd && nextUpSticky === undefined) { // show if nextUpSticky is unset
                 this.nextUpSticky = showUntilEnd;
-                this.toggle(showUntilEnd);
+                this.toggle(showUntilEnd, 'time');
             } else if (!showUntilEnd && nextUpSticky === false) { // reset if there was a backward seek
```
egreaves: This should be `!showUntilEnd && nextUpSticky === true` so that we reset if nextUp should not be displayed because the position is before the nextUpOffset.

src/js/view/controls/nextuptooltip.js
```diff
@@ -71,13 +75,26 @@ define([
...
 
-        toggle(show) {
+        toggle(show, reason) {
             if (!this.enabled) {
                 return;
             }
             dom.toggleClass(this.container, 'jw-nextup-sticky', !!this.nextUpSticky);
-            dom.toggleClass(this.container, 'jw-nextup-container-visible', show);
-            dom.toggleClass(this._playerElement, 'jw-flag-nextup', show);
+            if (this.shown !== show) {
```
johnBartos: Shouldn't we only enter this conditional if we have nextUp too? `if (this.shown !== show && this.nextUp)`

robwalch: We need to enter this conditional to toggle CSS classes off even when nextUp is undefined.

robwalch: If we are showing the UI and don't have nextUp, that is an error somewhere else in the code. The check for nextUp below, is stickly defensive, and intended to avoid from throwing an exception if that occurs.

PR #2076, opened 2017-05-24
Playback Rate UI

jw-kaurand:
This PR adds in a UI for selecting playback rates for providers which support this option.

This is needed because it exposes the playback rate API functionality to the user via a menu.  It allows selection of either custom rates defined by the publisher to a set of standard rates available when setting up the player with it as an option.

I have some concerns about the naming conventions involved with the SelectionDisplayMenu, especially its name.  The selection display menu may also be too narrowly tailored to the desired functionality, so perhaps looking at it to make it more simple or generic could help.  

There will be one in commercial to match this.  It will also have some tests related to this functionality.

#### Addresses Issue(s):

JW7-4038


src/js/api/config.js
```diff
@@ -105,6 +109,33 @@ define([
             config.skin = config.skin.replace('.xml', '');
         }
 
+        if (config.playbackRateControls) {
+            var playbackRateOptions;
+
+            var validRatesFilter = (value) => { return _.isNumber(value) && value >= 0.25 && value <= 4; };
+
+            if (_.isBoolean(config.playbackRateControls)) {
+                playbackRateOptions = [0.5, 1, 1.25, 1.5, 2];
```
egreaves: Initialize `playbackRateOptions` to the defaults. That way, you only need to check `_.isArray(config.playbackRateControls)`.

src/js/api/config.js
```diff
@@ -105,6 +109,33 @@ define([
...
+            } else if (_.isArray(config.playbackRateControls)) {
+                var filteredResults = config.playbackRateControls.filter(validRatesFilter);
+                playbackRateOptions = filteredResults.length ? filteredResults : false;
+            }
+
+            if (playbackRateOptions) {
```
egreaves: `playbackRateOptions` should be defined if `config.playbackRateControls` is set, so you don't need these conditions. I also don't understand why you'd reassign `config.playbackRateControls` AND set `config.playbackRateOptions`.

egreaves:
@jw-kaurand Either mutate playbackRateControls or use playbackRateOptions (if you want to keep track of what was in the config) for what goes into the menu, but there isn't a need for both to end up with the same array of options.

jw-kaurand:
A build of this in conjunction with the Commercial version of this can be found here: http://player-pr-test-jenkins.longtailvideo.com/builds/7425/archive/test/squash/

jw-kaurand:
@egreaves `playbackRateControls` and `playbackRateOptions` aren't necessarily always going to be the same. `playbackRateOptions` is the value that is used to populate the playback rate menu, so it will potentially fluctuate between playlist items which do not support playback rate (LIVE streams, unsupported providers, etc) while `playbackRateControls` should always be what we reset `playbackRateOptions` back to when in situations where playback rate is supported.

egreaves:
@jw-kaurand still seems like they're sharing responsibility. Use one property for keeping track of the menu and something else to determine if the menu should be visible or not based on the stream and provider.

src/js/controller/model.js
```diff
@@ -236,6 +234,7 @@ define([
 
             // Attempt setting the playback rate to be the user selected value
             this.setPlaybackRate(this.get('defaultPlaybackRate'));
+            this.setPlaybackRatesAvailable();
```
egreaves: The view should control whether the playbackRate controls are shown, not the model.

src/js/controller/model.js
```diff
@@ -417,6 +417,15 @@ define([
             }
         };
 
+        this.setPlaybackRatesAvailable = function() {
+            var supportsPlaybackRates =
+                this.get('playbackRateControls') &&
+                _provider.supportsPlaybackRate &&
+                this.get('streamType') !== 'LIVE';
+
+            this.set('playbackRatesAvailable', supportsPlaybackRates);
```
egreaves: This should be done in the view and doesn't require another model property.

src/js/view/controls/controlbar.js
```diff
@@ -145,6 +149,11 @@ define([
                 hd: menu('jw-icon-hd', this._localization.hd),
                 cc: menu('jw-icon-cc', this._localization.cc),
                 audiotracks: menu('jw-icon-audio-tracks', this._localization.audioTracks),
+                playbackrates: new SelectionDisplayMenu(
+                    'jw-icon-playback-rate',
+                    this._localization.playbackRates,
```
egreaves: Is this the right property name?

jw-kaurand: Yup.  Matches the localization key.

jw-kaurand:
The most recent build with the commercial version should be build number 7457.

jw-kaurand:
test this please

jw-kaurand:
The most recent build of this branch with the related commercial version should be available in build number 7506.

PR #2078, opened 2017-05-24
Add "viewable" event to the player API

robwalch:

Adds the "viewable" event to our public API. The event fires when:

- The active state of the browser tab containing the player changes, affecting a player's visibility (document "visibilitychange" event or `document.hidden`)
- %50 of the player area enters or exits the browser viewport

The event contains a `viewable` property that matches the value that would returned by `getViewable()`:

```js
{
    type: "viewable",
    viewable: 1 or 0
}
```

This allows developer to listen for a viewable change. Currently this is only available in the public API via `jwplayer().getViewable()`. This can be used to determine that the player is viewable to the user - for example: by scrolling into view or by switching browser tabs.

#### Addresses Issue(s):

JW7-4371

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -262,7 +262,7 @@ define([
                 }
 
                 _checkAutoStart();
-                _model.on('change:viewable', _checkPlayOnViewable);
+                _model.change('viewable', viewableChange);
             }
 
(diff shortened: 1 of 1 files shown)


egreaves:
Test failures are unrelated.

PR #2079, opened 2017-05-25
Use _provider instead of _currentProvider; delete _currentProvider

johnBartos:
Replaces all instances of `_currentProvider` with `_provider` and deletes `_currentProvider`

The two variables point to the same thing, and having two variables pointing to the same thing is confusing

Does `_currentProvider` have a legitimate use case?

No

#### Addresses Issue(s):
Tech debt

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -16,7 +16,6 @@ define([
         var _provider;
         var _beforecompleted = false;
         var _attached = true;
-        var _currentProvider = utils.noop;
 
         this.mediaController = _.extend({}, Events);
         this.mediaModel = new MediaModel();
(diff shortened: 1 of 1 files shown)


src/js/controller/model.js
```diff
@@ -207,28 +206,27 @@ define([
...
+            this.set('provider', _provider.getName());
 
-            if (_currentProvider.getName().name.indexOf('flash') === -1) {
+            if (_provider.getName().name.indexOf('flash') === -1) {
                 this.set('flashThrottle', undefined);
                 this.set('flashBlocked', false);
             }
 
-            _provider = _currentProvider;
+            _provider = _provider;
```
egreaves: Don't need this anymore

src/js/controller/model.js
```diff
@@ -207,28 +206,27 @@ define([
             }
 
             if (!Provider) {
-                _provider = _currentProvider = Provider;
```
egreaves: Still need `_provider = Provider;`

PR #2080, opened 2017-05-27
Simplify CSS rules for sharing tooltip

egreaves:
This simplifies the css rules that show the controls and title when the sharing tooltip is showing, all while maintaining the same level of specificity. It also hides the focus ring when `jw-reset` is applied to clickable elements.
To reduce css output and to ensure close buttons don't have focus rings.
No.
No.

### DIFF
--- a/src/css/jwplayer/imports/reset.less
+++ b/src/css/jwplayer/imports/reset.less
@@ -17,6 +17,7 @@
     font-stretch: inherit;
     /* non-standard prefixed style used in Chrome, Safari, and some MS browsers.  Not Autoprefixed */
     -webkit-tap-highlight-color: rgba(255, 255, 255, 0);
+    
     &:focus {
         outline: none;
     }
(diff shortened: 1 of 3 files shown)


src/css/jwplayer/imports/reset.less
```diff
@@ -17,4 +17,7 @@
     font-stretch: inherit;
     /* non-standard prefixed style used in Chrome, Safari, and some MS browsers.  Not Autoprefixed */
     -webkit-tap-highlight-color: rgba(255, 255, 255, 0);
```
robwalch: build failed. stylelint expects a new line here, before `&:focus {`.
```
> jwplayer@7.11.1 lint:styles /var/lib/jenkins-slave/workspace/jw7-opensource
> stylelint './src/css/**/*.less'


src/css/jwplayer/imports/reset.less
 20:5  ✖  Expected empty line before rule   rule-empty-line-before
```

src/css/jwplayer/imports/reset.less
```diff
@@ -17,4 +17,8 @@
...
-}
\ No newline at end of file
+    
+    &:focus {
+        outline: none;
```
robwalch: .jw-button-color and .jw-toggle do this (and also control color).

egreaves: I meant to ask you about that. The better solution here is to add `jw-button-color` in sharing so both focus and hover states are fixed along with the skin color.

PR #2083, opened 2017-05-31
Exit fullscreen on IOS before firing playlist complete 

johnBartos:
Exits fullscreen on IOS before firing PLAYLIST_COMPLETE

IOS uses it's native rendering for videos in fullscreen and cannot show HTML overlays. For a better UX we want to exit fullscreen so that our overlays are displayed.

No

No

#### Addresses Issue(s):

JW7-4377


johnBartos:
This isn't working properly; do not merge.

PR #2088, opened 2017-06-01
Fix "playbackRateChange" events dispatch issues

jw-kaurand:
This PR fixes issues where the "playbackRateChange" event is triggered too often or sometimes not at all.

This PR is needed because it normalizes some situations where the provider does not have enough knowledge to tell the model that the playback rate has changed

A commercial version of the PR also exists here: https://github.com/jwplayer/jwplayer-commercial/pull/3564

#### Addresses Issue(s):
JW7-4386 JW7-4387

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -80,9 +80,9 @@ define([
                     return;
                 case 'ratechange':
                     var rate = data.playbackRate;
-                    // Check if its a valid rate.  Shaka changes rate to 0 when pause or buffering
-                    if (rate >= 0.25 && rate <= 4) {
-                        this.set('playbackRate', data.playbackRate);
+                    // Check if its a generally usable rate.  Shaka changes rate to 0 when pause or buffering.
+                    if (rate > 0) {
+                        this.set('playbackRate', rate);
                     }
                     return;
                 case events.JWPLAYER_MEDIA_TYPE:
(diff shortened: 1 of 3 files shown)


jw-kaurand:
Build 7554 uses this and the commercial version of this PR for testing purposes.

src/js/view/controls/components/selection-display-menu.js
```diff
@@ -29,16 +29,15 @@ define([
         selectItem(selectedIndex) {
             super.selectItem(selectedIndex);
 
-            if (selectedIndex !== -1) {
-                var isDefaultOption = selectedIndex === this.defaultIndex;
+            // Show the default icon if we have the default value selected, or one that isn't in the menu
+            var isDefaultOption = selectedIndex === this.defaultIndex || selectedIndex === -1;
```
egreaves: Let's write this so the condition answers the question: should we show a text label

src/js/controller/model.js
```diff
@@ -230,10 +236,15 @@ define([
...
             _provider.mute(this.autoStartOnMobile() || _this.get('mute'));
 
+            _provider.on('all', _videoEventHandler, this);
+
             // Attempt setting the playback rate to be the user selected value
             this.setPlaybackRate(this.get('defaultPlaybackRate'));
 
-            _provider.on('all', _videoEventHandler, this);
+            // Trigger ratechange because the provider may already have the rate in the above setPlaybackRate()
+            if (_provider.supportsPlaybackRate !== previousProviderPlaybackRateSupported) {
```
egreaves: Let's always set the `playbackRate` with a comment so we know why it's necessary.

PR #2091, opened 2017-06-05
Updates the playback rate icon so its smaller and more vertically centered

jw-kaurand:
This PR updates the playback rate icon so its smaller and more centered.

This PR is needed because making the icon more aligned would look better.

JW7-4393


dannyfinks:
test this please

PR #2094, opened 2017-06-06
Trigger subtitlesChanged in non-native mode; reset _id when reusing a track

johnBartos:
### This PR will...
- Trigger `subtitlesTracks` from the mixin when rendering natively
- Reset track `_id` when reusing a track

Shaka assumes a mapping based on track order when mapping between it's internal tracks and our internal tracks. When moving between playlist items, if we reuse a track, then this assumption is broken - resetting `_id` forces it to be remade with the current track ordering.

If we do not trigger `subtitlesTracks` immediately when rendering natively, we may add cues to the wrong track because we have not yet selected the correct one for the current playlist. 

I tested the following cases:
- Changing tracks
- Preroll
- Midroll
- Changing to a playlist item with different tracks
- Changing back to the first playlist item after playing a second with different tracks

No

#### Addresses Issue(s):

JW7-4313

### DIFF
--- a/src/js/providers/tracks-mixin.js
+++ b/src/js/providers/tracks-mixin.js
@@ -452,8 +452,7 @@ define(['utils/underscore',
             }
         }
 
-        // We can setup the captions menu now since we're not rendering textTracks natively
-        if (!this.renderNatively && this._textTracks && this._textTracks.length) {
+        if (this._textTracks && this._textTracks.length) {
             this.trigger('subtitlesTracks', { tracks: this._textTracks });
         }
     }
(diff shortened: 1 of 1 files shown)


egreaves:
@dannyfinks I'd recommend a full regression on the captions suite because this change.

dannyfinks:
thanks for the heads up...should we perhaps run all our automated captions tests on some browsers before we merge this one in? I can kick that off tonight

egreaves:
@dannyfinks yeah. The kind of side effects I'm thinking of should be caught by the automated tests.

PR #2096, opened 2017-06-07
Chromecast improvements

karimJWP:
### This PR will...
prevent captions from displaying on player while casting

ChromeCast Captions ticket was rejected because captions would display on both video and cast screen

n/a

n/a

#### Addresses Issue(s):

JW7-3381

### DIFF
--- a/src/css/controls/flags/cast.less
+++ b/src/css/controls/flags/cast.less
@@ -9,9 +9,10 @@
         display: none;
     }
 
-    .jw-icon-audio-tracks,
     .jw-icon-hd,
-    .jw-icon-fullscreen {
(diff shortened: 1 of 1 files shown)


src/css/controls/flags/cast.less
```diff
@@ -9,9 +9,10 @@
         display: none;
     }
 
-    .jw-icon-audio-tracks,
     .jw-icon-hd,
-    .jw-icon-fullscreen {
+    .jw-captions,
```
egreaves: @karimJWP this will hide captions rendered by the player. Have you confirmed that browser-rendered captions don't display when we're casting?

PR #2102, opened 2017-06-09
Upgrade to Webpack2

johnBartos:
### This PR will...
Upgrade our build process to use Webpack2 🎉
(also removes lodash and refactors the config in es6)

Webpack1 is deprecated; additionally, we can take advantages of new features which could save us time/size

A passover of the config would be nice

https://github.com/jwplayer/jwplayer-commercial/pull/3604

#### Addresses Issue(s):
OKR


package.json
```diff
@@ -64,7 +63,7 @@
     "source-map": "0.5.6",
     "underscore": "1.8.3",
     "url-loader": "0.5.8",
-    "webpack": "1.13.1"
+    "webpack": "^2.6.1"
```
robwalch: Can we stick to a fixed version and update manually?

johnBartos: The package lock should do this for us

robwalch: Great. Thanks.

Gruntfile.js
```diff
@@ -275,24 +275,18 @@ module.exports = function(grunt) {
     });
```
robwalch: You need to include "strict mode" at the top of ES6 script that will run on node in jenkins. The build is failing because of this:

```
Loading "Gruntfile.js" tasks...ERROR
>> SyntaxError: Block-scoped declarations (let, const, function, class) not yet supported outside strict mode
```

After that make sure grunt on Jenkins will run it. You may need to stick to ES5 for this file unless we can get nvm updated.

hghazzi:
test this please

Gruntfile.js
johnBartos: Addressed

karma.conf.js
```diff
@@ -1,8 +1,7 @@
 /* eslint no-process-env: 0 */
 const webpack = require('webpack');
 const path = require('path');
-// Use the debug config
-const webpackConfig = require('./webpack.config.js')[0];
+const webpackConfig = require('./webpack.config.js')({ debug: true });
```
robwalch: If we test release files, we'll catch issues introduced by transpilers.

PR #2109, opened 2017-06-16
Preloading: Default to 'metadata'

hghazzi:
### This PR will...

set the default value of 'preload' to be 'metadata'

Unless the config explicitly is set to preload='none' or preload='auto', we should always preload the manifest

#### Addresses Issue(s):

JW8-68


dannyfinks:
@hghazzi @egreaves should we create a jw8 branch?

src/js/api/config.js
```diff
@@ -161,11 +161,21 @@ define([
...
 
         return config;
     };
 
+    function normalizePreload(config) {
+        const preloadValues = ['none', 'meta', 'auto'];
+        const preload = config.playlist[0].preload || config.preload;
+
+        return _.contains(preloadValues, preload) ? preload : 'meta';
+    }
```
karimJWP: @hghazzi  why are we checking that preload is contained in preloadValues ? Is this to prevent developers from typing a random preload value ?

hghazzi: Yes, we would want it to default to `meta` if an invalid input is entered

hghazzi:
Updated. Verified that there are no other references to `meta` for preload

robwalch:
@hghazzi please update the title and JIRA tickets as well

hghazzi:
@robwalch done

src/js/api/config.js
```diff
@@ -48,6 +48,7 @@ define([
             nextUpClose: 'Next Up Close',
             related: 'Discover',
             close: 'Close',
+            preload: 'metadata'
```
robwalch: Why is there a localization property for `preload`?

src/js/api/config.js
```diff
@@ -150,8 +151,7 @@ define([
                 'image',
                 'file',
                 'sources',
-                'tracks',
-                'preload'
+                'tracks'
```
robwalch: I wouldn't remove this.

src/js/api/config.js
```diff
@@ -161,11 +161,21 @@ define([
...
 
         return config;
     };
 
+    function normalizePreload(config) {
+        const preloadValues = ['none', 'metadata', 'auto'];
+        const preload = config.playlist[0].preload || config.preload;
+
+        return _.contains(preloadValues, preload) ? preload : 'metadata';
+    }
```
robwalch: This won't apply to feed loaded playlists. This kind of logic should be done in the controller and playlist/ playlist, item and source modules.

src/js/providers/html5.js
```diff
@@ -482,7 +482,7 @@ define([
                 // Playback rate is broken on Android HLS
                 _this.supportsPlaybackRate = false;
             }
-            if (source.preload && source.preload !== _videotag.getAttribute('preload')) {
+            if (source.preload !== _videotag.getAttribute('preload')) {
                 _setAttribute('preload', source.preload);
```
robwalch: Are you sure `source.preload` is set correctly? I see you setting it on the item level, not the source level.

hghazzi: Now being set correctly

src/js/api/config.js
```diff
@@ -47,10 +47,11 @@ define([
             nextUp: 'Next Up',
             nextUpClose: 'Next Up Close',
             related: 'Discover',
-            close: 'Close',
+            close: 'Close'
         },
         renderCaptionsNatively: true,
-        nextUpDisplay: true
+        nextUpDisplay: true,
+        preload: 'metadata'
```
robwalch: Let's not add a player default. Set the default on the item where this would be applied if no item preload setting was made explicitly.

robwalch: Looking at the code you have below, this line is unnecessary.

test/unit/preload-test.js
```diff
@@ -0,0 +1,40 @@
...
+        expect(preload).to.equal('auto');
+    });
+
+    it('should default to metadata value is invalid', function() {
+        const preload = getPreload('aut');
+
+        expect(preload).to.equal('metadata');
+    });
+
+    it('should be fallback when invalid', function() {
```
egreaves: should be fallback?

test/unit/preload-test.js
```diff
@@ -0,0 +1,40 @@
+import { getPreload } from 'utils/preload';
+
+describe('getPreload', function() {
+
+    it('should default to metadata when nothing set', function() {
```
egreaves: when undefined

test/unit/preload-test.js
```diff
@@ -0,0 +1,40 @@
...
+        expect(preload).to.equal('metadata');
+    });
+
+    it('should be set when value is valid', function() {
+        const preload = getPreload('auto');
+
+        expect(preload).to.equal('auto');
+    });
+
+    it('should default to metadata value is invalid', function() {
```
egreaves: if value is invalid

test/unit/preload-test.js
```diff
@@ -0,0 +1,40 @@
...
+        const preload = getPreload('aut');
+
+        expect(preload).to.equal('metadata');
+    });
+
+    it('should use the fallback value when invalid', function() {
```
egreaves: when what is invalid?

hghazzi:
@egreaves comments addressed

test/unit/preload-test.js
```diff
@@ -0,0 +1,40 @@
...
+        expect(preload).to.equal('metadata');
+    });
+
+    it('should return the fallback value when primary value is invalid', function() {
+        const preload = getPreload('aut', 'auto');
+
+        expect(preload).to.equal('auto');
+    });
+
+    it('should return if auto is primary and there is a valid fallback', function() {
```
egreaves: Should return what?

test/unit/preload-test.js
```diff
@@ -0,0 +1,40 @@
...
+        const preload = getPreload('auto', 'none');
+
+        expect(preload).to.equal('auto');
+    });
+
+    it('should return if none is primary and there is a valid fallback', function() {
```
egreaves: Should return what?

src/js/playlist/playlist.js
```diff
@@ -1,3 +1,5 @@
+import { getPreload } from '../utils/preload';
```
robwalch: put this module in playlist if that's the only place it's used. "utils" is a not a miscellaneous folder for javascript modules. It's main purpose is for code that implements helpers found in `jwplayer.utils`.

Modules that serve a single script should be placed next to that script or a sub directory.

test/unit/preload-test.js
```diff
@@ -0,0 +1,40 @@
...
+
+        expect(preload).to.equal('auto');
+    });
+
+    it('should default to metadata when primary value is invalid', function() {
+        const preload = getPreload('aut');
+
+        expect(preload).to.equal('metadata');
+    });
+
```
robwalch: You don't have a test for an invalid fallback value (only for unspecified fallback).

hghazzi:
@robwalch Addressed your comments in https://github.com/jwplayer/jwplayer/pull/2109/files

test/unit/preload-test.js
hghazzi: Done

src/js/providers/html5.js
```diff
@@ -483,6 +483,8 @@ define([
                 _this.supportsPlaybackRate = false;
             }
 
+            _setAttribute('preload', 'none');
```
hghazzi: @robwalch Regarding your comment in the other PR

"That depends on when the source is set relative to when preload it. This is also one of the reasons why preload is an item/source property - so that they can be set at the same time or correct order, but now the player needs be in control of preload so the source's setting is completely irrelevant (so why are we still setting preload on source?)."

robwalch: That depends on when the **`src`** is set relative to when preload it. This is also one of the reasons why preload is **an `item` or `item.source` property** - so that **video `src` and `preload` attributes** can be set at the same time or correct order, but now the player needs be in control of preloading. So **`item.source.preload`** is completely irrelevant (why are we still setting  **`item.source.preload`**?).

src/js/controller/controller.js
```diff
@@ -306,7 +307,7 @@ define([
             // Otherwise, it should try to preload the first player on the page,
             // which is the player that has a uniqueId of 1
             function shouldPreload(preloaded, viewable) {
-                return !preloaded && (viewable === 1 || _api.uniqueId === 1);
+                return !preloaded && (_api === window.jwplayer(0) || viewable === 1);
```
robwalch: Never use the`jwplayer` global in player source

src/js/api/global-api.js
```diff
@@ -53,6 +53,9 @@ define([
     var _addPlayer = function (api) {
         _uniqueIndex++;
         api.uniqueId = _uniqueIndex;
+        api.isFirstPlayer = function() {
+            return this === _instances[0];
+        };
```
hghazzi: @robwalch Let me know if adding this is cool. Can't require `global-api.js` because it would cause a circular dependency:

`controller/controller` => `api/global-api` => `api/api` => `controller/controller`

src/js/playlist/preload.js
```diff
@@ -0,0 +1,10 @@
+const _ = require('utils/underscore');
+
+const preloadValues = ['none', 'metadata', 'auto'];
```
johnBartos: If you use an object you don't need to do `_.contains`. An es6 `Map` is even better if we support it.

src/js/api/global-api.js
robwalch: I'm not a fan of adding API methods for internal use.  My suggestion we discussed was to make a module that exports _instances.

src/js/providers/html5.js
```diff
@@ -569,7 +571,7 @@ define([
         };
 
         this.preload = function(item) {
-            let preload = item.sources[_currentQuality] ? item.sources[_currentQuality].preload : 'metadata';
+            const preload = item.sources[_currentQuality] ? item.sources[_currentQuality].preload : 'metadata';
```
robwalch: Can we call `_setVideotagSource` first to ensure the src is set correctly?

src/js/providers/html5.js
```diff
@@ -569,7 +571,7 @@ define([
...
             _setAttribute('preload', preload);
```
robwalch: In the near-term we should insure that the source cannot be changed here. The "preload" is a player property. It will not longer be an item or source property. To ensure this preload should only accept preload values "auto", "metadata" and "none" ("none" would only be needed to stop preloading which is not something we need yet). Providers will be single item use - initialized or constructed with an item which cannot change (preload and load will not need to accept an item) https://github.com/jwplayer/jwplayer/issues/2130 .

src/js/controller/controller.js
```diff
@@ -263,7 +264,7 @@ define([
 
                 _checkAutoStart();
 
-                _model.set('preloaded', false);
+                _preloaded = false;
```
robwalch: `_preloaded` should be set to `false` inside `stop`. Since it's set to false on `setup` it does not need to be set to `false` again here.

PR #2114, opened 2017-06-19
Remove for-of loops

robwalch:
### This PR will...

- Add linting to prevent use of `for (...of)` loops
- Replace for-of loop in `setConfig` with Array.forEach method
- Remove code in set-config module that repeats logic in the controller
- Remove `setConfig` stub method in provider defaults (provider API should not have been changed for a player API feature)

Using  `for (...of)` loops requires `Symbol` and iterator polyfills to work on older browsers (see [babel/babel#1534](https://github.com/babel/babel/issues/1534)).
Including the polyfills for this would add code we don't need, since we can use `Array.forEach` or `_.each` in all places where `for (...of)` was used.

#### Addresses Issue(s):

JW7-4433


robwalch:
@dannyfinks We'll want to check the results of the "setConfig Api" automated tests ("api_set_config.feature").

src/js/api/set-config.js
```diff
@@ -6,57 +6,41 @@ const supportedFields = [
...
-    if (volumeSet) {
-        model.setVolume(volume);
-        if (!muteSet) {
-            model.setMute(volume === 0);
-        }
-    }
-
-    if (muteSet) {
-        model.setMute(mute);
-    }
```
hghazzi: The reason we did this was to handle the situation where `setConfig` can handle the case when both `volume` and `mute` are set at the same time. Though unlikely, one use case I can think of is if the user wants to mute the audio, but have the volume set a specific value when the user chooses to unmute it, instead of what was set before it was muted. 

Without this code, the UI can get into a weird state where audio is playing but the mute icon is muted.

.eslintrc
```diff
@@ -22,7 +22,11 @@
         "sourceType": "module"
     },
     "extends": "eslint:recommended",
+    "plugins": [
+        "no-for-of-loops"
+    ],
     "rules": {
+        "no-for-of-loops/no-for-of-loops": 2,
```
hghazzi: This is a nice add

src/js/api/set-config.js
robwalch: `forEach` ensures order so `setVolume` will be called before `setMute` if both volume and mute are set. This enforces the behavior where you can set the volume and mute the player at the same time.

PR #2121, opened 2017-06-20
Avoid setting read-only properties on the native Tracks object 

johnBartos:
### This PR will...
Avoid setting read-only properties on the native Tracks object 

For our webpack2 builds, which `use strict`, setting read-only properties throws exceptions. Language and kind area read-only: Avoid setting read-only properties on the native Tracks object 

No

No

#### Addresses Issue(s):

JW7-4438

### DIFF
--- a/src/js/providers/tracks-mixin.js
+++ b/src/js/providers/tracks-mixin.js
@@ -509,7 +509,7 @@ define(['utils/underscore',
                 if (utils.isIE() && renderNatively && /^(native|subtitle|cc)/.test(track._id)) {
                     return;
                 }
-                
+
                 // Cues are inaccessible if the track is disabled. While hidden,
                 // we can remove cues while the track is in a non-visible state
                 // Set to disabled before hidden to ensure active cues disappear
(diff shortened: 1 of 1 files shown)


src/js/providers/tracks-mixin.js
```diff
@@ -551,10 +551,7 @@ define(['utils/underscore',
             // already have one with the same label
             track = _.findWhere(tracks, { label: label });
 
-            if (track) {
-                track.kind = itemTrack.kind;
-                track.language = itemTrack.language || '';
-            } else {
```
robwalch: Do we depend on `kind` being updated properly in other places?

You should search the code base for places where we read `kind`. If necessary we should consider adding our own property using `Object.defineProperty`.

robwalch: My guess is that functionality is required for playlists where the video tag and it's tracks are reused.

💉  🙉  `removeTextTrack` we need you!

egreaves:
@johnBartos @robwalch We set `kind` and `language` to keep things consistent, but neither property is used visually, so it should be ok that they're not mutated when reusing a track. `label` is what ends up in the menu. It's likely that the `language` will be the same if the `label` hasn't changed and `kind` should only ever be `subtitles` or `captions` (never `metadata`).

+1 to needing `removeTextTrack`!

src/js/providers/tracks-mixin.js
egreaves: @johnBartos @robwalch We set `kind` and `language` to keep things consistent, but neither property is used visually, so it should be ok that they're not mutated when reusing a track. `label` is what ends up in the menu. It's likely that the `language` will be the same if the `label` hasn't changed and `kind` should only ever be `subtitles` or `captions` (never `metadata`).

+1 to needing `removeTextTrack`!

PR #2124, opened 2017-06-21
Bundle Promise and btoa polyfills in player embed

robwalch:
The Promise and btoa base64 polyfills are 2.6K and 884 bytes respectively. Adding these to the code base put a tiny bit of overhead on Chrome, but this is essential to IE compatibility.

Also fixed a strict mode issue with flash.js. `_swf.__commandQueue` is created as a getter and cannot be overwritten. All we need to do is empty this array rather than replace it.

Webpack 2 boilerplate loaders depend on Promises, so we can't load polyfills for them with webpack. IE 11 and lower does not include Promises.

No

[jwplayer-commercial] should be merged first, since this PR removes setup tasks currently required by commercial setup.

#### Addresses Issue(s):

JW8-118

### DIFF
--- a/src/js/controller/setup-steps.js
+++ b/src/js/controller/setup-steps.js
@@ -7,7 +7,9 @@ define([
     'utils/underscore',
     'utils/helpers',
     'events/events',
-    'controller/controls-loader'
+    'controller/controls-loader',
+    'polyfills/promise',
+    'polyfills/base64'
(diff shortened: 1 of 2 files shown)


johnBartos:
test this please

johnBartos:
test this please

robwalch:
Test this please

robwalch:
test this please

PR #2125, opened 2017-06-22
Preloading: Trigger When Player is Viewable

hghazzi:
### This PR will...

- Trigger providers to preload when a player is viewable or if it's the first player on the page if no players are viewable
- Update the HTML5 provider to handle preloading when player is viewable

To allow providers, such as Shaka & hls.js, to know that they should begin preloading

That any player should only ever have one provider triggered to preload

https://github.com/jwplayer/jwplayer-commercial/pull/3652
https://github.com/jwplayer/jwplayer-commercial/pull/2126

#### Addresses Issue(s):

JW8-23

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -285,7 +285,11 @@ define([
                     viewable: viewable
                 });
                 _checkPlayOnViewable(model, viewable);
-                _checkPreload(model, viewable);
+
+                if (shouldPreload(model.get('preloaded'), viewable)) {
+                    model.getVideo().preload();
+                    model.set('preloaded', true);
+                }
             }
 
             function _checkPlayOnViewable(model, viewable) {
@@ -298,15 +302,11 @@ define([
                 }
             }
 
-            function _checkPreload(model, viewable) {
-                const attemptPreload = (viewable === 1 || _api.uniqueId === 1);
-
-                if (model.get('preloaded') || !attemptPreload) {
-                    return;
-                }
-
-                model.getVideo().preload();
-                model.set('preloaded', true);
+            // Should only attempt to preload if the player is viewable.
+            // Otherwise, it should try to preload the first player on the page,
+            // which is the player that has a uniqueId of 1
+            function shouldPreload(preloaded, viewable) {
+                return !preloaded && (viewable === 1 || _api.uniqueId === 1);
             }
 
             this.triggerAfterReady = function(type, args) {
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -287,7 +287,7 @@ define([
                 _checkPlayOnViewable(model, viewable);
 
                 if (shouldPreload(model.get('preloaded'), viewable)) {
-                    model.getVideo().preload();
+                    model.getVideo().preload(model.get('playlistItem'));
                     model.set('preloaded', true);
                 }
             }
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -262,6 +262,8 @@ define([
                 }
 
                 _checkAutoStart();
+
+                _model.set('preloaded', false);
                 _model.change('viewable', viewableChange);
             }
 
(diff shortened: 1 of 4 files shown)


src/js/controller/controller.js
```diff
@@ -295,6 +298,17 @@ define([
                 }
             }
 
+            function _checkPreload(model, viewable) {
```
johnBartos: Can you leave comments explaining what this function does?

johnBartos: Would also prefer this as a pure function (+ unit tests)

dannyfinks:
@hghazzi https://jenkins.jwplayer.com/job/jw7-commercial-cucumber/6508/cucumber-html-reports/report-feature_features-dash_mixed_playlist-feature.html

johnBartos:
See: https://github.com/jwplayer/jwplayer/pull/2126

hghazzi:
flash.js comment addressed in #2126

src/js/controller/controller.js
```diff
@@ -262,6 +262,8 @@ define([
                 }
 
                 _checkAutoStart();
+
+                _model.set('preloaded', false);
```
robwalch: If this is a model default it should be set in model.js in the constructor. Is there an event listener on this that requires it be called here?

src/js/controller/controller.js
```diff
@@ -283,10 +285,15 @@ define([
                     viewable: viewable
                 });
                 _checkPlayOnViewable(model, viewable);
+
+                if (shouldPreload(model.get('preloaded'), viewable)) {
```
robwalch: If the only place model `model.preloaded` is read and written in here in the controller, don't use the model - add a property to the controller.

src/js/controller/controller.js
```diff
@@ -295,6 +302,13 @@ define([
                 }
             }
 
+            // Should only attempt to preload if the player is viewable.
+            // Otherwise, it should try to preload the first player on the page,
+            // which is the player that has a uniqueId of 1
+            function shouldPreload(preloaded, viewable) {
+                return !preloaded && (viewable === 1 || _api.uniqueId === 1);
```
robwalch: `api.uniqueId === 1` is the first player instance created with `jwplayer().setup()`, but there could be cases where old players are removed. So what we really want is to know if `_api === _instances[0]` in global-api.js. An easy what to do this is `_api === jwplayer(0)` or `selectPlayer(0)`.

src/js/providers/html5.js
```diff
@@ -482,9 +482,6 @@ define([
                 // Playback rate is broken on Android HLS
                 _this.supportsPlaybackRate = false;
             }
-            if (source.preload && source.preload !== _videotag.getAttribute('preload')) {
-                _setAttribute('preload', source.preload);
-            }
```
robwalch: Are you made sure that the `preload` attribute is being set on the video tag before `src` is set? If not, then the browser's default behavior will kick in before `provider. preload` is called. This would be bad in the case where preload should be set to none or metadata but the browser's default is to use auto.

hghazzi:
@robwalch will address your comments in https://github.com/jwplayer/jwplayer/pull/2125/files

src/js/controller/controller.js
hghazzi: you're right, no need to put it in the model.

src/js/controller/controller.js
hghazzi: Thanks for the clarification!

src/js/providers/html5.js
hghazzi: Good point. Since we no longer want to trigger preload on `init`, does it make sense to set it to `none` here, and then update it when the preloading conditions are met?

robwalch: That depends on when the source is set relative to when preload it. This is also one of the reasons why preload is an item/source property - so that they can be set at the same time or correct order, but now the player needs be in control of preload so the source's setting is completely irrelevant (so why are we still setting preload on source?).

Setting preload on the video to none here might be the best solution. It the only way to prevent any preloading until the player explicitly calls preload with the desired setting. This also means we do not need to call provider.preload() with a value of "none" since it would do nothing.

PR #2126, opened 2017-06-23
Call preload with item

johnBartos:
### This PR will...
- Call `preload` with the item to load

Makes the code more understandable

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -569,7 +569,8 @@ define([
         };
 
         this.preload = function(item) {
-            _setAttribute('preload', item.sources[_currentQuality]);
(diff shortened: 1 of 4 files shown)


src/js/providers/html5.js
```diff
@@ -565,16 +565,11 @@ define([
...
         };
 
-        this.preload = function() {
-            if (!this.preloadSource || this.preloadSource.preload === 'none') {
-                return;
-            }
-
-            _setAttribute('preload', this.preloadSource.preload);
+        this.preload = function(item) {
+            _setAttribute('preload', item.sources[_currentQuality]);
```
egreaves: Should this be `item.sources[_currentQuality].preload`?

PR #2131, opened 2017-06-28
Change to buffering when loading an ad

jw-ads-team:
### This PR will...
change the player state to buffering when loading a new ad.

When vpaid ad is closed in ad pod, and the next ad is an instream, the previous and the new state is 'playing', which results the play event to not fire when the instream begins. Play event only fires when there is a change in state.

https://github.com/jwplayer/jwplayer-ads-vast/pull/203

#### Addresses Issue(s):
ADS-336

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -208,12 +208,12 @@ define([
             var providersNeeded = providersManager.required(playlist, primary);
 
             _model.set('hideAdsControls', false);
+            _instream._adModel.set('state', states.BUFFERING);
             providersManager.load(providersNeeded)
                 .then(function() {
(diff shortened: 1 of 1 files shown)


src/js/controller/instream-adapter.js
```diff
@@ -213,6 +213,7 @@ define([
                     if (_instream === null) {
                         return;
                     }
+                    _instream._adModel.set('state', states.BUFFERING);
```
robwalch: Why not do this after line 210? That way the state is buffering while loading the needed provider.

robwalch:
Is this change intended for 8.0.0? 7.12.x work should be done agains the v7.12.x branch.

PR #2133, opened 2017-06-28
Share preload constants between providers

hghazzi:
### This PR will...

- Allow the buffer lengths to be shared between providers
- Block calling the `preload` provider method if preload is `none`

- So that hlsjs and shaka use the same buffer length values

Merge Order:
1. https://github.com/jwplayer/jwplayer/pull/2133 (Current)
2. https://github.com/jwplayer/jwplayer-commercial/pull/3672

#### Addresses Issue(s):

JW8-55

### DIFF
--- a/src/js/providers/constants.js
+++ b/src/js/providers/constants.js
@@ -1,4 +1,2 @@
-export const BufferLengths = {
-    METADATA: 1,
-    MAX: 25
-};
+export const MetaBufferLength = 1;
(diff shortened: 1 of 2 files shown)


src/js/providers/constants.js
```diff
@@ -0,0 +1,4 @@
+export const BufferLengths = {
```
johnBartos: Export individual consts not in all-caps

PR #2134, opened 2017-06-29
Fire 'nextHidden' when nextUp disappears

karimJWP:
### This PR will...
trigger 'nextHidden' when the nextUp mini pop-up disappears

in analytics we listen for nextHidden to reset the 'nextShownReason'.

n/a

https://github.com/jwplayer/jwplayer-analytics/pull/291

#### Addresses Issue(s):

JW7-4434

### DIFF
--- a/src/js/view/controls/nextuptooltip.js
+++ b/src/js/view/controls/nextuptooltip.js
@@ -93,6 +93,8 @@ define([
                         feedData: nextUp.feedData,
                         reason: reason,
                     });
+                } else if (!show && nextUp) {
+                    this.trigger('nextHidden');
                 }
             }
         }


src/js/view/controls/nextuptooltip.js
```diff
@@ -93,6 +93,8 @@ define([
                         feedData: nextUp.feedData,
                         reason: reason,
                     });
+                } else if (!show && nextUp) {
+                    this.trigger('nextHidden');
```
robwalch: Might be better written as:
```js
if (nextUp) {
  if (show) {
    // do a
  } else {
    // do b
  }
}

PR #2135, opened 2017-06-29
Check for mute before pausing a player scrolled out of visibility

karimJWP:
### This PR will...
pause the video after being scrolled off only if it is currently muted

To prevent pausing unmuted videos when scrolled out of view

n/a

n/a

#### Addresses Issue(s):

JW7-4442

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -289,7 +289,7 @@ define([
                 if (_model.get('playOnViewable')) {
                     if (viewable) {
                         _autoStart();
-                    } else if (utils.isMobile() && model.get('mute')) {
+                    } else if (utils.isMobile()) {
                         _this.pause({ reason: 'autostart' });
                     }
                 }
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -263,6 +263,7 @@ define([
 
                 _checkAutoStart();
                 _model.change('viewable', viewableChange);
+                _model.change('viewable', _checkPlayOnViewable);
             }
 
             function _updateViewable(model, visibility) {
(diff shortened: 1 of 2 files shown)


src/js/controller/controller.js
```diff
@@ -289,7 +289,7 @@ define([
                 if (_model.get('playOnViewable')) {
                     if (viewable) {
                         _autoStart();
-                    } else if (utils.isMobile()) {
+                    } else if (utils.isMobile() && model.get('mute')) {
```
robwalch: I'm not sure we want to use the mute setting for this. I thought when we use 'playOnViewable' there is a second mute setting that indicates that autostart mute is active. 'mute' just indicates the users preference, not whether or not we're dealing with this state where we are muted because of 'playOnViewable'.

karimJWP: @robwalch I'm still a bit confused. So it's not enough to check the volume? We would want to set playOnViewable to false if volume is no longer muted ?

robwalch: Maybe `_checkPlayOnViewable` should be handled by it own listener that is removed after the first unmute (see controls.js `unmuteAutoplay` and `unmuteCallback`).

src/js/controller/controller.js
```diff
@@ -263,6 +263,10 @@ define([
 
                 _checkAutoStart();
                 _model.change('viewable', viewableChange);
+                _model.change('viewable', _checkPlayOnViewable);
+                _model.on('change:autostartFailed change:autostartMuted change:mute', function(model) {
+                    model.off('change:viewable', _checkPlayOnViewable);
+                });
```
karimJWP: @egreaves I moved the removal of the listener to controller.js because _checkPlayOnViewable is undefined in controls.js

egreaves:
@dannyfinks @karimJWP note that if there are multiple players on the page with 1 or more still exhibiting muted autostart behavior, the player that is in view will cause the unmuted player to be paused.

PR #2137, opened 2017-06-30
Change to buffering when loading an ad

jw-ads-team:
### This PR will...
change the player state to buffering when loading a new ad.

When vpaid ad is closed in ad pod, and the next ad is an instream, the previous and the new state is 'playing', which results the play event to not fire when the instream begins. Play event only fires when there is a change in state.

https://github.com/jwplayer/jwplayer-ads-vast/pull/203

#### Addresses Issue(s):
ADS-336


jw-ads-team:
test this please

PR #2153, opened 2017-07-07
API backwards compatibility

karimJWP:
### This PR will...
create an optional script that can add methods to jwplayer

Enables API backwards compatibility for JW8

How should this be bundled ?

https://github.com/jwplayer/jwplayer-commercial/pull/3717

#### Addresses Issue(s):

JW8-110


src/js/compatibility.js
```diff
@@ -0,0 +1,75 @@
+(function(playerLibrary) {
+    // Check if the version of the player requires the compatibility shim
+    if (parseInt(playerLibrary.version, 10) >= 8) {
```
egreaves: This would throw an error if `window.jwplayer` is `undefined`. Though the compatibility script should ideally be added to the page after the jwplayer library, we should be more defensive and check that it's defined.

src/js/compatibility.js
```diff
@@ -0,0 +1,75 @@
...
+                onAdComplete: 'adComplete',
+                onAdCompanions: 'adCompanions',
+                onAdSkipped: 'adSkipped',
+                onAdPlay: 'adPlay',
+                onAdPause: 'adPause',
+                onAdMeta: 'adMeta',
+                onCast: 'cast',
+                onAudioTrackChange: 'audioTrackChanged',
+                onAudioTracks: 'audioTracks'
+            };
```
robwalch: All of these should be removed from api/api.js at the same time as they are added here. Included deprecated jsdocs would be useful too. See https://jwplayer.atlassian.net/browse/JW8-121

karimJWP:
@egreaves comments addressed
@robwalch comments addressed in https://github.com/jwplayer/jwplayer-commercial/pull/3717

src/js/compatibility.js
```diff
@@ -0,0 +1,79 @@
...
+                onAdTime: 'adTime',
+                onAdComplete: 'adComplete',
+                onAdCompanions: 'adCompanions',
+                onAdSkipped: 'adSkipped',
+                onAdPlay: 'adPlay',
+                onAdPause: 'adPause',
+                onAdMeta: 'adMeta',
+                onCast: 'cast',
+                onAudioTrackChange: 'audioTrackChanged',
+                onAudioTracks: 'audioTracks'
```
robwalch: All of these should be removed from api/api.js at the same time as they are added here. Included deprecated jsdocs would be useful too. See https://jwplayer.atlassian.net/browse/JW8-121

karimJWP: @robwalch  I was under the impression that on Friday we had agreed i would remove the callbacks from api.js in ticket [JW8-121](https://jwplayer.atlassian.net/browse/JW8-121). I may have misunderstood, but now that I think about it, it sounds like a good idea since removing them now would break integration tests until [JW8-121](https://jwplayer.atlassian.net/browse/JW8-121) is merged. If you disagree I'll remove the deprecated callbacks asap

src/js/compatibility.js
```diff
@@ -0,0 +1,79 @@
+(function(playerLibrary) {
+    // Verify JW Player is loaded
+    if (!window.jwplayer) {
```
robwalch: `if (!playerLibrary) {` would work just as well and not repeat `window.jwplayer`.

PR #2157, opened 2017-07-11
Remove deprecated callbacks

karimJWP:
### This PR will...
remove deprecated callbacks from the API

Deprecated callbacks are now defined in Compatibility script and should be excluded from api.js

n/a

https://github.com/jwplayer/jwplayer-plugin-gapro/pull/2

#### Addresses Issue(s):

JW8-121


johnBartos:
@robwalch I made that comment in my PR

PR #2159, opened 2017-07-12
Pass the controlbar argument through api.getSafeRegion

robwalch:
### This PR will...

Update the API to properly implement `getSafeRegion` which accepts an argument to exclude the controlbar from the region it returns (defaults to true).

This argument was removed breaking functionality that depended on it.

#### Addresses Issue(s):

JW8-179


dannyfinks:
👍

PR #2161, opened 2017-07-13
Unit test fixes and setup improvements

robwalch:
### This PR will...

- Fix the karma test runner setup by removing the mock-ensure module and importing modules into test-context that would otherwise be chunked into separate bundles not served by karma.
- Remove tests.js file - it was not used and completely unnecessary since test-context includes all test spec files.
- Attempt to make jQuery and sinon external. They should be but may still get bundled with test-context.
- Fix jwplayer-selectplayer-test skipped tests and fixture handling.

Tests were failing and being skipped because the test running was not being setup properly.

https://github.com/jwplayer/jwplayer-commercial/pull/3738


test/unit/jwplayer-selectplayer-test.js
```diff
@@ -4,28 +4,36 @@ define([
     'jwplayer'
 ], function (_, $, jwplayer) {
 
+    function testInstanceOfApi(assert, api) {
+        assert.isOk(_.isObject(api), 'jwplayer({dom id}) returned an object');
+        assert.isOk(_.isFunction(api.setup), 'object.setup is a function');
```
hghazzi: @robwalch when you write new or update unit tests, switch `assert` to `expect`. Follows BDD and makes easier to read. You need to pass around `expect` everywhere as well.

http://chaijs.com/api/bdd/

karma.conf.js
```diff
@@ -110,17 +111,22 @@ module.exports = function(config) {
...
                     __FLASH_VERSION__: 18.0
                 }),
-            ]
+            ],
+            externals: {
+                $: {
+                    commonjs: 'jquery',
+                    amd: 'jquery',
+                    root: '$'
+                },
```
hghazzi: Why don't we do this for `underscore` as well.

robwalch: You tell me. Not even sure this works.

PR #2162, opened 2017-07-13
Updated default width and height values

dYale:
### This PR will...
Change the default player size **from** 480 x 270 **to** 640 x 360.
Our default player size was not up to date with current standards of minimum resolutions. Updating this should lead to less resolution customization.
No
No
#### Addresses Issue(s):

JW8-173

### DIFF
--- a/test/unit/api-config-test.js
+++ b/test/unit/api-config-test.js
@@ -39,6 +39,13 @@ define([
                 return x;
             }
 
+            it('should have default width of 640 and height of 360', function() {
+                let x = testConfig();
+
+                expect(x.width).to.equal(640);
+                expect(x.height).to.equal(360);
+            });
+
             it('should accept widths in different formats', function () {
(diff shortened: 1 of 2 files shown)


test/unit/api-config-test.js
```diff
@@ -39,6 +39,13 @@ define([
                 return x;
             }
 
+            it('should have default width of 640 and height of 360', function() {
```
johnBartos: Include a space after a function declaration: `function ()`

test/unit/api-config-test.js
```diff
@@ -39,6 +39,13 @@ define([
...
+                let x = testConfig();
```
johnBartos: `const x`

dYale: I was going to do a `const` but wanted to be consistent with the rest of the file. 

Will update

PR #2167, opened 2017-07-17
Remove "time slider above"

johnBartos:
### This PR will...
- Remove the `timeSliderAbove` config option
- Merge `time-slider-above` styles into base styles, such that the player is always in "time slider above" mode

JW8 will only feature UIs with the time slider above. Removing the option to choose reduces complexity.

Did I merge the styles correctly? I verified states myself with the player. Let's see if cucumber turns anything up.

https://github.com/jwplayer/jwplayer-commercial/pull/3759
#### Addresses Issue(s):

JW8-89


dannyfinks:
@johnBartos the failing test here is not a bug, but our test method to always show control-bar not accommodating timeSliderAbove

robwalch:
@johnBartos @dannyfinks I added some styles to maintain the 36px tall cbar over ima ads required for the skip button to be clickable in our tests. Also removed redundant group style.

PR #2168, opened 2017-07-17
Consistent env names

johnBartos:
### This PR will...
- Consistentize environment names on a lowercase scheme
- Fix flipped conditional from original ticket

Consistent names are easier to understand & work with

No

No

#### Addresses Issue(s):

JW8-122

### DIFF
--- a/src/js/compatibility.js
+++ b/src/js/compatibility.js
@@ -110,7 +110,7 @@
             utils.isIOS = valueFn(environment.OS.iOS);
             utils.isIPad = valueFn(environment.OS.iPad);
             utils.isIPod = valueFn(environment.OS.iPhone);
-            utils.isMSIE = valueFn(environment.Browser.MSIE);
+            utils.isMSIE = valueFn(environment.Browser.msie);
             utils.isMobile = valueFn(environment.OS.mobile);
             utils.isOSX = valueFn(environment.OS.OSX);
             utils.isSafari = valueFn(environment.Browser.safari);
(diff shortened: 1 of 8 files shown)


robwalch:
What do you think about making `OS`, `Browser` and `Features` camelcase (all lowercase here)? When importing `* as Environment` or calling `getEnvironment()` the capitalized words look odd to me. I expect object properties to not be capitalized.

Currently it goes
`const isChrome = jwplayer().getEnvironment().Browser.chrome`
```js
jwplayer().getEnvironment() returns {
    Browser: {...},
    OS: {...},
    Feature: {...}
}
```

compared to

`const isChrome = jwplayer().getEnvironment().browser.chrome`
```js
jwplayer().getEnvironment() returns {
    browser: {...},
    os: {...},
    feature: {...}
}
```

src/js/compatibility.js
```diff
@@ -110,9 +110,9 @@
             utils.isIOS = valueFn(environment.OS.iOS);
             utils.isIPad = valueFn(environment.OS.iPad);
             utils.isIPod = valueFn(environment.OS.iPhone);
-            utils.isMSIE = valueFn(environment.Browser.MSIE);
+            utils.isMSIE = valueFn(environment.Browser.msie);
             utils.isMobile = valueFn(environment.OS.mobile);
-            utils.isOSX = valueFn(environment.OS.OSX);
+            utils.isOSX = valueFn(environment.OS.osx);
```
robwalch: I like actually like the idea of changing `osx` to `mac`.

PR #2169, opened 2017-07-17
Deprecate Skins

hghazzi:
### This PR will...

Deprecate all skins, including seven, as well as the `skin.name` config

We are no longer going to support any JW7 in JW8

https://github.com/jwplayer/jwplayer-commercial/pull/3757

#### Addresses Issue(s):

JW8-86


robwalch:
test this please

src/css/skins/seven.less
```diff
@@ -1,71 +0,0 @@
-@import "../shared-imports/vars";
-@import "../controls/imports/transitions";
-
-.jw-skin-seven {
-    @display-bkgd-color: @controlbar-background;
-
-    .jw-display-icon-container {
-        border-radius: 3.5em;
```
robwalch: You might consider moving some of these styles to the appropriate class defaults (like styles in display.less). Just make sure you're only adding or replacing (not duplicating) a style like `border-radius`.

src/css/skins/seven.less
```diff
@@ -1,71 +0,0 @@
...
-
-        & > .jw-icon {
-            color: rgba(255, 255, 255, 0.9);
```
robwalch: default

src/css/skins/seven.less
```diff
@@ -1,71 +0,0 @@
...
-        }
-    }
-
-    .jw-dock-button {
-        // Seven skin dock buttons do not highlight, they only show the tooltip
-        &:hover {
-            background: @controlbar-background;
-        }
-
-        border-radius: 2.5em;
```
robwalch: default

src/css/skins/seven.less
```diff
@@ -1,71 +0,0 @@
...
-    .jw-skip {
-        border-radius: @ui-padding;
-    }
-
-    .jw-text {
-        text-rendering: optimizeLegibility;
-        /* font-smoothing is a private setting not transformed by PostCSS */
-        -webkit-font-smoothing: antialiased;
-        /* OSX Firefox specific style for improving legibility */
-        -moz-osx-font-smoothing: grayscale;
```
robwalch: default

src/templates/error.js
```diff
@@ -1,6 +1,6 @@
-export default (id, skin = '', title = '', body = '') => {
+export default (id, title = '', body = '') => {
```
egreaves: What happens if there is a custom skin that has styles for the error template?

robwalch: I think we want to rename this one setup-error.js. I don't think it's necessary for us to apply the custom skin class here. There is not default so it doesn't belong in the template.

PR #2173, opened 2017-07-19
Simplify controlbar with flexbox

johnBartos:
### This PR will...
- Remove controlbar groups
- Wraps controlbar buttons in a new container
- Adds slider container and button container as children to `jw-controlbar`
- Use a flexible spacer to position elements
- Eliminate styles made redundant by Flexbox
- Upgrades Stylelint to latest

The controlbar DOM is confusing and difficult to customize

Everything

https://github.com/jwplayer/jwplayer-commercial/pull/3834 Upgrades stylelint in commercial

#### Addresses Issue(s):

JW8-127


src/css/controls/imports/controlbar.less
```diff
@@ -96,9 +75,17 @@
...
 }
 
-.jw-controlbar-left-group,
-.jw-controlbar-right-group {
-    white-space: nowrap;
+.jw-slider-horizontal {
+    background-color: transparent;
+}
+
+.jw-icon-cast {
```
radium-v: This collides with the declaration on [line 140](https://github.com/jwplayer/jwplayer/blob/cea51655b5f3a0c93a0fada53a3a6de66c46df29/src/css/controls/imports/controlbar.less#L140)

johnBartos:
Need to update the `browserlist` with what we now support.

src/css/controls/flags/ads.less
```diff
@@ -24,61 +24,19 @@
...
     .jw-controlbar {
-        padding-right: 5px;
-        pointer-events: none;
-
-        .jw-icon-inline,
-        .jw-icon-tooltip,
-        .jw-icon-cast,
-        .jw-text,
-        .jw-slider-horizontal {
+        .jw-slider-time,
```
egreaves: Seems odd that the classes are at the same indentation level as the rule.

src/css/controls/flags/audioplayer.less
```diff
@@ -22,43 +22,17 @@
...
-        bottom: 0;
-        margin: 0;
-        width: 100%;
-        min-width: 100%;
+        position: relative; // prevents children from exploding out of this container when set too small
 
-        .jw-icon-fullscreen,
-        .jw-icon-tooltip {
+        .jw-icon-tooltip,
+            .jw-icon-inline:not(.jw-icon-volume):not(.jw-icon-playback):not(.jw-icon-fullscreen):not(jw-icon-next) {
```
egreaves: Is there an alternative approach that would prevent us from having such high specificity for this rule? (`:not` adds specificity)

src/css/controls/imports/controlbar.less
```diff
@@ -12,13 +16,11 @@
     border-radius: 0;
     background-size: auto;
     box-shadow: none;
-    height: @mobile-touch-target * 1.5;
-    padding: @slider-fixed-height 0 0;
+    max-height: 72px;
+    //padding: @slider-fixed-height 0 0;
```
egreaves: The control bar needs left+right padding so that the slider knob doesn't go off the player and to make it usable on touch devices.

Before:
screen shot 2017-07-31 at 2 54 46 pm

After adding 14px:
screen shot 2017-07-31 at 2 54 26 pm

src/css/controls/imports/slider.less
```diff
@@ -88,49 +90,20 @@
...
-        right: 0;
-        left: (@def-rail-width - @thumb-size) / 2;
-        margin-bottom: @thumb-size/-2;
+        left: -50%;
+        margin-bottom: (-@thumb-size / 2);
     }
 }
 
 .jw-slider-time {
     width: 100%;
```
egreaves: `jw-slider-time` needs left+right padding so that the slider knob doesn't go off the player and to make it usable on touch devices.

Before:
screen shot 2017-07-31 at 2 54 46 pm

After adding 14px:
screen shot 2017-07-31 at 2 56 49 pm

src/css/controls/flags/audioplayer.less
radium-v: `:not(jw-icon-next)` should be `:not(.jw-icon-next)`

radium-v: `:not` itself doesn't add specificity, but the selectors inside of `:not()` do. Separating these would reduce the specificity from `0/0/80/0` to `0/0/50/0`.

egreaves:
@johnBartos I think we'll need a div to wrap the control bar buttons. This will make it possible to hide the button group while keeping the slider visible and change the `flex-direction` for ads.

src/css/controls/flags/cast-available.less
```diff
@@ -1,9 +1,9 @@
 .jwplayer.jw-flag-cast-available {
     .jw-icon-cast {
-        display: inline-block;
+        display: flex;
     }
 
     .jw-icon-airplay {
```
radium-v: These two selectors can be combined

src/css/controls/flags/audioplayer.less
egreaves: Thanks for the clarification. I don't think we'd ever use `:not()` without a class inside it, but still good to know.

src/css/controls/imports/controlbar.less
```diff
@@ -12,13 +16,11 @@
     border-radius: 0;
     background-size: auto;
     box-shadow: none;
-    height: @mobile-touch-target * 1.5;
-    padding: @slider-fixed-height 0 0;
+    max-height: 72px;
```
karimJWP: @johnBartos should this be hardcoded ? 72px would be a lot on mobile

src/css/controls/imports/controlbar.less
karimJWP: why are we keeping this ?

src/js/view/controls/controlbar.js
```diff
@@ -25,23 +25,38 @@ define([
...
+        return element;
+    }
+
     function menu(name, ariaText) {
         return new Menu(name, ariaText);
     }
 
     function createCastButton(castToggle, localization) {
-        if (Browser.chrome && OS.iOS) {
+        if (!Browser.chrome || OS.iOS) {
```
karimJWP: Thank you for fixing that! Wouldn't this prevent the chromecast button from appearing in the chrome iOS app?

src/css/controls/imports/controlbar.less
johnBartos: Subject to change when the new skin is implemented, but I'll double check

src/js/view/controls/controlbar.js
johnBartos: This reverts it to how it was before

src/css/controls/imports/slider.less
johnBartos: Addressed

src/css/controls/flags/ads.less
```diff
@@ -96,13 +56,8 @@
...
     }
 
-    .jw-group > .jw-text-alt {
-        display: inline-block;
-        margin: 0;
-        line-height: @mobile-touch-target;
-    }
-
-    .jw-text-duration {
+    .jw-text-duration,
```
egreaves: You can group these with the group of classes above ending with `.jw-autostart-mute`

.stylelintrc
```diff
@@ -68,7 +68,7 @@
         "selector-combinator-space-after": "always",
         "selector-combinator-space-before": "always",
         "selector-descendant-combinator-no-non-space": true,
-        "selector-no-id": true,
+        "selector-max-id": 1,
```
robwalch: What are we assigning ids to?

browserslist
```diff
@@ -1,12 +1,13 @@
...
+FirefoxAndroid >= 54
+Safari >= 9
 Last 4 Edge versions
-Last 4 ExplorerMobile versions
-IE 9-11
-Last 5 UCAndroid versions
\ No newline at end of file
+IE 11
+ExplorerMobile 11
+iOS >= 9.2
```
robwalch: Why not iOS 9.0?

src/js/view/controls/controlbar.js
```diff
@@ -171,35 +171,36 @@ define([
...
+                this.elements.volume,
+                this.elements.volumetooltip,
+                this.elements.fullscreen
+            ];
+
+
+            this.layout = [
+                this.elements.time,
+                this.elements.buttonContainer
+            ];
```
robwalch: `this.elements` should be assigned to a variable to improve minification. (Search "this.elements" in http://ssl.p.jwpcdn.com/player/v/7.11.2/jwplayer.controls.js)

.stylelintrc
robwalch: What are we assigning ids to?

browserslist
robwalch: What are we assigning ids to?

.stylelintrc
robwalch: Can this rule be added as a comment around namespaces, so we still prevent rules with ids from being added?

johnBartos: done

PR #2175, opened 2017-07-20
Events and states deprecation

karimJWP:
### This PR will...
Replace events and states with ES6 constants and move obsolete events to the compatibility script.

To deprecate Events and States for JW 8.

A sanity check would be much appreciated. Keep in mind Documentation for ES6 constants will be added in separate PR

https://github.com/jwplayer/jwplayer-commercial/pull/3768

#### Addresses Issue(s):
JW8-120


src/js/events/states.js
```diff
@@ -1,14 +1,43 @@
...
-        PAUSED: 'paused',
-        PLAYING: 'playing',
-        ERROR: 'error',
-
-        // These exist at the provider level, but are converted to BUFFERING at higher levels
-        LOADING: 'loading',
-        STALLED: 'stalled'
-    };
-});
+/**
```
robwalch: What do you think about putting all of these states in events and prefixing them with `STATE_`? These all are events and player states. I think it will make events.js more complete to include them:

```js
/**
* Triggered when the player attempts to play media, but is still loading media into the buffer or has stalled because there is not enough media in the buffer to continue.
*/
export const STATE_BUFFERING = 'buffering';
```

karimJWP: sounds good to me. Will address once I figure out why unit tests are failing

PR #2176, opened 2017-07-20
Fix issue with ClickHandler class not using options

jw-ads-team:
### This PR will...
Use options passed into the ClickHandler class.

The option that was passed into the ClickHandler was never used. This needs to be used for UI to work correctly.

#### Addresses Issue(s):
ADS-404


jw-ads-team:
@robwalch https://github.com/jwplayer/jwplayer/blob/master/src/js/view/view.js#L471
```{ useHover: true }``` is being passed in, but is never properly used.

PR #2177, opened 2017-07-20
Update Add/Remove Button API

hghazzi:
### This PR will...

Update the `addButton` & `removeButton` APIs to no longer add button to the dock and icon, but will be added to the control bar.

To support the new JW8 Skins, where we will no longer have a dock container and buttons.

- Removing dock html && classes will be handled in https://jwplayer.atlassian.net/browse/JW8-88
- Will need to move where the buttons are being added to the control bar in https://jwplayer.atlassian.net/browse/JW8-127
- Need to verify with Product on what should happen when button id is missing or image url is invalid or blank

#### Addresses Issue(s):

JW8-126

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -798,7 +798,6 @@ define([
                     }
                 }
 
-                console.log('Adding Buttons:', _.pluck(customButtons, 'id'));
                 _model.set('customButtons', customButtons);
             };
             this.removeButton = function(id) {
@@ -812,9 +811,7 @@ define([
                     (button) => button.id !== id
                 );
 
-                console.log('Removing Buttons:', _.pluck(customButtons, 'id'));
                 _model.set('customButtons', customButtons);
-
             };
             // Delegate trigger so we can run a middleware function before any event is bubbled through the API
             this.trigger = function (type, args) {
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -770,6 +770,8 @@ define([
                 return _model.checkComplete();
             };
             this.addButton = function(img, tooltip, callback, id, btnClass) {
+                let customButtons = _model.get('customButtons') || [];
+                let added = false;
                 const newButton = {
                     img: img,
                     tooltip: tooltip,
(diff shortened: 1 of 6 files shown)


src/js/view/controls/controlbar.js
```diff
@@ -113,10 +114,10 @@ define([
             }
             // Do not show the volume toggle in the mobile SDKs or <iOS10
             if (!_model.get('sdkplatform') && !(OS.iOS && OS.version.major < 10)) {
-                muteButton = button('jw-icon-volume', () => { _api.setMute(); }, vol);
+                muteButton = button('jw-icon-volume', () => _api.setMute(), vol);
             }
 
-            const nextButton = button('jw-icon-next', () => { _api.next(); }, next);
+            const nextButton = button('jw-icon-next', () => _api.next(), next);
```
robwalch: The brackets were intentional. Without them, the result of the statement is returned, we don't want to `return _api.setMute()` from the callback, we just want to call `_api.setMute()`.

src/js/view/controls/controlbar.js
```diff
@@ -158,8 +159,8 @@ define([
                 mute: muteButton,
                 volume: volumeSlider,
                 volumetooltip: volumeTooltip,
-                cast: createCastButton(() => { _api.castToggle(); }, this._localization),
-                fullscreen: button('jw-icon-fullscreen', () => { _api.setFullscreen(); }, this._localization.fullscreen)
+                cast: createCastButton(() => _api.castToggle(), this._localization),
+                fullscreen: button('jw-icon-fullscreen', () => _api.setFullscreen(), this._localization.fullscreen)
```
robwalch: ditto.

src/js/controller/controller.js
```diff
@@ -768,40 +768,20 @@ define([
...
-                }
-
-                _model.set('dock', dock);
+                _model.trigger('addButton', customButton);
             };
             this.removeButton = function(id) {
-                let dock = _model.get('dock') || [];
-                dock = _.reject(dock, _.matches({ id: id }));
-                _model.set('dock', dock);
+                _model.trigger('removeButton', id);
```
robwalch: Can you route these through the view and controls rather than the model?

Buttons should be pushed into a collection like they were previously for the doc. Buttons should added to the DOM when controls are rendered - not requiring controls to be active for buttons to be added.

src/js/view/controls/controlbar.js
```diff
@@ -238,6 +239,9 @@ define([
             _model.change('cues', this.addCues, this);
             _model.change('altText', this.setAltText, this);
 
+            _model.on('addButton', this.addButton, this);
+            _model.on('removeButton', this.removeButton, this);
```
robwalch: These would only get called when controls are active meaning you could not add buttons unless controls are true, loaded and the controlbar created.

The creation of buttons should be separate from the rendering of them in controlbar. (see my previous comment in controller.js).

PR #2179, opened 2017-07-21
Events and states deprecation

karimJWP:
### This PR will...
Replace events and states with ES6 constants and move obsolete events to the compatibility script.

To deprecate Events and States for JW 8.

A sanity check would be much appreciated. Keep in mind Documentation for ES6 constants will be added in separate PR

https://github.com/jwplayer/jwplayer-commercial/pull/3776
https://github.com/jwplayer/jwplayer-ads-vast/pull/220
jwplayer/jwplayer-ads-googima#231
jwplayer/jwplayer-ads-freewheel#60
https://github.com/jwplayer/jwplayer-vr-plugin/pull/42

#### Addresses Issue(s):
JW8-120


src/js/events/events.js
```diff
@@ -1,98 +1,383 @@
...
+/**
+ * The video is currently playing. No icon is visible in the display.
+*/
+export const STATE_PLAYING = 'playing';
+
+/**
+ * Playback was stopped due to an error.
+   In this state the error icon and a message are visible in the display.
+*/
+export const STATE_ERROR = 'error';
```
karimJWP: @robwalch  STATE_ERROR has the same value as ERROR_EVENT ('error'), and STATE_COMPLETE has the same value as MEDIA_COMPLETE ('complete'). Could this ever be a problem ? The problem with changing one is that some parts of the code use the value directly (i.e. 'error' and 'complete') for both the state and the event, so we would have to update those as well.

robwalch: The goal is to only have one string literal for each value, and use the constants through the projects (jwplayer and jwplayer-commercial). 

Having state and event constants makes sense. You use the state constants when setting state and event ones when listening to or triggering events. To avoid duplication define the state constants first and then set the events to those really every state should have its own event otherwise just keep the states and remove the duplicate events.

export const ERROR_STATE = 'error';
export const ERROR_EVENT = ERROR_STATE;

robwalch: The goal is to only have one string literal for each value, and use the constants through the projects (jwplayer and jwplayer-commercial). 

Having state and event constants makes sense. You use the state constants when setting state and event ones when listening to or triggering events. To avoid duplication define the state constants first and then set the events to those really every state should have its own event otherwise just keep the states and remove the duplicate events.

export const ERROR_STATE = 'error';
export const ERROR_EVENT = ERROR_STATE;

src/js/compatibility.js
```diff
@@ -131,5 +131,124 @@
         jwplayerCompatible.vid = playerLibrary.vid;
 
         window.jwplayer = jwplayerCompatible;
+
```
robwalch: Add `events` to `jwplayerCompatible` before the line above.

karimJWP:
@robwalch merge conflicts and comment addressed

robwalch:
test this please

PR #2190, opened 2017-07-29
Fix eslint issues

robwalch:
- Deprecate `triggerSafe` (use `catchExceptions` argument in `trigger(name, callback, context, catchExceptions)` instead
- Fix eslint issues in all files
- Convert backbone.events from AMD to ES6 module
- Convert VTT files from AMD to ES6 modules


johnBartos:
@robwalch Want to bump eslint to latest in this PR? I have it done in my controlbar PR but it makes more sense here. Doesn't seem to have any breaking changes for us.

robwalch:
@johnBartos I'm on it. eslint@4.3.0 coming right up.

PR #2191, opened 2017-07-31
Setup Git Push Prehook

hghazzi:
### This PR will...

Trigger `eslint` and `karma` to run before any push occurs. You can still force a push by running `git push -n (or --no-verify)`, though it is not recommended. 

To ensure we do not check in any code that could cause the build to fail

I created a separate npm script that allows karma tests to run more quickly. @johnBartos The slowest part now seems to be when webpack is compiling. Any insights on how we can speed it up?

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -272,6 +272,14 @@ module.exports = function(grunt) {
         });
     });
 
+    grunt.registerTask('prehook', 'Install Pre Push Hook', function() {
+        var command = 'cp .github/hooks/pre-push .git/hooks/pre-push';
+        execSync(command, {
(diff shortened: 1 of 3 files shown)


hghazzi:
test this please

johnBartos:
@hghazzi Should we make our own prehook script over downloading one? This adds time to the build, and the package itself hasn't had an update in a year. Seems easy enough to implement something like https://github.com/alykoshin/git-pre-push/blob/master/hook ourselves

hghazzi:
Considered it, though figured it would be fine since its a dev dependency. But you're right, will save us some build time, so will work on making our own.

PR #2198, opened 2017-08-02
Handle errors loading webpack chunks

robwalch:
### This PR will...

- Handle setup errors in core-shim (jwplayer.js embed)
- Only load one provider at a time, prior to playback rather than all providers required by a playlist before playback can start.
- Remove  try-catch blocks from the controller -> model.provider handling that did not provide clear error handling.
- Add an error handler for all webpack chunks that include the core player
- Failed request for webpack asset needed by setup will fail setup with the error message "Error loading player: Network error" (jwplayer.core*.js)
- Failed request for provider will error one of two ways:
  - A setup error caused by the provider required to play the first item in the playlist will have a message which reads "Error loading player: Network error"
  - A player error caused by a provider required to play a subsequent playlist item or playlist will have a message that reads "Could not play video: Network error"
- Failed request for assets not essential to playback will trigger a non-fatal error event
  - A controls error event caused by jwplayer.controls.js not loading after calling setControls(true) will have a message which reads "Controls failed to load" and a reason property whose value is the Error object with message "Network Error"
  - A captions error event caused by vttparser.js not loading will have a message which reads "Captions failed to load" and a reason property whose value is the Error object with message "Network Error"
  - A captions error event caused by polyfills.webvtt.js not loading will have a message which reads "Captions renderer failed to load" and a reason property whose value is the Error object with message "Network Error"
  - A casting error event caused by provider.cast.js or provider.airplay.js not loading will have a message which reads "Casting failed to load" and a reason property whose value is the Error object with message "Network Error"

These changes allow us to handle any exception in the setup promise chain as a setup error - whether it is thrown from jwplayer.js or core. We also will not have proper error events for asset loading failure non-essential to playback.

jwplayer-commercial

Addition changes to handle plugin loading errors:
https://github.com/jwplayer/jwplayer/pull/2209/commits

#### Addresses Issue(s):

JW8-112


src/js/api/core-shim.js
```diff
@@ -186,4 +194,47 @@ Object.assign(CoreShim.prototype, {
     }
 });
 
+function setupError(core, error) {
+    let message = error.message;
```
johnBartos: Is `message` always defined?

src/js/api/core-shim.js
```diff
@@ -186,4 +194,47 @@ Object.assign(CoreShim.prototype, {
...
+    if (message.indexOf(':') === -1) {
+        message = 'Error loading player: ' + message;
```
johnBartos: String template

src/js/api/core-shim.js
robwalch: Yes https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error

src/js/api/core-shim.js
```diff
@@ -186,4 +194,47 @@ Object.assign(CoreShim.prototype, {
...
+        padding: '.75em 1.5em',
+        width: width.toString().indexOf('%') > 0 ? width : (width + 'px'),
+        height: height.toString().indexOf('%') > 0 ? height : (height + 'px')
+    });
+
+    showView(core, errorElement);
+
+    setTimeout(() => {
+        core.trigger(SETUP_ERROR, {
+            message: message
```
johnBartos: Shorthand

src/js/controller/model.js
```diff
@@ -436,7 +445,12 @@ define([
         };
 
         this.playVideo = function() {
-            _provider.play();
+            const playPromise = _provider.play();
+            if (!playPromise) {
+                // TODO: return new Promise that resolves when playback starts, or rejects on error/timeout
```
johnBartos: TODO?

robwalch: Yeah. Provider work.

PR #2205, opened 2017-08-04
Replace icon fonts with SVGs

karimJWP:
### This PR will...
Enable inlining SVG in icons, add Live button which seeks to live edge in DVR, remove icon fonts

To replace icon fonts with SVGs

In button.js and tooltip.js the SVGs are added to the DOM.

### Known Issues:
- HD icon has not been updated because it is being moved to the menu. It has not been removed to avoid breaking tests until the menu is implemented.
- Play icon is not centered in the display icon; an updated SVG will be provided in the near future.
- Per Alicia's request, stop icon for Live has not been added to the control bar because its functionality has not been defined yet.
- Sharing and Related SVGs will be added in another ticket since they are not dependent on this branch and closing the feedback loop before/during implementation of Sharing and Related work is preferred.

https://github.com/jwplayer/jwplayer-commercial/pull/3859
related: https://github.com/jwplayer/jwplayer-plugin-related/pull/155
sharing: https://github.com/jwplayer/jwplayer-plugin-sharing/pull/51

#### Addresses Issue(s):
JW8-90


src/assets/SVG/playback-rate.svg
```diff
@@ -0,0 +1 @@
+<svg xmlns="http://www.w3.org/2000/svg" class="jw-svg-icon jw-svg-icon-playback-rate" viewBox="0 0 240 240"><path d="M158.83,48.83A71.17,71.17,0,1,0,230,120,71.163,71.163,0,0,0,158.83,48.83Zm45.293,77.632H152.34V74.708h12.952v38.83h38.83ZM35.878,74.708h38.83V87.66H35.878ZM10,113.538H61.755V126.49H10Zm25.878,38.83h38.83V165.32H35.878Z"/></svg>
```
karimJWP: @radium-v this is the new file for playback rate, for consistency with other icons

test/unit/controlbar-test.js
```diff
@@ -26,8 +29,8 @@ describe('Control Bar', function() {
         container.appendChild(spacer);
 
         controlBar = new ControlBar({}, model);
-        controlBar.el = container;
-        children = controlBar.el.children;
+        controlBar.elements.buttonContainer = container;
+        children = container.children;
```
hghazzi: Once the logo work is merged, you'll probably want to rebase against master

karimJWP:
@robwalch I removed the coloring of the background circle since we're ditching it for JW8. Still waiting on improved SVGs from Design.
screen shot 2017-08-08 at 12 25 22 pm

egreaves:
test this please

src/js/view/controls/controlbar.js
```diff
@@ -442,6 +473,13 @@ export default class Controlbar {
     onFullscreen(model, val) {
         utils.toggleClass(this.elements.fullscreen.element(), 'jw-off', val);
     }
+              
+    checkDvrLiveEdge() {
```
johnBartos: use `provider.isAtLiveEdge`. Every provider his this method

karimJWP: @johnBartos it always returns false. The original seek to live implementation https://github.com/jwplayer/jwplayer/pull/2205/files#diff-da6bb0847c43fc9abf65d5ce7b9829c5R356 is consistent with my check

src/js/view/controls/controlbar.js
```diff
@@ -246,6 +269,10 @@ export default class Controlbar {
         _model.change('cues', this.addCues, this);
         _model.change('altText', this.setAltText, this);
         _model.change('customButtons', this.updateButtons, this);
+        _model.change('state', () => {
+            // Check for change of position to counter race condition where state is updated before the current position
+            _model.once('change:position', this.checkDvrLiveEdge, this);
```
egreaves: Do we need to check this every time the position changes? I feel like we can check when the user pauses, plays or seeks or the player stalls. Thoughts?

karimJWP: @egreaves i used the once listener, not the on. We need to wait for the first change on position because state change occurs before position is updated therefore the check for isLive returns false. Is the comment i left in the code unclear ? If so i'll update it

karimJWP: Inshort we only check for atLiveEdge after the first position update after a state change. We could even remove the listener for onSeeked because state changes occur after seeking but i was worries there might be an edge case where this doesn't hold

package.json
```diff
@@ -5,7 +5,10 @@
   "repository": "git@github.com:jwplayer/jwplayer.git",
   "homepage": "https://github.com/jwplayer/jwplayer",
   "license": "CC-BY-NC-SA-3.0",
-  "dependencies": {},
+  "dependencies": {
```
johnBartos: We shouldn't have anything in `dependencies`

package.json
```diff
@@ -53,6 +56,8 @@
     "load-grunt-tasks": "3.5.2",
     "loader-utils": "1.1.0",
     "mocha": "3.5.0",
+    "module": "1.2.5",
```
johnBartos: Why do we have more dev dependencies? What are `module` and `natives`?

src/js/view/controls/controlbar.js
egreaves: No, it was an oversight on my part. I saw `on`, not `once`.

package.json
```diff
@@ -63,9 +63,10 @@
     "sinon": "3.0.0",
     "source-map": "0.5.6",
     "stylelint": "~8.0.0",
+    "svg-inline-loader": "0.8.0",
     "underscore": "1.8.3",
     "url-loader": "0.5.9",
-    "webpack": "3.4.1"
+    "webpack": "^3.5.2"
```
hghazzi: We should not have the `^` for our modules

src/js/view/controls/controlbar.js
```diff
@@ -442,6 +474,13 @@ export default class Controlbar {
     onFullscreen(model, val) {
         utils.toggleClass(this.elements.fullscreen.element(), 'jw-off', val);
     }
+              
+    checkDvrLiveEdge() {
+        if (this._model.get('streamType') === 'DVR') {
+            const currentPosition = this._model.get('position');
+            utils.toggleClass(this.elements.live.element(), 'jw-dvr-live', currentPosition > dvrSeekLimit);
```
johnBartos: `>=`

src/js/view/controls/components/tooltip.js
```diff
@@ -3,7 +3,7 @@ import ariaLabel from 'utils/aria';
 import utils from 'utils/helpers';
 
 export default class Tooltip {
-    constructor(name, ariaText, elementShown) {
+    constructor(name, ariaText, elementShown, svgIcons) {
```
hghazzi: lets default `svgIcons = []` in the arguments

src/js/view/controls/components/tooltip.js
```diff
@@ -20,6 +20,14 @@ export default class Tooltip {
         this.componentType = 'tooltip';
 
         this.el.appendChild(this.container);
+        if (svgIcons && svgIcons.length > 0) {
+            var parser = new DOMParser();
```
johnBartos: `const`

src/js/view/controls/components/tooltip.js
```diff
@@ -20,6 +20,14 @@ export default class Tooltip {
...
+            var self = this;
+            svgIcons.forEach(function(svgIcon) {
```
johnBartos: use an arrow function and you don't need `self`

src/js/view/controls/components/button.js
```diff
@@ -19,6 +19,14 @@ export default function (icon, apiAction, ariaText) {
         });
     }
 
+    if (svgIcons && svgIcons.length > 0) {
+        var parser = new DOMParser();
+        svgIcons.forEach(function(svgIcon) {
```
johnBartos: arrow function `(svgIcon) => `

src/js/view/controls/components/button.js
```diff
@@ -1,6 +1,6 @@
 import UI from 'utils/ui';
 
-export default function (icon, apiAction, ariaText) {
+export default function (icon, apiAction, ariaText, svgIcons) {
```
hghazzi: Let's revisit this later, about cleaning up the constructor to make it more readble and possibly letting `svgIcons` being and array or object:

const icons = isArray(svgIcons) ? svgIcons : [svgIcons];

That way we don't have to wrap a single icon arguments when called

src/js/view/view.js
```diff
@@ -750,6 +751,13 @@ function View(_api, _model) {
         _resizeMedia();
         _responsiveListener();
     }
+  
+    function _setLiveMode(model, streamType) {
+        if (!_instreamModel) {
+            const live = (streamType === 'LIVE');
+            utils.toggleClass(_playerElement, 'jw-flag-live', live);
```
johnBartos: Unimport `utils`, use `toggleClass`, which is already imported

karimJWP:
@aliciahurst merged and ready for your UI review

PR #2206, opened 2017-08-04
Ads mode controlbar

johnBartos:
### This PR will...
Implement the new controlbar design for Ads mode

The new UI looks better

Does the UI look fine at small breakpoints (0?). In the absence of a mock I implemented what I think looks good.

https://github.com/jwplayer/jwplayer-ads-googima/pull/236 is needed for the timeslider to appear during googima ads

#### Addresses Issue(s):

JW8-209

### DIFF
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -106,7 +106,6 @@ define([
             this.nextUpToolTip = null;
 
             const timeSlider = new TimeSlider(_model, _api);
-            let volumeSlider;
             let volumeTooltip;
             let muteButton;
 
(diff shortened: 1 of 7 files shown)


src/js/view/controls/controlbar.js
```diff
@@ -199,10 +197,7 @@ define([
                 elements.cc,
                 elements.audiotracks,
                 elements.playbackrates,
-                elements.mute,
                 elements.cast,
-                elements.volume,
-                elements.volumetooltip,
```
hghazzi: @bartos So why is `volume` no longer here? Does `mute` now represent both of them?

PR #2208, opened 2017-08-04
Support Logo in Control Bar

hghazzi:
### This PR will...

Add `control-bar` as a new config value for `logo.position`, which will set the given logo image as the rightmost button after the spacer

To give customers the ability to add their logo in the control bar with the new UI

#### Addresses Issue(s):

JW8-138

### DIFF
--- a/src/css/controls/imports/controlbar.less
+++ b/src/css/controls/imports/controlbar.less
@@ -31,6 +31,7 @@
         height: 100%;
         background-position: 50% 50%;
         background-repeat: no-repeat;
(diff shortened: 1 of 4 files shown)


src/css/controls/imports/controlbar.less
```diff
@@ -31,6 +31,7 @@
         height: 100%;
         background-position: 50% 50%;
         background-repeat: no-repeat;
+        background-size: 20px;
```
radium-v: This can conflict with other icons that use `background-image`

PR #2211, opened 2017-08-07
Git Pre Push Hook

hghazzi:
### This PR will...

Trigger `eslint` and `karma` to run before any push occurs. You can still force a push by running git `push -n (or --no-verify)`, though I would avoid doing so.

**NOTE: `grunt hooks` needs to be run locally before this takes effect.**

To ensure we do not check in any code that could cause the build to fail.

Would be great to bring down the runtime of both `eslint` and `karma` as much as possible. For `karma`, I'm only running unit tests on PhantomJS and not running istanbul. Maybe we can get some speedups from making changes like the ones @robwalch made in https://github.com/jwplayer/jwplayer-ads-googima/pull/241

### DIFF
--- a/.github/hooks/pre-push
+++ b/.github/hooks/pre-push
@@ -1,0 +1,17 @@
+#!/bin/bash
+
+echo "Running ESLint & Karma"
+
+npm run lint
+lint=$?
+if [ $lint -ne 0 ]; then
+    echo "ESLint Failed. Please fix the errors and push again. \n\n\n\n"
+  exit 1
+fi
+
+karma start --browsers=PhantomJS --reporters=mocha
+karma=$?
+if [ $karma -ne 0 ]; then
+    echo "Karma Failed. Please fix the failing unit tests and push again. \n\n\n\n"
+  exit 1
+fi
--- a/README.md
+++ b/README.md
@@ -113,6 +113,12 @@ grunt karma:{BROWSER} e.g. grunt karma:chrome
 npm run lint
 ````
 
+6. Setup git pre-push hook
+# Will lint and run unit tests on the branch before any push to remote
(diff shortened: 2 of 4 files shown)


.github/hooks/pre-push
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+echo "Running ESLint & Karma"
+
+grunt lint
```
robwalch: grunt is for human use. Script with underlying commands: run `eslint ./src/js` or the `npm run lint` if you want js and less linting.

hghazzi:
@robwalch Got it, done

hghazzi:
test this please

hghazzi:
test this please

hghazzi:
test this please

hghazzi:
Test this please

README.md
```diff
@@ -113,6 +113,12 @@ grunt karma:{BROWSER} e.g. grunt karma:chrome
 npm run lint
 ````
 
+6. Setup git pre-push hook
+# Will lint you code and run unit tests every time you push to remote
```
egreaves: your*

.github/hooks/pre-push
```diff
@@ -0,0 +1,17 @@
+#!/bin/bash
+
+echo "Running ESLint & Karma"
+
+npm run lint
+lint=$?
+if [ $lint -ne 0 ]; then
+    echo "ESLint Failed. Use --no-verify to force push. Please don't. \n\n\n\n"
```
egreaves: Use `--no-verify` or `-n`? Also, if we don't want people to do this, should we even include it?

hghazzi:
test this please

PR #2215, opened 2017-08-08
Ads ui improvements

johnBartos:
### This PR will...
- Increase the progress bar height to 8px
- Use the instream model to move the progress bar

- The progress bar wasn't visible enough
- Setting the model's position & duration during ad playback caused various side effects

Is what I did to sync instream time OK?

No

#### Addresses Issue(s):

JW8-209


src/js/view/controls/controlbar.js
```diff
@@ -568,5 +568,28 @@ define([
...
+            }
+
+            instreamModel
+                .change('position', timeSlider.onPosition, timeSlider)
+                .change('duration', timeSlider.onDuration, timeSlider);
+        }
+
+        syncPlaybackTime(model) {
+            // When resuming playback mode, trigger a change so that the slider immediately resumes it's original position
+            if (!timeSlider) {
```
egreaves: Is this correct? You're checking if `timeSlider` is truthy before assigning a value to it.

src/js/view/controls/controlbar.js
```diff
@@ -396,7 +396,7 @@ export default class Controlbar {
             utils.toggleClass(this.elements.mute.element(), 'jw-off', muted);
         }
         if (this.elements.volume) {
-            this.elements.volume.render(muted ? 0 : vol);
+            this.elements.volume.render(muted ? 30 : vol);
```
hghazzi: What was this change for?

PR #2217, opened 2017-08-08
Feature/timeout javascript assets

robwalch:
### This PR will...

- Replace all module exports and imports in out source code to harmony syntax (see https://github.com/jwplayer/jwplayer/pull/2210 for only those changes)
- Add better plugin loading and error handling (See https://github.com/jwplayer/jwplayer/pull/2209 for only those changes)
- Handle setup errors in core-shim (jwplayer.js embed)
- Only load one provider at a time, prior to playback rather than all providers required by a playlist before playback can start.
- Remove  try-catch blocks from the controller -> model.provider handling that did not provide clear error handling.
- Add an error handler for all webpack chunks that include the core player
- Failed request for webpack asset needed by setup will fail setup with the error message "Error loading player: Network error" (jwplayer.core*.js)
- Failed request for provider will error one of two ways:
  - A setup error caused by the provider required to play the first item in the playlist will have a message which reads "Error loading player: Network error"
  - A player error caused by a provider required to play a subsequent playlist item or playlist will have a message that reads "Could not play video: Network error"
- Failed request for assets not essential to playback will trigger a non-fatal error event
  - A controls error event caused by jwplayer.controls.js not loading after calling setControls(true) will have a message which reads "Controls failed to load" and a reason property whose value is the Error object with message "Network Error"
  - A captions error event caused by vttparser.js not loading will have a message which reads "Captions failed to load" and a reason property whose value is the Error object with message "Network Error"
  - A captions error event caused by polyfills.webvtt.js not loading will have a message which reads "Captions renderer failed to load" and a reason property whose value is the Error object with message "Network Error"
  - A casting error event caused by provider.cast.js or provider.airplay.js not loading will have a message which reads "Casting failed to load" and a reason property whose value is the Error object with message "Network Error"

These changes allow us to handle any exception in the setup promise chain as a setup error - whether it is thrown from jwplayer.js or core. We also will not have proper error events for asset loading failure non-essential to playback.

jwplayer-commercial

#### Addresses Issue(s):

JW8-112


src/js/utils/backbone.events.js
```diff
@@ -9,7 +9,7 @@
 // succession.
 //
 //     var object = {};
-//     _.extend(object, Backbone.Events);
+//     Object.assign(object, Backbone.Events);
 //     object.on('expand', function(){ alert('expanded'); });
 //     object.trigger('expand');
 //
```
hghazzi: Can this be removed?

hghazzi: Got it

hghazzi:
@robwalch I'm seeing a bunch of warning and exceptions in the unit tests jenkins build

https://jenkins.jwplayer.com/job/jw7-commercial-unit-tests/11135/consoleFull

PR #2219, opened 2017-08-09
Support setup options events with event names

robwalch:
### This PR will...

Support setup options `.events` with event names. Written this way, the `events` block is supported by the `on` method (`api.on(options.events, null, api)`).

```js
jwplayer().setup({
    events:  {
        ready: function() { this.play(); },
        play: function(e) { console.log('play event:', e); },
        levelsChanged: function(e) { console.log('levelsChanged:', e); },
    }
});
```
and add compatibility for deprecated event handler method keys...

JW7 and earlier used event listener registration method keys in the events block. These methods have been deprecated and are not compatible with the `on` method. The compatibility script will convert these to their corresponding method names.

```js
jwplayer().setup({
    events:  {
        onReady: function() { this.play(); },
        onPlay: function(e) { console.log('play event:', e); },
        onQualityChange: function(e) { console.log('levelsChanged:', e); },
    }
});
```

#### Addresses Issue(s):

JW8-206

### DIFF
--- a/src/js/compatibility.js
+++ b/src/js/compatibility.js
@@ -100,7 +100,7 @@
             playerInstance.setup = function(options) {
                 var eventMap = options.events;
                 if (eventMap) {
-                    Object.keys(options.events).forEach(eventKey => {
+                    Object.keys(eventMap).forEach(function(eventKey) {
                         var eventName = callbackMap[eventKey];
                         if (eventName) {
                             eventMap[eventName] = eventMap[eventKey];
--- a/src/js/compatibility.js
+++ b/src/js/compatibility.js
@@ -89,6 +89,28 @@
                 };
             });
 
+            /*
+             Event maps passed to the player on setup should use event names,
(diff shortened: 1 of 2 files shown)


src/js/compatibility.js
```diff
@@ -89,6 +89,28 @@
...
+             rather than onEventName. Here "onReady" is converted
+             to "ready" in the setup options events block. Ex:
+             `jwplayer().setup({ events: { onReady: fn } })` becomes
+             `jwplayer().setup({ events: { ready: fn } })`
+             */
+            var setup = playerInstance.setup;
+            playerInstance.setup = function(options) {
+                var eventMap = options.events;
+                if (eventMap) {
+                    Object.keys(options.events).forEach(eventKey => {
```
johnBartos: Arrow function 😡

robwalch: Oops!

PR #2221, opened 2017-08-10
Unit Tests Cleanup

hghazzi:
### This PR will...

- Remove exceptions in the karma output that were not causing tests to fail
- Unnecessary console logs in the karma output
- Removed `utils.log` tests that was doing nothing
- Removed `sinon` warning when karma is run


test/files/mock.js
```diff
@@ -0,0 +1 @@
+
```
karimJWP: is this empty file needed ?

hghazzi: Yes otherwise there will be 404 errors in when running karma

karma.conf.js
```diff
@@ -88,8 +88,7 @@ module.exports = function(config) {
         files: [
             { pattern: 'test-context.js' },
             { pattern: 'test/files/*', included: false },
-            { pattern: 'node_modules/jquery/dist/jquery.js' },
-            { pattern: 'node_modules/sinon/pkg/sinon.js' }
+            { pattern: 'node_modules/jquery/dist/jquery.js' }
```
robwalch: Do we really need jquery.js embedded on the page that runs tests? I thought we import it into tests that use it (which means it's bundled in test-context.js).

test/files/mock.js
robwalch: Are you sure that wasn't to test handling of 404s?

test/unit/api-test.js
```diff
@@ -58,43 +58,46 @@ describe('Api', function() {
...
+        api.on('x', invalidEvent);
+        api.on('x', validEvent);
+        api.on('x', invalidEvent);
 
-        api.trigger('x');
+        expect(function() {
+            api.trigger('x');
+        }).to.not.throw();
 
-        assert.isOk(check, 'When events blow up, handler continues');
```
robwalch: You've changed the functionality of this test by removing the update method. Why did you make `validEvent()` and `invalidEvent()` the same?

robwalch: We are checking two things, that it does throw AND that the second event handler has run.

test/unit/helpers-test.js
```diff
@@ -16,16 +16,5 @@ describe('helpers', function() {
...
-    it('log will not thow if console is cleared', function() {
-        var tmpConsole = window.console;
-
-        window.console = null;
-        // this should not break
-        utils.log('testing');
-
-        // restore actual window console
-        window.console = tmpConsole;
-    });
```
robwalch: Removing tests because they log to the console is not necessary. What's the big deal?

robwalch:
If you didn't change package.json you should not be pushing changes to package-lock.

test/unit/helpers-test.js
hghazzi: I removed this not because it was logging to the console, it's because it's not doing anything. Before this even runs, `utils.log` is already being set to `console.log`, so it doesn't matter if `console.log` is `null`, it'll still log to the console. This was relevant before you refactored the`utils.log` to simply be a passthrough.

test/unit/api-test.js
hghazzi: Fixed the typo and reintroduced the event check.

karma.conf.js
hghazzi: Must have been reintroduced in a rebase.

test/unit/api-test.js
```diff
@@ -2,21 +2,28 @@ import jwplayer from 'jwplayer';
...
 import apiMethods from 'data/api-methods';
 import apiMethodsChainable from 'data/api-methods-chainable';
 import apiMethodsDeprecated from 'data/api-methods-deprecated';
 import Events from 'utils/backbone.events';
+import utils from 'utils/helpers';
 
 describe('Api', function() {
 
     afterEach(function() {
+        utils.log = sinon.stub();
```
robwalch: `beforeEach`?

test/unit/api-test.js
```diff
@@ -58,43 +65,42 @@ describe('Api', function() {
         assert.equal(check, false, 'api.off works');
     });
 
-    it('bad events don\'t break player', function() {
-        jwplayer.debug = false;
+    it('bad events doesn\'t break player', function() {
```
robwalch: "events do not"

robwalch: Better yet, let's rename the test "event handlers that throw exceptions to not stop event propagation".

(Setting `jwplayer.debug === true` is the only way to disable catching these exceptions)

test/unit/helpers-test.js
robwalch: The test was there to make sure "log will not thow if console is cleared", not that it "will not log". We wanted the implementation of `utils.log` to guard from throwing during player code execution if `console.log` is tampered with.

The current implementation does this by  making sure `console.log` is a function before aliasing it and if not assigning a noop.

```js
export const log = (typeof console.log === 'function') ? console.log.bind(console) : function() {};
```
https://github.com/jwplayer/jwplayer/blob/v8.0.0-alpha.2/src/js/utils/helpers.js#L17

If the test were not here, I might have just done this `log = console.log.bind(console)` and that would cause this test to fail.

hghazzi: Synched up with @johnBartos about this and looks like there's no straightforward way to test this. Don't want to sink in anymore time right now.

hghazzi:
test this please

PR #2223, opened 2017-08-11
Replace icon fonts with SVG

karimJWP:
### This PR will...
Expose rewind SVG in display and enable next in display

add back functionality that was lost while clearing merge conflicts

n/a

view tests updated in https://github.com/jwplayer/jwplayer-commercial/pull/3940
https://github.com/jwplayer/jwplayer-plugin-related/pull/157

#### Addresses Issue(s):

JW8-90

### DIFF
--- a/src/js/view/controls/next-display-icon.js
+++ b/src/js/view/controls/next-display-icon.js
@@ -1,10 +1,15 @@
 import displayIconTemplate from 'view/controls/templates/display-icon';
 import NEXT_ICON from 'assets/SVG/next.svg';
 import utils from 'utils/helpers';
+import UI from 'utils/ui';
 
 export default class NextDisplayIcon {
-    constructor(model) {
+    constructor(model, api) {
         const element = utils.createElement(displayIconTemplate('next', model.get('localization').next, NEXT_ICON));
+
+        new UI(element).on('click tap', function() {
+            api.next();
+        });
 
         model.change('nextUp', function(nextUpChangeModel, nextUp) {
             element.style.display = nextUp ? '' : 'none';
--- a/src/css/controls/flags/touch.less
+++ b/src/css/controls/flags/touch.less
@@ -17,7 +17,7 @@
     }
 
     .jw-controlbar .jw-icon-volume {
-        display: none;
+        display: flex;
     }
 
     .jw-display,
--- a/src/css/controls/flags/touch.less
+++ b/src/css/controls/flags/touch.less
@@ -17,7 +17,7 @@
     }
 
     .jw-controlbar .jw-icon-volume {
-        display: flex;
+        display: none;
     }
 
     .jw-display,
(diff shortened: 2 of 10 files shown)


karimJWP:
test this please

src/js/view/controls/next-display-icon.js
```diff
@@ -1,11 +1,16 @@
...
 import utils from 'utils/helpers';
+import UI from 'utils/ui';
 
 export default class NextDisplayIcon {
-    constructor(model) {
+    constructor(model, api) {
         const element = utils.createElement(displayIconTemplate('next', model.get('localization').next, NEXT_ICON));
 
+        new UI(element).on('click tap', function() {
+            api.next();
```
egreaves: `next` takes a `meta` argument, which by default is `{ reason: 'external' }`. In this case, it should be: `{ reason: 'interaction' }`

egreaves: Also, I noticed that `this.iconUI = ...` was removed. Is it no longer needed anywhere? If so, can we get rid of it from the rewind display icon as well?

karimJWP: @egreaves there is a playlistNext() api call that takes in a meta, https://github.com/jwplayer/jwplayer/blob/master/src/js/api/api.js#L748 ; next does not take one
the difference between the 2 is at the last item of a playlist, playlistNext will set item1, whereas next will start playing related

egreaves: @karimJWP good observation. I was looking at my local repo which is out of date.

karimJWP:
test this please

karimJWP:
@msiddique integration tests fail bc of 'Interface method initPlugin in namespace com.longtailvideo.jwplayer.plugins:IPlugin is implemented with an incompatible signature in class com.longtailvideo.jwplayer.ads.vast:VAST.' , is this specific to this PR ?

karimJWP:
test this again

robwalch:
@karimJWP The view test should show what the auto-start muted state looks like. It looks like it's not. This could be a change in the player/model state required to show it, or view logic that changes and is not being executed. I would expect to not see the controlbar, here and only see the unmute button:

screen shot 2017-08-16 at 2 27 23 pm

http://player-pr-test-jenkins.longtailvideo.com/builds/8965/archive/test/view/

hghazzi:
@karimJWP Have the comments been addressed? Also, your branch needs a rebase

karimJWP:
@robwalch updated view test. Will be available when commercial build completes
Note: we have a line in states.less that removes the button on paused. Looks weird to me but a comment was left stating 'Hide autostart mute button when paused by the browser'.
screen shot 2017-08-17 at 2 42 38 pm

karimJWP:
@egreaves and @hghazzi  all comments addressed

egreaves:
@karimJWP something's broken with the player controls. Take a look at any player.

karimJWP:
@egreaves fixed

PR #2226, opened 2017-08-12
Modify html5 supports to account for hls ads

jw-ads-team:
### This PR will...
check the "mimeType" of the source with video.canPlayType function.

HLS ads already have mimeType, so we need to check the mimeType with canPlayType function.

#### Addresses Issue(s):
ADS-423

### DIFF
--- a/src/js/providers/providers-supported.js
+++ b/src/js/providers/providers-supported.js
@@ -3,6 +3,26 @@ import { isAndroidHls } from 'providers/html5-android-hls';
 import { isYouTube, isRtmp } from 'utils/validator';
 import video from 'utils/video';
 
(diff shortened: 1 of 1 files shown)


src/js/providers/providers-supported.js
```diff
@@ -35,6 +35,7 @@ const SupportsMatrix = [
 
             var file = source.file;
             var type = source.type;
+            var mimeType = source.mimeType;
```
robwalch: Declare mimeType properly:

`const mimeType  = MimeTypes[type] || source.mimeType;`

Then you can simplify the rest of this method by just using the mime type as intended.

PR #2228, opened 2017-08-14
Enable playbackRates for mgmt api

karimJWP:
### This PR will...
add playbackRates key to config and set its default to our default rates.
Papi wants to expose a field for playbackRateControls (bool) and playbackRates (array of rates) in the management API, without breaking JW7 behavior where playbackRateControls can be a bool or array.
p/d/w @dYale, management API will omit the playbackRates key from the config if the value is null. If the value is ever null however, our default will be overriden. Do we want to protect against this? 
n/a
#### Addresses Issue(s):
JW8-240

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -13,6 +13,7 @@ const Defaults = {
     mobilecontrols: false,
     defaultPlaybackRate: 1,
     playbackRateControls: false,
+    playbackRates: [0.5, 1, 1.25, 1.5, 2],
     repeat: false,
     castAvailable: false,
     skin: '',
@@ -104,7 +105,7 @@ const Config = function(options, persisted) {
     let rateControls = config.playbackRateControls;
 
     if (rateControls) {
-        let rates = [0.5, 1, 1.25, 1.5, 2];
+        let rates = config.playbackRates;
 
         if (_.isArray(rateControls)) {
             rates = rateControls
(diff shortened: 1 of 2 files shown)


egreaves:
@karimJWP my assumption is that the delivery API should ignore `playbackRates` if it is `undefined` or an empty array, so the config defaults would still be used. (cc: @dYale)

src/js/api/config.js
```diff
@@ -104,7 +105,7 @@ const Config = function(options, persisted) {
     let rateControls = config.playbackRateControls;
 
     if (rateControls) {
-        let rates = [0.5, 1, 1.25, 1.5, 2];
+        let rates = config.playbackRates;
```
egreaves: This takes care of setting the default rates, but would fail to sanitize the rates if `playbackRates` contains array items that aren't rounded or aren't within the min/max values allowed. Eg.:

```
config.playbackRates = [0.5, 1.25, 3.23, 5];
```

You'll want to check if `rateControls` is an array and assign it to `rates`, then filter `rates`.

egreaves: Also, `playbackRates` should be used to reflect the final list of rates moving forward:
```
config.playbackRateControls = rates;  // old
config.playbackRates = rates; // new
```

karimJWP: @egreaves comments addressed

src/js/api/config.js
```diff
@@ -104,25 +105,25 @@ const Config = function(options, persisted) {
...
 
         if (_.isArray(rateControls)) {
```
egreaves: Can you update this to use `Array.isArray` in lieu of `_.isArray`? We want to stop using underscore whenever there's a native API that does the same thing.

karimJWP: @egreaves  addressed

PR #2230, opened 2017-08-14
Check autostart after calling load at the end of current event loop

robwalch:
### This PR will...

Defer the `_checkAutoStart` call made after `load` on "itemReady" until after the current callstack is done executing. If `play` or `stop` is called, the deferred `_checkAutoStart` call is cancelled.

This is done by calling `autostartFallbackOnItemReady()` to defer the check  on "itemReady" with a Promise, and is cancelled by calling `cancelAutostartFallbackOnItemReady()`.

The player should autostart after calling `load` when autostart is enabled. Typically we track the reason for these plays as autostart plays in the "playAttempt" and "play" events' "playReason".

When `play()` is called after load, before the new item is auto-started, we want to track the "playAttempt" and "play" events' "playReason" as the the reason given by the call to play(), not autostart.

This allows us to provide accurate play reasons when UI elements load and play new content with autostart enabled, without turning off autostart which would change the expected behavior when `play()` is not called following a call to `load()`.

#### Addresses Issue(s):

JW7-4465

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -72,6 +72,7 @@ define([
             var _stopPlaylist = false;
             var _interruptPlay;
             var _this = this;
+            var checkAutoStartLastContext = {};
 
             var _video = function () {
                 return _model.getVideo();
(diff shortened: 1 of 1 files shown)


dannyfinks:
test this please

robwalch:
@dannyfinks @mikesalvia 
```
Running "squash" task
/var/lib/jenkins-slave/workspace/jw7-commercial/test/squash/feature-stats.js:52
            if (err) throw err;
                     ^

Error: ENOENT: no such file or directory, open '/var/lib/jenkins-slave/workspace/jw7-commercial/test/cucumber/features/analytics_play_reason_autostart.feature'
    at Error (native)
Warning: Command failed: node parse-squashed-files.js Use --force to continue.

Aborted due to warnings.
```

dannyfinks:
@mikesalvia @robwalch its missing the directory in front of it - I'm fixing it now

robwalch:
Thanks @dannyfinks. I restarted the PR build.

robwalch:
I got this one. Jenkins expects a grunt subtask for `karma:jenkins` that wasn't added to v7.12.x.

src/js/controller/controller.js
```diff
@@ -281,6 +282,28 @@ define([
...
+            }
+
+            function cancelAutostartFallbackOnItemReady() {
+                checkAutoStartLastContext.bail = true;
+                _model.off('itemReady', checkAutoStartLast);
+            }
+
+            function checkAutoStartLast() {
+                // Use promise as setImmediate() to allow synchonous calls to load() and play() set the playReason
+                Promise.resolve().then(() => {
```
egreaves: Interesting. I'd long wondered if there was a browser equivalent of `setImmediate`.  Seems only IE/Edge will ever implement it.

PR #2235, opened 2017-08-15
Implement Settings Menu

johnBartos:
### This PR will...
Implement the basic Settings Menu overlay:
- Gear icon
- DOM structure
- Basic styles
- Open/close behavior

Existing menus will be consolidated into one in order to save controlbar space

- Are there any points where settingsMenu can be undefined, causing exceptions?
- Are there any optimizations to the code I've wrote?

No

#### Addresses Issue(s):

 JW8-92


dannyfinks:
@johnBartos one of the A.C. missing here is to display a line below the gear when the menu is opened.

src/css/controls/imports/settings-menu.less
```diff
@@ -0,0 +1,61 @@
...
+
+.jw-settings-menu {
+    display: none;
+    position: absolute;
+    z-index: -1;
+    flex-flow: column wrap;
+    align-items: flex-start;
+    bottom: @controlbar-height;
+    right: 0;
+    height: ~"calc(100% - @{controlbar-height})";
```
egreaves: By setting `height`, `width` and `right` to these values, we'd be optimizing for the small-player breakpoints. Does the most common player size fall within breakpoint 0-2?

Alternatively, if you make the defaults the values for breakpoints 5-7, you'd reduce the css output.

src/js/view/view.js
```diff
@@ -838,6 +848,17 @@ function View(_api, _model) {
         }
     }
 
+    const settingsMenuVisible = () => {
+        let visible = false;
+        if (_controls) {
+            const settingsMenu = _controls.settingsMenu;
+            if (settingsMenu && settingsMenu.visible) {
```
egreaves:
```
const settingsMenu = _controls && _controls.settingsMenu;
return !!(settingsMenu && settingsMenu.visible);
```

src/js/view/view.js
```diff
@@ -482,23 +482,32 @@ function View(_api, _model) {
...
         const clickHandler = new ClickHandler(model, videoLayer, { useHover: true });
+        const controls = _model.get('controls');
         clickHandler.on({
             click: () => {
                 _this.trigger(DISPLAY_CLICK);
-                if (_model.get('controls')) {
-                    api.play(reasonInteraction());
+
+                if (_controls) {
+                    if (settingsMenuVisible()) {
```
egreaves: This would be a change to expected behavior. We can close the menu if it's open, similar to what you do for a tap, and always call `api.play(reasonInteraction());`.

src/js/view/controls/controls.js
```diff
@@ -308,12 +331,11 @@ export default class Controls {
 
     userInactive() {
         clearTimeout(this.activeTimeout);
-        this.showing = false;
-        if (this.controlbar) {
-            this.controlbar.closeMenus({
-                type: 'userInactive'
-            });
+        if (this.settingsMenu.visible) {
```
egreaves: What happens if the volume tooltip menu is open? 

Also, if you're hovering over the control bar, the `user-inactive` flag does not get added after the timeout.
screen shot 2017-08-16 at 1 20 42 am

src/js/view/view.js
johnBartos: The AC specifically says not to pause the player if the display is clicked and the menu is open.

`When the menu is open, anywhere the the click to pause functionality works should instead close the menu and not pause the player`

src/js/view/controls/controls.js
johnBartos: `What happens if the volume tooltip menu is open?`
Not sure what you mean by this

`Also, if you're hovering over the control bar, the user-inactive flag does not get added after the timeout.`

This is existing behavior (I verified with build 180).

src/css/controls/imports/settings-menu.less
johnBartos: I did this because it makes more sense from an organizational standpoint; and because we'd be applying less styles to mobile devices, which have less processing power.

johnBartos:
@dannyfinks Making a border as it is in the mock is pretty hard, and the simple approach doesn't look quite right. I don't want to waste time on something trivial right now.

src/js/view/controls/controls.js
```diff
@@ -107,6 +111,25 @@ export default class Controls {
         this.addActiveListeners(controlbar.element());
         this.div.appendChild(controlbar.element());
 
+        // Settings Menu
+        const visibilityChangeHandler = (visible) => {
+            utils.toggleClass(this.div, 'jw-settings-open', visible);
+            // Trigger userActive so that a dismissive click outside the player can hide the controlbar
+            // this.userActive();
```
hghazzi: @johnBartos Why is this commented out?

robwalch:
When we bring content down to above the controlbar, can we bring it down to the slider? It looks better and if there should be padding or a margin the thing we're placing above the slider can include that:

Making room for the dial looks like wasted space here
screen shot 2017-08-16 at 2 11 54 pm

This doesn't interfere with the slider it just looks better:
screen shot 2017-08-16 at 2 11 24 pm

johnBartos:
@robwalch I agree that looks better. Shouldn't be difficult to implement.

src/js/view/view.js
egreaves: Fair point.

PR #2241, opened 2017-08-17
Move MediaModel to Controls

hghazzi:
### This PR will...

Move the media mdoel change logic for quality and audio from the control bar and into controls

Since the quality and audio tracks menus are being moved out of the control bar, the media model logic should no longer be there. 

#### Addresses Issue(s):

JW8-311

### DIFF
--- a/src/js/view/controls/components/settings/settings-menu.js
+++ b/src/js/view/controls/components/settings/settings-menu.js
@@ -3,7 +3,7 @@ import button from 'view/controls/components/button';
 import SettingsMenuTemplate from 'view/controls/templates/settings/settings-menu';
 import { createElement } from 'utils/dom';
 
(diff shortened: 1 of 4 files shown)


src/js/view/controls/components/settings/settings-menu.js
```diff
@@ -3,7 +3,7 @@ import button from 'view/controls/components/button';
 import SettingsMenuTemplate from 'view/controls/templates/settings/settings-menu';
 import { createElement } from 'utils/dom';
 
-export default function SettingsMenu(visibilityChangeHandler) {
+export function SettingsMenu(visibilityChangeHandler) {
```
johnBartos: Why did you undefault this?

PR #2242, opened 2017-08-17
Settings submenu templates and components

johnBartos:
### This PR will...
- Create settings submenu templates
- Create settings submenu factories
- Refactor settings menu creation

- So that we can easily create submenus
- So that we can parallelize creation of submenus

No

No

#### Addresses Issue(s):

JW8-303


hghazzi:
A little pet peeve of mine. If we are gonna have a `settings` directory, no need to append the `settings-` to the file name.

PR #2244, opened 2017-08-17
Implement audioTracks submenu & default behavior

johnBartos:
### This PR will...
- Implement default behavior for submenus
- Implement `audioTracks` as an example

Audiotracks will be accessed from a settings submenu

A onceover on the logic would be nice

No

#### Addresses Issue(s):
Part of JW8-282


dannyfinks:
@egreaves @johnBartos should we be merging these WIP items into master? I think we should have a Do Not Merge so we can keep master clean.

dannyfinks:
Unable to close the menu from the close icon:
TypeError: menu__this2.close is not a function
 /builds/lastSuccessfulBuild/archive/bin-debug/jwplayer.core.controls.js:10308

Various streams that do not have audio tracks menu now show the audio track icon and show an empty menu.

PR #2258, opened 2017-08-21
Fix various settings menu bugs

johnBartos:
### This PR will...
- Close & destroy settings on controls `disable`
- Close & hide settings when entering instream mode
- Place the Settings button to the left of fullscreen

The settings menu must be closed and hidden when inaccessible. It's resources should be freed when destroyed.

No

No

#### Addresses Issue(s):

JW8-92


johnBartos:
Builds will fail until tests which rely on old menu functionality are removed

PR #2263, opened 2017-08-22
Make quality submenu the leftmost submenu

johnBartos:
### This PR will...
- Make the default submenu the leftmost submenu
- Make quality the default submenu
- A bit of code cleanup

It is better UX to have the default submenu visually first

No

No

#### Addresses Issue(s):

JW8-250

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -148,3 +148,7 @@ export function bounds(element) {
 
     return boundsRect;
 }
+
+export function prependChild(parentElement, childElement) {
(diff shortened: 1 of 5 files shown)


src/js/utils/dom.js
```diff
@@ -148,3 +148,7 @@ export function bounds(element) {
 
     return boundsRect;
 }
+
+export function prependChild(parentElement, childElement) {
+    parentElement.insertBefore(childElement, parentElement.firstChild);
```
hghazzi: Nice! Can use this for the custom buttons work and clean it up a bit :)

PR #2271, opened 2017-08-22
Remove youtube and RTMP support

karimJWP:
### This PR will...
* Error out when youtube or rtmp stream is loaded
* Updated youtube provider unit test to assert that a provider is not chosen for a youtube video
* Remove skins tests and assets left over from skins deprecation work
* Added manual tests for thumbnails and players with localized text

Removes youtube and rtmp support
Could I short-circuit at an earlier time ? 
https://github.com/jwplayer/jwplayer-commercial/pull/3962
#### Addresses Issue(s):
JW8-256


src/js/providers/providers.js
```diff
@@ -62,6 +55,9 @@ Object.assign(Providers.prototype, {
         // prevent throw on missing source
         source = (source === Object(source)) ? source : {};
 
+        if (source.type === 'youtube' || source.type === 'rtmp') {
+            return null;
```
robwalch: @karimJWP This `choose` method's job is to see if any providers support the source and return that provider. It should not make any decisions like this itself. Providers can be added to the player that have their own logic.

PR #2275, opened 2017-08-23
Prevent controls disappearing for ads flag

radium-v:
### This PR will...
Prevent the entire controls container from disappearing when in certain UI states.

#### Addresses Issue(s):
JW8-95

### DIFF
--- a/src/css/controls/imports/jwplayerlayout.less
+++ b/src/css/controls/imports/jwplayerlayout.less
@@ -17,7 +17,7 @@
     transition: opacity 250ms cubic-bezier(0, 0.25, 0.25, 1);
     z-index: 0;
 
-    .jwplayer.jw-state-playing.jw-flag-user-inactive:not(.jw-flag-casting) & {
+    .jwplayer.jw-state-playing.jw-flag-user-inactive:not(.jw-flag-casting):not(.jw-flag-ads) & {
         opacity: 0;
     }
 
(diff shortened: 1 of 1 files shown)


src/css/controls/imports/jwplayerlayout.less
```diff
@@ -17,7 +17,7 @@
     transition: opacity 250ms cubic-bezier(0, 0.25, 0.25, 1);
     z-index: 0;
 
-    .jwplayer.jw-state-playing.jw-flag-user-inactive:not(.jw-flag-casting) & {
+    .jwplayer.jw-state-playing.jw-flag-user-inactive:not(.jw-flag-casting):not(.jw-flag-ads) & {
```
egreaves: We should look into reducing the specificity by removing `.jwplayer`. 

Also, though the skip button should be visible, the other controls aren't visible during ad playback when `.jw-flag-user-inactive` is present.

radium-v: The `.jw-controlbar` element has its own transition state that's triggered when the `.jw-flag-user-inactive` is activated.

hghazzi:
test this please

egreaves:
test this please

src/css/controls/imports/jwplayerlayout.less
egreaves: Pulled it down and tested it. You're right.

PR #2278, opened 2017-08-23
Remove skin util from embed string

hghazzi:
### This PR will...

Normalize the skin config before customizing UI and not when loading the player

Decrease the size of the embed script since it will no longer include `utils/skin.js`


src/js/api/config.js
```diff
@@ -89,13 +87,8 @@ const Config = function(options, persisted) {
     config.aspectratio = _evaluateAspectRatio(config.aspectratio, config.width);
 
     if (_.isObject(config.skin)) {
-        config.skinColors = normalizeSkin(config.skin);
         config.skinUrl = config.skin.url; // get skin name if it exists
-        config.skin = _.isString(config.skin.name) ? config.skin.name : Defaults.skin; // get skin name if it exists
```
robwalch: `config.skin` is an object so why store `skinName` in the config and model? You always normalize the model skin object so this is unnecessary.

src/js/view/view.js
```diff
@@ -259,7 +259,8 @@ function View(_api, _model) {
             addClass(_playerElement, 'jw-ie');
         }
 
-        handleColorOverrides(_model.get('id'), _model.get('skinColors'));
+        const skinColors = normalizeSkin(_model.get('skin'));
```
robwalch: What happens when `_model.change('skin', onSkinChange, this);` is called and `_model.get('skin')` is undefined? Are we getting a `jw-skin-undefined` class added to the player?

PR #2285, opened 2017-08-28
Sharing Submenu

hghazzi:
### This PR will...

Change the Sharing UI into a submenu

To match the new UI design

No

https://github.com/jwplayer/jwplayer-commercial/pull/3971

#### Addresses Issue(s):

JW8-361


src/css/controls/imports/settings-menu.less
```diff
@@ -84,6 +84,7 @@
     line-height: 1;
     padding: 7px 0 7px 15px;
     cursor: pointer;
+    outline: none;
```
johnBartos: We want to leave the focus rings in

PR #2298, opened 2017-08-29
Recolor scrollbar for visibility 

johnBartos:
### This PR will...
Change scrollbar color to flat white

So that the scrollbar is clearly visible

No

No

#### Addresses Issue(s):

JW8-379

### DIFF
--- a/src/css/controls/imports/settings-menu.less
+++ b/src/css/controls/imports/settings-menu.less
@@ -73,9 +73,9 @@
     }
 
     &::-webkit-scrollbar-thumb {
-        color: #171717;
-        border: 1px solid #323232;
+        background-color: @white;
         border-radius: 6px;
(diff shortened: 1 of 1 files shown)


egreaves:
test this please

egreaves:
@aliciahurst take a look at [this](http://player-pr-test-jenkins.longtailvideo.com/builds/9252/archive/test/squash/bootstrap.html?primary=html5&edition=ads&feature=player_ui/skins_kitchen_sink&suite=Player%20UI%20and%20Customization). My concern is that it might be challenging to select a sub-menu item on mobile devices with each item being just 26 px tall.

aliciahurst:
@egreaves My suggestion would be, for touch devices, to increase line-height to 44px (or 3.25 to not use units) and increase the font size to 14px.

PR #2304, opened 2017-08-30
Don't use related to execute next action

johnBartos:
### This PR will...
Call `item()` internally with the next item when clicking the next button

Related executes actions async, which causes a the user gesture to be lost between playlist items, preventing autostart.

- Need to go through the related suite, checking the transition between playlist/related
- Need to make sure the `playReason` is correct (it should be `interaction`)

No

#### Addresses Issue(s):

JW8-390

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -670,8 +670,8 @@ Object.assign(Controller.prototype, {
                         itemsShown: [ nextUp ],
                         feedData: nextUp.feedData,
                     });
+                    _item(nextUp.index);
                 }
-                related.next();
             }
         }
 


src/js/controller/controller.js
```diff
@@ -670,8 +670,8 @@ Object.assign(Controller.prototype, {
                         itemsShown: [ nextUp ],
                         feedData: nextUp.feedData,
                     });
+                    _item(nextUp.index);
```
robwalch: This should work for playlist items where `nextUp.index` equals `jwplayer().getPlaylistIndex() + 1`.

If `nextUp` is a recommendation and not part of the playlist, then `_load(nextUp)` should be called to replace the playlist.

Make sure that `feedData` is also set correctly in this scenario, as well as any other logic that is part of related's `selectItem` "next" action that is exposed publicly (plugin events, etc...).

johnBartos:
JK its actually fine

PR #2308, opened 2017-08-31
Switch gradient back to pseudo element

radium-v:
### This PR will...
Switch the controls gradient to use an `::after` pseudoelement

We kept running into too many weird flag-specific bugs when the gradient was applied as a background image on the `.jw-controls` element and this method allows us to control the gradient and the controlbar states independently.

#### Addresses Issue(s):

JW8-397

### DIFF
--- a/src/css/controls/flags/ads.less
+++ b/src/css/controls/flags/ads.less
@@ -11,14 +11,6 @@
         display: none;
     }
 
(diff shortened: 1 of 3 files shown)


egreaves:
@radium-v let's endeavor to provide constructive information in the PR description.

PR #2309, opened 2017-08-31
Timeslider Tooltip improvements

egreaves:
### This PR will...
- Set width for Sprited images and text
- Don’t display thumbnails below breakpoint 2
- Make the display icons the same size at breakpoint 2&3 (unrelated bug)
When the player is resized and crosses the 1 <> 2 threshold, the width of the tooltip didn't change. This also addresses ensuring the tooltip isn't flush with the edge of the video's container. 

In future, we should address keeping the arrow directly above the pointer. 
No.
No.
#### Addresses Issue(s):

JW8-377

### DIFF
--- a/src/js/view/controls/components/thumbnails.mixin.js
+++ b/src/js/view/controls/components/thumbnails.mixin.js
@@ -61,6 +61,7 @@ const ThumbnailsMixin = {
                 url = matched[1];
                 style.backgroundPosition = (matched[2] * -1) + 'px ' + (matched[3] * -1) + 'px';
                 style.width = matched[4];
+                this.timeTip.setWidth(+style.width);
                 style.height = matched[5];
             } catch (e) {
                 // this.vttFailed('Could not parse thumbnail');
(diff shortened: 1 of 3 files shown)


src/js/view/controls/components/thumbnails.mixin.js
```diff
@@ -82,7 +83,8 @@ const ThumbnailsMixin = {
     },
 
     showThumbnail: function(seconds) {
-        if (this.thumbnails.length < 1) {
+        // Don't attempt to set thumbnail for small players or when a thumbnail doesn't exist
+        if (this._model.get('containerWidth') <= 420 || this.thumbnails.length < 1) {
```
radium-v: This only applies when the player is set up with a width smaller than 420, not if it's resized to a smaller breakpoint

egreaves: The model gets the updated player width when the player gets resized. The squash breakpoint toggles aren't working correctly atm.

johnBartos: What if the thumbnail is open while the player resizes?

egreaves: Highly unlikely that the user is hovering over the timeslider and resizing the player at the same time. Even then, there'll be eventual consistency once the model gets updated.

PR #2312, opened 2017-08-31
Move logo up when the control bar is visible

egreaves:
### This PR will...

- Move the logo up when the control bar is visible if it's positioned in the bottom-right or bottom-left
- Add var for transition function 
- Remove `jw-flag-casting` from control bar visibility rule since `jw-flag-cast-available` is always present whether casting or not

The logo may be a link and shouldn't obscure other controls.
No.
No.
#### Addresses Issue(s):
JW8-243

### DIFF
--- a/src/css/jwplayer/imports/logo.less
+++ b/src/css/jwplayer/imports/logo.less
@@ -39,6 +39,6 @@
 
 .jw-logo-bottom-left,
 .jw-logo-bottom-right {
-    bottom: 44px;
-    transition: bottom 150ms cubic-bezier(0, -0.25, 0.25, 1);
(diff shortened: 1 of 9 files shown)


src/css/jwplayer/imports/logo.less
```diff
@@ -30,11 +30,15 @@
...
 }
 
 .jw-logo-bottom-right {
-    bottom: 0;
     right: 0;
+}
+
+.jw-logo-bottom-left,
+.jw-logo-bottom-right {
+    bottom: 44px;
```
radium-v: This should be `@mobile-touch-target`

PR #2314, opened 2017-08-31
Airplay Tooltip

hghazzi:
### This PR will...

- Use the localized airplay text and not the chromecast text in the tooltip on hover.
- Only create the cast button in environments that support airplay or chromecast.

To do a browser check to verify casting device

#### Addresses Issue(s):

JW8-409


src/js/view/controls/controlbar.js
```diff
@@ -189,7 +189,9 @@ export default class Controlbar {
         const nextUpTip = SimpleTooltip(elements.next.element(), 'next', localization.nextUp);
         SimpleTooltip(elements.rewind.element(), 'rewind', localization.rewind);
         SimpleTooltip(elements.settingsButton.element(), 'settings', localization.settings);
-        SimpleTooltip(elements.cast.element(), 'chromecast', localization.cast);
+
+        const cast = !Browser.chrome || OS.iOS ? 'airplay' : 'cast';
```
egreaves: Chrome on iOS would show 'airplay'.

PR #2315, opened 2017-08-31
Fixed nextUp tooltip text 

dYale:
### This PR will...
Render a generic tooltip message when no title is found in nextUp object.

Right now the text renders a colon with empty whitespace in anticipation for the title which does not exist.

No

No

#### Addresses Issue(s):

JW8-413

### DIFF
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -257,7 +257,7 @@ export default class Controlbar {
         _model.on('change:captionsList', onCaptionsChanged, this);
         _model.change('nextUp', (model, nextUp) => {
             let tipText = localization.nextUp;
-            if (nextUp) {
+            if (nextUp && nextUp.title) {
                 tipText = (`NEXT: ${nextUp.title}`);
             }
             nextUpTip.setText(tipText);


src/js/view/controls/controlbar.js
```diff
@@ -257,8 +257,10 @@ export default class Controlbar {
         _model.on('change:captionsList', onCaptionsChanged, this);
         _model.change('nextUp', (model, nextUp) => {
             let tipText = localization.nextUp;
-            if (nextUp) {
+            if (nextUp && nextUp.title) {
                 tipText = (`NEXT: ${nextUp.title}`);
+            } else {
+                tipText = (`NEXT UP`);
```
johnBartos: Remove the else, L259 handles the default case with localization.

dYale: 👍 
=

johnBartos:
It may be wise to use the generic copy for when the string length is also 0.

dYale:
If string length is 0, wouldn't we want to use our default?

johnBartos:
Yes, but a 0 length string is truthy. Would have to do:

`if (nextUp && nextUp.title && nextUp.title.length > 0)`

dYale:
```
!!" "
true
!!""
false
```

johnBartos:
Oh ho ho, I must be getting rusty

PR #2316, opened 2017-08-31
Bring logo above controls

radium-v:
### This PR will...
Fix the z-index for display logos

Logos were being obscured by the controls element's background

#### Addresses Issue(s):
JW8-243

### DIFF
--- a/src/css/jwplayer/imports/logo.less
+++ b/src/css/jwplayer/imports/logo.less
@@ -13,6 +13,7 @@
     right: auto;
     left: auto;
     bottom: auto;
+    z-index: 1;
 }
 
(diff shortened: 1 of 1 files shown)


johnBartos:
Should we do this for the non-idle states too?

johnBartos:
screen shot 2017-08-31 at 16 50 26
screen shot 2017-08-31 at 16 50 10

egreaves:
@radium-v @johnBartos the logo (or bug as it's called on TV) is considered part of the video. It's present even when the controls aren't. It's hierarchy is the way it is for a reason because the controls aren't always present.

PR #2322, opened 2017-09-01
Audiotracks and levels now use .change instead of on.('change')

dYale:
### This PR will...
Ensure that setControls renders all desired control bar elements

Right now, setting setControls to false and then true doesn't render the quality levels setting or the audio tracks setting. 
No
No
#### Addresses Issue(s):

JW8-190


egreaves:
test this please

PR #2324, opened 2017-09-02
Include jwplayer.defaults in compatibility script

robwalch:
### This PR will...

Add `jwplayer.defaults` to the updated `jwplayer` library when the compatibility script is added after a libraries script.

libraries/ end-point includes jwplayer.js which defines `jwplayer` followed by `jwplayer.defaults`, default setup config settings. Since the compatibility script replaces `jwplayer`, `jwplayer.defaults` needs to be added to the replacement.

I noticed this was needed while doing some JW8 mapping of v7 libraries on websites using the player.


dannyfinks:
nice catch

PR #2328, opened 2017-09-04
Use background color for time slider cues

egreaves:
### This PR will...
- Use background color for time slider cues
- Add custom styles conditionally
- We want the player to be monochromatic by default and the cues were blue.
- Styles were being added for a player instance when there were no custom styles specified in the config. Also, `_.size` counts undefined object properties. The `addStyle` function does not add `undefined` styles, so this check was not necessary to begin with.
No.
No.
#### Addresses Issue(s):
JW8-428

### DIFF
--- a/src/css/controls/imports/slider.less
+++ b/src/css/controls/imports/slider.less
@@ -128,7 +128,7 @@
     }
 
     .jw-cue {
-        background-color: @accent-color;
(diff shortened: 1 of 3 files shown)


egreaves:
This broke some tests in commercial. I'll fix them shortly.

Issue #2054, opened 2017-05-10
Analyze impact of temporary DOM element creation for utils


johnBartos:
In `utils/dom` and `utils/css` (and perhaps elsewhere), we create a temporary `div` and read/write its `innerHTML` as a hack to decode strings. We should analyze the performance of this operation and if necessary, use only one "dummy" div per player.

robwalch:
What we need to do now is have a look at DOM creation in the controlbar and menu. By composing these elements from many smaller HTML components that each parse text into element instances we incur a lot of over head parsing each button and container. It would be great to have templates that generate more elements at once, or compose the HTML string another way before doing only one parse and then adding to controls.

robwalch:
@egreaves @johnBartos @radium-v @hghazzi @karimJWP 
Avoiding html/svg parsing altogether in some cases may also help. Both innerHTML and DOMParser create document in memory to perform the parsing and block until done.

While profiling, 80ms for controls creation makes a good target for optimization. Because these svg icons were loaded as fonts, their rendering did not block the completion of setup, and they do here:
<img width="786" alt="screen shot 2017-08-22 at 10 50 46 pm" src="https://user-images.githubusercontent.com/333258/29596750-90c0e7b0-878c-11e7-95c3-05d2ffb49598.png">

Controlbar, menu and sharing require a lot of work up front. Lazy DOM creation for menu and sharing which aren't even displayed at ready would also speed things up:
<img width="671" alt="screen shot 2017-08-22 at 10 51 20 pm" src="https://user-images.githubusercontent.com/333258/29596756-95327052-878c-11e7-81c2-14976da17669.png">

Here we can see the pattern for each button in the controlbar. Batching this DOM creation and parsing into a single template could have big benefits:
<img width="616" alt="screen shot 2017-08-22 at 10 51 39 pm" src="https://user-images.githubusercontent.com/333258/29596767-9a9d30fe-878c-11e7-8918-70068e588247.png">

robwalch:
I've already made a few passes at improving performance by grouping parsing of html. The last round was done as part of security enhancements around dealing with html input and merging related into jwplayer-commercial https://github.com/jwplayer/jwplayer-commercial/commit/3aaef55947b52c7be779b21638b18eb469d3b215

PR #2329, opened 2017-09-04
UI Optimizations

robwalch:
### This PR will...
- Combine display icon button templates in display container template
- Create DOM elements with `document.create` and `className =` instead of parsing small templates for simple tooltips and custom buttons
- Improve how controlbar handles updates to custom buttons. It was parsing and replacing the related playlist button on each state change from related. Now it only replaces buttons when a button with a new id or button class is introduced
- Combine parsing of multiple SVG elements into one call and extract icons with query selectors
- Don't add styles to the DOM when there are no color customization options set

`utils.createElement` (and any code that results in "Parse HTML" operations) is costly memory and processor wise. Reducing the number of calls by combining templates or creating 1-2 elements at a time with `document.createElement('div')` is better than using templates for simple DOM structures

There is only one display container and set of icons, so it doesn't make sense to parse the icon buttons outside of parsing the container - It's one component. Parsing it at once reduces the number of nodes and memory required to enable controls.

Simple tooltips and custom buttons are composed of two elements (or one container and an SVG document). The DOM elements for these components are created much faster and with less memory overhead by avoiding html parsing.

JW8-436


src/js/view/controls/controlbar.js
```diff
@@ -499,10 +519,17 @@ export default class Controlbar {
 
     updateButtons(model, newButtons = [], oldButtons = []) {
         const buttonContainer = this.elements.buttonContainer;
-        this.removeButtons(buttonContainer, oldButtons);
 
-        for (let i = newButtons.length - 1; i >= 0; i--) {
-            let buttonProps = newButtons[i];
+        const removedButtons = oldButtons.filter(oldButton =>
+            !newButtons.some(newButton => newButton.id + newButton.btnClass === oldButton.id + oldButton.btnClass));
```
karimJWP: `includes` definitely wouldn't work here ?

robwalch: Includes is ES6 and I'm not sure how or if Babel would transform it https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes#Specifications

Also the object instances are likely not the same, so comparing properties (`id` and `btnClass`) with `some` gets it done.

robwalch:
```js
[3].includes(3)
// true

[{a: 3}].includes({a: 3})
// false

[{a: 3}].some(o => o.a === {a: 3}.a)
// true
```

karimJWP: 👍

robwalch:
These changes address some issues in #2054

PR #2337, opened 2017-09-05
Update right-click styling

radium-v:
### This PR will...
Update the right-click styling to match the newest designs.

#### Addresses Issue(s):
JW8-182

### DIFF
--- a/src/css/controls/imports/rightclick.less
+++ b/src/css/controls/imports/rightclick.less
@@ -1,8 +1,7 @@
 @rightclick-primary-text: #fff;
-@rightclick-secondary-text: #777;
-@rightclick-bg: #000;
-@rightclick-light-bg: #252525;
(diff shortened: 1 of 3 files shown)


src/css/controls/imports/rightclick.less
```diff
@@ -14,57 +13,52 @@
...
+            .jw-rightclick-link {
+                border: none;
                 color: @rightclick-primary-text;
-                text-decoration: none;
-
-                // padding comes from the anchor, so that the whole item is clickable
-                padding: 1em;
                 display: block;
-                font-size: (11 / 16em);
+                font-size: 11px;
```
egreaves: Should this be `0.6875em`?

radium-v: IE and Edge chop values down to the nearest hundredth by default

PR #2338, opened 2017-09-06
Ensure the order of related, sharing and custom controlbar buttons

robwalch:
### This PR will...

Ensure that the related button is always added next to settings button, and that sharing is always added next to related if present, otherwise next to settings.

Fixes a regression with button order added by optimizations that no longer remove and recreate and add all buttons with each change in the button list.

#### Addresses Issue(s):

JW8-461


egreaves:
test this please

PR #2345, opened 2017-09-08
Map controlbar background color to gear menu

radium-v:
### This PR will...
Map custom styles for the settings menu so the top bar matches the controlbar background.

The forced opacity applied to the settings submenu caused visual artifacts in certain cases. This removes that opacity.

```
"skin": {
    "active": "red",
    "tooltips": {
      "text": "pink"
    },
    "menus": {
      "text": "rgba(0,255,255,1)",
      "textActive": "#FFFF00",
      "background": "#FFA500"
    },
    "timeslider": {
      "progress": "rgb(255,0,255)",
      "rail": "rgba(255,165,0,1)"
    },
    "controlbar": {
      "icons": "#008000",
      "iconsActive": "yellow",
      "text": "#00F",
      "background": "#c0c0c0"
    }
  }
```

#### Addresses Issue(s):

JW8-450

### DIFF
--- a/src/js/view/utils/skin.js
+++ b/src/js/view/utils/skin.js
@@ -1,5 +1,5 @@
 import { prefix } from 'utils/strings';
-import { css } from 'utils/css';
+import { css, getRgba } from 'utils/css';
 
 export function normalizeSkin(skinConfig = {}) {
 
@@ -156,7 +156,8 @@ export function handleColorOverrides(playerId, skin) {
         }
 
         addStyle([
-            '.jw-controlbar'
+            '.jw-controlbar',
+            '.jw-settings-topbar'
         ], 'background', config.background);
     }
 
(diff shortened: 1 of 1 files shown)


src/js/view/utils/skin.js
```diff
@@ -156,30 +156,30 @@ export function handleColorOverrides(playerId, skin) {
...
     }
 
     function styleTimeslider(config) {
 
         addStyle([
             '.jw-progress',
-            '.jw-buffer',
             '.jw-knob'
-        ], 'background', 'none ' + config.progress);
+        ], 'background-color', config.progress);
```
egreaves: By only setting `background-color`, this would not remove other styles applied to the background. The goal here is to reset the background and add the customization.

radium-v: It's unlikely that other background properties would be applied to the time slider. The customization options are intended for setting background color in general.

PR #2346, opened 2017-09-08
Alleviate text flicker during animations in Firefox

radium-v:
### This PR will...
Enforce `transform-style` to prevent flickering text on Firefox when animations/transitions are happening.

Firefox on MacOS on Retina screens has trouble switching between composite layers.

#### Addresses Issue(s):
JW8-438

### DIFF
--- a/src/css/controls/imports/controlbar.less
+++ b/src/css/controls/imports/controlbar.less
@@ -195,5 +195,9 @@
 .jwplayer:not(.jw-flag-user-inactive) {
     .jw-controlbar {
         will-change: transform;
+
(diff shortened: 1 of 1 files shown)


egreaves:
Something to be mindful of w.r.t. `will change`:

> Excessive use of will-change will result in excessive memory use and will cause more complex rendering to occur as the browser attempts to prepare for the possible change. This will lead to worse performance. (Src: [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/will-change#Browser_compatibility))

PR #2347, opened 2017-09-09
Click to Play Setup

### LINKED ISSUE #896: Determine if video tag requires interaction to start playback

robwalch:
### This PR will...

- Create/get media element on setup and call `load()` to unblock play when setup is called from a user-gesture event
- Make the model responsible for replacing video tags between provider instances
- Update html5 provider to use video mixins and return play promise
- Call `video.play()` on "playAttempt" rather than "bufferFull"
- Make rejected play promise trigger "playAttemptFailed" event (player should go to "pause" state when "playAttempt" fails)
- "playAttempt" is reattempted on next call to `controller.play()`, updating QoE
- Fix html5 provider bundling for iOS when first source is mpd
- Do not set "preload=none" attribute on video tag. It messes up play Promise on iOS. Preload "none" is achieved by not setting `video.src` on `provider.init()`
- Load additional providers in model where active provider is set
- LOADING state is set in providers when video tag fires "play" and on mediaModel when loading outside the provider occurs
- PAUSED state is set in provider when video tag fires "pause" per listener mixin rules (mediaModel should reflect `video.paused` when not loading or seeking)
- Added play promise polyfill to providers for older browsers
- Fixed initial visibility check on iOS
- Eliminate extended IDLE state between playlist items, while loading playlist/provider
    - Player state is set to BUFFERING while loading playlists and provider instead of leaving the player in IDLE state
    - Player state is reset to IDLE between playback items and modes
- Setup sideloaded tracks after setting the video source
- Remove tech-debt:
  - No longer require "bufferFull" event to initiate playback
  - Remove QoE timer clock hack
  - Remove state attributes added to video tag: "jw-gesture-required" and "jw-loaded"
  - Prevent calls to `preload()` on providers after "playAttempt" (not IDLE)
  - Simplify resetting of error state to idle state
  - Remove provider loading logic form instream handled by the model (or adModel in this case)
  - Use promise chain from `controller.load()` instead of listening for "ItemReady" allowing us to handle rejection for:
    - Playlist failing to load or parse
    - Provider failing to load or instantiate
    - Error setting playlist item

https://github.com/jwplayer/jwplayer-commercial/pull/4027

#### Addresses Issue(s):

JW8-19
Closes #896

### DIFF
--- a/src/js/api/core-loader.js
+++ b/src/js/api/core-loader.js
@@ -50,12 +50,13 @@ export function requiresPolyfills() {
 export function requiresProvider(model, providerName) {
     const playlist = model.get('playlist');
     if (Array.isArray(playlist) && playlist.length) {
-        const firstSource = Item(playlist[0]).sources[0];
-        if (firstSource) {
+        const sources = Item(playlist[0]).sources;
+        for (let i = 0; i < sources.length; i++) {
+            const source = sources[i];
             const providersManager = model.getProviders();
-            for (let i = 0; i < ProvidersSupported.length; i++) {
-                const provider = ProvidersSupported[i];
-                if (providersManager.providerSupports(provider, firstSource)) {
+            for (let j = 0; j < ProvidersSupported.length; j++) {
+                const provider = ProvidersSupported[j];
+                if (providersManager.providerSupports(provider, source)) {
                     return (provider.name === providerName);
                 }
             }
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -405,10 +405,9 @@ Object.assign(Controller.prototype, {
             const state = _model.get('state');
             if (state === STATE_IDLE || state === STATE_PAUSED) {
                 _play({ reason: 'autostart' }).catch(error => {
-                    _model.set('autostartFailed', true);
-                    _this.triggerError({
-                        message: `Could not play video: ${error.message}`
-                    });
+                    if (!_this._instreamAdapter || !_this._instreamAdapter._adModel) {
+                        _model.set('autostartFailed', true);
+                    }
                     _actionOnAttach = null;
                 });
             }
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -160,12 +160,7 @@ Object.assign(Controller.prototype, {
 
         _model.on('change:viewSetup', function(model, viewSetup) {
             if (viewSetup) {
-                const mediaElement = this.currentContainer.querySelector('video, audio');
                 showView(this, _view.element());
-                if (mediaElement) {
-                    const mediaContainer = _model.get('mediaContainer');
-                    mediaContainer.appendChild(mediaElement);
-                }
             }
         }, this);
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -409,7 +409,7 @@ Object.assign(Controller.prototype, {
 
         function _stop(internal) {
             checkAutoStartCancelable.cancel();
-            apiQueue.queue.length = 0;
+            apiQueue.empty();
 
             const fromApi = !internal;
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -304,43 +304,55 @@ Object.assign(Controller.prototype, {
 
             checkAutoStartCancelable.cancel();
             checkAutoStartCancelable = cancelable(_checkAutoStart);
-            _model.once('itemReady', checkAutoStartCancelable.async);
 
             switch (typeof item) {
                 case 'string':
-                    _loadPlaylist(item);
-                    break;
-                case 'object': {
-                    try {
-                        const playlist = Playlist(item);
-                        setPlaylist(_model, playlist, feedData);
-                    } catch (error) {
+                    _loadPlaylist(item).then(data => {
+                        return _updatePlaylist(data.playlist, data);
+                    }).then(checkAutoStartCancelable.async).catch(error => {
                         _this.triggerError({
                             message: `Error loading playlist: ${error.message}`
                         });
-                        return;
-                    }
-                    _setItem(0);
+                    });
                     break;
-                }
+                case 'object':
+                    _updatePlaylist(item, feedData).then(checkAutoStartCancelable.async);
+                    break;
                 case 'number':
-                    _setItem(item);
+                    _setItem(item).then(checkAutoStartCancelable.async);
                     break;
                 default:
                     break;
             }
         }
 
+        function _updatePlaylist(data, feedData) {
+            try {
+                const playlist = Playlist(data);
+                setPlaylist(_model, playlist, feedData);
+            } catch (error) {
+                _this.triggerError({
+                    message: `Playlist error: ${error.message}`
+                });
+                return;
+            }
+            return _setItem(0);
+        }
+
         function _loadPlaylist(toLoad) {
-            const loader = new PlaylistLoader();
-            loader.on(PLAYLIST_LOADED, function(data) {
-                _load(data.playlist, data);
+            return new Promise((resolve, reject) => {
+                const loader = new PlaylistLoader();
+                loader.on(PLAYLIST_LOADED, function(data) {
+                    resolve(data);
+                });
+                loader.on(ERROR, function(error) {
+                    _model.set('feedData', {
+                        error: error
+                    });
+                    reject(error);
+                }, this);
+                loader.load(toLoad);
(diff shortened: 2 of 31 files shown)


src/js/controller/model.js
```diff
@@ -101,17 +106,12 @@ const Model = function() {
...
-                mediaModel.change('playAttempt', (mediaModelChanged, playAttempt) => {
-                    if (playAttempt) {
-                        mediaModelChanged.off('playAttempt', null, this);
-                        this.playVideo();
-                    }
-                }, this);
-                this.setPlaybackRate(this.get('defaultPlaybackRate'));
-                break;
             case MEDIA_TIME:
+                // Don't sent time event on Android before real duration is known
```
egreaves: What's the side effect that this causes on Android? Seems like it'd be beneficial across the board.

robwalch: We should show the correct duration (and Live mode) as soon as possible. Android incorrectly shows live mode for HLS streams until it finishes manifest parsing. Other browsers would show duration of 0 instead of live mode before transitioning.

src/js/api/core-loader.js
```diff
@@ -50,12 +50,13 @@ export function requiresPolyfills() {
 export function requiresProvider(model, providerName) {
     const playlist = model.get('playlist');
     if (Array.isArray(playlist) && playlist.length) {
-        const firstSource = Item(playlist[0]).sources[0];
-        if (firstSource) {
+        const sources = Item(playlist[0]).sources;
+        for (let i = 0; i < sources.length; i++) {
```
robwalch: All sources need to be looped through in case unsupported sources are skipped (mpd on iOS).

src/js/controller/controller.js
```diff
@@ -229,13 +224,13 @@ Object.assign(Controller.prototype, {
...
 
-            _checkAutoStart();
-
             _model.change('viewable', viewableChange);
             _model.change('viewable', _checkPlayOnViewable);
             _model.once('change:autostartFailed change:autostartMuted change:mute', function(model) {
                 model.off('change:viewable', _checkPlayOnViewable);
             });
+
+            _checkAutoStart();
```
robwalch: Run `_checkAutoStart()` last since 'viewable' changes can result in `preload()` being called on the initial provider instance.

src/js/controller/controller.js
```diff
@@ -298,57 +293,61 @@ Object.assign(Controller.prototype, {
...
 
+        function _updatePlaylist(data, feedData) {
+            try {
+                const playlist = Playlist(data);
+                setPlaylist(_model, playlist, feedData);
+            } catch (error) {
+                _this.triggerError({
+                    message: `Playlist error: ${error.message}`
+                });
+                return;
```
johnBartos: If we get an error, code depending on this method returning a promise will throw exceptions

src/js/controller/controller.js
johnBartos: Worthy of a comment

johnBartos:
Looks like there's a regression autostarting on mobile (iOS) with the HTML5 provider. We're not entering autoplay muted mode.

johnBartos:
````
{
    "playlist": [
        {
            "file": "http://samplescdn.origin.mediaservices.windows.net/e0e820ec-f6a2-4ea2-afe3-1eed4e06ab2c/AzureMediaServices_Overview.ism/manifest(format=mpd-time-csf)",
            "title": "Azure - Purple Video",
            "type": "dash"
        }
    ],
    "preload": "none",
    "related": {
        "client": "../bin-debug/related.js",
        "file": "//playertest.longtailvideo.com/related-hls-feed.json",
        "onclick": "play"
    },
    "repeat": false,
    "stagevideo": false,
    "stretching": "uniform",
    "visualplaylist": true,
    "width": "100%",
    "aspectratio": "16:9"
}
````
Moving between the first and second items does not work - the second item is never loaded, because it is not in the idle state when play is called.

File: src/js/controller/controller.js (hunk 4 above)
Comment (robwalch) [reply src/js/controller/controller.js]: Can you provide an example?

Comment (robwalch) [reply src/js/controller/controller.js]: Got it!

**Comment by robwalch** [comment]
😮

File: src/js/controller/controller.js (hunk 4 above)
Comment (robwalch) [reply src/js/controller/controller.js]: 95e60bf3af4a71bf65d262f7b2e5b6e174822129

**Comment by hghazzi** [comment]
test this please

**Comment by hghazzi** [comment]
@robwalch Was testing this locally and came across an issue. If play a video stream and then try to call the stop API, I am unable to start playing the stream again. I either see it buffer infinitely or stuck in idle.

**Comment by robwalch** [comment]
thanks

File: src/js/providers/video-actions-mixin.js [hunk 5]
```diff
@@ -0,0 +1,59 @@
...
+        return false;
+    },
+
+    getContainer: function() {
+        return this.container;
+    },
+
+    setContainer: function(element) {
+        this.container = element;
+        if (this.video !== element.firstChild) {
```
robwalch: Calling this DOM operation unnecessarily, was resulting in a 'pause' event firing from the video tag while it resumed playback after a break or playlist item change.

src/js/controller/instream-html5.js
```diff
@@ -38,14 +44,20 @@ const InstreamHtml5 = function(_controller, _model) {
...
+        const mediaModelState = _model.mediaModel.attributes;
+        mediaModelState.setup = false;
+        mediaModelState.started = false;
 
-        _adModel.set('item', 0);
-        _adModel.set('playlistItem', item);
         // Make sure it chooses a provider
-        _adModel.setActiveItem(item);
-
-        // check provider after item change
```
robwalch: @cmulgannon @MaxMikheyenko @pajong 

The following changes will along with the changes in model.js should allow VAST to handle hls and mpd creative sources. The player will load the providers on the fly for creatives you pass it. We should look at the vast plugin creative filtering, and make sure it passes all video creatives as a prioritized playlist item source list the player can select from.

robwalch:
@hghazzi fixed `stop()` issue.

src/js/controller/controller.js
```diff
@@ -261,11 +257,10 @@ Object.assign(Controller.prototype, {
                 viewable: viewable
             });
 
-            if (shouldPreload(model, viewable)) {
-                const item = model.get('playlistItem');
 
-                model.getVideo().preload(item);
-                _preloaded = true;
+            // Only attempt to preload if this is the first player on the page and viewable
```
egreaves: first player on the page ~and~ or viewable

src/js/controller/model.js
```diff
@@ -83,42 +88,44 @@ const Model = function() {
...
+                mediaModel.set(PLAYER_STATE, data.newstate);
 
                 // This "return" is important because
                 //  we are choosing to not propagate this event.
                 //  Instead letting the master controller do so
                 return;
+            case MEDIA_ERROR:
+                mediaModel.set('setup', false);
+                mediaModel.set('started', false);
+                mediaModel.set('preloaded', false);
```
robwalch: turn this into a function

PR #2355, opened 2017-09-12
Update Next Up style

radium-v:
### This PR will...
Improve the Next Up overlay for JW8

#### Addresses Issue(s):

JW8-93

### DIFF
--- a/src/css/controls/flags/controls-hidden.less
+++ b/src/css/controls/flags/controls-hidden.less
@@ -9,9 +9,12 @@
         top: 0;
     }
 
-    .jw-plugin,
-    .jw-nextup-container {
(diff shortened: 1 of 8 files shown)


johnBartos:
@radium-v The hover color for next up text is always white, but should be set according to the menu colors:

screen shot 2017-09-12 at 11 31 25
screen shot 2017-09-12 at 11 31 21

aliciahurst:
Cool, looks good to me. At first it seemed the width was too much but then I found a different example where it scales. One thing: What does the X in the right corner map to? It didn't map to controlbar.icons or menus.text. I'd think it would map to one of those, ideally menus.text even though it's not text.

PR #2358, opened 2017-09-12
Hide controlbar customization on idle

karimJWP:
### This PR will...
add specificity to the controlbar customizations to exclude them on idle
to prevent the controlbar from being seen on idle when cast is available.

#### Addresses Issue(s):
JW8-503


johnBartos:
@karimJWP resolve conflicts

PR #2359, opened 2017-09-12
Firefox: Do not unobserve before observe for viewability

hghazzi:
### This PR will...

Keep track of observed players to avoid observing the same player multiple times, due to async calls.

Now that Firefox supports the Intersection Observer API, there exists a bug in v55 that breaks observing if you unobserve before ever observing

#### Addresses Issue(s):

JW8-476

### DIFF
--- a/src/js/view/utils/views-manager.js
+++ b/src/js/view/utils/views-manager.js
@@ -7,13 +7,9 @@ const observed = [];
 let intersectionObserver;
 let responsiveRepaintRequestId = -1;
 
-console.log('lazyInitIntersectionObservers');
-
 function lazyInitIntersectionObserver() {
     const IntersectionObserver = window.IntersectionObserver;
-
     if (!intersectionObserver) {
-
         // Fire the callback every time 25% of the player comes in/out of view
         intersectionObserver = new IntersectionObserver((entries) => {
             if (entries && entries.length) {
--- a/src/js/view/utils/views-manager.js
+++ b/src/js/view/utils/views-manager.js
@@ -2,7 +2,7 @@ import activeTab from 'utils/active-tab';
 import { requestAnimationFrame, cancelAnimationFrame } from 'utils/request-animation-frame';
 
 const views = [];
-const observed = [];
+const observed = {};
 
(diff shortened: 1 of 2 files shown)


src/js/view/utils/views-manager.js
```diff
@@ -2,6 +2,7 @@ import activeTab from 'utils/active-tab';
 import { requestAnimationFrame, cancelAnimationFrame } from 'utils/request-animation-frame';
 
 const views = [];
+const observed = [];
```
robwalch: This should be an object, not an array.

src/js/view/utils/views-manager.js
```diff
@@ -77,14 +78,21 @@ export default {
...
+            return;
+        }
+
+        observed[container.id] = true;
         intersectionObserver.observe(container);
     },
     unobserve(container) {
-        if (intersectionObserver) {
+        if (intersectionObserver && observed[container.id]) {
+            observed[container.id] = false;
```
robwalch: Use `delete`.

test/unit/views-manager-test.js
```diff
@@ -0,0 +1,65 @@
...
+        intersectionObserver = viewsManager.getIntersectionObsever();
+        intersectionObserver.observe = sinon.spy();
+        intersectionObserver.unobserve = sinon.spy();
+    });
+
+    describe('intersectionObserver', function() {
+
+        it('should observe & unobserve the same container', function() {
+            const container = document.createElement('div');
+            container.id = 1;
```
robwalch: Element.ids are strings not numbers

src/js/view/utils/views-manager.js
```diff
@@ -77,14 +78,21 @@ export default {
...
     },
     unobserve(container) {
-        if (intersectionObserver) {
+        if (intersectionObserver && observed[container.id]) {
+            delete observed[container.id];
             intersectionObserver.unobserve(container);
         }
+    },
+    getIntersectionObsever() {
+        return intersectionObserver;
```
dYale: Obsever 😨

PR #2364, opened 2017-09-12
Prevent control gradient from covering nonlinear ads

karimJWP:
### This PR will...
add a z-index of 1 to the ads plugins
to prevent the control gradient from making nonlinear ads less visible
https://github.com/jwplayer/jwplayer-ads-vast/pull/249
https://github.com/jwplayer/jwplayer-ads-googima/pull/250
#### Addresses Issue(s):
JW8-396

### DIFF
--- a/src/css/jwplayer/imports/jwplayerlayout.less
+++ b/src/css/jwplayer/imports/jwplayerlayout.less
@@ -77,6 +77,11 @@
     }
 }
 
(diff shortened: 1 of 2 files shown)


karimJWP:
@johnBartos i agree with you but how would you know the plugin is googima or ads in the LESS files ?

PR #2370, opened 2017-09-13
Remove unused submenus when casting 

johnBartos:
### This PR will...
- Remove the qualities, audio tracks, and playback rates submenus while casting, since these settings do not affect casting
- Add the respective submenus back when casting is finished

It is better UX to remove settings which do not work

To verify I haven't broken existing functionality I verified that submenus work between items with mixed settings. But this could use another look. Let's see what the automated tests dig up.

No

#### Addresses Issue(s):

JW8-394

### DIFF
--- a/src/js/view/controls/settings-menu.js
+++ b/src/js/view/controls/settings-menu.js
@@ -62,48 +62,35 @@ export function setupSubmenuListeners(settingsMenu, controlbar, model, api) {
         }
     };
 
-    model.change('mediaModel', (newModel, mediaModel) => {
-        // Quality Levels
-        mediaModel.change('levels', (changedModel, levels) => {
-            if (!levels || levels.length <= 1) {
-                removeQualitiesSubmenu(settingsMenu);
-                return;
-            }
+    const onAudiotracksChanged = (mediaModel, changedModel, audioTracks) => {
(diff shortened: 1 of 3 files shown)


johnBartos:
@hghazzi Does `castActive` not do what the name implies?

hghazzi:
it does, looks like the issue is you should be using `model.change` for the handler

src/js/view/controls/settings-menu.js
```diff
@@ -148,12 +126,52 @@ export function setupSubmenuListeners(settingsMenu, controlbar, model, api) {
...
+    model.change('playbackRates', onPlaybackRatesChange, this);
     model.change('playbackRate', (changedModel, playbackRate) => {
         const rates = model.get('playbackRates');
         if (rates) {
             activateSubmenuItem('playbackRates', rates.indexOf(playbackRate));
         }
     }, this);
+
+    // Remove the audio tracks, qualities, and playback rates submenus when casting
+    model.on('change:castActive', (changedModel, active) => {
```
hghazzi: This should be `model.change`. We I connect and then disconnect, the quality submenu shows up

hghazzi:
test this please

PR #2371, opened 2017-09-14
Add in autoadvance event for nextUp

jnatalzia:
### This PR will...
Add a new event to the player api which denotes when a player auto-advances to the next video after the `next up` ui is displayed.

We would like to add an API event when the player advances to the next video on its own in the event that the related plugin is being used to set up playlists/recs.

It would be good to ensure that the consolidation of the event firing makes sense. Instead of simple making them a single API event (`nextAdvance` or something similar) this approach makes the two different types of advancement individually trackable.

This can be merged with associated functionality in other PRs

#### Addresses Issue(s):
DATA948

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -539,7 +539,20 @@ Object.assign(Controller.prototype, {
             }
 
             // It wasn't the last item in the playlist,
-            //  so go to the next one
+            //  so go to the next one and trigger an autoadvance event
+            const related = _api.getPlugin('related');
+            if (related) {
+                const nextUp = _model.get('nextUp');
+                if (nextUp) {
+                    _this.trigger('nextAutoAdvance', {
+                        mode: nextUp.mode,
+                        ui: 'nextup',
+                        target: nextUp,
+                        itemsShown: [ nextUp ],
+                        feedData: nextUp.feedData,
+                    });
+                }
+            }
             _next({ reason: 'playlist' });
         }
 
@@ -850,7 +863,7 @@ Object.assign(Controller.prototype, {
         ], () => !_model.getVideo());
         // Add commands from CoreLoader to queue
         apiQueue.queue.push.apply(apiQueue.queue, commandQueue);
-        
+
         _view.setup();
     },
     get(property) {
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -541,18 +541,7 @@ Object.assign(Controller.prototype, {
             // It wasn't the last item in the playlist,
             //  so go to the next one and trigger an autoadvance event
             const related = _api.getPlugin('related');
-            if (related) {
-                const nextUp = _model.get('nextUp');
-                if (nextUp) {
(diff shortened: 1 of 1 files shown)


src/js/controller/controller.js
```diff
@@ -539,7 +539,9 @@ Object.assign(Controller.prototype, {
             }
 
             // It wasn't the last item in the playlist,
-            //  so go to the next one
+            //  so go to the next one and trigger an autoadvance event
+            const related = _api.getPlugin('related');
+            triggerAdvanceEvent(related, 'nextAutoAdvance');
             _next({ reason: 'playlist' });
```
johnBartos: Doesn't `reason: 'playlist'` imply an auto advance?

src/js/controller/controller.js
```diff
@@ -662,18 +664,25 @@ Object.assign(Controller.prototype, {
...
-                        itemsShown: [ nextUp ],
-                        feedData: nextUp.feedData,
-                    });
-                }
-                related.next();
+            triggerAdvanceEvent(related, 'nextClick', () => related.next());
+        }
+
+        function triggerAdvanceEvent(related, evt, cb) {
+            if (!related) {
```
johnBartos: I think it makes more sense to not have `related` as an argument because you're never using `related` at your callsites, outside of passing it to this function.

src/js/controller/controller.js
jnatalzia: Yeah but if I understand the setup correctly, the next event that would fire would be a `beforePlay` when the playlist item has already changed. This autoadvance event needs to track before the playlist switches, so triggering a separate event ensures we tracking the autoadvance at the correct time. See `feedClick` vs `relatedClick`

src/js/controller/controller.js
jnatalzia: I agree. I passed it in the avoid doing the plugin lookup twice (the callback for `nextClick` requires a reference to the related plugin). Another option is passing the related plugin into the callback and that way we keep `related` specific to this function but that's a little hard to follow when reading through the code imo. How do you feel about just doing the lookup twice in the case of the `nextclick` handler?

src/js/controller/controller.js
johnBartos: Gotcha

src/js/controller/controller.js
johnBartos: Oh I see what you mean actually. Fine the way it is.

robwalch: You're not calling it twice, but it's repeated twice, instead of only being written and called once in `triggerAdvanceEvent`.

robwalch: oh i see. passing in `() => related.next()`

src/js/controller/controller.js
robwalch: Is it critical that we track normal playlist progression? What about "beforePlaylistItem" for any playlist item change to track drop off? I'm OK with this but am very concerned that we're overloading the player with events for something that we want 'related' to track.

PR #2381, opened 2017-09-19
Use overflow-y: auto so that scrollbars do not display if content does not overflow 

johnBartos:
Scrollbars should only show when needed

No

No

#### Addresses Issue(s):

JW8-398

### DIFF
--- a/src/css/controls/imports/settings-menu.less
+++ b/src/css/controls/imports/settings-menu.less
@@ -81,7 +81,7 @@
     align-items: flex-start;
     align-content: flex-start;
     padding: 8px 20px 0 5px;
-    overflow-y: scroll;
+    overflow-y: auto;
     -webkit-overflow-scrolling: touch;
 
     &::-webkit-scrollbar {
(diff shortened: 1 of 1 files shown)


src/css/controls/imports/settings-menu.less
```diff
@@ -81,7 +81,7 @@
     align-items: flex-start;
     align-content: flex-start;
     padding: 8px 20px 0 5px;
-    overflow-y: scroll;
+    overflow-y: auto;
```
egreaves: `overflow-y: scroll` was intentionally done to improve the scrolling experience on iOS. (See: [momentum scrolling](https://css-tricks.com/snippets/css/momentum-scrolling-on-ios-overflow-elements/)).
If we change it to auto, then we might as well get rid of `-webkit-overflow-scrolling: touch;` as it's useless otherwise. Considering the list isn't likely to be that long that momentum scrolling would improve the scroll behavior on iOS, I'm ok with switching to auto and getting rid of  the webkit rule.

PR #2384, opened 2017-09-19
Click to play fixes

robwalch:
### This PR will...

- Cancel playlist update for async playlist loading when a new call to `load()` interrupts the process. A call to load invalidates previous calls.
- Always fire "playAttemptFailed" and include "item" for context. This will make it easier to trouble shoot ads behavior.
- Don't reset visualPlaylist when starting playback (It may have already been determined during preloading)
- Keep player and media model states in sync and pass ad state on to player after ads end. This helps keep the player state from getting out of sync when the main model is detached.

#### Addresses Issue(s):

JW8-532, JW8-569, JW8-132

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -43,7 +43,9 @@ const Model = function() {
     };
 
     this.getConfiguration = function() {
-        return _.omit(this.clone(), ['mediaModel']);
+        const config = this.clone();
+        delete config.mediaModel;
+        return config;
     };
 
     this.updateProviders = function() {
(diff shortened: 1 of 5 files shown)


src/js/controller/model.js
```diff
@@ -151,20 +157,18 @@ const Model = function() {
                 this.persistVideoSubtitleTrack(data.currentTrack, data.tracks);
                 break;
             case 'visualQuality':
-                var visualQuality = Object.assign({}, data);
-                mediaModel.set('visualQuality', visualQuality);
+                mediaModel.set('visualQuality', Object.assign({}, data));
```
dannyfinks: @robwalch do we polyfill this for IE11? I'm seeing this not trigger for html5 provider

robwalch: We do. I'll look at the bug you filed.

PR #2385, opened 2017-09-19
Make gradient separate element

karimJWP:
### This PR will...
* make a gradient element that moves from the back of the poster image to the back of the controls when linear ads play.
* hides the gradient when controls are not displayed and when related is open
* darkens the controls background on error, mimicking the behavior on idle and complete
* removes z-index on controls
To prevent the gradient from blocking UI elements
n/a
n/a
#### Addresses Issue(s):
JW8-396

### DIFF
--- a/src/css/controls/flags/ads.less
+++ b/src/css/controls/flags/ads.less
@@ -138,4 +138,8 @@
         /* stylelint-disable-next-line declaration-no-important */
         display: none !important;
     }
+
+    .jw-gradient {
+        opacity: 0;
+    }
 }
--- a/src/css/controls/flags/controls-hidden.less
+++ b/src/css/controls/flags/controls-hidden.less
@@ -27,6 +27,10 @@
         transition-delay: 0s, 250ms;
     }
 
+    .jw-gradient {
+        opacity: 0;
+    }
+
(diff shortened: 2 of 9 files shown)


karimJWP:
@egreaves sounds good to me. I checked the hierarchy using the view test page and it looked good to me. If you change the poster and media background to white, you can clearly see the effects across different states and configs. 
The gradient is positioned behind the poster because the poster is displayed on idle, complete and error, but we don't use a gradient in those states; a black background of opacity 0.4 is applied to jw-controls to make the icons stand-out.

robwalch:
We do show the control at in the Idle state when Chromecast is available.  This and several ads scenarios, setControl(toggle), and animation on the gradient need to be verified manually.

karimJWP:
@robwalch the cast button shouldn't be a problem on idle because jw-controls is given a background of rgba(0,0,0,0.4), but you raise a great point about chromecast since the poster is shown with controls during regular playback and my gradient would not be visible. I tested vast, freehweel, ima for both linear and non manually, and relied on the view test page for VPAID with both controls displayed and hidden. I'll do a manual check on VPAID. any other ad cases i should consider ?

PR #2388, opened 2017-09-20
Accessibility Support for New UI

hghazzi:
### This PR will...

Improve the user experience of navigating the video player and controls via the keyboard, using tab, enter, space and esc

To better support accessibility and screen readers

https://github.com/jwplayer/jwplayer-plugin-related/pull/163
https://github.com/jwplayer/jwplayer-commercial/pull/4076
https://github.com/jwplayer/jwplayer/pull/2388

#### Addresses Issue(s):

JW8-395

### DIFF
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -71,17 +71,17 @@ function createCastButton(castToggle, localization) {
         return;
     }
 
-    const ariaText = localization.cast;
 
     const castButton = document.createElement('button', 'google-cast-button');
-    ariaLabel(castButton, ariaText);
     castButton.setAttribute('type', 'button');
+    castButton.setAttribute('tabindex', '-1');
 
     const element = document.createElement('div');
     element.className = 'jw-reset jw-icon jw-icon-inline jw-icon-cast jw-button-color';
     element.style.display = 'none';
     element.style.cursor = 'pointer';
     element.appendChild(castButton);
+    ariaLabel(element, localization.cast);
 
     SimpleTooltip(element, 'chromecast', localization.cast);
 
@@ -340,7 +340,10 @@ export default class Controlbar {
         }
 
         if (elements.cast && elements.cast.button) {
-            new UI(elements.cast.button).on('click tap', function () {
+            new UI(elements.cast.element()).on('click tap enter', function(evt) {
+                if (evt.type === 'enter') {
+                    elements.cast.button.click();
+                }
                 this._model.set('castClicked', true);
             }, this);
         }
@@ -349,7 +352,7 @@ export default class Controlbar {
             _model.once('change:position', this.checkDvrLiveEdge, this);
         }, this);
 
-        new UI(elements.duration).on('click tap', function () {
+        new UI(elements.duration).on('click tap enter', function () {
             if (this._model.get('streamType') === 'DVR') {
                 // Seek to "Live" position within live buffer, but not before current position
                 const currentPosition = this._model.get('position');
--- a/src/js/view/logo.js
+++ b/src/js/view/logo.js
@@ -84,7 +84,13 @@ export default function Logo(_model) {
         _img.src = _settings.file;
 
         const logoInteractHandler = new UI(_logo);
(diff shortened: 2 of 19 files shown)


johnBartos:
@hghazzi Uncommit the lock

src/js/view/logo.js
```diff
@@ -84,7 +84,13 @@ export default function Logo(_model) {
         _img.src = _settings.file;
 
         const logoInteractHandler = new UI(_logo);
-        logoInteractHandler.on('click tap', function (evt) {
+
+        if (_settings.link) {
+            _logo.setAttribute('tabindex', '0');
+            _logo.setAttribute('aria-label', 'Logo');
```
karimJWP: even if the logo does not have a redirect, would we want the user to know it's there, for branding? Branding is one of the reasons for having a logo on your video. I could go either way on this, @aliciahurst any thoughts ?

src/js/view/utils/submenu-factory.js
```diff
@@ -19,10 +19,15 @@ export const makeSubmenu = (settingsMenu, name, contentItems, icon) => {
     } else {
         const categoryButton = button(`jw-settings-${name}`, () => {
             settingsMenu.activateSubmenu(name);
+            console.log('submenu enter', submenu.element().children[0]);
```
karimJWP: @hghazzi are we purposely leaving this console.log?

src/js/view/logo.js
hghazzi: @karimJWP that was part of the acceptance criteria

src/js/view/controls/controlbar.js
```diff
@@ -71,17 +71,17 @@ function createCastButton(castToggle, localization) {
         return;
     }
 
-    const ariaText = localization.cast;
 
     const castButton = document.createElement('button', 'google-cast-button');
-    ariaLabel(castButton, ariaText);
     castButton.setAttribute('type', 'button');
+    castButton.setAttribute('tabindex', '-1');
```
karimJWP: wouldn't this prevent users from tabbing to the cast button ? Why do we want that ?

hghazzi: There's a div around the cast button, so we only want the div to be tabbed to?

src/js/view/utils/submenu-factory.js
hghazzi: Removed

src/js/view/controls/controlbar.js
karimJWP: 👍

aliciahurst:
Tested on PR build 9762, here's everything I found.
* When I tabbed to discovery and hit enter to open it, and hit tab again, I was out of the player.  Expected is when discovery opens the first item is in focus.
* When I hit escape on the discovery overlay it did not close. Expected is the overlay closed.
* I had the whole player focused and the controls faded. Had to pause the video to get them back. Is this how we want it to behave, or when focused, do we want controls to stay up if the entire player is in focus?
* When the menu first opens, I imagine the first submenu is selected with the first item focused as per the requirements.
* "Tabbing advances through the settings menu top nav" I think this may have been unclear. Tabbing should advance the _focus_ through the icons in the submenu. But currently when you hit tab, it changes to that tab, which is a break from the click behavior (it's not on hover, it's on click). You should still have to hit enter to get to that tab, and when you hit enter, the first item in that submenu should be selected. Make sense? 
* In fullscreen, pressing space didn't seem to pause/unpause, and I don't seem to be able to control anything via keyboard at all.
* When I was in the options of one submenu and I kept tabbing through the options, when it reached the end of the text options, instead of going back up to the icons in the menu bar, the focus left the player. So then it was impossible to even close the menu or navigate to other tabs. Expected is it goes back up to the next icon in the submenu.
* When you focus on an icon in the control bar there's no way to tell that you've selected it. So it should have a focus outline.
* When you open the sharing menu or the CC menu the first item should already be in focus.

hghazzi:
@aliciahurst looks like the related changes were not included in this build because they were in the related repo

hghazzi:
test this please

PR #2389, opened 2017-09-20
Fix extra player being created by the compatibility script

johnBartos:
With the compatibility script, calls to `jwplayer()` (without an id argument) will return an unsetup player.

No

No

### DIFF
--- a/src/js/compatibility.js
+++ b/src/js/compatibility.js
@@ -119,7 +119,10 @@
          an Environment object. This object details the environment in which the player thinks it's in. Refer to our
          API docs for more information.
          */
-        var environment = playerLibrary(document.createElement('div')).getEnvironment();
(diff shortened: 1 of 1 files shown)


src/js/compatibility.js
```diff
@@ -119,7 +119,10 @@
          an Environment object. This object details the environment in which the player thinks it's in. Refer to our
          API docs for more information.
          */
-        var environment = playerLibrary(document.createElement('div')).getEnvironment();
+        var tempPlayer = playerLibrary(document.createElement('div'));
+        var environment = tempPlayer.getEnvironment();
+        tempPlayer.remove();
```
robwalch: We should just include `getEnvironment` in failed query calls too - so `playerLibrary().getEnvironment()` would work no matter what and because we're not passing in an element, we will not create a new Api instance.

PR #2392, opened 2017-09-20
Always flip skipButton state

robwalch:
...for commercial controls to receive change event. The UI listens for `adMode` 'change:skipButton' events to reattach click listeners to the skip button. Freewheel was essentially using the same skip button instance for multiple pods before. Now it's reset for each item and removed if one of the items in the pod is not skippable.

#### Addresses Issue(s):

JW8-584


pajong:
@dannyfinks Let me know if when we are safe to merge this

PR #2398, opened 2017-09-21
Setup skip button after loading instream media

robwalch:

This prevents the skip button on subsequent ad pods from picking up the duration and position of the previous pod's media.

#### Addresses Issue(s):

JW8-606

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -208,14 +208,16 @@ var InstreamAdapter = function(_controller, _model, _view) {
 
         _this.addClickHandler();
 
+        adModel.set('skipButton', false);
+
+        const playPromise = _instream.load(item);
+
         const skipoffset = item.skipoffset || _options.skipoffset;
         if (skipoffset) {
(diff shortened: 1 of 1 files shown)


src/js/controller/instream-adapter.js
```diff
@@ -208,14 +208,16 @@ var InstreamAdapter = function(_controller, _model, _view) {
 
         _this.addClickHandler();
 
+        adModel.set('skipButton', false);
+
+        const playPromise = _instream.load(item);
```
johnBartos: This variable seems unnecessary

pajong: We need to return the play promise we get from ```_instream.load```

pajong: You might be right. Seems like we do not use the promise anywhere.

robwalch: You should use it to capture ad start failures in vast.

robwalch: And to only fire ad impression on `then`

PR #2404, opened 2017-09-22
Separate gradient from controls

karimJWP:
### This PR will...
* make a gradient element that moves from the back of the poster image to the back of the controls when linear ads play.
* hides the gradient when controls are not displayed and when related is open
* darkens the controls background on error, mimicking the behavior on idle and complete
* removes z-index on controls
To prevent the gradient from blocking UI elements
n/a
n/a
#### Addresses Issue(s):
JW8-396


src/css/controls/states.less
```diff
@@ -126,10 +126,8 @@
     }
 
     .jw-controls {
```
johnBartos: Do you need to set this gain? Looks like it's handled in https://github.com/jwplayer/jwplayer/pull/2404/files#diff-ed218305fc3cedaf045e81cf450e0087R166

src/js/view/view.js
```diff
@@ -345,6 +347,20 @@ function View(_api, _model) {
     function addControls() {
         const controls = new ControlsModule(document, _this.element());
         _this.addControls(controls);
+        addGradient(_playerElement.className.indexOf('jw-flag-ads') >= 0);
+    }
+
+    function addGradient(adsMode) {
+
```
johnBartos: Remove this blank space

src/js/view/view.js
```diff
@@ -749,9 +766,12 @@ function View(_api, _model) {
         this.instreamModel = _instreamModel = instreamModel;
         _instreamModel.on('change:controls', _onChangeControls, this);
         _instreamModel.on('change:state', _stateHandler, this);
-
+        
         addClass(_playerElement, 'jw-flag-ads');
         removeClass(_playerElement, 'jw-flag-live');
+        if (_controls) {
+            addGradient(true);
```
johnBartos: Looks like there's a potential to add the gradient multiple times

src/css/controls/states.less
karimJWP: @johnBartos one is for state complete, the other is state error. we could probably group them but i know @radium-v  will be changing the background effect on idle/complete states so i preferred to defer that tech debt for now

src/js/view/view.js
```diff
@@ -783,6 +802,9 @@ function View(_api, _model) {
         this.setAltText('');
         removeClass(_playerElement, ['jw-flag-ads', 'jw-flag-ads-hide-controls']);
         _model.set('hideAdsControls', false);
+        if (_controls) {
```
egreaves: There's already a check to see if controls exist in line 798. You can call `addGradient` there.

src/js/view/view.js
```diff
@@ -752,6 +768,9 @@ function View(_api, _model) {
 
         addClass(_playerElement, 'jw-flag-ads');
         removeClass(_playerElement, 'jw-flag-live');
+        if (_controls) {
```
egreaves: There's already a check to see if controls exist in line 776. You can call `addGradient` there before `_controls.userActive();`.

robwalch:
This looks really good @karimJWP! I made a few changes I think we needed to make sure that the gradient element's display is set to none when it is for controls, and that we keep the order of poster gradient and captions consistent: 607e623

src/css/controls/flags/overlay.less
```diff
@@ -5,9 +5,5 @@
         .jw-logo {
             display: none;
         }
-
-        .jw-gradient {
```
karimJWP: @robwalch this was put because when overlay is open, if you look very carefully you might notice a slight gradient. I'll move it up to the display none block on top

PR #2407, opened 2017-09-22
Handle logo insertion correctly for controls true|false

egreaves:
### This PR will...
Insert the logo before the dock icons when controls are enabled. Append the logo otherwise.
When `controls: false`, `.insertBefore` causes an exception because of the dom structure.
No.
No.
#### Addresses Issue(s):
JW7-4494

### DIFF
--- a/src/js/view/logo.js
+++ b/src/js/view/logo.js
@@ -81,11 +81,13 @@ define([
         this.setContainer = function(container) {
             if (_logo) {
                 const dock = container.querySelector('.jw-dock');
-                if (dock) {
+
+                if (_model.get('controls') && dock) {
                     container.insertBefore(_logo, dock);
-                } else {
-                    container.appendChild(_logo);
+                    return;
                 }
+
+                container.appendChild(_logo);
             }
         };
 


src/js/view/logo.js
```diff
@@ -81,11 +81,13 @@ define([
         this.setContainer = function(container) {
             if (_logo) {
                 const dock = container.querySelector('.jw-dock');
-                if (dock) {
+
+                if (_model.get('controls') && dock) {
                     container.insertBefore(_logo, dock);
-                } else {
-                    container.appendChild(_logo);
+                    return;
```
karimJWP: @egreaves  why would you prefer a return statement over an else block? is it more performant? personally i find if else blocks easier to read

egreaves: @karimJWP The strongest argument for early returns is to mitigate condition nesting. Here's a post that makes some points I agree with: http://blog.timoxley.com/post/47041269194/avoid-else-return-early

robwalch: Since the browser requires the element in the second argument to be a child of the element `insertBefore` is called on, instead of checking for controls, you should be checking that (don't rely on the model for something the browser expects):

```js
if (dock && dock.parentNode === container) {
    container.insertBefore(_logo, dock);
```

PR #2409, opened 2017-09-25
Improve Error state design and layout

### LINKED ISSUE #2336: SVG icon module
It would be great to have a module that intelligently groups parsing of SVG assets (controlbar, sharing, etc...), lazy parses groups when getting an icon, and returns a deep clone of the parsed icon so that the SVG strings are never reparsed.

We could also benefit by doing this for templates as well.

JW8-536

radium-v:
### This PR will...
Adjust the error state view to match the designs for JW8.
I also removed the old SVG/font icons from the project.

This only affects the `jw-error` view, not the `jw-state-error` view. These are two different approaches in the player, and we should approach the `jw-state-error` view from a different angle.

jwplayer/jwplayer-commercial#4092

#### Addresses Issue(s):
JW8-98
JW8-427
JW8-536
Resolves #2336


src/css/jwplayer/imports/title.less
```diff
@@ -10,34 +9,33 @@
 // if we set color on the parent, it is overwritten by the reset
 .jw-title-primary,
 .jw-title-secondary {
+    color: @white;
```
karimJWP: would this prevent titles from being customizable ?

src/js/api/core-shim.js
```diff
@@ -226,15 +226,17 @@ Object.assign(CoreShim.prototype, {
 
 function setupError(core, error) {
     let message = error.message;
-    if (message.indexOf(':') === -1) {
-        message = `Error loading player: ${message}`;
+    const index = message.indexOf(':');
+    let title = 'Error loading player:';
+    if (message.indexOf(':') > 0) {
```
karimJWP: if (index > 0)

src/css/jwplayer/imports/title.less
radium-v: Nope, this is just sorting the existing props and swapping the `#fff` with the Less variable `@white`, which is already `#fff`

src/js/api/core-shim.js
radium-v: d'oh

src/templates/error.js
```diff
@@ -1,13 +1,13 @@
+import ERROR_ICON from 'assets/SVG/playback-error.svg';
```
robwalch: This results in the error icon being added to jwplayer.js. We should not add the icon here. The only setup errors where we have loaded controls with icons would be an invalid playlist error. For that type of error we should add the icon in another part of the code (view.js or `core._view.somethingThatReturnsIconDOMElement`)

src/js/api/core-shim.js
```diff
@@ -226,15 +226,17 @@ Object.assign(CoreShim.prototype, {
 
 function setupError(core, error) {
     let message = error.message;
-    if (message.indexOf(':') === -1) {
-        message = `Error loading player: ${message}`;
+    const index = message.indexOf(':');
+    let title = 'Error loading player:';
+    if (index > 0) {
+        title = error.message.substr(0, index + 1);
+        message = error.message.substr(index + 1);
```
robwalch: Improve readability by ordering variable declaration, reuse `message` and only add `1` once:

```js
let title = 'Error loading player:';
let message = error.message;
const index = message.indexOf(':') + 1;
if (index > 0) {
    title = message.substr(0, index);
    message = message.substr(index);
```

aliciahurst:
Questions:
1. When is the icon displayed vs not? I can [see](http://player-pr-test-jenkins.longtailvideo.com/builds/9876/archive/test/squash/bootstrap.html?primary=html5&edition=ads&feature=playlists_errors&suite=Playlist) that the first case has no icon, but the rest do.
2. In [scenario 11](http://player-pr-test-jenkins.longtailvideo.com/builds/9876/archive/test/squash/bootstrap.html?primary=html5&edition=ads&feature=chromeless/chromeless_components_unaccessible&suite=undefined) it seems like the error message should be displayed even when controls are false.

robwalch:
@aliciahurst The icon is only displayed when the error occurs after jwplayer.controls.js has loaded. Setup errors that occur before that do not have the icon. I can do a little extra work to display it if that file finishes loading following the error. These setup errors are rarer than playback errors and I want to avoid graphics in the player embed.

PR #2410, opened 2017-09-26
Further Accessibility Improvements

hghazzi:
### This PR will...

- Closing the related or settings overlay will keep focus on the respective control bar icons
- In Settings
   - Enter on a submenu icon will open the submenu and go to the first item
   - Tab after the close button will loop back to the first submenu icon
- Logo in control bar now has a focus ring
- After toggling fullscreen, focus goes to the play button
- When tabbing through the player, keep the control bar displayed

To improve the experience when tabbing through elements

https://github.com/jwplayer/jwplayer-plugin-related/pull/168

#### Addresses Issue(s):

JW8-603

### DIFF
--- a/src/js/view/controls/components/settings/menu.js
+++ b/src/js/view/controls/components/settings/menu.js
@@ -24,27 +24,46 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
         instance.close();
     }, 'Close Settings', [cloneIcon('close')]);
     closeButton.show();
+    closeButton.element().addEventListener('keydown', function(evt) {
+        if (evt.keyCode !== 9 || evt.shiftKey) {
+            return;
+        }
+
+        instance.close();
+    });
+
     topbarElement.appendChild(closeButton.element());
-    settingsMenuElement.addEventListener('keydown', function(evt) {
+
+    const keyHandler = function(evt) {
         if (evt && evt.keyCode === 27) {
             instance.close();
             evt.stopPropagation();
         }
-    });
+    };
+
+    settingsMenuElement.addEventListener('keydown', keyHandler);
 
     const instance = {
-        open() {
+        open(isDefault) {
             visible = true;
             onVisibility(visible);
             settingsMenuElement.setAttribute('aria-expanded', 'true');
             addDocumentListeners(documentClickHandler);
-            active.categoryButtonElement.focus();
+
+            if (isDefault) {
+                active.categoryButtonElement.focus();
+            } else {
+                active.element().firstChild.focus();
+            }
+
         },
         close() {
             visible = false;
             onVisibility(visible);
+
             active = null;
             deactivateAllSubmenus(submenus);
+
             settingsMenuElement.setAttribute('aria-expanded', 'false');
             removeDocumentListeners(documentClickHandler);
         },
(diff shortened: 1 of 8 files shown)


hghazzi:
Test this please

src/js/view/controls/components/settings/submenu.js
```diff
@@ -12,6 +12,16 @@ export default function SettingsSubmenu(name, categoryButton, isDefault) {
     categoryButtonElement.className += ' jw-submenu-' + name;
     categoryButton.show();
 
+    const nextSubmenu = function(evt) {
```
johnBartos: This function name is confusing. Is this a function go to the next submenu? Each submenu should not know about other submenus (and currently do not).

src/js/view/controls/controls.js
```diff
@@ -22,6 +22,25 @@ const reasonInteraction = function() {
...
 
+const keepActive = function(player) {
+    const focused = document.activeElement;
+
+    if (player === focused || !player.contains(focused)) {
+        return false;
+    }
+
+    // remove focus if button has an overlay
+    const hasOverlay = ['jw-icon-volume', 'jw-icon-settings', 'jw-related-btn', 'jw-playlist-btn'];
```
johnBartos: I don't think an array should be named with a verb (has)

src/js/view/controls/controls.js
```diff
@@ -127,6 +146,10 @@ export default class Controls {
             // Trigger userActive so that a dismissive click outside the player can hide the controlbar
             this.userActive();
             lastState = state;
+
+            if (!visible) {
+                this.controlbar.elements.settingsButton.element().focus();
```
johnBartos: Can we run into any undefined elements in this chain?

johnBartos:
Worth doing a performance evaluation on this to make sure we're not introducing UI deoptimizations

src/js/view/controls/components/settings/submenu.js
hghazzi: I'll rename it. What this does is that when you click tab after the last submenu item, the focus goes back to the submenu's category element. then the default tab behaviour kicks in, which would move the focus to the next item in the topbar.

hghazzi:
comments addressed

src/js/view/controls/components/settings/menu.js
```diff
@@ -24,28 +24,49 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
...
             visible = true;
             onVisibility(visible);
             settingsMenuElement.setAttribute('aria-expanded', 'true');
             addDocumentListeners(documentClickHandler);
             active.categoryButtonElement.focus();
+
+            if (isDefault) {
+                active.categoryButtonElement.focus();
+            } else {
+                active.element().children[0].focus();
```
egreaves: `active.element().firstChild.focus()`

dannyfinks:
@hghazzi the user experience when opening up the menu should be fixed:
Go to the first player on http://player-pr-test-jenkins.longtailvideo.com/builds/9944/archive/test/public/moviepilot/
1. Open up the settings menu and switch from the quality menu to the sharing menu and observe that the first option (facebook) has an active style applied

PR #2411, opened 2017-09-26
Shrink and add padding to breakpoint-1

hghazzi:
### This PR will...

- Changed the size of the display icons to be the default touch target size
- Added padding to the sides of the display icons to add more space between the buttons 

To improve the UI at small player sizes

@aliciahurst I feel like the the space between the display icons could be smaller. Right now, its the same size as the icon width. I think it would still be good at half that.

screenshot 2017-09-26 13 22 36
screenshot 2017-09-26 13 27 00

#### Addresses Issue(s):

JW8-527

### DIFF
--- a/src/css/controls/imports/display.less
+++ b/src/css/controls/imports/display.less
@@ -92,6 +92,7 @@ display icons
 .jw-breakpoint-1 {
     .jw-display-icon-container {
         margin: 0 (@mobile-touch-target * 0.5);
+        padding: (@mobile-touch-target * 0.125);
     }
 
     .jw-display .jw-icon,
--- a/src/css/controls/imports/display.less
+++ b/src/css/controls/imports/display.less
@@ -92,9 +92,9 @@ display icons
 .jw-breakpoint-1 {
     .jw-display .jw-icon,
     .jw-display .jw-svg-icon {
-        height: (@mobile-touch-target * 1.25);
-        line-height: (@mobile-touch-target * 1.25);
-        width: (@mobile-touch-target * 1.25);
+        height: (@mobile-touch-target);
+        line-height: (@mobile-touch-target);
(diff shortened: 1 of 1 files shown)


src/css/controls/imports/display.less
```diff
@@ -103,6 +103,8 @@ display icons
     }
 
     .jw-display .jw-icon {
+        padding: 0 (@mobile-touch-target * 0.5);
```
robwalch: Can you put the padding on `jw-display-icon-container` instead of `jw-icon` so that applying a background and border-radius to `jw-display-icon-container` results in a circle or square with symmetrical width and height.

hghazzi:
Makes sense to me. Let me just confirm if @aliciahurst thinks just adding the margins will be enough.

src/css/controls/imports/display.less
```diff
@@ -90,11 +90,15 @@ display icons
...
+    }
+
     .jw-display .jw-icon,
     .jw-display .jw-svg-icon {
-        height: (@mobile-touch-target * 1.25);
-        line-height: (@mobile-touch-target * 1.25);
-        width: (@mobile-touch-target * 1.25);
+        height: (@mobile-touch-target);
+        line-height: (@mobile-touch-target);
+        width: (@mobile-touch-target);
```
robwalch: `padding: (@mobile-touch-target * 0.25);`

robwalch: and please put width and height together. idiomatic CSS property ordering is better than alphabetical https://css-tricks.com/poll-results-how-do-you-order-your-css-properties/

hghazzi: Agreed 👍

robwalch:
^ @johnBartos @hghazzi what are the dimensions in the compiled CSS?

PR #2413, opened 2017-09-26
Show tooltips in submenus

karimJWP:
### This PR will...
- add tooltips to submenus
- add specificity the tooltip captions in the control bar to enable the tooltip captions in the settings menu
To show tooltips on submenus
Upon opening the menu, since an element is focused, a tooltip pops up automatically. 
https://github.com/jwplayer/jwplayer-commercial/pull/4099
#### Addresses Issue(s):

JW8-531

### DIFF
--- a/src/js/view/utils/submenu-factory.js
+++ b/src/js/view/utils/submenu-factory.js
@@ -1,6 +1,7 @@
 import SettingsSubmenu from 'view/controls/components/settings/submenu';
 import SettingsContentItem from 'view/controls/components/settings/content-item';
 import button from 'view/controls/components/button';
+import SimpleTooltip from 'view/controls/components/simple-tooltip';
 import CAPTIONS_OFF_ICON from 'assets/SVG/captions-off.svg';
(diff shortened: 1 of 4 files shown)


src/js/view/utils/submenu-factory.js
```diff
@@ -12,7 +13,7 @@ const QUALITIES_SUBMENU = 'quality';
 const PLAYBACK_RATE_SUBMENU = 'playbackRates';
 const DEFAULT_SUBMENU = QUALITIES_SUBMENU;
 
-export const makeSubmenu = (settingsMenu, name, contentItems, icon) => {
+export const makeSubmenu = (settingsMenu, name, contentItems, icon, localizedText) => {
```
johnBartos: I'd call `localizedText` something like `tooltipText`. I think that it's more clear to name the argument after how it will be used, rather than what it is. When I read it, I say it in my head as "the {argument}", like "the name", "the icon", etc. and "the tooltip text" sounds a lot more meaningful than "the localized text".

PR #2415, opened 2017-09-27
Show display icon when play attempt fails

robwalch:
### This PR will...
- Show display icon when play attempt fails using `jw-flag-play-rejected` CSS class
- Fix play promise handler so that media model 'started' is set

JW8-611 Showing the display icon when a play attempt fails is a call to action making it clearer to the user that the player is paused and must be interacted with to start playback.

JW8-624 The mediaModel was never getting updated with 'started' set to `true` when the play promise resolved. This caused subsequent calls to `play()` to fire additional "playAttempt" events and to put the player in a buffering state when the video was already playing.

### DIFF
--- a/src/css/controls/states.less
+++ b/src/css/controls/states.less
@@ -235,12 +235,18 @@ body .jwplayer.jw-state-error,
 }
 
 .jwplayer.jw-state-playing:not(.jw-flag-touch):not(.jw-flag-small-player):not(.jw-flag-casting),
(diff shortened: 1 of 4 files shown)


robwalch:
@dannyfinks address the buffering state issue in https://github.com/jwplayer/jwplayer/pull/2415/commits/4c8b23c6560cb9110de99ae7772fbe44aa47a8a9

PR #2418, opened 2017-09-27
	Clear video width and height when resizing video element

robwalch:
Sometimes external scripts like VPAID js creatives add inline width and height style settings to the video element. These should be cleared when the player is resized.

https://github.com/jwplayer/jwplayer/pull/2418

#### Addresses Issue(s):
JW8-621

### DIFF
--- a/src/js/providers/video-actions-mixin.js
+++ b/src/js/providers/video-actions-mixin.js
@@ -30,7 +30,7 @@ const VideoActionsMixin = {
                 objectFit = 'fill';
             }
             style(this.video, {
-                objectFit: objectFit,
+                objectFit,
                 width: null,
                 height: null
             });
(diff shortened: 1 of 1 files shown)


src/js/providers/video-actions-mixin.js
```diff
@@ -30,7 +30,9 @@ const VideoActionsMixin = {
                 objectFit = 'fill';
             }
             style(this.video, {
-                objectFit: objectFit
+                objectFit: objectFit,
```
johnBartos: You can shorthand this

robwalch: true. It seems weird to mix shorthand and not...

robwalch: thanks for updating it.

PR #2420, opened 2017-09-27
Guard check tracksById, which can be null 

johnBartos:
`stop()`, which `null`s `_tracksById`, can be called before the `MEDIA_DETACHED` handler, which tries to access `_tracksById`. This causes an exception.

No

https://github.com/jwplayer/hls.js/pull/116

#### Addresses Issue(s):

JW8-465


dannyfinks:
test this please

PR #2424, opened 2017-09-28
Show casting in audio player

hghazzi:
### This PR will...

Not hide the controls in audio player mode if casting is available

So users can play and cast audio-only streams

@radium-v @johnBartos When there is no cast icon there was padding on the right side so that the elapsed time was not flushed again the player. I used a negative margin to offset it when the cast icon is there. Let me know if there's a better way to go about it.

#### Addresses Issue(s):

JW8-620

### DIFF
--- a/src/css/controls/flags/cast-available.less
+++ b/src/css/controls/flags/cast-available.less
@@ -7,7 +7,7 @@
     &.jw-state-idle:not(.jw-flag-audio-player) {
         .jw-controlbar {
             .jw-slider-time,
-            .jw-icon:not(.jw-icon-cast):not(.jw-icon-airplay) {
+            .jw-icon:not(.jw-icon-cast):not(.jw-icon-airplay):not(.jw-icon-cardboard) {
                 display: none;
             }
         }
--- a/src/css/controls/flags/cast-available.less
+++ b/src/css/controls/flags/cast-available.less
@@ -3,4 +3,13 @@
     .jw-icon-airplay {
         display: flex;
     }
+
+    &.jw-state-idle:not(.jw-flag-audio-player) {
+        .jw-controlbar {
+            .jw-slider-time,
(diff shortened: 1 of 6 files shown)


src/css/controls/flags/cast.less
```diff
@@ -27,8 +27,8 @@
     }
 }
 
-.jwplayer.jw-state-playing.jw-flag-casting,
-.jwplayer.jw-state-paused.jw-flag-casting {
+.jwplayer.jw-state-playing.jw-flag-casting:not(.jw-flag-audio-player),
```
egreaves: Since we added specificity with `:not(.jw-flag-audio-player)`, we can remove some specificity by getting rid of `.jwplayer`

johnBartos:
@hghazzi The negative right margin does work for aligning with the controlbar (to the right),  but it does not create the correct amount of space between the time and the cast icon (to the left). The spacing between each icon should be 20px, and there should be a ~10px padding insetting the left/right of the controlbar. I think what you need to do is a) adjust the right padding of the button container (notice the left is already set to `0px`) and b) adjust the left-padding of the chromecast icon to have 20px inbetween it and the time. It may help to adjust the cast icon size to be 24x24 instead of 20x20 to match the size of our other icons.

hghazzi:
@johnBartos updated the css, let me know what you think. I made the icon 24x24, not sure if its too big now. @radium-v what do you think?

screenshot 2017-09-29 15 39 12
screenshot 2017-09-29 15 39 23
screenshot 2017-09-29 15 39 29

src/css/controls/flags/cast-available.less
```diff
@@ -3,4 +3,13 @@
     .jw-icon-airplay {
         display: flex;
     }
+
+    &.jw-state-idle:not(.jw-flag-audio-player) {
```
karimJWP: @hghazzi this has been moved to states.less under idle-state, and includes cardboard icon. you can delete this

PR #2425, opened 2017-09-28
Controls backdrop fixes (aka "Gradient Layer")

robwalch:
### This PR will...
- Fix exception with backdrop removal when controls are disabled
- Rename jw-gradient to jw-controls-backdrop
- Move backdrop element to controls
- Pass instream model to controls when controls are enabled during ads playback

#### Addresses Issue(s):
JW8-644
JW8-635

### DIFF
--- a/src/css/controls/flags/ads.less
+++ b/src/css/controls/flags/ads.less
@@ -137,7 +137,7 @@
 }
 
 .jwplayer.jw-flag-ads-hide-controls {
-    .jw-gradient,
+    .jw-controls-backdrop,
     .jw-controls {
         /* Important is required to controls are never displayed */
         /* stylelint-disable-next-line declaration-no-important */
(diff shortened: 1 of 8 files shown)


src/css/controls/flags/ads.less
```diff
@@ -71,7 +71,7 @@
 
     // Control bar should always be visible during ad playback on touch devices
     // except when jw-flag-ads-vpaid is set
-    &.jw-flag-ads.jw-flag-touch:not(.jw-flag-ads-vpaid) {
+    &.jw-flag-ads.jw-state-playing.jw-flag-touch:not(.jw-flag-ads-vpaid) {
```
karimJWP: @robwalch why are we specifying state-playing ? shouldn't the backdrop always be displayed on ads on touch devices ?

egreaves: @karimJWP this is so that the shorter gradient is only shown during the playing state. In other states where the gradient is present, it extends from the control bar up to the display icons to provide a backdrop for all visible control elements.

PR #2427, opened 2017-09-28
Enable 360 icon on idle

karimJWP:
### This PR will...
create a carboard available flag 
to display 360 icon on idle
is this overkill ?
https://github.com/jwplayer/jwplayer-vr-plugin/pull/45
#### Addresses Issue(s):
JW8-643

### DIFF
--- a/src/css/controls/flags/cardboard-available.less
+++ b/src/css/controls/flags/cardboard-available.less
@@ -2,13 +2,4 @@
     .jw-icon-cardboard {
         display: flex;
     }
-
-    &.jw-state-idle {
-        .jw-controlbar {
-            .jw-slider-time,
-            .jw-icon:not(.jw-icon-cardboard):not(.jw-icon-cast):not(.jw-icon-airplay) {
-                display: none;
-            }
-        }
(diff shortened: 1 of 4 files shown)


src/css/controls/flags/cardboard-available.less
```diff
@@ -0,0 +1,14 @@
+.jwplayer.jw-flag-cardboard-available {
+    .jw-icon-cardboard {
+        display: flex;
+    }
+
+    &.jw-state-idle {
```
egreaves: this duplicates the rule in `cast-available.less`. We should remove it from both and put it in `states.less`.

johnBartos:
@karimJWP Can you screenshot or provide a config where we can see what this looks like?

karimJWP:
@johnBartos on android chrome http://0.0.0.0:3000/jwplayer-commercial/test/squash/bootstrap.html?primary=html5&edition=ads&feature=vr_360/360_playlists&suite=VR_360

karimJWP:
@johnBartos unnamed

PR #2431, opened 2017-09-29
Display tooltips above submenu

karimJWP:
### This PR will...
allow tooltips to be displayed over the submenu when hovered over
esthetics 
#### Addresses Issue(s):
JW8-524

### DIFF
--- a/src/css/controls/imports/settings-menu.less
+++ b/src/css/controls/imports/settings-menu.less
@@ -155,8 +155,13 @@
         }
     }
 
+    .jw-tooltip-fullscreen,
     .jw-tooltip-settings,
-    .jw-settings-submenu-button .jw-tooltip-captions {
+    .jw-tooltip-related,
+    .jw-settings-submenu-button .jw-tooltip,
+    .jw-tooltip-chromecast,
+    .jw-tooltip-cardboard,
+    .jw-tooltip-airplay {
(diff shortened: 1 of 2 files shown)


src/css/controls/imports/settings-menu.less
```diff
@@ -155,10 +155,15 @@
         }
     }
 
-    .jw-tooltip-settings,
-    .jw-settings-submenu-button .jw-tooltip-captions {
+    .jw-tooltip-settings {
         display: none;
     }
+
+    .jwplayer:not(.jw-flag-small-player) & {
```
egreaves: Currently, the settings menu is after the controlbar in the DOM and the last element in the controls div. If we insert it before the control bar instead of appending it, this bug fixes itself. 
screen shot 2017-10-02 at 5 54 49 pm

karimJWP: @egreaves i had considered that option but we would still need css for breakpoints 0 and 1 because the menu covers the entire player on flag-small-player, and the controlbar would be layered on top

egreaves: @karimJWP the controlbar can be set to `display: none` when the settings menu is open at those player sizes.

karimJWP:
@egreaves comments addressed

PR #2434, opened 2017-09-29
Don't Show Tooltips on Touch Events

hghazzi:
### This PR will...

Does not show tooltips on touch events

So we dont show tooltips after tapping a control bar icon on iPad

#### Addresses Issue(s):

JW8-616

### DIFF
--- a/src/js/view/controls/components/simple-tooltip.js
+++ b/src/js/view/controls/components/simple-tooltip.js
@@ -1,4 +1,5 @@
 import { addClass, removeClass } from 'utils/dom';
+import UI from 'utils/ui';
 
 export function SimpleTooltip(attachToElement, name, text, openCallback) {
     const tooltipElement = document.createElement('div');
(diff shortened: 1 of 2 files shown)


src/js/view/controls/components/simple-tooltip.js
```diff
@@ -29,13 +40,11 @@ export function SimpleTooltip(attachToElement, name, text, openCallback) {
...
-        attachToElement.addEventListener('pointerover', instance.open);
-        attachToElement.addEventListener('pointerout', instance.close);
-    } else {
-        attachToElement.addEventListener('mouseover', instance.open);
-        attachToElement.addEventListener('mouseout', instance.close);
-    }
+    attachToElement.addEventListener('mouseover', instance.open);
+    attachToElement.addEventListener('mouseout', instance.close);
+    attachToElement.addEventListener('touchstart', (evt) => {
+        instance.pointerType = getPointerType(evt);
```
johnBartos: I think it's better to do `instance.touch` rather than check `instance.pointerType === 'touch'`

hghazzi: coolio

PR #2438, opened 2017-10-02
Hide tooltips when button is in focus

hghazzi:
### This PR will...

- Not create tooltips for the topnav icons in the settings menu
- Not focus on the settings button after closing if its not triggered by a key event

So that the tooltips do not show up on mobile devices after button is tapped

https://github.com/jwplayer/jwplayer-plugin-related/pull/170

#### Addresses Issue(s):

JW8-616

### DIFF
--- a/src/js/view/controls/components/settings/menu.js
+++ b/src/js/view/controls/components/settings/menu.js
@@ -29,14 +29,14 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
             return;
         }
 
-        instance.close();
+        instance.close(true);
     });
 
     topbarElement.appendChild(closeButton.element());
 
     const keyHandler = function(evt) {
         if (evt && evt.keyCode === 27) {
-            instance.close();
+            instance.close(true);
             evt.stopPropagation();
         }
     };
@@ -57,9 +57,9 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
             }
 
         },
-        close() {
+        close(isKeyEvent = false) {
             visible = false;
-            onVisibility(visible);
+            onVisibility(visible, isKeyEvent);
 
             active = null;
             deactivateAllSubmenus(submenus);
--- a/src/js/view/controls/components/settings/menu.js
+++ b/src/js/view/controls/components/settings/menu.js
@@ -17,31 +17,31 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
     let visible;
     let active = null;
     const submenus = {};
+
     const settingsMenuElement = createElement(SettingsMenuTemplate());
-    const topbarElement = settingsMenuElement.querySelector('.jw-settings-topbar');
 
-    const closeButton = button('jw-settings-close', () => {
-        instance.close();
-    }, 'Close Settings', [cloneIcon('close')]);
-    closeButton.show();
-    closeButton.element().addEventListener('keydown', function(evt) {
-        if (evt.keyCode !== 9 || evt.shiftKey) {
-            return;
-        }
-
-        instance.close(true);
-    });
-
-    topbarElement.appendChild(closeButton.element());
-
-    const keyHandler = function(evt) {
+    const closeOnEnter = function(evt) {
         if (evt && evt.keyCode === 27) {
             instance.close(true);
             evt.stopPropagation();
         }
     };
+    settingsMenuElement.addEventListener('keydown', closeOnEnter);
 
-    settingsMenuElement.addEventListener('keydown', keyHandler);
+    const closeButton = button('jw-settings-close', () => {
+        instance.close();
+    }, 'Close Settings', [cloneIcon('close')]);
+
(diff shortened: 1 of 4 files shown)


hghazzi:
@johnBartos Happy to brainstorm but some thoughts on the current solution:

- Would you consider the flag `instance.open(isDefault)` different from `instance.close(isKeyEvent)`? We could just have a private method `closeOnKey` since its only used internally.

- For `visibilityChangeHandler`, I see it similar to a simplified event handler, which takes an argument of the new state, so it would make sense to me that it could take in the trigger as well.

src/js/view/utils/submenu-factory.js
```diff
@@ -26,7 +26,9 @@ export const makeSubmenu = (settingsMenu, name, contentItems, icon, tooltipText)
         // Qualities submenu is the default submenu
         submenu = SettingsSubmenu(name, categoryButton, name === DEFAULT_SUBMENU);
         submenu.addContent(contentItems);
-        SimpleTooltip(categoryButtonElement, name, tooltipText);
+        if (!('ontouchstart' in window)) {
```
karimJWP: @hghazzi is this to check that the device is mobile ? 'ontouchstart' is supported in chrome https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/ontouchstart

hghazzi: `'ontouchstart' in window` is `false` in Chrome

johnBartos: Will this always be the case? We may open ourselves up to a regression if they decide to change that in the future. I could see that happening since Chrome wants to unify their mobile/desktop platforms.

src/js/view/controls/components/settings/menu.js
```diff
@@ -17,31 +17,31 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
...
-    closeButton.element().addEventListener('keydown', function(evt) {
-        if (evt.keyCode !== 9 || evt.shiftKey) {
-            return;
+    const closeOnEnter = function(evt) {
+        if (evt && evt.keyCode === 27) {
+            instance.close(true);
+            evt.stopPropagation();
         }
+    };
+    settingsMenuElement.addEventListener('keydown', closeOnEnter);
```
egreaves: We should use `keyup` for the following reasons:
1) Keyup acts as a signal that the user deliberately wants the end result of the keypress.
2) Keydown keeps firing if the user still has the key depressed, so you'd potentially have `closeOnEnter` get called several times.

src/js/view/controls/components/settings/menu.js
```diff
@@ -17,31 +17,31 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
...
-        if (evt && evt.keyCode === 27) {
-            instance.close();
-            evt.stopPropagation();
+    const closeOnButton = function(evt) {
+        if (evt.keyCode === 13 || (evt.keyCode === 9 && !evt.shiftKey)) {
+            instance.close(true);
         }
     };
+    closeButton.show();
+    closeButton.element().addEventListener('keydown', closeOnButton);
```
egreaves: Same feedback as above.

src/js/view/controls/components/settings/menu.js
```diff
@@ -17,31 +17,31 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
...
-
-    topbarElement.appendChild(closeButton.element());
+    }, 'Close Settings', [cloneIcon('close')]);
 
-    const keyHandler = function(evt) {
-        if (evt && evt.keyCode === 27) {
-            instance.close();
-            evt.stopPropagation();
+    const closeOnButton = function(evt) {
+        if (evt.keyCode === 13 || (evt.keyCode === 9 && !evt.shiftKey)) {
```
karimJWP: @hghazzi this is to close the menu if tabbing past the X icon ? @aliciahurst what are you thoughts ? I like it but i feel like it might be confusing for some users

src/js/view/controls/components/settings/menu.js
egreaves: I found this link helpful in understanding the differences: https://www.quirksmode.org/dom/events/keys.html

src/js/view/controls/components/settings/menu.js
hghazzi: that was part of the AC

egreaves: A comment here would be helpful to know what this is checking

src/js/view/controls/components/settings/menu.js
hghazzi: I used `keydown` since that's what is being used in `ui.js` and `controls.js` before any of this work was started, so was keeping it consistent. 

@egreaves Should I still change it?

src/js/view/controls/components/settings/menu.js
hghazzi: ok

PR #2439, opened 2017-10-02
Fix vpaid2 nonlinear from not firing events

pajong:
### This PR will...
Fix vpaid2 nonlinear from not firing events

Vpaid nonlinear sets the provider to null so that event listeners are not destroyed when instreamDestroy is called. This is neeeded since nonlinear ads play with the content, so the listeners need to stay active when blocking instream is destroyed.

#### Addresses Issue(s):
JW8-657


robwalch:
Replaced by #2440

PR #2441, opened 2017-10-03
Checks if menu has childNode before removing a submenu

karimJWP:
### This PR will...
prevent the removal of a node if it is not a child element of the menu
to prevent the player from crashing when controls are off and a new item is loaded
according to [js perf](https://jsperf.com/haschildnodes-vs-childnodes-length) calling `hasChildNodes` is the most performant way of verifying.
#### Addresses Issue(s):
JW8-658

### DIFF
--- a/src/js/view/controls/components/settings/menu.js
+++ b/src/js/view/controls/components/settings/menu.js
@@ -101,7 +101,7 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
         },
         removeSubmenu(name) {
             const submenu = submenus[name];
-            if (!submenu || !settingsMenuElement.hasChildNodes()) {
+            if (!submenu || submenu.element().parentNode !== settingsMenuElement) {
                 return;
             }
             settingsMenuElement.removeChild(submenu.element());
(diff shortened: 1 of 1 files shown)


src/js/view/controls/components/settings/menu.js
```diff
@@ -101,7 +101,7 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
         },
         removeSubmenu(name) {
             const submenu = submenus[name];
-            if (!submenu) {
+            if (!submenu || !settingsMenuElement.hasChildNodes()) {
```
egreaves: Since we're specifically removing the submenu, it's better to use:
```
if (!submenu || !settingsMenuElement.contains(submenu.element())
```

karimJWP: @egreaves agreed but `includes` doesn't work on IE

karimJWP: the crash happens only when controls are false, not aware yet of a case where it crashes when controls are displayed

robwalch: We do `submenu.element().parentNode === settingsMenuElement ` in other places.

robwalch: That is the fastest and fail proof. The submenu's parent must be the menu if you expect removeChild to work. Neither function above guarantees that.

PR #2442, opened 2017-10-03
Check visualQuality on timeupdate for IE11

robwalch:
IE11 doesn't fire resize events for video width and height so `mediaElement.videoWidth` and height need to be check while video is playing.

#### Addresses Issue(s):
JW8-656


johnBartos:
@robwalch What do you think about adding an adaptation method to HTML5, so that the code would be consistent across providers? A bit of modification to `checkVisualQuality` should be enough

PR #2443, opened 2017-10-03
Clean up error state view for controls only mode

radium-v:
### This PR will...
Fix the error state for controls only mode. It also adjusts the error state backdrop.

#### Addresses Issue(s):
JW8-98

### DIFF
--- a/src/css/controls/states.less
+++ b/src/css/controls/states.less
@@ -173,9 +173,8 @@ body .jwplayer.jw-state-error {
         }
     }
 
-    .jw-controls {
-        background: @controlbar-background;
-        height: 100%;
+    .jw-controls-backdrop {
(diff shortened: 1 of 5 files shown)


src/css/controls/states.less
```diff
@@ -11,8 +11,8 @@
         }
     }
 
-    .jw-controls {
-        background: @controlbar-background;
+    .jw-controls-backdrop {
+        background: linear-gradient(fade(@black, 40%), transparent) 0 0 / 100% 120px no-repeat transparent;
```
karimJWP: @radium-v why aren't we using the controlbar-background ? What happens if we have a title, the cast icon, and the center play button ? The cast icon is displayed on the idle state when casting is enabled to allow users to start casting before playback begins. Cast icon appears at the bottom right, title at the top left, and center play button in the middle

PR #2448, opened 2017-10-04
Only fire "visualQuality" events while playing

robwalch:
### This PR will...

Remove the media element "resize" listener that fires "visualQuality" events.

"visualQuality" events are fired before playback. Our test expect them to only be fired while playing. `getVisualQuality` will not be available for preloaded playlist items. It will only be available after the first time event when the player state is playing.

#### Addresses Issue(s):

JW8-656


robwalch:
@dannyfinks I understand that vq was fired before playback, but was the second time it fired the same quality? See the notes above.

PR #2449, opened 2017-10-04
Fix issues with VPAID midrolls replaying content

pajong:
### This PR will...
Reset the mediaModel when detaching the media.

When vpaid ads are played as midrolls, there is nothing that sets the mediaModel's setup back to false. This causes the providers to start the content from the beginning after the midrolls, because it never loads the video again. 
See line: https://github.com/jwplayer/jwplayer/blob/master/src/js/controller/model.js#L585

#### Addresses Issue(s):
ADS-583 ADS-579

### DIFF
--- a/src/js/controller/instream-html5.js
+++ b/src/js/controller/instream-html5.js
@@ -51,6 +51,8 @@ const InstreamHtml5 = function(_controller, _model) {
 
     /** Load an instream item and initialize playback **/
     _this.load = function() {
+        // Let the player media model know we're using it's video tag
+        _model.mediaModel.srcReset();
 
         // Make sure it chooses a provider
         _adModel.stopVideo();
@@ -159,10 +161,7 @@ const InstreamHtml5 = function(_controller, _model) {
             return;
         }
 
-        // Let the player media model know we're using it's video tag
-        _model.mediaModel.srcReset();
-
-        const isVpaidProvider = provider.type === 'vpaid';
+        var isVpaidProvider = provider.type === 'vpaid';
 
         provider.off();
         provider.on('all', function(type, data) {
--- a/src/js/controller/instream-html5.js
+++ b/src/js/controller/instream-html5.js
@@ -51,8 +51,6 @@ const InstreamHtml5 = function(_controller, _model) {
 
     /** Load an instream item and initialize playback **/
     _this.load = function() {
-        // Let the player media model know we're using it's video tag
-        _model.mediaModel.srcReset();
 
         // Make sure it chooses a provider
         _adModel.stopVideo();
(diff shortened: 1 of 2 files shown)


pajong:
Another solution will be to seek before starting. Something like:
```else {
            if (item.starttime) {
                _provider.seek(item.starttime);
            }
            playPromise = _provider.play() || resolved;
            if (!this.mediaModel.get('started')) {
                playAttempt(this, playPromise, playReason);
            }
        }
```
Let me know your thoughts

src/js/controller/model.js
```diff
@@ -587,6 +586,9 @@ const Model = function() {
             playPromise = loadAndPlay(this, item);
             playAttempt(this, playPromise, playReason);
         } else {
+            if (item.starttime) {
+                _provider.seek(item.starttime);
+            }
             playPromise = _provider.play() || resolved;
```
robwalch: Providers should handle seeking to item.starttime. Doing this in the model will have side effects on different providers. 

If, and only if, the video’s arc is changed externally, then the detached mediamodel’s setup state should be set back to false.

robwalch: *src (autocorrected to arc)

src/js/controller/instream-html5.js
```diff
@@ -161,7 +159,10 @@ const InstreamHtml5 = function(_controller, _model) {
             return;
         }
 
-        var isVpaidProvider = provider.type === 'vpaid';
+        // Let the player media model know we're using it's video tag
+        _model.mediaModel.srcReset();
+
+        const isVpaidProvider = provider.type === 'vpaid';
```
robwalch: What’s this for?

src/js/controller/instream-html5.js
```diff
@@ -51,8 +51,6 @@ const InstreamHtml5 = function(_controller, _model) {
 
     /** Load an instream item and initialize playback **/
     _this.load = function() {
-        // Let the player media model know we're using it's video tag
-        _model.mediaModel.srcReset();
```
robwalch: Please do not remove. See my comment below.

src/js/controller/instream-html5.js
pajong: Changing ```var``` to ```const```

src/js/controller/instream-html5.js
```diff
@@ -161,7 +159,10 @@ const InstreamHtml5 = function(_controller, _model) {
             return;
         }
 
-        var isVpaidProvider = provider.type === 'vpaid';
+        // Let the player media model know we're using it's video tag
+        _model.mediaModel.srcReset();
```
robwalch: Check provider is also called by apply listeners. It should not reset the media model. Only instream load(), and external clearing of the video.src should reset the mediaModel.

You should listen for video “unload” event to determine that a new src was loaded.

You could also move the source comparison code in html5 into a module and use it to compare the item source to to the video on destroy to determine if we need to do a reset outside of instream load.

src/js/controller/model.js
```diff
@@ -250,18 +251,34 @@ const Model = function() {
         return _beforecompleted;
     };
 
+    function sourceChanged(event) {
+        _provider.video.removeEventListener('loadeddata', sourceChanged);
```
robwalch: @pajong did you try using 'emptied'?

PR #2451, opened 2017-10-05
Only attempt to preload an idle player

egreaves:
### This PR will...
Only preload an idle player.
We were preloading when in the complete state after a postroll, caused by the player becoming viewable after switching away from then back to the tab it was in.
No.
No.
#### Addresses Issue(s):
JW8-683

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -259,8 +259,8 @@ Object.assign(Controller.prototype, {
             });
 
 
-            // Only attempt to preload an idle player if it's the first player on the page or viewable
-            if (model.get('state') === 'idle' && (instances[0] === _api || viewable === 1)) {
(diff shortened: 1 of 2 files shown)


src/js/controller/controller.js
```diff
@@ -259,8 +259,8 @@ Object.assign(Controller.prototype, {
             });
 
 
-            // Only attempt to preload if this is the first player on the page or viewable
-            if (instances[0] === _api || viewable === 1) {
+            // Only attempt to preload an idle player if it's the first player on the page or viewable
+            if (model.get('state') === 'idle' && (instances[0] === _api || viewable === 1)) {
                 model.preloadVideo();
```
robwalch: I prefer we make this check in the `model.preloadVideo()` so all conditions outside of the api instances are in one place.

PR #2460, opened 2017-10-09
Show controls when navigating via keyboard

hghazzi:
### This PR will...

Keep controls visible when navigating the player using the keyboard, but will fade out when clicking or tabbing out of the player.

To be able to continue navigating the player while using keys without the controls fading out and focus is lost

#### Addresses Issue(s):

JW8-603

### DIFF
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -46,6 +46,8 @@ export default class Controls {
         this.enabled = true;
         this.instreamState = null;
         this.keydownCallback = null;
+        this.keyupCallback = null;
+        this.blurCallback = null;
         this.mute = null;
         this.nextUpToolTip = null;
         this.playerContainer = playerContainer;
@@ -140,7 +142,7 @@ export default class Controls {
             }
 
             // Trigger userActive so that a dismissive click outside the player can hide the controlbar
-            this.userActive();
+            this.userActive(null, visible || isKeyEvent);
             lastState = state;
 
             const settingsButton = this.controlbar.elements.settingsButton;
@@ -198,24 +200,11 @@ export default class Controls {
                 return true;
             }
 
-            // keep controls active when navigating inside the player
-            if (!this.instreamState) {
-                const container = this.playerContainer;
-                const fullscreenButton = this.controlbar.elements.fullscreen;
-                const insideContainer = container !== evt.target && container.contains(evt.target);
-                const isTab = evt.keyCode === 9;
-
-                // check if we tab after fullscreen or shift-tab from player
-                const exitingContainer =
-                    (isTab && evt.shiftKey && evt.target === container) ||
-                    (isTab && evt.target === fullscreenButton.element());
-
-                this.userActive(null, insideContainer && !exitingContainer);
-            }
-
             switch (evt.keyCode) {
                 case 27: // Esc
                     api.setFullscreen(false);
+                    this.playerContainer.blur();
+                    this.userInactive();
                     break;
                 case 13: // enter
                 case 32: // space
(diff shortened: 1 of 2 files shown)


robwalch:
@hghazzi can you include links to past PRs ("Further Accessibility Improvements") that this builds on? I'd like to review both for context.

src/js/view/controls/controls.js
```diff
@@ -368,8 +400,9 @@ export default class Controls {
     }
 
     userActive(timeout = ACTIVE_TIMEOUT, isKeyDown) {
```
robwalch: Please remove the default `timeout = ACTIVE_TIMEOUT`. We are already using `timeout || ACTIVE_TIMEOUT` which is less strict and doesn't get transpiled by babel into more code.

src/js/view/controls/controls.js
```diff
@@ -197,10 +197,22 @@ export default class Controls {
...
+                const fullscreenButton = this.controlbar.elements.fullscreen;
+                const insideContainer = container !== evt.target && container.contains(evt.target);
+                const isTab = evt.keyCode === 9;
+
+                // check if we tab after fullscreen or shift-tab from player
+                const exitingContainer =
+                    (isTab && evt.shiftKey && evt.target === container) ||
+                    (isTab && evt.target === fullscreenButton.element());
+
+                this.userActive(null, insideContainer && !exitingContainer);
```
robwalch: This should be done on key up because then the tabbing would be done and we only need to check `insideContainer`.

src/js/view/controls/controls.js
```diff
@@ -259,6 +271,22 @@ export default class Controls {
         this.playerContainer.addEventListener('keydown', handleKeydown);
         this.keydownCallback = handleKeydown;
 
+        // If not mobile, show controls after closing overlay
+        if (!OS.mobile) {
+            const blurHandler = (evt) => {
```
robwalch: How does `playerContainer` emitting a "blur" event equate to the overlay being closed?

src/js/view/controls/controls.js
```diff
@@ -259,6 +271,22 @@ export default class Controls {
...
+                const insidePlayer = evt.currentTarget.contains(evt.relatedTarget);
+                const overlayClosed =
+                    evt.currentTarget.classList.contains('jw-flag-overlay-open-related') ||
+                    settingsMenu.element().contains(evt.target);
+
+                // show controls if inside the player and after overlay is closed
+                this.userActive(null, insidePlayer || overlayClosed);
+            };
+
+            this.playerContainer.addEventListener('blur', blurHandler, true);
```
robwalch: So anytime an element in the player looses focus, we check if the `relatedTarget` is inside the player (isn't it always if the listener is on the player)?

Please clarify on blur:
`evt.relatedTarget` (getting focus?)
`evt.currentTarget` (loosing focus?)
`evt.target` (shouldn't this be `playerContainer` or the same as `currentTarget`?)

What does `settingsMenu.element()` have to do with this? It's not covered in your comments.

robwalch:
@hghazzi 

I made changes to make this functionality less dependent on DOM structure and state:
- Use keyup to handle tabbing into the player
- Use blur to handle tabbing out
- Pressing escape key hides controls (if not intercepted by menu or overlay handler)
- Pass `isKeyEvent` argument to `userActive` in menu's `visibilityChangeHandler` so that using escape to close the menu does not hide controls.

For related, there isn't anything that will show the controls when related is closed. We will need to do that at a later point. That along with this PR to stop propagation of the escape keydown event https://github.com/jwplayer/jwplayer-plugin-related/pull/171 will make the related overlay work the same way the menu does.

JW8-603

hghazzi:
@robwalch thanks!

PR #2461, opened 2017-10-09
Disable fitVideoUsingTransforms on Android

karimJWP:
### This PR will...
Disable fitVideoUsingTransforms on Android
To allow side-loaded captions to reposition themselves when controls are shown
n/a
n/a
#### Addresses Issue(s):
JW8-618

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -553,7 +553,7 @@ function VideoProvider(_playerId, _playerConfig) {
         // object-fit is not implemented in IE or Android Browser in 4.4 and lower
         // http://caniuse.com/#feat=object-fit
         // feature detection may work for IE but not for browsers where object-fit works for images only
-        const fitVideoUsingTransforms = Browser.ie || (OS.iOS && OS.version.major < 9) || (Browser.androidNative && !Browser.firefox);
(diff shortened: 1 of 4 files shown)


src/js/providers/html5.js
```diff
@@ -553,7 +553,7 @@ function VideoProvider(_playerId, _playerConfig) {
         // object-fit is not implemented in IE or Android Browser in 4.4 and lower
         // http://caniuse.com/#feat=object-fit
         // feature detection may work for IE but not for browsers where object-fit works for images only
-        const fitVideoUsingTransforms = Browser.ie || (OS.iOS && OS.version.major < 9) || (OS.android && !Browser.firefox);
+        const fitVideoUsingTransforms = Browser.ie || (OS.iOS && OS.version.major < 9);
```
robwalch: @egreaves looking at the comment and http://caniuse.com/#feat=object-fit I think we should use ` || (OS.androidNative && !Browser.firefox);`.

We could leave our firefox if `androidNative` excludes it.  If it doesn't now, we should make it.

Also as a follow up `androidNative` should be on `Browser` not `OS`

robwalch: @karimJWP we have Galaxy S devices that you can test this on.

PR #2464, opened 2017-10-10
Prevent firefox window event exception

karimJWP:
### This PR will...
check if window event exists before getting its pointer type. I chose to always focus the default element on firefox instead of never focusing it because I believe focusing is necessary for accessibility ( @hghazzi please correct me if i'm wrong) and the only reason we don't focus on click is to avoid the tooltip popping up.
To prevent a firefox exception
n/a
n/a
#### Addresses Issue(s):
JW8-732

### DIFF
--- a/src/js/view/controls/components/settings/menu.js
+++ b/src/js/view/controls/components/settings/menu.js
@@ -53,6 +53,7 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
             addDocumentListeners(documentClickHandler);
 
             if (isDefault) {
+                // TODO: remove window.event; instead, pass the event from the button interaction or use 'keyup' in controls.js
                 if (!window.event || !window.event.pointerType) {
                     active.categoryButtonElement.focus();
                 }
--- a/src/js/view/controls/components/settings/menu.js
+++ b/src/js/view/controls/components/settings/menu.js
@@ -53,7 +53,7 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
             addDocumentListeners(documentClickHandler);
 
             if (isDefault) {
-                if (!window.event.pointerType) {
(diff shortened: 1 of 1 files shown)


src/js/view/controls/components/settings/menu.js
```diff
@@ -53,7 +53,7 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
             addDocumentListeners(documentClickHandler);
 
             if (isDefault) {
-                if (!window.event.pointerType) {
```
robwalch: We should pass the event (`evt`) into open and close from the methods calling them. The logic on line 38 to determine `isKeyEvent` or `isDefault` should be in these methods.

robwalch: Do not use `window.event`. It is not a reliable way to get event context for code that is part of an event hander chain.

robwalch: @karimJWP @hghazzi you guys should pass this to @johnBartos to address. The event handling needs quick fix for release.

robwalch: If we can't address padding click listener events into here than let's remove the check. What's wrong with always focusing? Is this only supposed to happen on key press? If so you can do that using 'keyup' in controls.js

hghazzi:
@robwalch By commenting this out, you can no longer enter the submenu via the keyboard when hitting enter on the gear icon, only with CC and sharing.

src/js/view/controls/components/settings/menu.js
karimJWP: @robwalch  the reason we don't want to pass events is because of a comment left at https://github.com/jwplayer/jwplayer/blob/master/src/js/view/controls/components/button.js#L17 . It was left last year by Evol, do you think this is still relevant ?

PR #2466, opened 2017-10-16
[AB-1254] Fixed an issue where autostart would be muted in the SDK for Android

paulmandal:
### This PR will... 

Autostart unmuted (unless mute requested) in the SDK for Android.

Regression where autostart playback would be muted in the SDK for Android.

@karimJWP is autostart disallowed in iOS apps?  This will only autostart unmuted for Android.

No

#### Addresses Issue(s):

[AB-1254]

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -229,8 +229,9 @@ const Model = function() {
 
         _provider.volume(_this.get('volume'));
 
-        // Mute the video if autostarting on mobile. Otherwise, honor the model's mute value
-        _provider.mute(_this.autoStartOnMobile() || _this.get('mute'));
+        // Mute the video if autostarting on mobile, except for Android SDK. Otherwise, honor the model's mute value
+        const isAndroidSdk = _this.get('sdkplatform') === 1;
+        _provider.mute((this.autoStartOnMobile() && !isAndroidSdk) || _this.get('mute'));
 
         _provider.on('all', _videoEventHandler, this);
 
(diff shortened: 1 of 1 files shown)


src/js/controller/model.js
```diff
@@ -229,8 +229,9 @@ const Model = function() {
 
         _provider.volume(_this.get('volume'));
 
-        // Mute the video if autostarting on mobile. Otherwise, honor the model's mute value
-        _provider.mute(_this.autoStartOnMobile() || _this.get('mute'));
+        // Mute the video if autostarting on mobile, except for Android SDK. Otherwise, honor the model's mute value
+        const isAndroidSdk = _this.get('sdkplatform') === 1;
+        _provider.mute((this.autoStartOnMobile() && !isAndroidSdk) || _this.get('mute'));
```
egreaves: `this.autoStartOnMobile()` calls  `platformCanAutostart()`, which checks `isAndroidSdk`. You can make the change there and leave this condition as it was.

src/js/controller/model.js
```diff
@@ -655,7 +656,8 @@ const Model = function() {
         }
 
         const autoStartOnMobile = this.autoStartOnMobile();
-        if (autoStartOnMobile) {
+        const isAndroidSdk = _this.get('sdkplatform') === 1;
+        if (autoStartOnMobile && !isAndroidSdk) {
```
egreaves: If you make the change in `platformCanAutostart()`, you won't need this change.

src/js/controller/model.js
paulmandal: Not sure if I follow here -- won't changing `platformCanAutostart()` disable autostart in the SDK for Android?

egreaves: You're right - I was suggesting conflating autostart behavior with 'muted' autostart behavior. Your logic is correct.

PR #2467, opened 2017-10-16
Pressing enter on a submenu item triggers click event

hghazzi:
### This PR will...

reintroduce being able to hit enter on submenu items

To make submenus more accessible

#### Addresses Issue(s):

JW8-754

### DIFF
--- a/src/js/view/controls/components/settings/menu.js
+++ b/src/js/view/controls/components/settings/menu.js
@@ -22,7 +22,7 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
 
     const closeOnEnter = function(evt) {
         if (evt && evt.keyCode === 27) {
-            instance.close(true);
+            instance.close(evt);
             evt.stopPropagation();
         }
     };
@@ -36,7 +36,7 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
         // Close settings menu when enter is pressed on the close button
         // or when tab key is pressed since it is the last element in topbar
         if (evt.keyCode === 13 || (evt.keyCode === 9 && !evt.shiftKey)) {
-            instance.close(true);
+            instance.close(evt);
         }
     };
     closeButton.show();
@@ -62,9 +62,9 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
             }
 
         },
-        close(isKeyEvent = false) {
+        close(event = {}) {
             visible = false;
-            onVisibility(visible, isKeyEvent);
+            onVisibility(visible, event);
 
             active = null;
             deactivateAllSubmenus(submenus);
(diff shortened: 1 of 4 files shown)


src/js/view/utils/submenu-factory.js
```diff
@@ -37,14 +37,9 @@ export const makeSubmenu = (settingsMenu, name, contentItems, icon, tooltipText)
...
-        const contentItemElement = SettingsContentItem(track.id, track.label, () => {
+        const contentItemElement = SettingsContentItem(track.id, track.label, (evt) => {
             action(index);
-            settingsMenu.close();
-        });
-
-        contentItemElement.uiElement().on('enter', () => {
-            action(index);
-            settingsMenu.close(true);
+            settingsMenu.close(evt.type === 'enter');
```
egreaves: Since this is being done in multiple places, I'd suggest passing the event in and have the `close` function determine how it should react based on the event type.

src/js/view/controls/components/settings/menu.js
```diff
@@ -62,9 +62,9 @@ export function SettingsMenu(onVisibility, onSubmenuAdded, onMenuEmpty) {
             }
 
         },
-        close(isKeyEvent = false) {
+        close(event = {}) {
```
egreaves: You don't need a default value for `event` here. The only time a property on the event object is checked is in `visibilityChangeHandler `.

src/js/view/controls/controls.js
```diff
@@ -126,9 +126,10 @@ export default class Controls {
 
         // Settings Menu
         let lastState = null;
-        const visibilityChangeHandler = (visible, isKeyEvent) => {
+        const visibilityChangeHandler = (visible, evt = {}) => {
```
egreaves: Not a required change, but I still prefer doing an `undefined` check manually because of the output when a default argument is transpiled.

src/js/view/controls/controls.js
```diff
@@ -126,9 +126,10 @@ export default class Controls {
...
             const state = model.get('state');
             const settingsInteraction = { reason: 'settingsInteraction' };
+            const isKeyEvent = evt.sourceEvent ? evt.sourceEvent.type === 'keydown' : evt.type === 'keydown';
```
robwalch: Here is a short replacement for the default parameter that also removes duplication in your ternary statement
`const isKeyEvent = (evt && evt.sourceEvent || evt || {}).type === 'keydown';`

src/js/view/controls/controls.js
```diff
@@ -141,6 +142,7 @@ export default class Controls {
                 }
             }
 
+
```
robwalch: remove extra line

src/js/view/controls/controls.js
robwalch: Here is a short replacement for the default parameter that also removes duplication in your ternary statement
`const isKeyEvent = (evt && evt.sourceEvent || evt || {}).type === 'keydown';`

PR #2476, opened 2017-10-17
Remove build number from pop up window on right click of beta players

DanFerrer:
### This PR will...
Remove the build number on beta players when the user right clicks the player and sees the pop up window.

The build number means nothing to do the user and should only be seen in development.

#### Addresses Issue(s):

JW8-525

### DIFF
--- a/src/js/view/controls/rightclick.js
+++ b/src/js/view/controls/rightclick.js
@@ -19,7 +19,6 @@ export default class RightClick {
     buildArray() {
         var semverParts = version.split('+');
         var majorMinorPatchPre = semverParts[0];
-        var isBeta = majorMinorPatchPre.indexOf('beta') > 0;
 
         var menu = {
             items: [{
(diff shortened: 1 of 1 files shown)


src/js/view/controls/rightclick.js
```diff
@@ -33,8 +34,10 @@ export default class RightClick {
         var versionMeta = semverParts[1];
```
robwalch: We want to remove `versionMeta` from all prerelease players. There is no need to look for the word 'beta'. That was probably a miscommunication in the ticket requirements.

Just remove the use of  `isPrerelease` and `versionMeta` completely.

robwalch: For more clarity see the definition of "pre-release" at http://semver.org/

DanFerrer: @robwalch Ah, thanks! Removed the variables and the conditionals.

PR #2478, opened 2017-10-18
v7 Safari 11 fixes

robwalch:
### This PR will...

- Call play in html5 provider to start playback from an idle state even when preload is set to metadata.
- Only show play promise warning on 'NotAllowedError' error
- Make sure state is set back to paused if play fails

#### Addresses Issue(s):

JW8-736

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -69,6 +69,7 @@ define([
             loadeddata: _onLoadedData, // we have video tracks (text, audio, metadata)
             loadedmetadata: _loadedMetadataHandler, // we have video dimensions
             canplay: _canPlayHandler,
+            play: _loading,
             playing: _playingHandler,
             progress: _progressHandler,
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -446,29 +450,15 @@ define([
...
                 }
-
-                _play();
             }
 
             _position = _videotag.currentTime;
 
-            if (_isMobile && !hasPlayed) {
-                // results in html5.controller calling video.play()
-                _sendBufferFull();
```
egreaves: Since we only call `_sendBufferFull` after the `canPlay` event, we can get rid of some code in the `_sendBufferFull` function. We no longer need to check if we're on iOS and we may also be able to get rid of the `_canPlay` flag:
```
if (!_bufferFull && (!utils.isIOS() || _canPlay)) {
    ...
}
```

PR #2482, opened 2017-10-18
Custom controlbar background shouldn't hide captions on mobile

hghazzi:
### This PR will...

Only apply the custom controlbar background colour up to the time slider and not the entire touch target on mobile devices

So that the background colour does not obscure captions

#### Addresses Issue(s):

JW8-685

### DIFF
--- a/src/css/controls/imports/slider.less
+++ b/src/css/controls/imports/slider.less
@@ -160,11 +160,6 @@
         }
     }
 
-    .jw-flag-touch & {
-        align-items: flex-end;
-        padding-bottom: 5px;
-    }
-
     .jw-flag-touch &::before {
         height: 44px;
         width: 100%;
--- a/src/css/controls/imports/slider.less
+++ b/src/css/controls/imports/slider.less
@@ -161,8 +161,17 @@
     }
 
     .jw-flag-touch & {
-        .rect(100%, 44px);
         align-items: flex-end;
         padding-bottom: 5px;
+    }
+
+    .jw-flag-touch &::before {
+        height: 44px;
+        width: 100%;
+        content: "";
+        position: absolute;
+        display: block;
+        bottom: calc(100% - 17px);
+        left: 0;
     }
 }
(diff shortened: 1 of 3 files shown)


src/js/view/utils/skin.js
```diff
@@ -165,6 +165,12 @@ export function handleColorOverrides(playerId, skin) {
             ':not(.jw-state-idle) .jw-controlbar',
             '.jw-flag-audio-player .jw-controlbar'
         ], 'background', config.background, true);
+
+        // For touch devices, we want the height of the controlbar to be 88px so it would have a larger touch target.
+        // But we want the background color to only affect the area up to the time slider, which has a height of 61px.
+        // Using a sharp linear gradient to handle this, so need hard-code the transparent area to 27px (88px - 61px)
+        css(`#${playerId}.jw-flag-touch:not(.jw-flag-audio-player):not(.jw-state-idle) .jw-controlbar`,
+            `{background: linear-gradient(transparent 27px, ${config.background} 27px, ${config.background})}`);
```
robwalch: Since this code is far removed from the change in controlbar that sets the 88px height, please include a github.com link to that change specifically in your comment. Ex:

```js
// For touch devices, we set the height of the controlbar to 88px https://github.com/...```

Is it here? https://github.com/jwplayer/jwplayer/blob/b5da824a4b81b29d66bee9e3722abeb8a557dbf9/src/css/controls/imports/controlbar.less#L39

Comment (hghazzi) [reply src/js/view/utils/skin.js]: Good idea, thanks!

**Comment by karimJWP** [comment]
@hghazzi why is the settings tooltip empty in your build? does it need a rebase ? 
screen shot 2017-10-23 at 11 01 01 am

**Comment by radium-v** [comment]
This would be better done with a `::before` that doesn't affect the height of the container

**Comment by egreaves** [comment]
@hghazzi can you follow up with @radium-v re: his feedback?

**Comment by hghazzi** [comment]
@radium-v updated the change to use `::before`

File: src/css/controls/imports/slider.less [hunk 2]
```diff
@@ -161,8 +161,17 @@
     }
 
     .jw-flag-touch & {
-        .rect(100%, 44px);
         align-items: flex-end;
         padding-bottom: 5px;
```
robwalch: These styles are stragglers. They don't seem to do anything now. Their purpose was to push the slider back down and that isn't necessary if the `::before` is absolute.

```css
align-items: flex-end;
padding-bottom: 5px;
```

hghazzi:
@robwalch removed the unused css and tested it on both android and ios phones and was able to verify scrubbing works.

PR #2485, opened 2017-10-20
pass event instead of window.event in menu.js

DanFerrer:
### This PR will...

Remove the check for window.event and pass in the event from the element event listener in button.js to menu.js to determine if the tooltip should be focused or not.

Window.event doesn't exist in Firefox and Safari. Passing in the initial event from the button makes it easier to determine if the event type is appropriate to show the tooltip .

Passing in the event to the apiAction in button.js

#### Addresses Issue(s):
JW8-741


egreaves:
test this please

PR #2486, opened 2017-10-23
Set native captions tracks on attach media in html5 provider

robwalch:
### This PR will...

- Reset text tracks player model on attachMedia
- Revert `setupSideloadedTracks` pushed to v7.12.x branch

Because ads plugins don't fully load the video and any 608 captions before the ad starts, textTracks need to be set on attach.

#### Addresses Issue(s):

JW8-797


robwalch:
test this please

PR #2487, opened 2017-10-24
Fire "seek" event when currentTime is set externally (iOS fullscreen)

### LINKED ISSUE #2456: iOS fullscreen missing seek event
### Expected Behavior

Seek event occurs before seeked

### Actual Behavior

Seek event not present

### Steps to reproduce

1. Create basic html page with JW Player
2. Open it in WebView in iOS app 
3. Play content (it should open if fullscreen mode) and then seek back and forth.
4. Seek event is missing while seeked works correctly.

### Environment

- *iPhone 7*
- *iOS 10*
- *iOS app with WebView*
- *JW Player version - 7.12.8*

robwalch:
### This PR will...

Fire "seek" event when currentTime is set externally (iOS fullscreen)

To track seeking from start to end ("seek" event to "seeked" event) when `video.currentTime` is set externally, or through native media element controls.

In order to keep the "seek" event `position` (currentTime before seeking) and `offset` (currentTime requested) properties consistent, I had to store `_seekOffset` and update `_position` before the seeking handler runs as a result of currentTime being set. Otherwise `position` would not be up to date at the moment seek is called and `offset` would reflect the time requested modified by the media element's currentTime setter which caps time at duration and may by affected in other ways when writing and reading asynchronously.

I also added a guard so that we don't attempt to seek to `-1` in live streams or streams assumed to be live while loading (this may be common on Android with HLS where duration is not reported properly until a rendition is ready to play).

#### Addresses Issue(s):

JW8-783

Fixes #2456 in v7

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -73,6 +73,7 @@ define([
             playing: _playingHandler,
             progress: _progressHandler,
             pause: _pauseHandler,
+            seeking: _seekingHandler,
             seeked: _seekedHandler,
             timeupdate: _timeUpdateHandler,
             ratechange: _playbackRateHandler,
@@ -86,6 +87,7 @@ define([
         var _canSeek = false;
         var _bufferFull;
         var _delayedSeek = 0;
+        var _seekOffset = null;
         var _playbackTimeout = -1;
         var _buffered = -1;
         var _levels;
@@ -248,7 +250,8 @@ define([
 
         function _updateDuration(duration) {
             _duration = duration;
-            if (_delayedSeek && duration && duration !== Infinity) {
+            // Don't seek when _delayedSeek is set to -1 in _completeLoad
+            if (_delayedSeek && _delayedSeek !== -1 && duration && duration !== Infinity) {
                 _this.seek(_delayedSeek);
             }
         }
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -248,7 +250,8 @@ define([
 
         function _updateDuration(duration) {
             _duration = duration;
-            if (_delayedSeek && duration && duration !== Infinity) {
+            // Don't seek when _delayedSeek is set to -1 in _completeLoad
+            if (_delayedSeek && _delayedSeek !== -1 && duration && duration !== Infinity) {
                 _this.seek(_delayedSeek);
```
karimJWP: @robwalch would this work with [HAVE_METADATA](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/readyState)? Description says: 'Seeking will no longer raise an exception.'

robwalch: Which description?

The issue is that this hack was added, without consideration of the code above:
https://github.com/jwplayer/jwplayer/blob/d30fee0e70ef211b4a0700787f65830fbe758892/src/js/providers/html5.js#L459

Since we're using `-1` to prevent seeking there, we should not allow it in `_updateDuration`.

robwalch: Regarding " HAVE_METADATA" readyState, we don't use it to determine when we can seek - that is not reliable. We use seekable ranges, duration and the "canplay" or "timeupdate" media element events (whichever comes first).

PR #2489, opened 2017-10-24
Add new API to replace one playlist item

pajong:
### This PR will...
Add a new API that replaces a playlist item at the given index, instead of replacing the whole playlist.

DAI playlist support needs this function so that the stream that we get from google DAI only replaces one playlist item.

#### Addresses Issue(s):
ADS-523

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -669,6 +669,18 @@ export default function Api(element) {
         },
 
         /**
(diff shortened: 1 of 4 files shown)


robwalch:
Let's discuss this API decision and alternatives before implementing anything.

robwalch:
@pajong did you try replacing the item's sources? How about changing the instream API instead of the player API?

PR #2491, opened 2017-10-25
Adjust bottom on Ad Skip Button so it isn't obscured by a customized control bar

DanFerrer:
### This PR will...

Change the position of the skip ad overlay

On custom coloured control bars, the skip ad overlay rests on top of the control bar so there needs to be more space between the skip ad overlay and control bar. Adjusting the bottom by .5em fixes it.

#### Addresses Issue(s):

JW8-412

### DIFF
--- a/src/css/controls/imports/skipad.less
+++ b/src/css/controls/imports/skipad.less
@@ -6,7 +6,7 @@
     position: absolute;
     display: flex;
     right: 0.75em;
-    bottom: 3em;
+    bottom: 3.5em;
     padding: @ui-padding;
     border: 1px solid lighten(@black, 20%);
     background-color: @black;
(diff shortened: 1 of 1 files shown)


src/css/controls/imports/skipad.less
```diff
@@ -6,7 +6,7 @@
     position: absolute;
     display: flex;
     right: 0.75em;
-    bottom: 3em;
+    bottom: 3.5em;
```
egreaves: I'd prefer migrating to using pixel values. I believe this would be `3.5x16=56`. (ref: [default em size](https://github.com/jwplayer/jwplayer/blob/a935629f36a5c40448e1aa3d9c957e80abe8259e/src/css/shared-imports/vars.less#L1)) (also: [just use pixels](https://benfrain.com/just-use-pixels/))

PR #2492, opened 2017-10-25
Make function to replace current source

pajong:
### This PR will...
Add a new function that replaces a playlist item at the given index, instead of replacing the whole playlist.

DAI playlist support needs this function so that the stream that we get from google DAI only replaces one playlist item.

#### Addresses Issue(s):
ADS-523

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -276,6 +276,14 @@ var InstreamAdapter = function(_controller, _model, _view) {
         }
     };
 
+    this.replacePlaylistItem = function(url) {
+        const item = _model.get('playlistItem');
+        item.sources[0].file = url;
+        item.file = url;
+
+        _model.mediaModel.set('setup', false);
+    };
+
     this.destroy = function() {
         this.off();
 
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -276,12 +276,8 @@ var InstreamAdapter = function(_controller, _model, _view) {
         }
     };
 
-    this.replacePlaylistItem = function(url) {
-        const item = _model.get('playlistItem');
-        item.sources[0].file = url;
-        item.file = url;
-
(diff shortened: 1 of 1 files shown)


src/js/controller/instream-adapter.js
```diff
@@ -276,6 +276,14 @@ var InstreamAdapter = function(_controller, _model, _view) {
         }
     };
 
+    this.replacePlaylistItem = function(url) {
+        const item = _model.get('playlistItem');
+        item.sources[0].file = url;
```
robwalch: What do you think about replacing `sources` with a single source?

`item.sources = [Object.assign(item.sources[0], {file: url})]`

robwalch: If we want to leave the playlist unmodified, then we need to set 'playlistItem'. That would ensure that if we left this playlist item, and came back to it, that the same replacement would happen - otherwise we would feed the modified url back to the SDK.

pajong: Thank you for looking into this! I wanted to run by you on this actually. If we request stream to the SDK on every playlistItem, it would re-request stream when we do set 'playlistItem'. Do you have any thoughts on how we could prevent that?

robwalch: If you don't modify the original object using my example above, then you can use `allSources`. We should be using `allSources` to find the correct media type to pass to the SDK anyway.

src/js/controller/instream-adapter.js
```diff
@@ -276,6 +276,14 @@ var InstreamAdapter = function(_controller, _model, _view) {
         }
     };
 
+    this.replacePlaylistItem = function(url) {
```
robwalch: How about `this.setPlaylistItem(item)`? So in the plugin you can use get the item with `api.getPlaylistItem()` make these changes, and this just does `_model.set('playlistItem', item)`.

pajong:
@robwalch Changed to that it only replaces the playlistItem (short and simple). Thank you.

PR #2493, opened 2017-10-25
Enable seeking event on ios fullscreen

karimJWP:
### This PR will...
listen to the html5 seeking event before firing our seek event instead of firing it when the seek command is executed.
To enable our seeking events to be fired when seeking in iOS's native player during fullscreen.
1) I had to store the position in a 'positionBeforeSeek' variable to have an accurate initial position when seeking in iOS fullscreen. Storing the position was removed after version 7. Currently it is set as time updates and after seeking since scrubbing in ios fullscreen will fire multiple seek events.
2) this.seeking = true is set twice; do we need it to be set on the seek command ? It is also set twice in [v7's fix for this bug](https://github.com/jwplayer/jwplayer/blob/d30fee0e70ef211b4a0700787f65830fbe758892/src/js/providers/html5.js#L638)

#### Addresses Issue(s):
JW8-708

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -114,6 +114,17 @@ function VideoProvider(_playerId, _playerConfig) {
             VideoEvents.canplay.call(_this);
         },
 
+        seeking() {
+            var offset = _seekOffset !== null ? _seekOffset : _videotag.currentTime;
+            _seekOffset = null;
+            _delayedSeek = 0;
+            _this.seeking = true;
+            _this.trigger(events.JWPLAYER_MEDIA_SEEK, {
+                position: _position,
+                offset: offset
+            });
+        }
+
         webkitbeginfullscreen(e) {
             _fullscreenState = true;
             _sendFullscreen(e);
@@ -182,6 +193,7 @@ function VideoProvider(_playerId, _playerConfig) {
 
     let _canSeek = false; // true on valid time event
     let _delayedSeek = 0;
+    let _seekOffset = null;
     let _levels;
     let _currentQuality = -1;
     let _fullscreenState = false;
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -229,13 +244,20 @@ function VideoProvider(_playerId, _playerConfig) {
...
+
     _this.getCurrentTime = function() {
-        let currentTime = _videotag.currentTime;
+        return _convertTime(_videotag.currentTime);
+    };
+
+    function _convertTime(position) {
         if (_this.getDuration() < 0) {
-            currentTime = -(_getSeekableEnd() - currentTime);
+            position = -(_getSeekableEnd() - position);
```
egreaves: this is essentially 
`position = position -_getSeekableEnd();` or 
`position -= _getSeekableEnd()`;

egreaves:
@karimJWP LGTM. One suggestion - I prefer keeping the old name, `setPosition`, which is more reflective of what's being done in the `timeUpdate` event handler, for instance, where the position is being tracked not just for the purpose of seeking.

karimJWP:
@egreaves I would prefer to keep _setPositionBeforeSeek instead of just _setPosition because I think it communicates clearly why we are storing the position. My concern is that since setPosition was removed in 8, it is probably being handled somewhere else; having a clear name would prevent someone from thinking that storing the position is redundant and removing it. Even though that person is diligent he may overlook this because the benefit of storing the position is only seen during fullscreen iOS since we are rendering on iOS's native player and the only way to track the time before a seek event is with the timeUpdate.

karimJWP:
@egreaves my personal preference is clear function names instead of comments, but if you feel strongly about renaming it, I can do so.

PR #2495, opened 2017-10-27
Prevent hover state on buffer icon  for custom skins

DanFerrer:
### This PR will...

Prevent hover state from being triggered when the user has a custom player skin and hovers on the buffer icon when the player is buffering.

Buffer icon should remain stateless when hovered on.

#### Addresses Issue(s):

JW8-659

### DIFF
--- a/src/js/view/utils/skin.js
+++ b/src/js/view/utils/skin.js
@@ -142,13 +142,14 @@ export function handleColorOverrides(playerId, skin) {
             addStyle([
                 // Toggle and menu button active colors
                 '.jw-button-color.jw-toggle:not(.jw-icon-cast)',
+                '.jw-button-color:hover:not(.jw-icon-cast)',
                 '.jw-button-color:focus:not(.jw-icon-cast)',
                 '.jw-button-color.jw-toggle.jw-off:hover:not(.jw-icon-cast)'
             ], 'color', config.iconsActive);
 
             addStyle([
-                ':not(.jw-state-buffering) .jw-button-color:hover:not(.jw-icon-cast)'
-            ], 'color', config.iconsActive, true);
+                '.jw-svg-icon-buffer',
+            ], 'fill', config.icons);
 
             // Chromecast overrides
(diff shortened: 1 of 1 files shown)


src/js/view/utils/skin.js
```diff
@@ -142,11 +142,14 @@ export function handleColorOverrides(playerId, skin) {
             addStyle([
                 // Toggle and menu button active colors
                 '.jw-button-color.jw-toggle:not(.jw-icon-cast)',
-                '.jw-button-color:hover:not(.jw-icon-cast)',
                 '.jw-button-color:focus:not(.jw-icon-cast)',
                 '.jw-button-color.jw-toggle.jw-off:hover:not(.jw-icon-cast)'
             ], 'color', config.iconsActive);
 
+            addStyle([
+                ':not(.jw-state-buffering) .jw-button-color:hover:not(.jw-icon-cast)'
```
karimJWP: @DanFerrer would this prevent all icons from going to active state when hovered over ? I think the buffer icon is the only one that should be prevented

DanFerrer: @karimJWP Possibly, but the buffering icon is the only one that's shown when the player state is buffering. All other svgs with the class jw-button-color are hidden.

egreaves: @karimJWP - @DanFerrer is right in that it wouldn't. `:not(.jw-state-buffering)` ensures no hover color is applied when the player is in a buffering state.

egreaves:
@DanFerrer - this is also a problem for the default player, but it's less noticeable because of the transition from `rgba(255,255,255,0.8)` to `#fff`. Take a look [here](http://player-pr-test-jenkins.longtailvideo.com/builds/10374/archive/test/view/).

src/js/view/utils/skin.js
egreaves: @karimJWP you're right, actually. Dan's addressing your feedback now.

PR #2496, opened 2017-10-27
Display buffer icon in Safari when seeking outside of buffer

karimJWP:
### This PR will...
- listen for 'waiting' to trigger buffer state, if seeking outside the buffered area
- ignore timeupdates when seeking
- ignore timeupdates if time does not change
- prevent setting playing state on playing event; playing event means play was requested, not that playback is actually occurring. 
To display buffer icon when seeking outside of buffer area in safari for HLS Streams
- removed setting the playing state on the playing event because it gets triggered when play is requested, not when able. For example, it gets triggered before seek is complete, see attachment:
screen shot 2017-10-26 at 4 56 22 pm
- does not work for mp4s on Firefox since waiting event is never triggered, see attachment:
screen shot 2017-10-27 at 5 30 01 pm
- removed playing since it does not mean playback is occurring and gets fired while seeking, see attachment: 
screen shot 2017-10-27 at 5 33 28 pm

#### Addresses Issue(s):
JW8-725

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -55,67 +55,6 @@ function VideoProvider(_playerId, _playerConfig) {
 
     const _this = this;
 
-    // karim >
-
-    function observe(media) {
-        const observer = new MutationObserver(function(mutations) {
-            mutations.forEach(function(mutation) {
-                console.warn('mutation', mutation);
-            });
-        });
-        observer.observe(media, {
-            attributes: true
-        });
-        const load = media.load;
-        const pause = media.pause;
-        const play = media.play;
-        media.load = function() {
-            console.error('media.load()');
-            return load.call(this);
-        };
-        media.pause = function() {
-            console.error('media.pause()');
-            return pause.call(this);
-        };
-        media.play = function() {
-            console.error('media.play()');
-            return play.call(this);
-        };
-        function videoEventHandler(e) {
-            console.warn('>> "' + e.type + '"');
-        }
-        function videoEventHandlerSpecial(e) {
-            console.error('>> "' + e.type + '"');
-        }
-        media.addEventListener('loadstart', videoEventHandler);
-        media.addEventListener('progress', videoEventHandler);
-        media.addEventListener('suspend', videoEventHandler);
-        media.addEventListener('abort', videoEventHandler);
-        media.addEventListener('error', videoEventHandler);
-        media.addEventListener('emptied', videoEventHandler);
-        media.addEventListener('stalled', videoEventHandler);
-        media.addEventListener('loadedmetadata', videoEventHandler);
-        media.addEventListener('loadeddata', videoEventHandler);
-        media.addEventListener('canplay', videoEventHandler);
-        media.addEventListener('canplaythrough', videoEventHandler);
-        media.addEventListener('playing', videoEventHandler);
-        media.addEventListener('waiting', videoEventHandler);
-        media.addEventListener('seeking', videoEventHandlerSpecial);
-        media.addEventListener('seeked', videoEventHandlerSpecial);
-        media.addEventListener('ended', videoEventHandler);
-        media.addEventListener('durationchange', videoEventHandler);
-        media.addEventListener('timeupdate', videoEventHandler);
-        media.addEventListener('play', videoEventHandler);
-        media.addEventListener('pause', videoEventHandler);
-        media.addEventListener('ratechange', videoEventHandler);
-        media.addEventListener('resize', videoEventHandler);
-        media.addEventListener('volumechange', videoEventHandler);
-    }
-
-//    observe(media);
-
-    // < karim
-
     const MediaEvents = {
         progress() {
             VideoEvents.progress.call(_this);
@@ -123,6 +62,9 @@ function VideoProvider(_playerId, _playerConfig) {
         },
         
         timeupdate() {
+            if (_positionBeforeSeek === _videotag.currentTime) {
+                return;
+            }
             _setPositionBeforeSeek(_videotag.currentTime);
             VideoEvents.timeupdate.call(_this);
             checkStaleStream();
@@ -191,7 +133,6 @@ function VideoProvider(_playerId, _playerConfig) {
 
         seeked() {
             VideoEvents.seeked.call(_this);
-//            _videotag.removeEventListener('waiting', () => _this.setState(STATE_BUFFERING));
             _videotag.removeEventListener('waiting', setBufferingState);
         },
 
@@ -330,7 +271,6 @@ function VideoProvider(_playerId, _playerConfig) {
         }
 
         if (outOfBufferRange) {
-            console.error('outOfBufferRange');
             _videotag.addEventListener('waiting', setBufferingState);
         }
     }
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -55,6 +55,67 @@ function VideoProvider(_playerId, _playerConfig) {
 
     const _this = this;
 
+    // karim >
+
+    function observe(media) {
+        const observer = new MutationObserver(function(mutations) {
+            mutations.forEach(function(mutation) {
+                console.warn('mutation', mutation);
+            });
+        });
+        observer.observe(media, {
+            attributes: true
+        });
+        const load = media.load;
+        const pause = media.pause;
+        const play = media.play;
+        media.load = function() {
+            console.error('media.load()');
+            return load.call(this);
+        };
+        media.pause = function() {
+            console.error('media.pause()');
+            return pause.call(this);
+        };
+        media.play = function() {
+            console.error('media.play()');
+            return play.call(this);
+        };
+        function videoEventHandler(e) {
+            console.warn('>> "' + e.type + '"');
+        }
+        function videoEventHandlerSpecial(e) {
+            console.error('>> "' + e.type + '"');
+        }
+        media.addEventListener('loadstart', videoEventHandler);
+        media.addEventListener('progress', videoEventHandler);
+        media.addEventListener('suspend', videoEventHandler);
+        media.addEventListener('abort', videoEventHandler);
+        media.addEventListener('error', videoEventHandler);
+        media.addEventListener('emptied', videoEventHandler);
+        media.addEventListener('stalled', videoEventHandler);
+        media.addEventListener('loadedmetadata', videoEventHandler);
+        media.addEventListener('loadeddata', videoEventHandler);
+        media.addEventListener('canplay', videoEventHandler);
+        media.addEventListener('canplaythrough', videoEventHandler);
+        media.addEventListener('playing', videoEventHandler);
+        media.addEventListener('waiting', videoEventHandler);
+        media.addEventListener('seeking', videoEventHandlerSpecial);
+        media.addEventListener('seeked', videoEventHandlerSpecial);
+        media.addEventListener('ended', videoEventHandler);
+        media.addEventListener('durationchange', videoEventHandler);
+        media.addEventListener('timeupdate', videoEventHandler);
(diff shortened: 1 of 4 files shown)


src/js/providers/html5.js
```diff
@@ -244,6 +253,30 @@ function VideoProvider(_playerId, _playerConfig) {
...
+            for (let i = 0; i < bufferedRange.length; i++) {
+                withinBuffer = position > bufferedRange.start(i) && position < bufferedRange.end(i);
+                if (withinBuffer) {
+                    break;
+                }
+            }
+        }
+
+        if (!withinBuffer) {
+            _videotag.addEventListener('waiting', setBufferingState);
```
egreaves: Remove any `'waiting'` listener before adding one here to be safe.

src/js/providers/video-listener-mixin.js
```diff
@@ -105,7 +108,6 @@ const VideoListenerMixin = {
     },
 
     playing() {
-        this.setState(STATE_PLAYING);
```
robwalch: This makes sense for iOS, but not for Chrome and Firefox as the behavior you are observing is iOS/Safari specific behavior.

src/js/providers/video-listener-mixin.js
```diff
@@ -56,7 +59,7 @@ const VideoListenerMixin = {
             return;
         }
 
-        if (!this.video.paused && (this.state === STATE_STALLED || this.state === STATE_LOADING)) {
```
robwalch: Provider state should never be set to buffering. Providers use `STATE_LOADING` and `STATE_STALLED` which both map to `STATE_BUFFERING` in the player.

src/js/providers/html5.js
```diff
@@ -2,7 +2,7 @@ import { qualityLevel } from 'providers/data-normalizer';
 import { Browser, OS } from 'environment/environment';
 import { isAndroidHls } from 'providers/html5-android-hls';
 import { STATE_IDLE, STATE_PLAYING, MEDIA_META, MEDIA_ERROR,
-    MEDIA_LEVELS, MEDIA_LEVEL_CHANGED, MEDIA_SEEK } from 'events/events';
+    MEDIA_LEVELS, MEDIA_LEVEL_CHANGED, MEDIA_SEEK, STATE_BUFFERING } from 'events/events';
```
robwalch: Remove `STATE_BUFFERING` and use `STATE_LOADING` (or stalled for under-buffering).

src/js/providers/html5.js
```diff
@@ -244,6 +253,31 @@ function VideoProvider(_playerId, _playerConfig) {
...
+        if (!withinBuffer) {
+            _videotag.addEventListener('waiting', setBufferingState);
+        }
+    }
+
+    function setBufferingState() {
+        _this.setState(STATE_BUFFERING);
+    }
+
     function _setPositionBeforeSeek(position) {
```
robwalch: This doesn't make sense as a function - it's only called in one place. Just update `_position` or `_positionBeforeSeek`  in `timeupdate`.

src/js/providers/html5.js
```diff
@@ -125,6 +128,12 @@ function VideoProvider(_playerId, _playerConfig) {
                 offset: offset
             });
             _setPositionBeforeSeek(offset);
```
robwalch: At this point where "seeking" is fired, `currentTime` has already changed. This line should be removed.

src/js/providers/html5.js
```diff
@@ -62,6 +62,9 @@ function VideoProvider(_playerId, _playerConfig) {
         },
         
         timeupdate() {
+            if (_positionBeforeSeek === _videotag.currentTime) {
+                return;
+            }
```
robwalch: Can this check be avoided? I noticed in the mixin we exit if `this.seeking` is true.

src/js/providers/html5.js
karimJWP: @robwalch this is used when seeking in fullscreen iOS. On iOS fullscreen you can only scrub so you get multiple seeking events as you scrub. if we don't update the position after each seek event, the position in the callback is never updated.

src/js/providers/html5.js
karimJWP: @robwalch once seeking has ended, we're still in a buffering state but we get timeupdates where the current time has not changed. What that causes is the state being set to playing even though you're not done buffering, so the video stays frozen without  a buffer icon

src/js/providers/html5.js
karimJWP: so you end up having onSeek {position 0 - offset 1} , onSeek {position 0 - offset 2} , onSeek {position 0 - offset 3}
instead of onSeek {position 0 - offset 1} , onSeek {position 1 - offset 2} , onSeek {position 2 - offset 3}

src/js/providers/video-listener-mixin.js
karimJWP: makes sense, I could prevent the state change if seeking === true, would you be ok with that ? ( context:  this was removed because a playing event is fired in between seeking and seeked, causing the player to exit the buffering state).
so `playing() {
         if (seeking) {
                return;
         }
         this.setState(STATE_PLAYING);
}`

src/js/providers/html5.js
karimJWP: 👍

src/js/providers/video-listener-mixin.js
karimJWP: 👍

src/js/providers/html5.js
robwalch: Shouldn't we update it on "seeked" then?

src/js/providers/html5.js
```diff
@@ -125,6 +128,12 @@ function VideoProvider(_playerId, _playerConfig) {
                 offset: offset
             });
             _setPositionBeforeSeek(offset);
+            _checkForBuffering(offset);
+        },
+
+        seeked() {
+            VideoEvents.seeked.call(_this);
+            _videotag.removeEventListener('waiting', setBufferingState);
```
robwalch: Remove 'waiting' before `VideoEvents.seeked`. `VideoEvents.seeked` triggers `MEDIA_SEEKED` which could result in additional code running in an event handler. We should always remove event listeners first, and run code that dispatches events as late as possible.

src/js/providers/html5.js
```diff
@@ -244,6 +253,31 @@ function VideoProvider(_playerId, _playerConfig) {
         }
     }
 
+    function _checkForBuffering(position) {
+        const bufferedRange = _videotag.buffered;
+        if (bufferedRange.length <= 0) {
```
robwalch: length cannot be < 0. `!bufferedRange.length` will work.

src/js/providers/html5.js
```diff
@@ -244,6 +253,31 @@ function VideoProvider(_playerId, _playerConfig) {
...
+            return;
+        }
+        let withinBuffer;
+        if (position <= bufferedRange.end(bufferedRange.length - 1)) {
+            for (let i = 0; i < bufferedRange.length; i++) {
+                withinBuffer = position > bufferedRange.start(i) && position < bufferedRange.end(i);
+                if (withinBuffer) {
+                    break;
```
robwalch: If you move `_videotag.removeEventListener('waiting', setBufferingState);` before this loop, we can replace the `withinBuffer` variable with a return statement. Avoiding declaration of variables whenever possible:

```js
_videotag.removeEventListener('waiting', setBufferingStat
if (position <= bufferedRange.end(bufferedRange.length - 1)) {
    for (let i = 0; i < bufferedRange.length; i++) {
        if(position > bufferedRange.start(i) && position < bufferedRange.end(i);
            return;
}}}
_videotag.addEventListener('waiting', setBufferingState);
```

I don't see the point of adding this event listener only when seeking outside the buffer range. I would either always emit a `STATE_LOADING` event when seeking outside the buffer range OR always listen for 'waiting' to trigger `STATE_LOADING` regardless of `_videotag.buffered` state. Putting these two together just makes the solution more complicated and less performant.

src/js/providers/video-listener-mixin.js
robwalch: Yes. That makes sense.

src/js/providers/html5.js
robwalch: Seeking and buffering should be done on "seeked" where we set `this.seeking` to false.

src/js/providers/html5.js
```diff
@@ -244,6 +253,31 @@ function VideoProvider(_playerId, _playerConfig) {
...
+        }
+
+        _videotag.removeEventListener('waiting', setBufferingState);
+        if (!withinBuffer) {
+            _videotag.addEventListener('waiting', setBufferingState);
+        }
+    }
+
+    function setBufferingState() {
+        _this.setState(STATE_BUFFERING);
```
robwalch: The state should be changed to `STATE_LOADING`.

src/js/providers/html5.js
robwalch: If you want to add support for Firefox we could also add an additional listener for "progress" but I would prefer we open a bug with mozilla to get the "waiting" event to match other browsers. "progress" could mean that the other parts of the video are still loading not essential to playback of the part we're seeking to, and "stalled" means that the browser data from the server is not forthcoming - but that data could also be not essential to playback.

src/js/providers/html5.js
karimJWP: it's called in pause, seeking, seek and time update

src/js/providers/video-listener-mixin.js
```diff
@@ -37,6 +37,9 @@ const VideoListenerMixin = {
 
     timeupdate() {
         this.stopStallCheck();
+        if (this.seeking) {
+            return;
+        }
```
robwalch: Let's move this to after `this._helperLastVideoHeight = height` so we still get adaption (visual quality) events.

src/js/providers/html5.js
karimJWP: @robwalch seeked is only fired at the end of scrubbing, so when removing `_setPositionBeforeSeek(offset);` we're left with
screen shot 2017-10-31 at 12 27 25 pm

it seemed confusing to me but if you think it's fine i don't mind removing the line

src/js/providers/html5.js
karimJWP: @robwalch buffering begins upon seeking, and continues after seeked
screen shot 2017-10-31 at 12 35 30 pm

waiting is fired right after seek, and the time update fired after seeked has not changed, therefore video has not progressed so video is still buffering.

src/js/providers/video-listener-mixin.js
```diff
@@ -105,6 +109,9 @@ const VideoListenerMixin = {
     },
 
     playing() {
+        if (this.seeking) {
+            return;
+        }
```
robwalch: Make this a guard around `STATE_PLAYING` so that we don't miss firing `PROVIDER_FIRST_FRAME`:

```js
playing() {
    if (!this.seeking) {
         this.setState(STATE_PLAYING);
    }
    this.trigger(PROVIDER_FIRST_FRAME);
},
```

src/js/providers/video-listener-mixin.js
```diff
@@ -50,6 +50,10 @@ const VideoListenerMixin = {
         }
         this._helperLastVideoHeight = height;
 
+        if (this.seeking) {
+            return;
+        }
```
robwalch: This conflicts with the statement below `if (this.state === STATE_PLAYING || this.seeking) {`. 

I think what you want is to add !seeking as a guard around here, but we might not need it anymore.

`if (!this.video.paused && (this.state === STATE_STALLED || this.state === STATE_LOADING)) {`

PR #2497, opened 2017-10-29
Use getBoundingClientRect on player element for right click menu position

DanFerrer:
### This PR will...

Use getBoundingClientRect on player element to determine where the right click menu pops up on the player rather than traversing the DOM upwards until the player element is found. 

If the target  of the mouse event is too deeply nested, the right click menus left and top position isn't correct. It's better to base the position of the right click menu relative to the player rather than the events target. 

#### Addresses Issue(s):

JW8-632

### DIFF
--- a/src/js/view/controls/rightclick.js
+++ b/src/js/view/controls/rightclick.js
@@ -2,7 +2,7 @@ import rightclickTemplate from 'view/controls/templates/rightclick';
 import { cloneIcon } from 'view/controls/icons';
 import { version } from 'version';
 import { flashVersion } from 'utils/browser';
-import { createElement, emptyElement, addClass, removeClass } from 'utils/dom';
+import { createElement, emptyElement, addClass, removeClass, bounds } from 'utils/dom';
 import UI from 'utils/ui';
 
 function createDomElement(html) {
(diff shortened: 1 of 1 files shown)


src/js/view/controls/rightclick.js
```diff
@@ -56,17 +56,10 @@ export default class RightClick {
...
-        var x = evt.offsetX || evt.layerX;
-        var y = evt.offsetY || evt.layerY;
-        while (target !== this.playerElement) {
-            x += target.offsetLeft;
-            y += target.offsetTop;
-
-            target = target.parentNode;
-        }
-
+        var bounds = this.playerElement.getBoundingClientRect();
```
egreaves: We have a utility method that does this safely. Import it from `dom`:
```
import { createElement, emptyElement, addClass, removeClass, bounds } from 'utils/dom';
...
const playerBounds = bounds(this.playerElement);
```

src/js/view/controls/rightclick.js
```diff
@@ -56,16 +56,9 @@ export default class RightClick {
...
-        var x = evt.offsetX || evt.layerX;
-        var y = evt.offsetY || evt.layerY;
-        while (target !== this.playerElement) {
-            x += target.offsetLeft;
-            y += target.offsetTop;
-
-            target = target.parentNode;
-        }
+        var playerBounds = bounds(this.playerElement);
+        var x = evt.clientX - playerBounds.left;
```
egreaves: This doesn't work once you're scrolled down on the page. Try `pageX` and `pageY`.

egreaves:
test this please

PR #2498, opened 2017-10-30
Do not disable selected 'Unknown CC' after midroll on iOS

hghazzi:
### This PR will...

Correctly label captions with no name in the manifest so that it is persisted after a midroll on iOS and not turn off captions.

On iOS, we reload the captions after a midroll. The logic around giving captions with no labels a unique `Unknown CC...` label has a bug that caused it to not reset the count correctly. So after a midroll, `Unknown CC` became `Unknown CC [2]` which was not being matched to the caption name stored in local storage. Since there was no match, captions would default to `Off` and be disabled. 

#### Addresses Issue(s):

JW8-752

### DIFF
--- a/src/js/providers/tracks-mixin.js
+++ b/src/js/providers/tracks-mixin.js
@@ -49,9 +49,9 @@ function setTextTracks(tracks) {
         this._initTextTracks();
     } else {
         // Remove the 608 captions track that was mutated by the browser
+        this._unknownCount = 0;
         this._textTracks = _.reject(this._textTracks, function(track) {
             const trackId = track._id;
-            this._unknownCount = 0;
             if (this.renderNatively && trackId && trackId.indexOf('nativecaptions') === 0) {
                 delete this._tracksById[trackId];
                 return true;
--- a/src/js/providers/tracks-mixin.js
+++ b/src/js/providers/tracks-mixin.js
@@ -51,12 +51,12 @@ function setTextTracks(tracks) {
         // Remove the 608 captions track that was mutated by the browser
         this._textTracks = _.reject(this._textTracks, function(track) {
             const trackId = track._id;
+            this._unknownCount = 0;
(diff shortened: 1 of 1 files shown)


hghazzi:
@robwalch done

src/js/providers/tracks-mixin.js
```diff
@@ -51,9 +51,12 @@ function setTextTracks(tracks) {
         // Remove the 608 captions track that was mutated by the browser
         this._textTracks = _.reject(this._textTracks, function(track) {
             const trackId = track._id;
+            this._unknownCount = 0;
             if (this.renderNatively && trackId && trackId.indexOf('nativecaptions') === 0) {
                 delete this._tracksById[trackId];
                 return true;
+            } else if (track.name.indexOf('Unknown') === 0) {
```
robwalch: Why the `else if`?

robwalch: @egreaves I think there's another problem with this unknown count thing if we assign the count when we label the tracks and then set names on the tracks in track-helper:

```js
    if (!label) {
        label = 'Unknown CC';
        unknownCount += 1;
        if (unknownCount > 1) {
            label += ' [' + unknownCount + ']';
        }
    }
```
You basically shouldn't be counting these at all, you should be parsing `unknownCount` from the name. (if not found, it's 1, otherwise it's this pattern`/^Unknown CC(?: \[(\d+)\])?$/`.

Something like:
```js
if (track.name.indexOf('Unknown') === 0) {
    this._unknownCount = `/^Unknown CC(?: \[(\d+)\])?$/.match(0) || 1;`
}
```

I would like the description to explain things from the context of the bug fix a little "because the captions meny will say X instead of Y when there are Z 608 tracks".

johnBartos:
> I would like the description to explain things from the context of the bug fix 

Please add more details to your description

PR #2499, opened 2017-10-31
Use padding instead of left/right for captions positioning

egreaves:
### This PR will...

- Use `padding-left` and `padding-right` in lieu of `left` and`right`  css values for setting the position of captions.
- Disable native rendering in Chrome so that publishers can immediately take advantage of better window styling of captions.
- Wait until we've added a text track to the menu before loading the renderer.
When rendering captions natively in Chrome, window styles applied to VTTCues span the beginning of the cue to the end of the video container. Publishers expect the behavior to be similar to Safari, where the captions window margin is the same on the left and right, regardless of where the text within it is positioned.
No.
No.
#### Addresses Issue(s):
JW8-475

### DIFF
--- a/src/js/polyfills/webvtt.js
+++ b/src/js/polyfills/webvtt.js
@@ -407,8 +407,8 @@ function CueStyleBox(window, cue) {
     var styles = {
         textShadow: '',
         position: 'relative',
-        left: 0,
-        right: 0,
+        'padding-left': 0,
+        'padding-right': 0,
         top: 0,
         bottom: 0,
         display: 'inline',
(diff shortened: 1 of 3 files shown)


src/js/polyfills/webvtt.js
```diff
@@ -407,8 +407,8 @@ function CueStyleBox(window, cue) {
     var styles = {
         textShadow: '',
         position: 'relative',
-        left: 0,
-        right: 0,
```
robwalch: Use camelCase for JavaScript style properties (`padddingRight`).

dannyfinks:
@egreaves @robwalch disregard the failing VTT test...we will need to update our captions validation for this change

egreaves:
@dannyfinks thanks!

PR #2500, opened 2017-10-31
Base skin unit tests on DOM modification rather than spying on css util methods

robwalch:
### This PR will...

Replace the use of sinon in our skin customization tests with code that looks for diffs in the player style sheet elements.

Testing that skin customization results in only our css utils running or not does not assert affect on the DOM. These changes do not affect coverage of our code. They make the tests better reflect the intent of the assertions, and remove the expectation that a certain utility interface is used to make the style changes (with the exception of style sheet element attributes added by simple-style-loader).

There is some cleanup on the skin module itself, and changes to the tests "beforeEach" and "after" handlers to remove styles added by simple-style-loader.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -3052,6 +3052,12 @@
       "integrity": "sha1-liVqO8l1WV6zbYLpkp0GDYk0Of8=",
       "dev": true
     },
+    "fast-diff": {
+      "version": "1.1.2",
+      "resolved": "https://registry.npmjs.org/fast-diff/-/fast-diff-1.1.2.tgz",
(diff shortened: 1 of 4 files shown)


hghazzi:
@robwalch The purpose of unit tests should be to test the logic in the module to be test and not the dependencies. In this case, we should be be mocking out `cssUtils.css` since if there is any change in `css.js` it shouldn't break tests in here. 

We should be able to trust that `css.js` is working as it is supposed to be, assuming it has good code coverage. By adding more unit tests in `css-test.js` would accomplish that, and `fast-diff` seems to be a great tool to do so. 

These are integration tests. I would keep `skin-test.js` as is and increase coverage for `css.js` (not neccessarily right now) and move these tests into its own folder, `test/integration`.

robwalch:
@hghazzi this is not the css.js test, it's the skin.js test, and the tests assert that styles have or have not been added to the DOM.

dannyfinks:
@robwalch the captions failures here appear to be from before we updated our captions tests in chrome to validate for player rendered captions.

PR #2501, opened 2017-11-01
Fix issues with getting state from adModel

pajong:
### This PR will...
Add check on whether if ```_instream._adModel``` before getting ```_instream._adModel.get('state')```

For DAI, ```_instream._adModel``` does not exist and causes console error when calling ```get('state')```

#### Addresses Issue(s):
ADS-646


egreaves:
@pajong can you add a fix version (milestone) to this PR?

PR #2504, opened 2017-11-02
Add playlistItem event listener in preview.js

DanFerrer:
### This PR will...

Change the preview image when a new playlist is loaded via the load api when the player state is idle.

The preview image would only change if a new playlist was loaded while the video was playing, paused, or buffering, but not when the player state is idle.

The setImage function in preview.js may be called multiple times now because of my event listener in the setup function. May need to put a conditional in my new function to make sure the setImage function doesn't fire if the preview image on the new item is the same is old one (by the time my function is hit, the image on the item and the model are the same)

#### Addresses Issue(s):

JW8-803

### DIFF
--- a/src/js/view/preview.js
+++ b/src/js/view/preview.js
@@ -12,8 +12,6 @@ function validState(state) {
 Object.assign(Preview.prototype, {
     setup: function(element) {
         this.el = element;
-
-        this.model.on('change:playlistItem', this.playlistItem, this);
     },
     setImage: function(img) {
         // Remove onload function from previous image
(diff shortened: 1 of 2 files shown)


src/js/view/preview.js
```diff
@@ -65,6 +67,13 @@ Object.assign(Preview.prototype, {
             });
         }
     },
+    playlistItem: function(model, item) {
+        if (!item) {
```
robwalch: Is it possible for playlistItem to be `undefined`?

PR #2506, opened 2017-11-02
Clear cuepoints when empty array is passed in

pajong:
### This PR will...
Allow ```setCues``` api to clear cue points if empty array ```[]``` is passed in.

There is a public api ```setCues``` to add cue points, but there is no way to remove the cue points. This will allow cue points to be removed. Clearing cue points is needed for DAI, as we get cue point changed event whenever an ad ends. Putting same cue points places them at random location. 

#### Addresses Issue(s):
ADS-643

### DIFF
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -426,12 +426,18 @@ export default class Controlbar {
     }
 
     addCues(model, cues) {
-        if (this.elements.time) {
+        if (!this.elements.time) {
+            return;
+        }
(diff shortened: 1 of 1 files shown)


pajong:
@egreaves Tried to create a new api, but it looks like it is going to be going through the same steps as ```setCues```. Thought it was too repetitive, so modified the existing ```setCues``` function to account for empty array. Let me know if you think this can cause issues.

robwalch:
@pajong what do you mean by "Putting same cue points places them at random location" ?

PR #2507, opened 2017-11-03
Replace assert with expect

hghazzi:
### This PR will...

- Convert all `assert` statements to `expect`
- Fixes all skipped tests
- Adding more structure to the unit test folder to be more consistent with `src` (`test/unit/controls/settings-menu-test.js`)
- Refactored `autostart-mute-test.js` to `volume-test.js` maintaining the save coverage
- Removes `intersection-observer-test.js` since it was failing and skipped, and depends on a global variable that can't be modified `views-manager.js#11`

remove all references to `assert` in unit tests, and using `expect` instead. This will make our unit tests be completely BDD oriented and a consistent.

chaijs.com/api/bdd/

#### Addresses Issue(s):

JW8-835


robwalch:
FAILED TESTS:
  **trycatch**
    ✖ should throw custom Error on catch
      IE 11.0.0 (Windows 7 0.0.0)
    expected undefined to equal ''
```
    Error
       at AssertionError (node_modules/chai/chai.js:9320:7)
       at Assertion.prototype.assert (node_modules/chai/chai.js:239:7)
       at assertEqual (node_modules/chai/chai.js:1383:7)
       at methodWrapper (node_modules/chai/chai.js:7709:5)
       at Anonymous function (test-context.js:19515:9)
```

test/unit/trycatch-test.js
```diff
@@ -1,50 +1,51 @@
+import ApiSettings from 'api/api-settings';
 import { tryCatch, JwError } from 'utils/trycatch';
 
 describe('trycatch', function() {
 
     it('defines', function() {
-        assert.equal(typeof tryCatch, 'function', 'trycatch function is defined');
-        assert.equal(typeof JwError, 'function', 'Error function is defined');
+        expect(typeof tryCatch, 'trycatch function is defined').to.equal('function');
+        expect(typeof JwError, 'Error function is defined').to.equal('function');
```
robwalch: `expect(x).to.be.a('function')`

test/unit/trycatch-test.js
```diff
@@ -1,50 +1,51 @@
...
-            });
-        };
+    it('should throw custom Error on catch', function() {
+        const jwError = tryCatch(function() {
+            throw new Error('Danger, Danger, Will Robinson!');
+        });
 
-        expect(trycatchThrow).to.throw(JwError);
+        expect(jwError.name).to.equal('');
+        expect(jwError.message).to.equal('Danger, Danger, Will Robinson!');
```
robwalch: This change is failing IE 11. `expect(function).to.throw(class);` is what we want. `name` is an Error property not present in IE11 and not something we need to test.

robwalch: Verifying the message prop is fine since that's something we use.

hghazzi: @robwalch using a flag to test is the exception is thrown since it is caught so `to.throw` would not work. Rather going to add a check that shows that it is not being thrown.

robwalch: I see. It throws in the next test when the `debug` is enabled, but returns a `JwError` when `debug` is not enabled. The `JwError` instance was not assigned a name because `function.name` is undefined in IE 11.

PR #2509, opened 2017-11-03
Make the hlsjsdefault flag be true by default

hghazzi:
### This PR will...

If `hlsjsdefault` is not explicitly set to false in the config, it will use the `hls.js` provider for Android devices.

To support the larger set that HLS provides compared to simply using the native HTML player on Android devices.

We will want to run the Android cucumber tests and do extensive manual testing to verify that `hls.js` is robust enough to now be used.

https://github.com/jwplayer/jwplayer-commercial/pull/4268

#### Addresses Issue(s):

JW8-264


hghazzi:
test this please

hghazzi:
test this please

hghazzi:
test this please

PR #2512, opened 2017-11-06
Modularize Provider Creation Flow

johnBartos:
### This PR will...
Split out provider loading, creation, changing, and synchronization into discrete steps which can be called outside of changing a provider.

The bulk of this work separates provider control into a promise chain. The chain roughly goes:

Load providers -> check providers -> change provider OR reload current provider

But we need to prime the next video tag (if necessary) synchronously, so that's found outside of it. The code right now looks a little bit messy but that's fine - we'll be migrating this logic into a new module soon. The `provider-controller` looks a bit slim so I may delete it, but it's purpose was to remove the other provider modules from the Model.

In the near future we'll need to be able to do some of these steps, but not all of them at once. Refactoring out the methods will give us the flexibility to background load a provider without changing the active one.

Everything

https://github.com/jwplayer/jwplayer-commercial/pull/4272

#### Addresses Issue(s):

JW8-42

### DIFF
--- a/src/js/providers/provider-controller.js
+++ b/src/js/providers/provider-controller.js
@@ -1,0 +1,53 @@
+import Providers from 'providers/providers';
+
+export default function ProviderController(initialConfig) {
+    let config = initialConfig;
+    let providers = new Providers(config);
+
+    return {
+        choose(source) {
+            return providers.choose(source).provider;
+        },
+        make(id, source) {
+            const Provider = this.choose(source);
+            if (!Provider) {
+                return null;
+            }
+            return new Provider(id, config);
+        },
+        loadProviders(playlist) {
+            return providers.load(providers.required(playlist));
+        },
+        allProviders() {
+            return providers;
+        },
+        updateConfig(updatedConfig) {
+            providers = new Providers(updatedConfig);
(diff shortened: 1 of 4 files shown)


johnBartos:
test this please

johnBartos:
The build failed due to a jenkins error but the cucumber job ran without error

egreaves:
test this please

src/js/providers/provider-controller.js
```diff
@@ -0,0 +1,53 @@
...
+
+const syncModelToProvider = (model, provider) => {
+    if (provider.getName().name.indexOf('flash') === -1) {
+        model.set('flashThrottle', undefined);
+        model.set('flashBlocked', false);
+    }
+    // Set playbackRate because provider support for playbackRate may have changed and not sent an update
+    model.set('playbackRate', provider.getPlaybackRate());
+    model.set('renderCaptionsNatively', provider.renderNatively);
+};
```
hghazzi: Let's add unit tests for this.

johnBartos:
Closed in favor of https://github.com/jwplayer/jwplayer/pull/2520

PR #2515, opened 2017-11-09
Set visualQuality on loadeddata event in html5 provider

DanFerrer:
### This PR will...

Trigger the visualQuality event in the html5 provider when the video is preloaded. 

To improve analytics data for play events, we should call getVisualQuality() as soon as possible. We can get data such as width, height, reason, etc when the player is preloaded.

visualQuality event triggers twice on videos with only one quality level: when the player is loaded and again when the you press play. Both times return the same info Need to add a check in checkVisualQuality() in html5.js to compare the last level with the new one to make sure we dont call event with the same info.

https://github.com/jwplayer/jwplayer-commercial/pull/4288
https://github.com/jwplayer/hls.js/pull/125

#### Addresses Issue(s):

JW8-823

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -102,6 +102,7 @@ function VideoProvider(_playerId, _playerConfig) {
             VideoEvents.loadeddata.call(_this);
             _setAudioTracks(_videotag.audioTracks);
             _checkDelayedSeek(_this.getDuration());
+            checkVisualQuality();
         },
 
         canplay() {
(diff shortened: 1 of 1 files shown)


dannyfinks:
click to play test page - http://player-develop-test-jenkins.longtailvideo.com/builds/lastSuccessfulBuild/archive/test/public/click-to-play/

DanFerrer:
Working fine on iOS

PR #2516, opened 2017-11-09
getMediaElement private API

karimJWP:
### This PR will...
expose a getter to obtain the videoTag
to unblock Ads Team @pajong @jw-ads-team with their background loading work
n/a
n/a
#### Addresses Issue(s):
JW8-857

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -968,7 +968,7 @@ Object.assign(Api.prototype, /** @lends Api.prototype */ {
 
     /**
      * Extracts the video tag in the foreground.
-     * @returns {object|null} videoTag - the HTML <video> element in the foreground.
+     * @returns {Element|null} videoTag - the HTML <video> element in the foreground.
      */
     getVideoTag() {
         const container = this.getContainer();
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -967,6 +967,14 @@ Object.assign(Api.prototype, /** @lends Api.prototype */ {
     },
 
     /**
+     * Extracts the video tag in the foreground.
+     * @returns {string} videoTag - the video tag in the foreground.
+     */
+    getVideoTag() {
+        return this.getContainer().getElementsByClassName('jw-media')[0].getElementsByTagName('video')[0];
+    },
+
+    /**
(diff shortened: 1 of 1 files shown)


src/js/api/api.js
```diff
@@ -967,6 +967,14 @@ Object.assign(Api.prototype, /** @lends Api.prototype */ {
     },
 
     /**
+     * Extracts the video tag in the foreground.
+     * @returns {string} videoTag - the video tag in the foreground.
+     */
+    getVideoTag() {
+        return this.getContainer().getElementsByClassName('jw-media')[0].getElementsByTagName('video')[0];
```
johnBartos: What happens if there's no container, or `jw-media` containers, or a `video` within the container?

egreaves: The following works: 
```
const container = this.getContainer();
return container && container.querySelector('video'); // returns the 1st video element or null if there's none
```

karimJWP: addressed

karimJWP:
test this please

src/js/api/api.js
```diff
@@ -967,6 +967,15 @@ Object.assign(Api.prototype, /** @lends Api.prototype */ {
     },
 
     /**
+     * Extracts the video tag in the foreground.
+     * @returns {string} videoTag - the video tag in the foreground.
```
johnBartos: It doesn't return a string

src/js/api/api.js
```diff
@@ -967,6 +967,15 @@ Object.assign(Api.prototype, /** @lends Api.prototype */ {
     },
 
     /**
+     * Extracts the video tag in the foreground.
+     * @returns {object|null} videoTag - the HTML <video> element in the foreground.
```
johnBartos: It's an `Element`, not object

https://developer.mozilla.org/en-US/docs/Web/API/Document/querySelector

src/js/api/api.js
```diff
@@ -967,6 +967,15 @@ Object.assign(Api.prototype, /** @lends Api.prototype */ {
     },
 
     /**
+     * Extracts the video tag in the foreground.
+     * @returns {Element|null} videoTag - the HTML <video> element in the foreground.
```
johnBartos: It can also return undefined

PR #2519, opened 2017-11-13
Add buffering case to setPreviewImage for autostart streams

DanFerrer:
### This PR will...

Show poster images on audio streams that auto start.

setPosterImage wasn't being called when autostart was set to true because the stream was never in the valid states to set the poster image (idle, complete)

#### Addresses Issue(s):

JW8-859


johnBartos:
Does this mean that the poster image shows whenever we're playing (and the stream happens to buffer)?

karimJWP:
@DanFerrer though this may work, the buffering state can occur repeatedly at any time (i.e. after a seek event); it's not very efficient to check for poster image at that time. i think it would be best to set it at the first play event or play attempt

DanFerrer:
@johnBartos if setPosterImage is called by itself outside of _stateUpdate then yeah but we only use it in the _stateUpdate function and it wont run unless autostart is true. I can modify Preview.setImage so it wont run unless the state is buffering and autostart is true

DanFerrer:
@karimJWP Ah i see that makes sense

PR #2520, opened 2017-11-13
Program Controller

johnBartos:
### This PR will...
Builds on top of https://github.com/jwplayer/jwplayer/pull/2512, but goes further. This PR takes the Provider loading/switching logic out of the Model and into the Program Controller. It manages the provider promise. Additionally, this PR introduces the Media controller, which is responsible for controlling the provider. It manages the play promise.

The bulk of this work separates provider control into a promise chain. The chain roughly goes:

Load providers -> check providers -> change provider OR reload current provider

But we need to prime the next video tag (if necessary) synchronously, so that's found outside of it. The code right now looks a little bit messy but that's fine - we'll be migrating this logic into a new module soon. The `provider-controller` looks a bit slim so I may delete it, but it's purpose was to remove the other provider modules from the Model.

The Program Controller will be doing the background loading work in the future. In order to background load, we need to be able to load a Provider independent of setting it as active. While the current iteration of this module does not do that, it makes it much more straightforward - the promise chains needed to load and have been separated. Playback was tangled up as a child of loading - in order to flatten it out (and thus making the promise chain extensible) I separated playback logic into the Media Controller.

Everything

https://github.com/jwplayer/jwplayer-commercial/pull/4299

#### Addresses Issue(s):

JW8-42

### DIFF
--- a/src/js/api/set-playlist.js
+++ b/src/js/api/set-playlist.js
@@ -16,12 +16,4 @@ const setPlaylist = function(model, playlist, feedData = {}) {
     }
 };
 
-export function loadProvidersForPlaylist(model) {
-    const playlist = model.get('playlist');
-    const providersManager = model.getProviders();
-    const providersNeeded = providersManager.required(playlist);
-
-    return providersManager.load(providersNeeded);
-}
-
 export default setPlaylist;
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -733,11 +733,11 @@ Object.assign(Controller.prototype, {
         this.setConfig = (newConfig) => setConfig(_this, newConfig);
         this.setItemIndex = _setItem;
 
-        // Program passthroughs
-        this.playVideo = _programController.playVideo.bind(_programController);
-        this.stopVideo = _programController.stopVideo.bind(_programController);
-        this.castVideo = _programController.castVideo.bind(_programController);
-        this.stopCast = _programController.stopCast.bind(_programController);
+        // Program Controller passthroughs
+        this.playVideo = (playReason) => _programController.playVideo(playReason);
+        this.stopVideo = () => _programController.stopVideo();
+        this.castVideo = (castProvider, item) => _programController.castVideo(castProvider, item);
+        this.stopCast = () => _programController.stopCast();
 
         // Model passthroughs
         this.setVolume = _model.setVolume.bind(_model);
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -529,8 +529,10 @@ Object.assign(Controller.prototype, {
             const length = playlist.length;
 
             // If looping past the end, or before the beginning
-            index = parseInt(index, 10) || 0;
-            index = (index + length) % length;
+            index = (parseInt(index, 10) || 0) % length;
+            if (index < 0) {
+                index += length;
+            }
 
             const item = playlist[index];
             return _programController.setActiveItem(item, index);
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -633,7 +633,6 @@ Object.assign(Controller.prototype, {
             return _programController.audioTracks;
         }
 
-        // TODO mediaController
         function _setCurrentCaptions(index) {
             index = parseInt(index, 10) || 0;
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -21,6 +21,7 @@ import { PLAYER_STATE, STATE_BUFFERING, STATE_IDLE, STATE_COMPLETE, STATE_PAUSED
     STATE_STALLED, MEDIA_BEFOREPLAY, PLAYLIST_LOADED, ERROR, PLAYLIST_COMPLETE, CAPTIONS_CHANGED, READY,
     MEDIA_ERROR, MEDIA_COMPLETE, CAST_SESSION, FULLSCREEN, PLAYLIST_ITEM, MEDIA_VOLUME, MEDIA_MUTE, PLAYBACK_RATE_CHANGED,
     CAPTIONS_LIST, CONTROLS, RESIZE } from 'events/events';
+import ProgramController from 'program/program-controller';
 
 // The model stores a different state than the provider
 function normalizeState(newstate) {
@@ -54,6 +55,7 @@ Object.assign(Controller.prototype, {
         _model.setup(config);
         _view = this._view = new View(_api, _model);
         _view.on('all', _triggerAfterReady, _this);
+        const _programController = new ProgramController(_model);
 
         _model.mediaController.on('all', _triggerAfterReady, _this);
         _model.mediaController.on(MEDIA_COMPLETE, () => {
@@ -147,10 +149,6 @@ Object.assign(Controller.prototype, {
         // Ensure captionsList event is raised after playlistItem
         _captions = new Captions(_model);
         _captions.on('all', _triggerAfterReady);
-
-        function _video() {
-            return _model.getVideo();
-        }
 
         function _triggerAfterReady(type, e) {
             _this.triggerAfterReady(type, e);
@@ -265,7 +263,7 @@ Object.assign(Controller.prototype, {
 
             // Only attempt to preload if this is the first player on the page or viewable
             if (instances[0] === _api || viewable === 1) {
-                model.preloadVideo();
+                _programController.preloadVideo();
             }
         }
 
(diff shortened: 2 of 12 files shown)


johnBartos:
test this please

src/js/controller/instream-html5.js
```diff
@@ -54,25 +56,22 @@ const InstreamHtml5 = function(_controller, _model) {
...
         // Make sure it chooses a provider
-        _adModel.stopVideo();
-        _adModel.setItemIndex(0).then(() => {
-            if (!_adModel) {
-                return;
-            }
-            _checkProvider(_adModel.getVideo());
-        });
+        _programController.stopVideo();
         _checkProvider();
```
karimJWP: @johnBartos is it still necessary to checkProvider here ? Shouldn't this be moved to program controller or its dependencies ?

johnBartos: Seems like a lot of bespoke ads logic that shouldn't be mixed with the program controller. Instream-html5 will change when doing the BGL ticket itself

src/js/program/program-controller.js
```diff
@@ -0,0 +1,287 @@
...
+            .then(() => {
+                ProviderConstructor = providerController.choose(source);
+                // The provider we need couldn't be loaded
+                if (!ProviderConstructor) {
+                    if (mediaController) {
+                        mediaController.destroy();
+                        model.resetProvider();
+                        this.mediaController = null;
+                    }
+                    const message = 'Could not load provider for playlist';
```
johnBartos: Use old error message

src/js/program/program-controller.js
```diff
@@ -0,0 +1,287 @@
...
+export default class ProgramController {
+    constructor(model) {
+        this.mediaController = null;
+        this.model = model;
+        this.providerController = ProviderController(model.getConfiguration());
+        this.providerPromise = resolved;
+    }
+
+    setActiveItem(item, index) {
+        const { mediaController, model } = this;
```
johnBartos: ensure state is synchronized before playlistItem fires

src/js/program/media-controller.js
```diff
@@ -0,0 +1,190 @@
...
+            this.provider.setSubtitlesTrack(index);
+        }
+    }
+}
+
+function loadAndPlay(item, provider, model) {
+    // Calling load() on Shaka may return a player setup promise
+    const providerSetupPromise = provider.load(item);
+    if (providerSetupPromise) {
+        const thenPlayPromise = model.thenPlayPromise = cancelable(() => {
```
johnBartos: Cancel any existing thenPlayPromise

src/js/program/program-controller.js
```diff
@@ -0,0 +1,287 @@
...
+            return;
+        }
+
+        const item = model.get('playlistItem');
+        if (!item || (item && item.preload === 'none')) {
+            return;
+        }
+
+        // Only attempt to preload if media hasn't been loaded and we haven't started
+        if (model.get('state') === 'idle' && model.get('autostart') === false && !mediaController.setup) {
```
johnBartos: Double check attach/detach exclusion

src/js/api/set-playlist.js
```diff
@@ -16,12 +16,4 @@ const setPlaylist = function(model, playlist, feedData = {}) {
     }
 };
 
-export function loadProvidersForPlaylist(model) {
```
robwalch: You should visit this comment (and perhaps the implementation) https://github.com/jwplayer/jwplayer/blob/v8.0.2/src/js/api/setup-steps.js#L42-L52 since you're moving the code it references.

src/js/controller/controller.js
```diff
@@ -746,6 +731,13 @@ Object.assign(Controller.prototype, {
         this.getState = _getState;
         this.next = _nextUp;
         this.setConfig = (newConfig) => setConfig(_this, newConfig);
+        this.setItemIndex = _setItem;
+
+        // Program passthroughs
+        this.playVideo = _programController.playVideo.bind(_programController);
+        this.stopVideo = _programController.stopVideo.bind(_programController);
+        this.castVideo = _programController.castVideo.bind(_programController);
+        this.stopCast = _programController.stopCast.bind(_programController);
```
robwalch: Prefer arrow functions over `.bind` since we've seen examples of `function.prototype.bind` tampering.

src/js/api/set-playlist.js
johnBartos: Addressed

src/js/providers/providers.js
```diff
@@ -40,11 +40,14 @@ Object.assign(Providers.prototype, {
             let loadProvider = false;
             for (let i = playlist.length; i--;) {
                 const item = playlist[i];
-                const supported = this.providerSupports(provider, item.sources[0]);
-                if (supported) {
-                    playlist.splice(i, 1);
+                const source = item.sources[0];
+                if (source) {
+                    const supported = this.providerSupports(provider, item.sources[0]);
```
robwalch: `item.sources[0]` is called twice. I don't see why this `source` check is necessary.

src/js/program/media-controller.js
```diff
@@ -0,0 +1,215 @@
...
+        if (mediaModelContext === model.mediaModel) {
+            syncPlayerWithMediaModel(mediaModelContext);
+        }
+    }).catch(error => {
+        model.set('playRejected', true);
+        const videoTagPaused = provider && provider.video && provider.video.paused;
+        if (videoTagPaused) {
+            mediaModelContext.set(PLAYER_STATE, STATE_PAUSED);
+        }
+        model.mediaController.trigger(MEDIA_PLAY_ATTEMPT_FAILED, {
```
robwalch: I think it would be a little easier if `playAttempt` was on mediaController, or if we just passed `this` (mediaController) into this method. Seeing `model.mediaController` here is a little confusing 🌀 .

src/js/controller/model.js
```diff
@@ -304,114 +232,8 @@ const Model = function() {
...
-        mediaElement.load();
-    }
-
     this.getProviders = function() {
-        return _providers;
-    };
-
-    this.resetProvider = function() {
-        _provider = null;
+        return providerController.allProviders();
```
robwalch: We should see if we can get rid of this is setup.

src/js/controller/controller.js
```diff
@@ -530,7 +523,17 @@ Object.assign(Controller.prototype, {
...
         function _setItem(index) {
-            return _model.setItemIndex(index);
+            _programController.stopVideo();
+
+            const playlist = _model.get('playlist');
+            const length = playlist.length;
+
+            // If looping past the end, or before the beginning
+            index = parseInt(index, 10) || 0;
+            index = (index + length) % length;
```
egreaves: `item = playlist[index]` would return an undefined value for negative values of index > length * 2. eg. index = -7 when length = 3.

Though this shouldn't happen, here's a safe way to handle  either extreme:
```
index %= length;
if (index < 0) {
    index += length;
}
```

src/js/program/media-controller.js
```diff
@@ -0,0 +1,215 @@
...
+            playReason = model.get('playReason');
+        }
+
+        model.set('playRejected', false);
+        let playPromise = resolved;
+        if (mediaModel.get('setup')) {
+            playPromise = provider.play();
+        } else {
+            playPromise = loadAndPlay(item, provider, model);
+            mediaModel.set('setup', true);
```
robwalch: `mediaModel.set('setup', true);` should be done before `loadAndPlay`

src/js/controller/controller.js
johnBartos: This was existing logic but I can update it

src/js/program/media-controller.js
johnBartos: Will address in the next PR

robwalch: Oh! I get it now. `model.mediaController` is the standing event dispatcher in the model for provider related events. It is not this program/media-controller.

src/js/program/media-controller.js
```diff
@@ -0,0 +1,216 @@
...
+            if (!mediaModel.get('started')) {
+                playAttempt(playPromise, model, playReason, provider);
+            }
+        }
+        return playPromise;
+    }
+
+    stop() {
+        const { provider } = this;
+        provider.stop();
```
egreaves: For consistency: `this.provider.stop();`

johnBartos: Addressed

PR #2524, opened 2017-11-14
Tech Talk Unit Tests

hghazzi:
During the Unit Testing Lunch & Learn, team members wrote some unit tests for different modules. 

JW8-835


test/unit/model/config-test.js
```diff
@@ -0,0 +1,28 @@
...
+    };
+
+    let model;
+
+    before(function () {
+        model = new Model();
+        model.setup(config);
+    });
+
+    describe('return value', function() {
```
hghazzi: @DanFerrer I would say that `describe` should be used for grouping related tests, usually the function being tested. Here, this could be `describe('Setup')`.

test/unit/model/config-test.js
```diff
@@ -0,0 +1,28 @@
...
+        it.only('should return an object on setup', function() {
+            expect(model).to.be.an('object');
```
hghazzi: Since you're instantiating `model`, it would be an object. A better test would be to confirm that the configs passed in were added to the model.

test/unit/model/config-test.js
```diff
@@ -0,0 +1,28 @@
...
+        });
+    });
+
+    describe('state', function() {
+        it.only('should return an idle state', function() {
+            expect(model.get('state')).to.equal('idle');
```
hghazzi: I'd put this test in the same describe block as the previous test. And reword the `it` to say something like that it is testing the default value of state. Another useful test would be to confirm that the default state is overwritten if passed in the setup config.

test/unit/model/config-test.js
```diff
@@ -0,0 +1,28 @@
...
+    });
+
+    describe('return value', function() {
+        it.only('should return an object on setup', function() {
+            expect(model).to.be.an('object');
+        });
+    });
+
+    describe('state', function() {
+        it.only('should return an idle state', function() {
```
hghazzi: Make sure to remove `.only` here and in the above test

hghazzi:
@DanFerrer Added some comments to your unit tests

@boyntoni @egreaves @karimJWP Can you push your tests to this branch?

robwalch:
@hghazzi This branch needs to be rebased. It is out of sync with jwplayer-commercial.

test/unit/model/config-test.js
```diff
@@ -0,0 +1,26 @@
+import Model from 'controller/model';
```
robwalch: This file should be called "model-test". We already have "api-config" and "api-get-gonfig" which test the config module, and the `getConfig()` method which returns a clone of the model attributes after `api.setup()`.

hghazzi:
test this please

PR #2528, opened 2017-11-16
Add pinning test for api.getConfig()

robwalch:
### This PR will...

Test all model properties presence and type returned by `playerApi.getConfig()` after an api instance is created using `jwplayer().setup({})`;

So that we can document/control any changes to API behavior in v8.1 after background loading changes.

### What else?

Created exports for transient model properties not set by default on player setup, and those set only in instream ad model.

See https://github.com/jwplayer/jwplayer/pull/2529 mock-model has not been kept in sync with config changes - it should at the very least contain all the properties we look for here, as well as some deeper comparison for objects like `localization`.

A commercial PR is required for the build to pass since commercial API instances will have additional model properties.

#### Addresses Issue(s):

JW8-819


robwalch:
test this please

robwalch:
test this please

egreaves:
test this please

karimJWP:
test this please

karimJWP:
test this please

PR #2531, opened 2017-11-17
Enable reusing getConfig test in Commercial

karimJWP:
### This PR will...
make getConfig test reusable
to reuse this test in commercial
This works, but I'm not sure if it's the best approach to making a test reusable
Needed for https://github.com/jwplayer/jwplayer-commercial/pull/4314
#### Addresses Issue(s):
JW8-819


karimJWP:
@robwalch that worked, but it breaks the tests that expect data to come from OS. Is there a way to specify a fallback path ?

PR #2532, opened 2017-11-17
Set poster image on audio only streams where autostart is true

DanFerrer:
### This PR will...

Prevent poster images from being loaded on video streams where autostart is true.

Poster images should only be loaded for audio only streams where autostart is true.

Had to set "buffering" as a valid state in the validState method in preview.js because thats the earliest the isAudioFile variable is set to true in the _onMediaTypeChange method and the setImage method in preview.js will exit if the state of the player isnt valid.

#### Addresses Issue(s):

JW8-875

### DIFF
--- a/src/js/view/preview.js
+++ b/src/js/view/preview.js
@@ -6,7 +6,7 @@ const Preview = function(_model) {
 };
 
 function validState(state) {
-    return state === 'complete' || state === 'idle' || state === 'error';
+    return state === 'complete' || state === 'idle' || state === 'error' || state === 'buffering';
 }
 
 Object.assign(Preview.prototype, {
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -679,6 +679,12 @@ function View(_api, _model) {
         // Put the preview element before the media element in order to display browser captions
         // otherwise keep it on top of the media element to display captions with the captions renderer
         _playerElement.insertBefore(_preview.el, element);
+
+        if (isAudioFile && _model.get('autostart')) {
+            setPosterImage(_model);
+        }
+
+
     }
 
     function _errorHandler(model, evt) {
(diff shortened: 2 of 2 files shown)


src/js/view/view.js
```diff
@@ -679,6 +679,12 @@ function View(_api, _model) {
         // Put the preview element before the media element in order to display browser captions
         // otherwise keep it on top of the media element to display captions with the captions renderer
         _playerElement.insertBefore(_preview.el, element);
+
+        if (isAudioFile && _model.get('autostart')) {
+            setPosterImage(_model);
+        }
+
+
```
egreaves: Remove extra new line

src/js/view/preview.js
```diff
@@ -6,7 +6,7 @@ const Preview = function(_model) {
 };
 
 function validState(state) {
-    return state === 'complete' || state === 'idle' || state === 'error';
+    return state === 'complete' || state === 'idle' || state === 'error' || state === 'buffering';
```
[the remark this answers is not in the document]
karimJWP: @egreaves i had forgotten to hit publish before you merged. do you disagree with my comments ?

src/js/view/view.js
```diff
@@ -679,6 +679,10 @@ function View(_api, _model) {
         // Put the preview element before the media element in order to display browser captions
         // otherwise keep it on top of the media element to display captions with the captions renderer
         _playerElement.insertBefore(_preview.el, element);
+
+        if (isAudioFile && _model.get('autostart')) {
+            setPosterImage(_model);
+        }
```
robwalch: We don't care about autostart in this case. It needs to load either way.

src/js/view/preview.js
robwalch: Adding buffering here means we will load a poster while loading a video, which is what we were trying to avoid.

PR #2534, opened 2017-11-20
Create unit tests for transient model properties

DanFerrer:
### This PR will...

Create unit tests for getting position, duration, and buffer on the model and mediaModel

To pin existing behavior of these properties on the models so they can be refactored and only be set and gotten from the mediaModel.

#### Addresses Issue(s):

JW8-818

### DIFF
--- a/test/unit/model/model-properties-test.js
+++ b/test/unit/model/model-properties-test.js
@@ -50,7 +50,7 @@ describe('Transient Model Properties', function () {
         });
     });
 
-    it('should set position and position on model and mediaModel', function() {
+    it('should set position and duration on model and mediaModel', function() {
         expect(model.get('position')).to.equal(0);
         expect(model.get('duration')).to.equal(0);
         expect(model.mediaModel.get('position')).to.equal(0);
--- a/test/unit/model/model-properties-test.js
+++ b/test/unit/model/model-properties-test.js
@@ -30,9 +30,9 @@ describe('Transient Model Properties', function () {
     });
 
     it('should have the same properties as the config', function() {
-        expect(model.get('volume')).to.eq(config.volume);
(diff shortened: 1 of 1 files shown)


test/unit/model/model-properties-test.js
```diff
@@ -0,0 +1,65 @@
...
+describe('Transient Model Properties', function () {
+    let model;
+
+    before(function () {
+        model = new Model().setup(config);
+        model.set('playlist', playlist);
+    });
+
+    it('should have the same properties as the config', function() {
+        expect(model.get('volume')).to.eq(config.volume);
```
egreaves: `.to.eq`?

test/unit/model/model-properties-test.js
```diff
@@ -0,0 +1,65 @@
...
+        expect(model.get('mute')).to.eq(config.mute);
+        expect(model.get('edition')).to.eq(config.edition);
+    });
+
+    it('should properly load playlist', function() {
+        return new Promise(function (resolve, reject) {
+            model.setItemIndex(0)
```
egreaves: This is no longer a model method. See: https://github.com/jwplayer/jwplayer/pull/2520/files#diff-a181c6e20596563a55e5710b8f4dbab7R736

test/unit/model/model-properties-test.js
```diff
@@ -0,0 +1,65 @@
...
+                    expect(model.get('state')).to.equal('idle');
+                    resolve();
+                })
+                .catch(function (err) {
+                    reject(err);
+                });
+        });
+    });
+
+    it('should set position and position on model and mediaModel', function() {
```
egreaves: should set position and duration

PR #2539, opened 2017-11-21
Enhance media-controller with behavior from the model

johnBartos:
### This PR will...
- Have the media-controller listen to and forward Provider events (migrated from the Model)
- Have the program-controller act as a single source for media-controller events
- Removes the media-controller from the model
- Remove media-controller references in View and QOE (and replaces them with events from the Model)
- Have the controller listen to the program-controller for Provider events

Background loading requires the program-controller to be able to control which events are propagated to the rest of the player. It now accomplishes this through the media-controller, which is solely responsible for receiving and propagating state updates. In this manner we'll be able to hot swap media-controller instances to control which provider influences the player.

### Anything Else?
I still have a bit of cleanup to do. Let's see what bugs automation uncovers.

https://github.com/jwplayer/jwplayer-commercial/pull/4332

#### Addresses Issue(s):
JW8-896


src/js/program/media-controller.js
```diff
@@ -25,6 +33,7 @@ export default class MediaController {
         mediaModelState.duration = duration;
         model.setProvider(provider);
         model.setMediaModel(mediaModel);
+        model.mediaController = this;
```
robwalch: Is this a good idea considering `model.mediaController` is used in other places and expects `mediaController` not to change? I think we should try to only set static properties on `model`. Anything else should be a model attribute.

src/js/program/provider-listener.js
```diff
@@ -0,0 +1,147 @@
...
+            case MEDIA_RATE_CHANGE: {
+                const rate = data.playbackRate;
+                // Check if its a generally usable rate.  Shaka changes rate to 0 when pause or buffering.
+                if (rate > 0) {
+                    model.set('playbackRate', rate);
+                }
+            }
+                return;
+            case MEDIA_TYPE:
+                if (mediaModel.get('mediaType') !== data.mediaType) {
```
egreaves: We can use `MEDIA_TYPE` or `type` in lieu of `'mediaType'` for better minification. There are other types above and below for which we could make this change.

src/js/program/provider-listener.js
```diff
@@ -0,0 +1,147 @@
+import { _isNaN, _isNumber } from 'utils/underscore';
+import { PLAYER_STATE, STATE_IDLE, MEDIA_VOLUME, MEDIA_MUTE,
+    MEDIA_TYPE, PROVIDER_CHANGED, AUDIO_TRACKS, AUDIO_TRACK_CHANGED,
+    MEDIA_RATE_CHANGE, MEDIA_BUFFER, MEDIA_TIME, MEDIA_LEVELS, MEDIA_LEVEL_CHANGED, MEDIA_ERROR,
+    MEDIA_BEFORECOMPLETE, MEDIA_COMPLETE, MEDIA_META, MEDIA_SEEK, MEDIA_SEEKED,
+    NATIVE_FULLSCREEN } from 'events/events';
+
+export default function ProviderEventListener(mediaController) {
```
robwalch: This is called `ProviderListener` (not `ProviderEventListener`) externally.

robwalch:
@johnBartos @dannyfinks there appears to be a fair number of [test failures](https://jenkins.jwplayer.com/view/JW7/job/jw7-commercial-cucumber/9047/). I'm going to hold off on merging until you get a chance to look at the captions and test failures.

> [:cucumber:](https://jenkins.jwplayer.com/view/JW7/job/jw7-commercial-cucumber/9047/) Automated Tests UNSTABLE

src/js/program/media-controller.js
```diff
@@ -79,30 +87,84 @@ export default class MediaController {
...
+                item,
+                playReason
+            });
+        });
+    }
+
+    _playbackComplete() {
+        const { provider } = this;
+        this.beforeComplete = false;
+        provider.setState(STATE_COMPLETE);
```
robwalch: Where's the attached check? That's what allows postrolls to play without complete firing until after the postroll is done.

PR #2540, opened 2017-11-22
View Model

robwalch:
### This PR will...

- Implement the view-model
  - Forwards all events from the player model, instream model and current media-model
    - Listeners for media-model changes only need to be added once thanks to event forwarding
  - Gets attributes from the current media-model if it has the attribute
    - Otherwise, from the instream model when in ads mode
    - Otherwise, from the player model
  - Sets attributes on the player model only (view sets "width", "height", and captions renderer updates "captions" styling)
- Removed all references to mediaModel and mediaController from view code
- Removed all references to instreamModel from view code
- Forward mediaController (provider) events through the model for now (will require follow up with https://github.com/jwplayer/jwplayer/pull/2539)
- Fix default provider fullscreen method stubs (`get` returns `false`, `set` is `noop`)
- Prevent `provider.remove` from throwing an error if it's container has not been set
- Update unit tests:
  - setting-menu no longer depends on mediaModel changes
  - CaptionsRenderer needs viewModel 
- Remove several instances where `model.getVideo()` is called to get access to the provider from the view. There are still four places where the view interacts directly with the provider. These are related to resizing the media within the player, setting and polling fullscreen in iOS via the media element, and making sure the media element is re-added to the player after instream mode.

https://github.com/jwplayer/jwplayer-commercial/pull/4333

#### Addresses Issue(s):

JW8-807


src/js/controller/controller.js
```diff
@@ -593,6 +601,7 @@ Object.assign(Controller.prototype, {
         }
 
         function _getConfig() {
+            // TODO: merge viewModel config or implement complete config in ViewModel?
```
egreaves: My vote is to merge the ViewModel and not conflate it with the complete config.

src/js/view/view-model.js
```diff
@@ -0,0 +1,115 @@
+import SimpleModel from 'model/simplemodel';
+
+const INITIAL_VIEW_STATE = {
+    viewSetup: false,
```
egreaves: Can we alphabetize these?

src/js/view/view.js
```diff
@@ -41,44 +40,19 @@ let ControlsModule;
 const _isMobile = OS.mobile;
 const _isIE = Browser.ie;
 
-function View(_api, _model) {
+function View(_api, _model, _captionsRenderer) {
```
egreaves: Can we change `_model` to `_viewModel` in the view for disambiguation?

robwalch: I avoided doing this because model in all files under view/ (except for view-model) refer to the viewModel, and I want the initial diff for this to be as small as possible.

src/js/view/view-model.js
robwalch: I'm going to remove these and just keep all state in the model. Storing these attributes in the view-model requires that the view code only write to the view-model properties and not be able to write to the model, which it currently does - "width", "height"...

src/js/controller/controller.js
robwalch: What I'm saying is should the controller do the merge, or should view-model, since view-model has everything.

egreaves:
@robwalch seems there's 1 legit failure. `visualQuality` doesn't fire when you manually change qualities. [Take a look](http://player-pr-test-jenkins.longtailvideo.com/builds/10853/archive/test/harness.html?edition=ads&config=hlsjs/hls-visual-quality). The other failure relates to viewability, which seems to be working fine.

robwalch:
@egreaves looks like the menu itself is not working - I had to tweak the unit tests for settings menu but fixing the tests does equate with the feature actually working. Looking into it...

dannyfinks:
@egreaves @robwalch the viewability issue looks like something I may have introduced with a change I made to be able to set javascript configs in our cucumber framework. I'm going to look into that. The visual quality failure is from Bartos's background loading work where the setCurrentLevel is broken so you can disregard that.

dannyfinks:
Resolved the issue with the viewability test

robwalch:
@dannyfinks no. The issue is that `jwplayer().setCurrentQuality(n)` is broken in development. The changes in the PR from using model to using the API to change qualities exposes the regression.

@johnBartos When `jwplayer().setCurrentQuality(n)` is called, we get to the program-controller calling `mediaController.currentQuality =`, but the setter is `quality` not `currentQuality`: https://github.com/jwplayer/jwplayer/commit/d426ebcf8bc1ea70f5fe1c2d33c23296d8ce196c#diff-dd041e064487f60f90cb3ce13aeb440dR291

I'll add a fix here.

dannyfinks:
@robwalch thats what I meant. Their is a sub-bug created for @johnBartos to fix the issue.

robwalch:
I fixed it in this PR.

src/js/view/view.js
```diff
@@ -627,8 +583,8 @@ function View(_api, _model) {
             return !!(fsElement && fsElement.id === _model.get('id'));
         }
         // if player element view fullscreen not available, return video fullscreen state
-        return _instreamModel ? _instreamModel.getVideo().getFullScreen() :
-            _model.getVideo().getFullScreen();
+        const provider = _model.getVideo();
```
johnBartos: Can we do `model.get('provider')` here? Would like to remove the getVideo api

robwalch: That will return `{ name: 'provider' }`. The provider's fullscreen state, should update the model. Then we can remove this.

There is also another place in the view where we set fullscreen on the provider (iOS) if DOM fullscreen is not available. I would like to do that without accessing the provider in the view. Maybe a view-model event that doesn't get forwarded to the API (view events do get forwarded by API).

egreaves: @johnBartos `model.get('provider')` currently returns an object with the provider name (eg. `{name: "html5"}`), which is what gets returned when the public api calls `jwplayer().getProvider()`. This would result in a breaking change to the public api.

PR #2541, opened 2017-11-22
Optimised images.

grischard:
### This PR will...

For every file, [ImageOptim](https://github.com/ImageOptim/ImageOptim) compared the different tools and picked the smallest file that was produced by svgo or by a combination of `advpng`, `pngout` and `zopfli`.

No related issues, no code change, no functionality change, no visual differences in images (compression is lossless).

Make jwplayer faster for free

No

No

#### Addresses Issue(s):

None.

JW8-###


grischard:
I'd also like to run this on the built-in themes, but can't find a github repo for them. Any ideas?

robwalch:
Thanks @grischard. The png assets optimized here are not used in the player. On the SVG's I only see a one char change from `<path opacity="0.2"` to `<path opacity=".2"`.

Maybe @radium-v can comment on why these path's have opacity set on them at all. It seems like we'd be better off with opacity only set on CSS styles, removing it from SVG rendering but it could be a component of a part of the graphic or button state.

grischard:
Thanks! There's also a semicolon that can be dropped. Agreed, that's not a huge change, and I understand if it can't be accepted.

Omitting the leading 0 is allowed by the SVG spec, and is a pretty common optimisation. It is discussed in for example [this bug in OpenOffice](https://bz.apache.org/ooo/show_bug.cgi?id=125447) or [this edge case in Firefox](https://github.com/emilbjorklund/svg-weirdness/issues/29).

Would you know where I could submit a similar PR for the built-in themes? It looks like they're not on github.

robwalch:
@grischard by built-in themes do you mean CSS skins? We've removed them from JW8 in favor of simpler UI customization with the player config: https://developer.jwplayer.com/jw-player/docs/developer-guide/customization/configuration-reference/#skin

To make a PR against JW7 use the v7.12.x branch https://github.com/jwplayer/jwplayer/tree/v7.12.x

JW7 uses webfonts which are pretty well optimized, and skins are loaded externally so we're not accepting changes to them since you can make a copy, optimize it, and then configure the player on your page to load your optimized copy.

PR #2542, opened 2017-11-22
Set the poster image when necessary

egreaves:
### This PR will...
- Always set the poster image for audio streams
- Only set the poster image for videos when we're in the idle, error or complete state longer than it takes for RAF. This typically happens when playback starts (idle), after playback ends for the last item in a playlist (complete), or when we enter the error state. 
In states other than `playing`, we delay calling `_stateUpdate` until the next animation frame. This resulted in the poster image not being set for all audio items in a playlist when a player is set to autostart or between playlist items where the player is typically not in the `idle` state long enough. 
No.
No.
#### Addresses Issue(s):
JW8-875

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -739,7 +739,6 @@ function View(_api, _model) {
         }
         replaceClass(_playerElement, /jw-state-\S+/, 'jw-state-' + state);
 
-        _model.off('change:playlistItem', setPosterImage);
         switch (state) {
             case STATE_IDLE:
             case STATE_ERROR:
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -739,7 +744,13 @@ function View(_api, _model) {
             case STATE_IDLE:
             case STATE_ERROR:
             case STATE_COMPLETE:
-                _model.change('playlistItem', setPosterImage);                
+                // Set the poster image for videos before playback starts (idle), when the playlist ends (complete),
+                // or when an error is encountered. We don't get to the idle state between playlist items because of RAF
+
+                if (_model.mediaModel.get('mediaType') === 'video') {
+                    _model.change('playlistItem', setPosterImage);
```
robwalch: Why don't we just make this `setPosterImage(_model);`? The "change" listener will result in the poster being set on on the next playlist item regardless of state.

PR #2554, opened 2017-11-28
Remove behavior that seeks on "seeked" event while scrubbing

robwalch:
### This PR will...

Remove a listener that caused seek to be called on "seeked" events. It also moves logic that handles the buffering state in the html5 provider to run before the "seek" event is fired.

The seeked listener was used early in JW7 to ensure seeking to the final dragging point on the time slider. It is no longer needed, probably hasn't worked until we recently forwarded provider events through the model, and now causes many redundant calls to seek while scrubbing.

Moving the logic in html5 is a best practice. State can change during event propagation so you want to handle as much logic as possible before events are dispatched - dispatching events last in a function unless there is an important reason to do otherwise.

#### Addresses Issue(s):

JW8-914

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -128,8 +128,8 @@ function VideoProvider(_playerId, _playerConfig) {
             _videotag.removeEventListener('waiting', setLoadingState);
             _videotag.addEventListener('waiting', setLoadingState);
             _this.trigger(MEDIA_SEEK, {
-                position: position,
-                offset: offset
+                position,
+                offset
             });
         },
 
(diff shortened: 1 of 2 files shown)


src/js/providers/html5.js
```diff
@@ -119,17 +119,18 @@ function VideoProvider(_playerId, _playerConfig) {
...
+            const position = _positionBeforeSeek;
+            _setPositionBeforeSeek(offset);
             _seekOffset = null;
             _delayedSeek = 0;
             _this.seeking = true;
+            _videotag.removeEventListener('waiting', setLoadingState);
+            _videotag.addEventListener('waiting', setLoadingState);
             _this.trigger(MEDIA_SEEK, {
-                position: _positionBeforeSeek,
+                position: position,
```
johnBartos: Can use object shorthand here

PR #2556, opened 2017-11-28
Improve muted autoplay mobile controls

DanFerrer:
### This PR will...

Display controbar immediately when the player autostarts and is muted. Tapping the player will display the control bar while the mute icon is shown. Also changed bottom property of jw-autostart-mute class so the control bar is fully visible while the user is tapping on the player and controlbar shows.

Controls should be visible immediately on the player without interacting with it

#### Addresses Issue(s):

JW8-290

### DIFF
--- a/src/css/controls/flags/user-inactive.less
+++ b/src/css/controls/flags/user-inactive.less
@@ -39,3 +39,7 @@
         }
     }
 }
+
+.jwplayer.jw-flag-user-inactive .jw-autostart-mute {
+    display: flex;
+}
--- a/src/css/controls/flags/user-inactive.less
+++ b/src/css/controls/flags/user-inactive.less
@@ -40,6 +40,6 @@
     }
 }
 
-.jwplayer.jw-state-playing.jw-flag-user-inactive .jw-autostart-mute {
+.jw-state-playing.jw-flag-user-inactive .jw-autostart-mute {
     display: flex;
 }
--- a/src/css/controls/flags/user-inactive.less
+++ b/src/css/controls/flags/user-inactive.less
@@ -40,6 +40,6 @@
     }
 }
 
-.jwplayer.jw-flag-user-inactive .jw-autostart-mute {
+.jwplayer.jw-state-playing.jw-flag-user-inactive .jw-autostart-mute {
     display: flex;
 }
--- a/src/css/controls/imports/autostartmute.less
+++ b/src/css/controls/imports/autostartmute.less
@@ -3,7 +3,7 @@
 .jw-autostart-mute {
     position: absolute;
     bottom: 0;
-    right: 0;
+    right: 14px;
     height: @mobile-touch-target;
     width: @mobile-touch-target;
     background-color: rgba(33, 33, 33, 0.4);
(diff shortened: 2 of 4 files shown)


src/js/view/controls/controls.js
```diff
@@ -384,6 +384,7 @@ export default class Controls {
         this.controlbar.renderVolume(mute, model.get('volume'));
         this.mute.hide();
         utils.removeClass(this.playerContainer, 'jw-flag-autostart');
+        utils.removeClass(this.playerContainer, 'jw-flag-user-inactive');
```
robwalch: This class should only be changed by the `userActive` and `userInactive` methods. It probably makes more sense to call `userActive()` here so that the class is removed **and** the inactive timer is restarted.

src/css/controls/imports/autostartmute.less
```diff
@@ -2,7 +2,7 @@
 
 .jw-autostart-mute {
     position: absolute;
-    bottom: 0.5em;
+    bottom: 4em;
```
robwalch: What's this for (please cover in the PR description)?

src/js/view/controls/controls.js
```diff
@@ -408,6 +409,11 @@ export default class Controls {
                 timeout || ACTIVE_TIMEOUT);
         }
         if (!this.showing) {
+            const mute = this.mute;
+
+            if (mute && mute.element().style.display === '') {
+                utils.removeClass(this.playerContainer, 'jw-flag-autostart');
+            }
```
robwalch: Implementing logic based on DOM state is not a good practice - if we change how mute is displayed this breaks.

What's the thinking here?

There is existing code ("autostartMuted" logic) that I would expect to have removed or changed as part of this improvement, instead this PR just adds more code on top.

robwalch: @egreaves can you provide an overview of the code you added for the initial implementation (what these props are and if they are needed given the scope of the changes)?

I would like us to remove these if they are no longer needed:

- model "autostartMuted"
- CSS .jw-flag-autostart
- button .jw-autostart-mute

src/css/controls/imports/autostartmute.less
DanFerrer: @robwalch Added it. I adjusted the position so the mute icon so the control bar is fully visible while the user is tapping the player while the mute button is displayed (it covers up the fullscreen icon)

src/css/controls/flags/user-inactive.less
```diff
@@ -39,3 +39,7 @@
         }
     }
 }
+
+.jwplayer.jw-state-playing.jw-flag-user-inactive .jw-autostart-mute {
```
egreaves: We may not need `.jwplayer` if the other 3 classes provide enough specificity.

src/css/controls/imports/autostartmute.less
```diff
@@ -2,12 +2,13 @@
 
 .jw-autostart-mute {
     position: absolute;
-    bottom: 0.5em;
-    right: 0.5em;
+    bottom: 0;
+    right: 0;
```
robwalch: I don't really like that the button is flushed left button. I think you should apply the same padding that buttons in the controlbar have for consistency.

PR #2562, opened 2017-11-29
Update cues after instream mode

robwalch:
### This PR will...

- Remove default `instream` model property so that it's on set in the player model and never set in the instream model
- Remove `_` from chapters.mixin.js
- Move cues logic from controlbar to timeslider
- Remove a redundant call to `resetChapters`
- Remove controls `syncPlaybackTime` since the view-model "diff" functionality handles changes between instream and player model properties
- Fix view-model instream/player "diff" functionality so that it runs properly when exiting instream

Fixes the display of instream cues after a preroll.

#### Addresses Issue(s):

JW8-915

### DIFF
--- a/src/js/view/controls/components/chapters.mixin.js
+++ b/src/js/view/controls/components/chapters.mixin.js
@@ -1,4 +1,3 @@
-import _ from 'utils/underscore';
 import utils from 'utils/helpers';
 import srt from 'parsers/captions/srt';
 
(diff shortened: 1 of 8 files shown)


src/js/view/controls/components/chapters.mixin.js
```diff
@@ -53,7 +52,7 @@ const ChaptersMixin = {
             return;
         }
 
-        _.each(this.cues, (cue) => {
+        this.cues.forEach((cue) => {
```
johnBartos: Will `this.cues` always be defined? I know `_.each` had guard checks against this.

robwalch: It is set `[]` in TimeSlider constructor.

PR #2569, opened 2017-12-01
Add player only view-model

robwalch:
Adds a player-only sub view-model used by components that should only get state from the player (never from instream).

This is used primarily by settings-menu and next-up components. It prevents a bug where next-up is displayed after prerolls because of duration changes between instream and player state.

Also adds unit tests for view-model.

#### Addresses Issue(s):

JW8-807


robwalch:
Going to change the approach to making two instances more OOP.

PR #2570, opened 2017-12-04
Add hlsjs flag for Safari

hghazzi:
### This PR will...

Check if the `safarihlsjs` flag is set in the model and add it as a flag in the source. This will allow `ProvidersSupported` to verify if it should use `hlsjs`

Since `ProvidersSupported` only takes in the `source` and not `model`, the flag needs to be passed through.

https://github.com/jwplayer/jwplayer-commercial/pull/4370

#### Addresses Issue(s):

JW8-856


hghazzi:
test this please

hghazzi:
test this please

PR #2571, opened 2017-12-04
Rename or remove transient and duplicate model properties

robwalch:
### This PR will...

- Remove `position`, `duration` and `buffer` from the player model, and only write these in the media-model.
- Rename media-model `state` to `mediaState` so that it does not collide with `state` in the player (or view) model.

This places transient state that is reset between playlist items and ad breaks in the media-model. We can worry less about state from previous items affecting the current player state since each item gets it's own media-model.

Test updates https://github.com/jwplayer/jwplayer-commercial/pull/4376

#### Addresses Issue(s):

JW8-806

### DIFF
--- a/test/unit/model-qoe-test.js
+++ b/test/unit/model-qoe-test.js
@@ -16,29 +16,29 @@ describe('Model QoE', function() {
         initQoe(model, program);
         const qoeItem = model._qoeItem;
 
-        mediaModel.set('state', STATE_IDLE);
+        mediaModel.set('mediaState', STATE_IDLE);
         program.trigger(MEDIA_PLAY_ATTEMPT);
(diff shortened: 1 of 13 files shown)


test/unit/model-qoe-test.js
```diff
@@ -16,29 +16,29 @@ describe('Model QoE', function() {
         initQoe(model, program);
         const qoeItem = model._qoeItem;
 
-        mediaModel.set('state', STATE_IDLE);
+        mediaModel.set('mediaState', STATE_IDLE);
```
johnBartos: Can you add a constant for `mediaState`?

robwalch: Sure, but the constants we have now are really only for events, not for model state properties.

PR #2575, opened 2017-12-05
Added styling for fullscreen tooltip

boyntoni:
### This PR will...

Contain the Fullscreen tooltip within the bounds of the player, allowing for longer, localized text to be used. 

Localized fullscreen tooltips were exceeding player bounds.

N/A

no

#### Addresses Issue(s):

JW8-878

### DIFF
--- a/src/css/controls/imports/tooltip.less
+++ b/src/css/controls/imports/tooltip.less
@@ -109,6 +109,21 @@
             transition-delay: @tooltip-popup-delay, 0s, @tooltip-popup-delay;
             visibility: visible;
         }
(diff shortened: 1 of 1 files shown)


src/css/controls/imports/tooltip.less
```diff
@@ -134,7 +149,7 @@
 .jw-tooltip {
     bottom: @controlbar-height;
     display: none;
-    position: absolute;
+    position: relative;
```
egreaves: Is this change needed for all tooltips?

PR #2582, opened 2017-12-06
Show the selected captions track when playback starts and after a midroll

egreaves:
### This PR will...
- Set the captions track when the initial index > 0
- Make `setCaptionsList` private since it's only used in `captions.js`.

The model no longer controls syncing up the provider with the default captions track. This is now the responsibility of the controller. As such, the controller should `setCurrentCaptions` when the provider notifies the program controller that subtitles exist.
No
No
#### Addresses Issue(s):
JW8-934


egreaves:
test this please

PR #2583, opened 2017-12-06
Support querying and setup of multiple elements without ids

### LINKED ISSUE #2382: unable to initialize multiple players on elements without IDs
### Expected Behavior
multiple players initialized

### Actual Behavior
only one player is initialized

### Steps to reproduce
no id code: https://codepen.io/trinonsense/pen/KXzVNZ
with id code: https://codepen.io/trinonsense/pen/JrXXOE

### Environment
- *hardware | device*: n/a
- *operating system | version*: n/a
- *browser | version*: n/a
- *JW Player version | url*: v7/8

robwalch:
### This PR will...

Support querying and setup of multiple elements without ids.

The data attribute "data-jwplayer-id" is assigned to elements queried with jwplayer(domElememt) which do not have an id. The attribute is removed when jwplayer().remove() is called, and it is not present on the player element after setup, because the player always assigns an id to elements it creates.

Our API supports querying of a single element without an id. This extends to support multiple elements:

```js
const element1 = document.createElement('div');
const element2 = document.createElement('div');

document.body.appendChild(element1);
document.body.appendChild(element2);

jwplayer(element1).setup({ file: 'video.mp4' });
jwplayer(element2).setup({ file: 'video.mp4' });

jwplayer(element1).play();
jwplayer(element2).seek(30);
```

Here's an example that generates three elements to setup players with provided by @trinonsense 
 in #2382 https://codepen.io/trinonsense/pen/KXzVNZ Currently only one player is setup.

#### Addresses Issue(s):

JW8-801
Fixes #2382

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -40,6 +40,7 @@ function coreFactory(api, element) {
 function resetPlayer(api, core) {
     api.off();
     core.playerDestroy();
+    core.getContainer().removeAttribute('data-jwplayer-id');
 }
 
 /**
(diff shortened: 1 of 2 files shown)


egreaves:
test this please

trinonsense:
This is great! Thanks so much

PR #2585, opened 2017-12-06
Check if device is touchScreen

karimJWP:
### This PR will...
Add a section in the environment module indicating if a device supports touchScreen 
our mobile check excludes microsoft touch devices and therefore we need a better indicator to determine if swiping will be supported in related
this check does not work for FireFox on microsoft touch screen devices.
blocker for [related PR](https://github.com/jwplayer/jwplayer-plugin-related/pull/186)
#### Addresses Issue(s):
JW8-942

### DIFF
--- a/src/js/environment/environment.js
+++ b/src/js/environment/environment.js
@@ -10,6 +10,7 @@ import {
     isAndroidNative,
     isIOS,
     isMobile,
+    isTouchScreen,
     isOSX,
     isIPad,
     isIPod,
@@ -65,6 +66,12 @@ export const OS = {};
  * @property {boolean} iframe - Is the session in an iframe?
  */
 export const Features = {};
+
+/**
+ * @typedef {object} HardwareCapabilities
+ * @property {boolean} touchScreen - Is the browser being rendered on a touch screen device?
+ */
+export const Hardware = {};
 
 const isWindows = () => {
     return userAgent.indexOf('Windows') > -1;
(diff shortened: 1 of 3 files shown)


egreaves:
test this please

src/js/utils/browser.js
```diff
@@ -70,6 +70,11 @@ export function isMobile() {
     return isIOS() || isAndroid();
 }
 
+export function isTouchScreen() {
+    // Note: does not work for FireFox on Surface.
+    return isMobile() || 'ontouchstart' in window || navigator.msMaxTouchPoints || navigator.maxTouchPoints;
```
robwalch: This check should match the check in related, or what we do in UI.js

src/js/utils/browser.js
```diff
@@ -70,6 +70,11 @@ export function isMobile() {
     return isIOS() || isAndroid();
 }
 
+export function isTouchScreen() {
```
robwalch: Is this a check for touch-only devices? Or devices with multi pointer support (mouse, touch, pointer)? (`hasTouch` vs `isTouch` vs `feature.touch` `feature.touch`)

Why `isMobile`? What do we want for Windows?

Remove the word 'Screen'.

src/js/environment/environment.js
```diff
@@ -158,3 +165,10 @@ Object.defineProperties(Features, {
         enumerable: true
     }
 });
+
+Object.defineProperties(Hardware, {
```
robwalch: Please put `touch` in `Features`. I think we should also add `mouse`, so touch-only is `Features.touch && !Features.mouse`.

robwalch: Even if we do that, it doesn't map to the checks we use in most of our UI https://github.com/jwplayer/jwplayer/blob/ed44063687a787c7e6daf5b8fe19fd240909fb85/src/js/utils/ui.js#L6

PR #2588, opened 2017-12-07
Background load media for prerolls, midrolls

johnBartos:
### This PR will...
- Background load preroll and midroll ads
-- Background loading acts as an alternative to attach/detch for browsers which support it (everything but iOS)
- Restore backgrounded content when resuming from ad playback
- Synchronize volume and mute across all media elements (background, foreground, and pool)
- Provide `instream` with an enhanced `getMediaElement` method so that it doesn't kill background loaded content
- Maintain a primed pool of media elements to ensure we can continue to autoplay after a gesture
- Replace`instream-html5` with the `ad-program-controller`
- Add BGL unit tests
- Document and cleanup a ton of code

So that we more quickly resume playback from an ad. With background loading, we never throw away content when switching to ad playback - when resuming, we swap the background loaded media element into the foreground. This reduces the TTFF between ad and content by two orders of magnitude.

Everything

https://github.com/jwplayer/jwplayer-commercial/pull/4393
https://github.com/jwplayer/hls.js/pull/135

#### Addresses Issue(s):

JW8-31


src/js/api/core-shim.js
```diff
@@ -85,9 +85,9 @@ Object.assign(CoreShim.prototype, {
         model.setProvider = function() {};
 
         // Create/get click-to-play media element, and call .load() to unblock user-gesture to play requirement
-        const mediaElement =
-            model.attributes.mediaElement = getMediaElement(this.originalContainer);
-        mediaElement.load();
+
+        const mediaPool = MediaElementPool();
+        mediaPool.prime();
```
robwalch: This isn't a singleton. Every time `MediaElementPool()` is called, a new set of three tags is returned.

PR #2591, opened 2017-12-08
Prime the media element used by instream on iOS

robwalch:
### This PR will...

Prime model "mediaElement" when not background loading at the same time we prime elements in the pool, so that it is primed when we play a preroll.

Also removes "sharedVideoTag" are from instream (no longer in use), and cleans up `pool.prime()` implementation.

This allows us to remove code from ads plugins that attempts to prime video elements for playback https://github.com/jwplayer/jwplayer-ads-vast/pull/294

#### Addresses Issue(s):

JW8-770


robwalch:
test this please

egreaves:
test this please

PR #2593, opened 2017-12-08
Always pass playlist to update playlist

robwalch:
### This PR will...

Always pass a normalized playlist to `updatePlaylist`.

This simplifies logic in the extension of `updatePlaylist` added in https://github.com/jwplayer/jwplayer-commercial/pull/4382. The playlist argument will always be an array and any items and their sources will be normalized. 

https://github.com/jwplayer/jwplayer-commercial/pull/4382

#### Addresses Issue(s):

JW8-757


hghazzi:
Ahh, a `Playlist` object, not a playlist array. Got it.

PR #2594, opened 2017-12-08
Fix alignment of chapter markers

boyntoni:
### This PR will...

Fix the alignment of chapter markers

Chapter cues were significantly off their position

N/A

No

#### Addresses Issue(s):

JW8-800

### DIFF
--- a/src/js/view/controls/components/chapters.mixin.js
+++ b/src/js/view/controls/components/chapters.mixin.js
@@ -2,9 +2,10 @@ import utils from 'utils/helpers';
 import srt from 'parsers/captions/srt';
 
 class Cue {
-    constructor (time, text) {
+    constructor (time, text, playerWidth) {
         this.time = time;
         this.text = text;
+        this.playerWidth = playerWidth;
         this.el = document.createElement('div');
         this.el.className = 'jw-cue jw-reset';
     }
@@ -14,7 +15,8 @@ class Cue {
         if (this.time.toString().slice(-1) === '%') {
             this.pct = this.time;
         } else {
-            const percentage = Math.floor((this.time / duration) * 100);
+            let markerWidth = 6;
+            const percentage = (100 * this.time / duration) - (100 * markerWidth / this.playerWidth);
             this.pct = percentage + '%';
         }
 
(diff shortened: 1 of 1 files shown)


src/js/view/controls/components/chapters.mixin.js
```diff
@@ -14,7 +14,7 @@ class Cue {
         if (this.time.toString().slice(-1) === '%') {
             this.pct = this.time;
         } else {
-            const percentage = (this.time / duration) * 100;
+            const percentage = Math.floor((this.time / duration) * 100);
```
robwalch: This just floors the % to a whole value. It doesn't center the markers based on the precise position they should be placed at (which would be more precise than a whole percentage).

boyntoni: I originally thought the same, and will certainly take another look, but in manual testing this was yielding the expected results.

robwalch: 1. Multiply then divide (no need for parens and order of operations ensures the highest precision) `100 * this.time / duration`.

2. You want to subtract the percentage of player width that half a cue marker takes up from the percentage already calculated. That will move the marker to the left by half it's width, which is the real fix.

`(100 * this.time / duration) - (50 * markerWidth / playerWidth)`

In your manual testing you should make sure you're not just dealing with positions and durations that seem to work when floored. If they always do please share screenshots that show how the alignment of markers works when floored.

src/js/view/controls/components/chapters.mixin.js
```diff
@@ -2,9 +2,10 @@ import utils from 'utils/helpers';
 import srt from 'parsers/captions/srt';
 
 class Cue {
-    constructor (time, text) {
+    constructor (time, text, playerWidth) {
         this.time = time;
         this.text = text;
+        this.playerWidth = playerWidth;
```
robwalch: @boyntoni you should pass in the model so that in align you can get the latest "containerWidth". That will make it more responsive. Note that the slider width is not the full player width so you might want to do a little more math in there too to match the CSS like the hardcoded `markerWidth`.

PR #2599, opened 2017-12-11
Extract fitToBounds method used in html5 to be used across providers for IE/Edge stretching

boyntoni:
### This PR will...

Allow IE / Edge users to use 'stretching' configuration option.

Stretching on IE/Edge had not been working with HLS and DASH.

Ensure the logic for stepping in to fitToBounds() in video-actions-mixins is not overly restrictive, or not restrictive enough.

#### Addresses Issue(s):

JW8-772


src/js/providers/video-actions-mixin.js
```diff
@@ -1,4 +1,6 @@
 import { style } from 'utils/css';
+import { Browser, OS } from 'environment/environment';
+import fitToBounds from '../utils/fit-to-bounds';
```
egreaves: `from 'utils...`

src/js/providers/video-actions-mixin.js
```diff
@@ -18,23 +20,28 @@ const VideoActionsMixin = {
     },
 
     resize: function(width, height, stretching) {
-        if (!width || !height || !this.video.videoWidth || !this.video.videoHeight) {
+        const fitVideoUsingTransforms = Browser.ie || (OS.iOS && OS.version.major < 9) || Browser.androidNative;
```
egreaves: Since you're using this in 2 places, create a module for it in your new util.

src/js/providers/video-actions-mixin.js
```diff
@@ -18,23 +20,28 @@ const VideoActionsMixin = {
     },
 
     resize: function(width, height, stretching) {
-        if (!width || !height || !this.video.videoWidth || !this.video.videoHeight) {
+        const fitVideoUsingTransforms = Browser.ie || (OS.iOS && OS.version.major < 9) || Browser.androidNative;        
+        if ((!width || !height || !this.video.videoWidth || !this.video.videoHeight) && !fitVideoUsingTransforms) {
```
egreaves: Create a local const in the resize method for `this.video`.

src/js/providers/video-actions-mixin.js
```diff
@@ -18,23 +20,28 @@ const VideoActionsMixin = {
...
+                styles.objectFit = 'fill';
+                stretching = 'exactfit';
             }
-            style(this.video, {
-                objectFit,
-                width: null,
-                height: null
-            });
         }
+        if (fitVideoUsingTransforms && !this.video.videoWidth < width && !this.video.videoHeight < height) {
```
egreaves: These conditions are not logically correct. You're converting `videoWidth` and `videoHeight` to a boolean, then comparing with a number. What is the objective here?

boyntoni: The bang operator was a mistake; but this essentially is designed to achieve two things:

(1) is the user in an environment that requires the fitToBounds method, as opposed to one that can utilize object-fit
(2) Ensure that the video should be fit to the bounds of the player -- in an HLS/Dash stream, if we up switch quality, then the height/width of the tag can exceed the player bounds. These two additional checks ensure that the video remains within the player bounds.

src/js/utils/fit-to-bounds.js
```diff
@@ -0,0 +1,29 @@
+import { transform } from 'utils/css';
+import { Browser, OS } from 'utils/browser'; 
+
+export function fitVideoUsingTransforms() {
```
egreaves: `export const fitVideoUsingTransforms = Browser.ie || (OS.iOS && OS.version.major < 9) || Browser.androidNative;`

src/js/providers/video-actions-mixin.js
```diff
@@ -1,4 +1,7 @@
 import { style } from 'utils/css';
+import fitToBounds from 'utils/fit-to-bounds';
```
egreaves: This is fine and works because it's the default exported module.

src/js/providers/video-actions-mixin.js
```diff
@@ -1,4 +1,7 @@
...
+import fitVideoUsingTransforms from 'utils/fit-to-bounds';
```
egreaves: This does not do what you think it does. You need curly braces in order to import a module that's not the default exported module, so: 

`import { fitVideoUsingTransforms } from 'utils/fit-to-bounds';`

src/js/utils/fit-to-bounds.js
```diff
@@ -0,0 +1,29 @@
...
+    return Browser.ie || (OS.iOS && OS.version.major < 9) || Browser.androidNative;
+}
+
+export default function fitToBounds(_videotag, width, height, stretching, styles) {
```
egreaves: I'd suggest renaming this file to better describe the modules it exports (eg. `video-fit.js`). Also, remove `default` and do explicit imports where needed.

PR #2601, opened 2017-12-12
Use Passive Event Listeners to Improve Scrolling Performance

robwalch:
### This PR will...

Mark all of our listeners, except those that implement drag behavior on volume and time sliders, as "passive".

Making this change should improve page scroll performance on Android Chrome, and will prevent tens to hundreds of console warnings (depending on number of players or embeds) from being seen by website developers using our player. These warning also factor into page performance audit scores rendered by Chrome 63+ and sites like these:
https://www.webpagetest.org/
https://developers.google.com/speed/pagespeed/insights/
https://gtmetrix.com/

### Misc

There may be other "touchstart" and "touchmove" listeners in the player or plugins that should be marked as passive. This only covers those that produced warnings immediately upon setup and playback.

There is still a warning for "touchstart" and "touchmove" on the timeslider. That's to be expected, but solutions to suppressing these without breaking the sliders are welcome!

#### Addresses Issue(s):

JW8-978


egreaves:
@karimJWP @hghazzi - we should also make this change in related for the swipe event handlers and anywhere else you can think of that doesn't use `ui.js`.

PR #2615, opened 2017-12-19
Don't load preview images for autostarted videos

egreaves:
### This PR will...
Use the `_stateHandler` for loading the preview image when necessary.
This [change](https://github.com/jwplayer/jwplayer/pull/2552/files#diff-adecb551656b12ea38bd16aad3077585R350) - which first gets called when the player state is IDLE, even for autostarted players -  loads the poster image immediately. Before, the player would be in a buffering or playing state for an autostarted video because of the raf check in `_stateHandler`, which calls `_stateUpdate`. 
No - the other changes in https://github.com/jwplayer/jwplayer/pull/2552 are still good.
No.
#### Addresses Issue(s):
JW8-1005

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -353,7 +353,7 @@ function View(_api, _model) {
 
     function redraw(model, visibility, lastVisibility) {
         if (visibility && !lastVisibility) {
-            _stateHandler(model);
+            _stateHandler(model, model.get('state'));
             _this.updateStyles();
         }
     }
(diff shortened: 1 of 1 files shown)


egreaves:
@robwalch yeah, it does break posters when stop() is called. I'm gonna go back to using `_stateHandler` and confirm that it still holds up with the other changes from https://github.com/jwplayer/jwplayer/pull/2552 in place.

PR #2616, opened 2017-12-20
Add new adBreakStart and adBreakEnd events

pajong:
### This PR will...
Create new events called adBreakStart and adBreakEnd, which are fired whenever instream adapter's init and destroy functions are called. Rename the old 'adBreakEnd' used for vast to 'vastAdBreakEnd', so that we can use 'adBreakEnd' for the new event.

To track when ad break starts and end.

https://github.com/jwplayer/jwplayer-ads-vast/pull/304

#### Addresses Issue(s):

ADS-792

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -24,6 +24,8 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
     let _backgroundLoadTriggered = false;
     let _oldpos;
     let _backgroundLoadPosition;
+    const _advertisingBlock = _controller.getConfig().advertising || {};
+    const _adclient = utils.getAdClient(_advertisingBlock.client);
 
     const _clickHandler = (evt) => {
         evt = evt || {};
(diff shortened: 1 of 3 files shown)


src/js/controller/instream-adapter.js
```diff
@@ -72,6 +74,8 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         // Make sure the original player's provider stops broadcasting events (pseudo-lock...)
         _controller.detachMedia();
 
+        this.trigger('adBreakStart', _getAdBreakData('adBreakStart'));
```
robwalch: The add plugins are calling `instream.init`. They should also trigger 'adBreakStart' at the time most appropriate to their implementation.

src/js/controller/instream-adapter.js
```diff
@@ -146,7 +168,7 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
             _loadNextItem();
         } else {
             // notify vast of breakEnd
-            this.trigger('adBreakEnd', {});
+            this.trigger('vastAdBreakEnd', {});
```
robwalch: Why not always fire `PLAYLIST_COMPLETE` (if we skipped an ad and there are no more ads to play then we're done)? You could also change the state to COMPLETE and listen for that.

src/js/controller/instream-adapter.js
pajong: @egreaves Please comment on this. It was a conclusion that we made after discussion.

PR #2617, opened 2017-12-20
Fire "itemReady" after provider.init()

robwalch:
### This PR will...

Trigger "itemReady" at the point it was originally intended to be fired - when the provider is ready.

If you look at all the places where we listen for "itemReady", it is expected that there is a provider with a media element ready for playback. The most common use is for captions track logic that requires a media element and provider after "renderCaptionsNatively" has been sent, or we're casting.

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -13,9 +13,6 @@ const Captions = function(_model) {
     // Listen for captions menu index changes from the view
     _model.on('change:captionsIndex', _captionsIndexHandler, this);
 
-    // Listen for item ready to determine which provider is in use
-    _model.on('itemReady', _itemReadyHandler, this);
-
     function _setSubtitlesTracks(tracks) {
         if (!tracks.length) {
             return;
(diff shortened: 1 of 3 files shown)


src/js/controller/captions.js
```diff
@@ -42,36 +39,41 @@ const Captions = function(_model) {
...
+                                    _addVTTCuesToTrack(track, vttCues);
+                                },
+                                (error) => {
+                                    this.trigger(ERROR, {
+                                        message: 'Captions failed to load',
+                                        reason: error
+                                    });
+                                });
+                        }
+                        _updateMenu();
```
egreaves: Is there a reason behind updating the menu inside the for loop? Calling `_updateMenu()` after should set the menu up correctly, unless the desired behavior is an update for each track that gets added.

robwalch: This should be on the next like, outside of the loop for tracks added natively after the provider is inited (what other reason is there to wait for “itemReady”?)

src/js/controller/captions.js
```diff
@@ -42,36 +39,41 @@ const Captions = function(_model) {
...
+                                },
+                                (error) => {
+                                    this.trigger(ERROR, {
+                                        message: 'Captions failed to load',
+                                        reason: error
+                                    });
+                                });
+                        }
+                    }
+                    _updateMenu();
```
robwalch: @egreaves I moved `_updateMenu` it outside the for loop. Any reason this native rendering `_addTrack` code can't be added to the provider's tracks mixin? It would be great to get rid of as many "itemReady" listeners as possible.

egreaves: Yeah, that should work since the menu would be setup [here](https://github.com/jwplayer/jwplayer/blob/f88b2dd2564e54b189a6ab310959c8b16980fd23/src/js/controller/controller.js#L687) after they get added by the mixin. Only risk I can think of is duplication, but I'm pretty sure there's logic to safeguard against that.

PR #2619, opened 2017-12-26
A/B Test Recs Localization

hghazzi:
### This PR will...

- Take the selected variant in the `playerConfig.ab` block and either display the discovery icon inside or above the controlbar
- Adds the new `More Videos` svg icon.

- For the first step of the new Recs Display UI, we can now A/B test the position of the discovery icon, making it clearer and more visible to users when the controlbar is showing.

https://github.com/jwplayer/ab-test/pull/19

#### Addresses Issue(s):

JW8-1014

### DIFF
--- a/src/assets/SVG/more-videos.svg
+++ b/src/assets/SVG/more-videos.svg
@@ -1,0 +1,1 @@
+<svg class="jw-svg-icon jw-svg-icon-more-videos" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 240 240"><path d="M25,109.8h48.8V51.2H25V109.8z M25,191.2h48.8v-58.6H25V191.2z M96.6,191.2h48.8v-58.6H96.6V191.2z M168.2,191.2H217v-58.6 h-48.8V191.2z M96.6,109.8h48.8V51.2H96.6V109.8z M168.2,51.2v58.6H217V51.2H168.2z"/></svg>
--- a/src/js/view/controls/components/custom-button.js
+++ b/src/js/view/controls/components/custom-button.js
@@ -15,7 +15,9 @@ export default class CustomButton {
         }
 
         let iconElement;
-        if (img && img.substring(0, 4) === '<svg') {
+        if (img instanceof SVGSVGElement) {
+            iconElement = img;
+        } else if (img && img.substring(0, 4) === '<svg') {
             iconElement = svgParse(img);
         } else {
             iconElement = document.createElement('div');
--- a/src/js/view/controls/components/custom-button.js
+++ b/src/js/view/controls/components/custom-button.js
@@ -15,7 +15,7 @@ export default class CustomButton {
         }
 
         let iconElement;
-        if (img instanceof SVGSVGElement) {
+        if (img instanceof SVGElement) {
             iconElement = img;
         } else if (img && img.substring(0, 4) === '<svg') {
             iconElement = svgParse(img);
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -39,9 +39,8 @@ function div(classes) {
 }
 
 function isRecsUpDisplay(abConfig) {
-    const tests = abConfig ? abConfig.tests : null;
-
-    if (tests && tests.discoverIcon && tests.discoverIcon) {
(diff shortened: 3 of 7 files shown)


src/assets/SVG/more-videos.svg
```diff
@@ -0,0 +1 @@
+<svg class="jw-svg-icon jw-svg-icon-more-videos" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 240 240"><path d="M25,109.8h48.8V51.2H25V109.8z M25,191.2h48.8v-58.6H25V191.2z M96.6,191.2h48.8v-58.6H96.6V191.2z M168.2,191.2H217v-58.6 h-48.8V191.2z M96.6,109.8h48.8V51.2H96.6V109.8z M168.2,51.2v58.6H217V51.2H168.2z"/></svg>
```
hghazzi: @radium-v Are there any optimizations we can do here?

src/js/view/controls/components/custom-button.js
```diff
@@ -15,7 +15,9 @@ export default class CustomButton {
         }
 
         let iconElement;
-        if (img && img.substring(0, 4) === '<svg') {
+        if (img instanceof SVGSVGElement) {
```
egreaves: Is `SVGSVGElement` a thing? or should it be `SVGElement`?

src/js/view/controls/controlbar.js
```diff
@@ -37,6 +38,21 @@ function div(classes) {
     return element;
 }
 
+function isRecsUpDisplay(abConfig) {
+    const tests = abConfig ? abConfig.tests : null;
+
+    if (tests && tests.discoverIcon && tests.discoverIcon) {
```
egreaves: You're checking  if `tests.discoverIcon` is truthy twice.

src/js/view/controls/controlbar.js
```diff
@@ -37,6 +38,21 @@ function div(classes) {
     return element;
 }
 
+function isRecsUpDisplay(abConfig) {
+    const tests = abConfig ? abConfig.tests : null;
```
egreaves: Returning early would simplify the logic of checking if `abConfig.tests.discoverIcon` is truthy.

src/js/view/controls/components/custom-button.js
hghazzi: It is thing which also returns true, but it should be SVGElement

robwalch:
Aren’t the recs buttons in the controlbar added by the related plugin? Why not include this new code in related so you can branch the logic there? The related overlay container is always present and a good place to put the alternative button.

hghazzi:
@robwalch The recs button is added to the control bar via the `addButton` API call, so the DOM manipulation for the buttons is happening here, so makes sense to have the logic here as well.

robwalch:
@hgazzi I’m thinking that addButton is called from related. The Svg is in that project and the test is for UI defined in that project.

One of our goals should be to minimize changes to the player and to multiple projects for a single goal. You have full access to the related plugin div in the plugin and that’s a better place to do DOM manipulation for a test, and for a feature that would not be needed if related was never loaded.

hghazzi:
I see where you're coming from, but I think for the time being we should keep it here. We want to get a simple A/B test out for the new localization and position that we can release in the next week. I think moving it to related right now would require more work since we would be keeping the overlay always visible and swapping the contents on click, and the implications of those changes. 

We are still refining and planning this epic, and I will be working on the design/arch of the recs up display.

robwalch:
OK. Yeah, I'm trying to prevent us from adding tech-debt to the player. If the player change was to extend `addButton` to add buttons outside of the controlbar, then we'd be getting a feature, and the test code could still go in related. Instead, you're adding a lot of one off code in the core player that will need to be cleaned up or will live as tech-debt affecting the timeline of ending this test and iterating on related and the controlbar.

esboyd:
any reason why " Adds the new More Videos svg icon." is an svg instead of text that can be localized? Now to localize you have to create an svg instead of using the existing functionality of related.heading?

hghazzi:
@esboyd The new svg is just the icon. The text is being appended to the icon, so it is still being localized.

PR #2622, opened 2017-12-28
 A/B Test Recs UI

hghazzi:
### This PR will...

- Support colour customization for the new 'More Videos' button in related

To keep the styling consistent across the A/B tests

https://github.com/jwplayer/ab-test/pull/19
https://github.com/jwplayer/jwplayer-plugin-related/pull/192

#### Addresses Issue(s):

JW8-1014

### DIFF
--- a/src/js/view/utils/skin.js
+++ b/src/js/view/utils/skin.js
@@ -150,7 +150,6 @@ export function handleColorOverrides(playerId, skin) {
                 '.jw-related-more-btn:focus .jw-text',
             ], 'color', config.iconsActive);
 
-
             addStyle([
                 '.jw-svg-icon-buffer',
             ], 'fill', config.icons);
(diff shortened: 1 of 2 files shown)


src/js/view/utils/skin.js
```diff
@@ -144,7 +145,9 @@ export function handleColorOverrides(playerId, skin) {
                 '.jw-button-color.jw-toggle:not(.jw-icon-cast)',
                 '.jw-button-color:hover:not(.jw-icon-cast)',
                 '.jw-button-color:focus:not(.jw-icon-cast)',
-                '.jw-button-color.jw-toggle.jw-off:hover:not(.jw-icon-cast)'
+                '.jw-button-color.jw-toggle.jw-off:hover:not(.jw-icon-cast)',
+                '.jw-related-more-btn:hover .jw-text',
+                '.jw-related-more-btn:focus .jw-text',
```
robwalch: So far we've tried to be very generic with the button classes so that these rules can be reused. Can you think or the new requirement added by the "More Videos" button (like applying color to a .jw-text and :focus state) and then rewrite this rule so that we can use it for all buttons? If done right you might only need to replace "jw-related-more-btn" with "jw-button-color" in these changes and add "jw-button-color" to the "More Videos" button.

hghazzi: This was to avoid styling the text in the tooltips. I'll add a new class `jw-button-text` that can be used to apply the color to anything text that is within `jw-button-color`. Otherwise, I could apply it to just direct descendants.

PR #2627, opened 2017-12-29
Apply focus event listener to simple tooltip 

boyntoni:
### This PR will...

Apply same styles to Simple Tooltip when focused as when it is moused over. 

Recent changes to styles were not applied when user tabbed over Fullscreen icon.

N/A

No. 

#### Addresses Issue(s):

JW8-878

### DIFF
--- a/src/js/view/controls/components/simple-tooltip.js
+++ b/src/js/view/controls/components/simple-tooltip.js
@@ -40,6 +40,7 @@ export function SimpleTooltip(attachToElement, name, text, openCallback) {
 
     attachToElement.addEventListener('mouseover', instance.open);
     attachToElement.addEventListener('focus', instance.open);
+    attachToElement.addEventListener('blur', instance.close);
     attachToElement.addEventListener('mouseout', instance.close);
     attachToElement.addEventListener('touchstart', (evt) => {
         instance.touchEvent = getPointerType(evt) === 'touch';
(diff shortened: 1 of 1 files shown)


egreaves:
@boyntoni can you rebase this PR?

karimJWP:
@boyntoni why was this only happening on the fullscreen icon though ? it seems like your solution changes the behavior on all icons to fix something wrong with one specific icon.

src/js/view/controls/components/simple-tooltip.js
```diff
@@ -39,6 +39,7 @@ export function SimpleTooltip(attachToElement, name, text, openCallback) {
     };
 
     attachToElement.addEventListener('mouseover', instance.open);
+    attachToElement.addEventListener('focus', instance.open);
     attachToElement.addEventListener('mouseout', instance.close);
     attachToElement.addEventListener('touchstart', (evt) => {
         instance.touchEvent = getPointerType(evt) === 'touch';
```
robwalch: On an unrelated note, this use of `getPointerType` is completely unnecessary. This statement will always return `true`, because this is a 'touchstart' (TouchEvent) listener and `getPointerType` will always return `true` for TouchEvents. @hghazzi added this code.

egreaves:
test this please

PR #2635, opened 2018-01-02
Casting: Persist media position on connect & disconnect

hghazzi:
### This PR will...

Before connecting and disconnecting from the chromecast device during playback, the current position needs to be persisted. Now that the position is maintained in the media model, we get and set the start time based on that value.

It will also move the logic within the chromecast controller to now be in the program controller.

So when we connect or disconnect, the media continues to play and does not start from the beginning

https://github.com/jwplayer/jwplayer-commercial/pull/4463

#### Addresses Issue(s):

JW8-986

### DIFF
--- a/src/js/program/program-controller.js
+++ b/src/js/program/program-controller.js
@@ -215,17 +215,13 @@ export default class ProgramController extends Eventable {
      * @returns {undefined}
      */
     stopCast() {
-        const { mediaController, model } = this;
+        const { model } = this;
         const index = model.get('item');
         const item = model.get('playlist')[index];
 
         item.starttime = model.mediaModel.get('position');
-        model.resetItem(item);
 
-        if (mediaController) {
-            mediaController.stop();
-            this.mediaController = null;
-        }
+        this.stopVideo();
 
         this.setActiveItem(index);
     }
--- a/src/js/program/program-controller.js
+++ b/src/js/program/program-controller.js
@@ -198,12 +198,16 @@ export default class ProgramController extends Eventable {
     castVideo(castProvider, item) {
         const { model } = this;
 
+        const playlistItem = Object.assign({}, item);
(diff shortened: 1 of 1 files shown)


src/js/program/program-controller.js
```diff
@@ -198,21 +198,36 @@ export default class ProgramController extends Eventable {
...
-        this.stopVideo();
-        this.mediaController = null;
+        const { mediaController, model } = this;
+        const index = model.get('item');
+        const item = model.get('playlist')[index];
+
+        item.starttime = model.mediaModel.get('position');
+        model.resetItem(item);
+
+        if (mediaController) {
```
johnBartos: `stopVideo` already checks if there's a mediaController. Why not just call that?

src/js/program/program-controller.js
```diff
@@ -198,21 +198,36 @@ export default class ProgramController extends Eventable {
...
      * @returns {undefined}
      */
     stopCast() {
-        this.stopVideo();
-        this.mediaController = null;
+        const { mediaController, model } = this;
+        const index = model.get('item');
+        const item = model.get('playlist')[index];
+
+        item.starttime = model.mediaModel.get('position');
```
johnBartos: you can consolidate this into one variable

src/js/program/program-controller.js
```diff
@@ -198,21 +198,36 @@ export default class ProgramController extends Eventable {
...
+        model.resetItem(item);
+
+        if (mediaController) {
+            mediaController.stop();
+            this.mediaController = null;
+        }
+
+        this.setActiveItem(index);
```
robwalch: > // The Player is expected to restore video playback afterwards
`setActiveItem` returns a promise which you could use to resume playback. See it's other uses in instream-adapter and the controller.

PR #2638, opened 2018-01-02
Added UA Checks for Windows Phone

waxidiotic:
### This PR will...

Add user-agent checks for Windows Phone.

This will enable customers to workaround the player not working properly in Edge for Windows Phone.

If you want. Probably a good idea.

No.

#### Addresses Issue(s):

JW8-932

### DIFF
--- a/src/js/environment/environment.js
+++ b/src/js/environment/environment.js
@@ -8,6 +8,7 @@ import {
     isSafari,
     isAndroid,
     isAndroidNative,
+    isWindowsMobile,
     isIOS,
     isMobile,
     isOSX,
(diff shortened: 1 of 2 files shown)


waxidiotic:
@robwalch Are you saying don't change anything and tell the customer to check for `OS.windows && OS.mobile`? I want to make sure.

robwalch:
@waxidiotic no. I’m saying we should only extend isMobile to include the “Windows Phone” UA check, without adding more util methods or OS.windowsPhone since we have no use or plans to officially support this environment.

What bugs are being reported specifically on that platform that we can’t support with feature detection?

PR #2645, opened 2018-01-04
Sync model volume and mute on initial setup 

johnBartos:
We want to set `mute` and `volume` across all media elements on initial setup; previously, only the active element had these properties set.

No

No

#### Addresses Issue(s):

JW8-1045 JW8-1046

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -63,6 +63,7 @@ Object.assign(Controller.prototype, {
         _view.on('all', _triggerAfterReady, _this);
 
         const _programController = new ProgramController(_model, mediaPool);
+        syncInitialModelState();
         addProgramControllerListeners();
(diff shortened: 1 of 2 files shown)


robwalch:
Check cuke failures and merge

robwalch:
@johnBartos this actually broke your provider integration unit tests because they expect volume and mute to be updated by simply calling `programController.setActiveItem`.
```
✖ Sets a Provider when no Provider is currently set
      Firefox 57.0.0 (Mac OS X 10.10.0)
      Chrome 63.0.3239 (Mac OS X 10.10.5)
    AssertionError: videoVolume: expected 1 to equal 0.2
```

I'll have a fix for the test as I'm updating that test in commercial and adding program-controller unit tests to this project as well. Ideally only the tests that integrate hls functionality would be in commercial.

PR #2648, opened 2018-01-04
BGL Play after load fixes

karimJWP:
### This PR will...
- convert string `skipOffsets` to number when updating `_backgroundLoadPosition`
- call `model.setActiveItem(index)` right before returning the provider promise
- remove offset percentage conversion logic of [commercial](https://github.com/jwplayer/jwplayer-commercial/pull/4481) and implement it in `utils/strings`
- check if position > duration -5 when ad time is updated since at that point we should no the duration.
- enables preloading during  a preRoll triggered after a playlist load (only for vast)
- prevents playing the wrong item when `play` is called immediately after `load`
- if a playlist item is loaded with a preRoll on IMA, the playlist item will never preLoad because we never get the skipoffset from IMA; IMA never calls the instream adapter's `loadItem`.
https://github.com/jwplayer/jwplayer-commercial/pull/4481
#### Addresses Issue(s):
JW8-930

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -50,7 +50,8 @@ const Defaults = {
         unmute: 'Unmute'
     },
     renderCaptionsNatively: false,
-    nextUpDisplay: true
+    nextUpDisplay: true,
+    itemReady: false
 };
 
 function _deserialize(options) {
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -50,8 +50,7 @@ const Defaults = {
         unmute: 'Unmute'
     },
     renderCaptionsNatively: false,
-    nextUpDisplay: true,
-    itemReady: false
+    nextUpDisplay: true
 };
 
 function _deserialize(options) {
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -26,7 +26,7 @@ const Captions = function(_model) {
     }, this);
 
     // Update tracks once we know "renderCaptionsNatively" based on provider
-    _model.on('itemReady', () => {
+    _model.on('change:itemReady', () => {
         // Sideload tracks when not rendering natively
         const item = _model.get('playlistItem');
         const tracks = item.tracks;
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -202,7 +202,7 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         if (skipoffset) {
             // Start background loading once the skip button is clickable
             _this.setupSkipButton(skipoffset, _options);
-            _backgroundLoadPosition = skipoffset;
+            _backgroundLoadPosition = typeof skipoffset === 'string' ? utils.seconds(skipoffset) : skipoffset;
         } else {
             // If no skipoffset is set, default to background loading 5 seconds before the end
             _backgroundLoadPosition = item.duration - BACKGROUND_LOAD_OFFSET;
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -4,7 +4,7 @@ import { STATE_BUFFERING, STATE_COMPLETE, STATE_PAUSED,
     INSTREAM_CLICK, AD_SKIPPED } from 'events/events';
 import { BACKGROUND_LOAD_OFFSET, BACKGROUND_LOAD_MIN_OFFSET } from '../program/program-constants';
 import Promise from 'polyfills/promise';
-import { offsetSeconds } from 'utils/strings';
+import { offsetToSeconds } from 'utils/strings';
 import Events from 'utils/backbone.events';
 import _ from 'utils/underscore';
 import AdProgramController from 'program/ad-program-controller';
@@ -139,7 +139,7 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         // Start background loading once the skip button is clickable
         // If no skipoffset is set, default to background loading 5 seconds before the end
         if (!_backgroundLoadPosition) {
-            _backgroundLoadPosition = offsetSeconds(_skipOffset, duration) || duration - BACKGROUND_LOAD_OFFSET;
+            _backgroundLoadPosition = offsetToSeconds(_skipOffset, duration) || duration - BACKGROUND_LOAD_OFFSET;
         }
         if (!_backgroundLoadTriggered && position >= _backgroundLoadPosition) {
             _controller.preloadNextItem();
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -26,7 +26,7 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
     let _backgroundLoadTriggered = false;
     let _oldpos;
     let _skipOffset;
-    let _backgroundLoadPosition;
+    let _backgroundLoadStart;
     let _destroyed = false;
     let _inited = false;
 
@@ -138,10 +138,11 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
 
         // Start background loading once the skip button is clickable
         // If no skipoffset is set, default to background loading 5 seconds before the end
-        if (!_backgroundLoadPosition) {
-            _backgroundLoadPosition = offsetToSeconds(_skipOffset, duration) || duration - BACKGROUND_LOAD_OFFSET;
+        if (!_backgroundLoadStart) {
+            // Ensure background loading doesn't degrade ad performance by starting too early
+            _backgroundLoadStart = offsetToSeconds(_skipOffset, duration) || duration - BACKGROUND_LOAD_OFFSET;
         }
-        if (!_backgroundLoadTriggered && position >= _backgroundLoadPosition) {
+        if (!_backgroundLoadTriggered && position >= Math.max(_backgroundLoadStart, BACKGROUND_LOAD_MIN_OFFSET)) {
             _controller.preloadNextItem();
             _backgroundLoadTriggered = true;
         }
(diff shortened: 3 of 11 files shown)


src/js/controller/instream-adapter.js
```diff
@@ -201,7 +201,7 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         if (skipoffset) {
             // Start background loading once the skip button is clickable
             _this.setupSkipButton(skipoffset, _options);
-            _backgroundLoadPosition = skipoffset;
+            _backgroundLoadPosition = typeof skipoffset === 'string' ? utils.seconds(skipoffset) : skipoffset;
```
robwalch: - `utils.seconds(skipoffset)` has a type check do you should just use it, don't add a check around it.
- skipOffset can be a percentage in which case this won't work - look at asktipbutton.js and consider updating and reading adModel 'skipOffset' after the time has been updated
- When skipOffset is a percent it will not be calculated until the ad duration is known so this should be async

src/js/controller/instream-adapter.js
```diff
@@ -201,7 +201,7 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
...
         } else {
             // If no skipoffset is set, default to background loading 5 seconds before the end
             _backgroundLoadPosition = item.duration - 5;
```
robwalch: Ad duration may not be known until ad playback starts. This logic should really go into a time handler.

src/js/program/program-controller.js
```diff
@@ -58,6 +56,7 @@ export default class ProgramController extends Eventable {
                 this._setActiveMedia(mediaController);
                 // Initialize the provider last so it's setting properties on the (newly) active media model
                 mediaController.provider.init(item);
+                model.setActiveItem(index);
```
robwalch: Setting of playlist item and index should be synchronous. This would make `setPlaylistItem()` and `setPlaylistIndex` asynchronous.

robwalch: oh nvrmind!

robwalch: Can this come after `this.providerPromise` ? If that doesn't work, the one on line 84 won't work either.

robwalch: It should work if play() always follows `this.providerPromise`.

robwalch: Just after `this._setActiveMedia` is good enough.

karimJWP:
test this please

src/js/controller/instream-adapter.js
```diff
@@ -3,6 +3,7 @@ import { STATE_BUFFERING, STATE_COMPLETE, STATE_PAUSED,
     ERROR, MEDIA_META, MEDIA_TIME, MEDIA_COMPLETE,
     PLAYLIST_ITEM, PLAYLIST_COMPLETE, INSTREAM_CLICK, AD_SKIPPED } from 'events/events';
 import utils from 'utils/helpers';
+import { offsetSeconds } from 'utils/strings';
```
johnBartos: `offsetSeconds` isn't a name which describes what the function does. What about something like `toSeconds`?

karimJWP: @johnBartos  agreed, its not the best name but we already have a `seconds` function. The added benefit of this function is that it handles offsets that are percentages as well (i.e. '30%' ) . So this is specific to offsets. `offsetToSeconds`?

src/js/utils/strings.js
```diff
@@ -99,6 +99,19 @@ export function seconds(str, frameRate) {
...
 
+// Convert an offset string to a number; supports conversion of percentage offsets
+export function offsetSeconds(offset, duration, frameRate) {
+    if (_.isString(offset) && offset.slice(-1) === '%') {
+        const percent = parseFloat(offset);
+        if (duration && !isNaN(percent)) {
+            return duration * percent / 100;
+        }
+        return 0;
+    } else if (!isNaN(Number(offset))) {
```
johnBartos: Can we split this so that we only call the `Number(offset`) constructor once?

src/js/utils/strings.js
```diff
@@ -99,6 +99,19 @@ export function seconds(str, frameRate) {
...
+        return Number(offset);
```
johnBartos: Can we use `parseInt` or `parseFloat` instead? The `Number` constructor can cast all sorts of stuff, including hex, which we may not want.

src/js/utils/strings.js
```diff
@@ -99,6 +99,19 @@ export function seconds(str, frameRate) {
     return sec;
 }
 
+// Convert an offset string to a number; supports conversion of percentage offsets
+export function offsetSeconds(offset, duration, frameRate) {
```
johnBartos: Add unit tests

src/js/utils/strings.js
```diff
@@ -99,6 +99,17 @@ export function seconds(str, frameRate) {
...
 }
 
+// Convert an offset string to a number; supports conversion of percentage offsets
+export function offsetToSeconds(offset, duration, frameRate) {
+    if (_.isString(offset) && offset.slice(-1) === '%') {
+        const percent = parseFloat(offset);
+        if (duration && !isNaN(percent)) {
+            return duration * percent / 100;
+        }
+        return 0;
```
robwalch: We should not return 0 if duration is unknown. The result is not an offset of `0` seconds. We should return `null` and then so code using this util knows to wait for a valid duration.

src/js/utils/strings.js
```diff
@@ -99,6 +99,17 @@ export function seconds(str, frameRate) {
     return sec;
 }
 
+// Convert an offset string to a number; supports conversion of percentage offsets
+export function offsetToSeconds(offset, duration, frameRate) {
+    if (_.isString(offset) && offset.slice(-1) === '%') {
+        const percent = parseFloat(offset);
+        if (isNaN(duration) || isNaN(percent)) {
```
robwalch: If duration is `0` (or falsy), then we also do not want to return a final result:

`if (!duration || isNaN(duration) || isNaN(percent)) {`

johnBartos:
test this please

karimJWP:
@johnBartos and @robwalch , pdw @dannyfinks integration tests are flaky for IE 10. @dannyfinks do we have your blessing ?

dannyfinks:
@karimJWP good to go

src/js/program/program-controller.js
```diff
@@ -89,6 +89,7 @@ class ProgramController extends Eventable {
                     return nextMediaController;
                 }
             });
+        model.setActiveItem(index);
```
robwalch: This change breaks the test I added over the weekend that ensure that "playlistItem" is triggered before other events.

I think we should remove this change to when `model.setActiveItem(index);`, and instead add an "itemReady" property to the model, that is set to `false` at the beginning of `setActiveItem(index)` and then set to `true` where we trigger "itemReady". This property would be used as the condition for queuing commands on the controller so that `play` when called on playlistItem, does not execute until everything in this method is resolved.

robwalch: @karimJWP @johnBartos 
https://github.com/jwplayer/jwplayer/blob/master/src/js/controller/controller.js#L908-L919

This is the condition under which we wait to execute `play`. It seems like since `activeProvider` is set in between items, play can be run while `setActiveItem` is running. Having an "itemReady" flag would help us prevent this.

```js
        // Setup ApiQueueDecorator after instance methods have been assigned 
        const apiQueue = new ApiQueueDecorator(this, [
            'play',
            'pause',
            'seek',
            'setCurrentAudioTrack',
            'setCurrentCaptions',
            'setCurrentQuality',
            'setFullscreen',
        ], () => !_programController.activeProvider);
```

karimJWP:
test this please

src/js/api/config.js
```diff
@@ -50,7 +50,8 @@ const Defaults = {
         unmute: 'Unmute'
     },
     renderCaptionsNatively: false,
-    nextUpDisplay: true
+    nextUpDisplay: true,
+    itemReady: false
```
robwalch: This goes in `INITIAL_PLAYER_STATE`, not here in the default config (it's not configurable).

src/js/controller/captions.js
```diff
@@ -26,7 +26,7 @@ const Captions = function(_model) {
     }, this);
 
     // Update tracks once we know "renderCaptionsNatively" based on provider
-    _model.on('itemReady', () => {
+    _model.on('change:itemReady', () => {
```
robwalch: This will fire when it's changed to false and to true. It should only run when true. You can prevent change:itemReady with false if you set false directly on the `model.attributes`.

src/js/program/program-controller.js
```diff
@@ -45,6 +45,7 @@ class ProgramController extends Eventable {
         const item = model.get('playlist')[index];
 
         model.setActiveItem(index);
+        model.set('itemReady', false);
```
robwalch: This should be set before `setActiveItem`. To prevent this from trigging a change event, set it directly on the `model.attributes`:
  
```js
model.attributes.itemReady = false;
model.setActiveItem(index);
```

src/js/program/program-controller.js
```diff
@@ -213,8 +213,7 @@ class ProgramController extends Eventable {
         this._setActiveMedia(castMediaController);
         // Initialize the provider last so it's setting properties on the (newly) active media model
         castMediaController.provider.init(playlistItem);
-        model.trigger('itemReady', playlistItem);
-
+        model.set('itemReady', true);
```
robwalch: Since we don't call `model.setActiveItem` there is no "playlistItem" event here. If we do need the "change: itemReady" event the you do have to set "itemReady" to false first (before `_setActiveMedia`)

src/js/controller/model.js
```diff
@@ -13,6 +13,7 @@ const Model = function() {
     let _provider;
     this.mediaModel = new MediaModel();
 
+
```
robwalch: extra line

karimJWP:
test this please

johnBartos:
`
Expected that the event on('adClick') has triggered the correct number of times

	expected:	1
	got:		1
`

Hmmmm. Results look fine to me

PR #2651, opened 2018-01-05
Fix Freewheel midroll playback

johnBartos:
### This PR will...
- Allocate a single tag for ad plugins to use. This tag will be used for all ad playback during the session, and will not be used for regular media playback
- Remove `ad-program-controller`
- Refactor `shared-media-pool` to pull a tag from the `media-pool`
- Add a constants file to eliminate magic numbers

Freewheel initializes once per playlist item, and uses this tag for all media elements. If we don't always give it the same tag, it will sometimes play it's ad in the background.

No

No

#### Addresses Issue(s):

JW8-1042


src/js/controller/instream-adapter.js
```diff
@@ -350,7 +351,8 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         if (_destroyed) {
             return null;
         }
-        return _adProgram.primedElement;
+        // window.primedTag = _adProgram.primedElement;
+        // return _adProgram.primedElement;
```
robwalch: @johnBartos what happened here? VAST and IMA need a video tag.

src/js/program/ad-program-controller.js
```diff
@@ -132,21 +135,17 @@ export default class AdProgramController extends ProgramController {
 
 
     destroy() {
-        const { model } = this;
+        const { model, mediaElement, mediaPool } = this;
 
         model.off();
-        this.mediaPool.recycle();
+        mediaPool.clean(this.mediaElement);
```
robwalch: We should only call `mediaPool.clean` if `Features.backgroundLoading`, since the operation is not necessary when we're only using a single video tag, and when that's the case it would have a noticeable effect.

robwalch:
@johnBartos you might want to rollback the merge with master and then rebase - there are still mere conflicts.

PR #2654, opened 2018-01-05
Only trigger captionsList event once when item is ready

robwalch:
### This PR will...

Reset captions in the model without triggering a captionsList event on playlistItem. The captionsList event will be triggered on itemReady and contain sideloaded tracks.

This fixes the `cucumber features/api/api_captions.feature` automated test (https://github.com/jwplayer/jwplayer-commercial/pull/4484) that has been failing since captions.js item handling was changed for background loading.

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -66,11 +66,19 @@ const Captions = function(_model) {
                                 });
                         }
                     }
-                    _updateMenu();
                 }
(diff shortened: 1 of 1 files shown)


dannyfinks:
@robwalch looks like captions tests failed because were getting the previous playlist items cues in the next playlist item, but this may just be because this built before the corresponding commercial PR

PR #2657, opened 2018-01-06
Guard against undefined window.localStorage

### LINKED ISSUE #2538: cannot read property 'jwplayer.volume' of undefined
### Expected Behavior

`jwplayer.js` can be loaded into `jsdom` for rudimentary testing of api.

### Actual Behavior

`error: cannot read property 'jwplayer.volume' of undefined`

image

### Steps to reproduce

1.  setup project as described here https://github.com/barsh/jwplayer-jest
2.  run `yarn test`

### Environment

- laptop/desktop
- windows 10
- https://www.npmjs.com/package/jsdom
- https://github.com/barsh/jwplayer-jest/blob/master/jwplayer/debug-v8.0.2/jwplayer.js

### Related
- https://github.com/barsh/jwplayer-jest/issues/1

robwalch:
According to #2538 there's a case where `window.localStorage` could be undefined which breaks player setup. We were only dealing with the case where accessing `window.localStorage` throws an error (Safari), but not where it's undefined.

Fixes #2538

### DIFF
--- a/src/js/model/storage.js
+++ b/src/js/model/storage.js
@@ -6,7 +6,7 @@ let storage = {
 };
 
 try {
-    storage = window.localStorage;
+    storage = window.localStorage || storage;
 } catch (e) {/* ignore */}
 
 function Storage(namespace, persistItems) {


robwalch:
@egreaves let's hold on merging until the npm outage is resolved. I've added a yarn lock file and config for travis which gets this project back up. We'll need to get yarn on Jenkins on Monday to avoid these types of outages.

PR #2667, opened 2018-01-08
Stopped mouseover event from triggering when coming from nextUp

waxidiotic:
### This PR will...

Stop the (mouse)OVER event from triggering when coming from any element within the nextup overlay

Without it, if the controls hide/user goes inactive while hovering the nextup overlay, a mouseover event will fire on the video element causing the controls to show again. This causes the illusion that the nextup display is flickering.

Is this the best way of doing this?

No.

#### Addresses Issue(s):

JW8-1004

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -129,7 +129,7 @@ const UI = function (elem, options) {
 
     // overHandler and outHandler not assigned in touch situations
     function overHandler(evt) {
-        if (evt.pointerType !== 'touch' && evt.fromElement.classList[0].indexOf('jw-nextup') === -1) {
+        if (evt.pointerType !== 'touch') {
             triggerEvent(OVER, evt);
         }
     }
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -129,7 +129,7 @@ const UI = function (elem, options) {
 
     // overHandler and outHandler not assigned in touch situations
     function overHandler(evt) {
-        if (evt.pointerType !== 'touch') {
+        if (evt.pointerType !== 'touch' && evt.fromElement.classList[0].indexOf('jw-nextup') === -1) {
             triggerEvent(OVER, evt);
         }
     }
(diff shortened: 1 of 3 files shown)


src/js/utils/ui.js
```diff
@@ -129,7 +129,7 @@ const UI = function (elem, options) {
 
     // overHandler and outHandler not assigned in touch situations
     function overHandler(evt) {
-        if (evt.pointerType !== 'touch') {
+        if (evt.pointerType !== 'touch' && evt.fromElement.classList[0].indexOf('jw-nextup') === -1) {
```
robwalch: Ui.js is a generic util. Custom logic for nextup should not go into it. What is listening to `OVER` that causes controls to toggle? This logic should be handled there - likely in view.js or controls.js or a submodule of controls.

src/js/view/view.js
```diff
@@ -404,7 +404,11 @@ function View(_api, _model) {
             },
             doubleClick: () => _controls && api.setFullscreen(),
             move: () => _controls && _controls.userActive(),
-            over: () => _controls && _controls.userActive()
+            over: (evt) => {
+                if (_controls && (evt.sourceEvent.fromElement.classList.length === 0 || evt.sourceEvent.fromElement.classList[0].indexOf('jw-nextup') === -1)) {
```
waxidiotic: Had to also add check for if the classList array was empty, otherwise there was an exception being thrown if the mouse event was sourcing from an element without any classes due to `classList[0]`.

robwalch: Use `evt.sourceEvent.fromElement.className.indexOf('jw-nextup') === -1`. That's just a string of all classes, and it's an empty string when there are no classes - unless we're dealing with a different nodeType that doesn't where classes are undefined. In that case `(evt.sourceEvent.fromElement.className || '').indexOf('jw-nextup') === -1`

robwalch:
We shul

PR #2672, opened 2018-01-09
Implement a small time padding around BGL triggers

johnBartos:
To ensure we start background loading before being able to transition to that item. This ensures that we always (attempt to) load something. With the current implementation tying BGL to when the skip button becomes clickable, automated tests work too quickly to allow BGL to begin.

No

No


robwalch:
@johnBartos please resolve conflicts

PR #2680, opened 2018-01-12
Add background tag priming for Android 

johnBartos:
A player which did not set up within a gesture will not start with primed tags. If we preload and then play a preroll, we have background loaded content which cannot be played without further gesture on Android. This PR will call load again on Android if the media has set a src, but has not yet begun playback (and is thus safe to reload).

Sure

No

#### Addresses Issue(s):

JW8-1076

### DIFF
--- a/src/js/program/media-element-pool.js
+++ b/src/js/program/media-element-pool.js
@@ -65,11 +65,14 @@ function primeMediaElementForPlayback(mediaElement) {
     // If we're in a user-gesture event call load() on video to allow async playback
     if (!mediaElement.src) {
         mediaElement.load();
-    } else if (OS.android && !mediaElement.parentNode && !mediaElement.currentTime) {
+    } else if (OS.android && !mediaElement.parentNode) {
         // If the player sets up without a gesture and preloads, the background tag may not be primed for playback.
         // We need to load again on Android in order to play without another gesture. But make sure we're only reloading
-        // a tag which hasn't begun playback yet.
-        mediaElement.load();
+        // a tag which hasn't begun playback yet
+        const played = mediaElement.played;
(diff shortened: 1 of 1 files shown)


src/js/program/media-element-pool.js
```diff
@@ -64,6 +65,12 @@ function primeMediaElementForPlayback(mediaElement) {
     // If we're in a user-gesture event call load() on video to allow async playback
     if (!mediaElement.src) {
         mediaElement.load();
+    } else if (OS.android && !mediaElement.parentNode) {
+        // If the player sets up without a gesture and preloads, the background tag may not be primed for playback.
+        // We need to load again on Android in order to play without another gesture
+        if (!mediaElement.played.length) {
```
egreaves: Combine this if with the else if:
`else if (OS.android && !mediaElement.parentNode && !mediaElement.played.length)`

src/js/program/media-element-pool.js
```diff
@@ -64,6 +65,10 @@ function primeMediaElementForPlayback(mediaElement) {
     // If we're in a user-gesture event call load() on video to allow async playback
     if (!mediaElement.src) {
         mediaElement.load();
+    } else if (OS.android && !mediaElement.parentNode && !mediaElement.played.length) {
```
robwalch: `mediaElement.played` could be `undefined`.

This could be the case on older android devices or browsers. I think just calling `load()` in these cases would be best otherwise we risk old gesture blocking behavior.

johnBartos: What about if `!currentTime`?

johnBartos:
Failure is a flaky IE test

PR #2682, opened 2018-01-12
Restore stretching functionality in IE/Edge with HLS/DASH streams

boyntoni:
### This PR will...

Enables stretching settings to be respected in IE/Edge. Because IE/Edge do not support object-fit, moved existing logic from HTML5 provider into the VideoActionsMixin. Refactored further to have HTML5 provider rely on this mix-in, as opposed to its own method. 

Stretching settings had not been functioning on IE/Edge with HLS/DASH streams. 

N/A

No

#### Addresses Issue(s):

JW8-772

### DIFF
--- a/src/js/providers/video-actions-mixin.js
+++ b/src/js/providers/video-actions-mixin.js
@@ -23,21 +23,21 @@ const VideoActionsMixin = {
             return false;
         }
         const _videotag = this.video;
-        const styles = {
-            objectFit: '',
-            width: '',
-            height: '',
+        let styles = {
+            objectFit: null,
+            width: null,
+            height: null,
         };
         if (stretching === 'uniform') {
             // snap video to edges when the difference in aspect ratio is less than 9%
-            var playerAspectRatio = width / height;
-            var videoAspectRatio = _videotag.videoWidth / _videotag.videoHeight;
+            let playerAspectRatio = width / height;
+            let videoAspectRatio = _videotag.videoWidth / _videotag.videoHeight;
             if (Math.abs(playerAspectRatio - videoAspectRatio) < 0.09) {
                 styles.objectFit = 'fill';
             }
         }
         if (fitVideoUsingTransforms) {
-            fitToBounds(_videotag, width, height, stretching, styles);  
+            styles = fitToBounds(_videotag, width, height, stretching, styles);  
         } 
         style(_videotag, styles);
         return false;
--- a/src/js/providers/video-actions-mixin.js
+++ b/src/js/providers/video-actions-mixin.js
@@ -1,4 +1,5 @@
 import { style } from 'utils/css';
+import { fitToBounds, fitVideoUsingTransforms } from 'utils/video-fit';
 
 const VideoActionsMixin = {
     container: null,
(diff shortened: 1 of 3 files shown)


src/js/providers/video-actions-mixin.js
```diff
@@ -21,20 +22,24 @@ const VideoActionsMixin = {
         if (!width || !height || !this.video.videoWidth || !this.video.videoHeight) {
             return false;
         }
+        const _videotag = this.video;
+        const styles = {
+            objectFit: '',
+            width: '',
```
johnBartos: Are there any side effects of using `''` instead of `null` for width and height?

src/js/providers/video-actions-mixin.js
```diff
@@ -21,20 +22,24 @@ const VideoActionsMixin = {
...
+            height: '',
+        };
         if (stretching === 'uniform') {
             // snap video to edges when the difference in aspect ratio is less than 9%
             var playerAspectRatio = width / height;
-            var videoAspectRatio = this.video.videoWidth / this.video.videoHeight;
-            var objectFit = null;
+            var videoAspectRatio = _videotag.videoWidth / _videotag.videoHeight;
```
johnBartos: Let's use `const` and `let` where applicable

src/js/utils/video-fit.js
```diff
@@ -0,0 +1,26 @@
...
+    } else if (stretching === 'uniform') {
+        scaleX = scaleY = Math.min(scaleX, scaleY);
+    }
+    styles.width = _videotag.videoWidth;
+    styles.height = _videotag.videoHeight;
+    styles.top = styles.left = '50%';
+    styles.margin = 0;
+    transform(_videotag,
+        'translate(' + x + 'px, ' + y + 'px) scale(' + scaleX.toFixed(2) + ', ' + scaleY.toFixed(2) + ')');
+    return styles;
```
johnBartos: Either use this return or remove it. I think it would be better to make our code with with a return rather than the modification of an argument

src/js/providers/video-actions-mixin.js
boyntoni: This doesn't; have a few more things to fix, but have made the update for my next commit.

src/js/providers/video-actions-mixin.js
boyntoni: Changed to let in next commit

src/js/utils/video-fit.js
boyntoni: Used the return in next commit;

src/js/providers/video-actions-mixin.js
robwalch:
```js
function _styleValue(property, value) {
    if (value === '' || value === undefined || value === null) {
        return '';
    }
```

src/js/providers/video-actions-mixin.js
robwalch:
```js
function _styleValue(property, value) {
    if (value === '' || value === undefined || value === null) {
        return '';
    }
```

src/js/providers/video-actions-mixin.js
```diff
@@ -21,20 +22,24 @@ const VideoActionsMixin = {
...
-            style(this.video, {
-                objectFit,
-                width: null,
-                height: null
-            });
         }
+        if (fitVideoUsingTransforms) {
+            fitToBounds(_videotag, width, height, stretching, styles);  
+        } 
+        style(_videotag, styles);
```
robwalch: @boyntoni @johnBartos We should never apply "objectFit" styles, and a hack that attempts to mimic "objectFit" at the same time.

It should always be one of the other. If some versions or Edge or native Android browser actually render "objectFit" properly - then you add the hack on top, it would become over stretched.

src/js/utils/video-fit.js
```diff
@@ -0,0 +1,26 @@
+import { transform } from 'utils/css';
```
robwalch: utils is for common utils used across the code base.

This is code specific to providers. It belongs in the providers folder, next to video-actions-mixin.js or just put this code in that mixin.

PR #2685, opened 2018-01-16
Use attach detach with providers that do not use a video tag

robwalch:
### This PR will...

Use media-controller's attach/detach methods when background is set on providers that do use a video tag (`provider.video`).

The program-controller either uses backgrounding logic or attach/detach based on a global Feature setting. However, backgrounding only works with providers that have a video property. The Flash provider does not have a video property, use a video tag, or have the ability to be removed from the DOM without completely resetting it's state (plugins only run while in the DOM).

Because of this limitation, the flash provider was exempt from backgrounding, but always reported as backgrounded. None of it's events were ever forwarded through the program-controller.

To get around this, attach/detach is used, which has the same affect.

Events appear to come through, and I was able to run through a simple VAST config with an flv playlist item. There's also a new test with mock "videoless" provider to assert this fix.

I have not run through cucumber tests for Flash, but would like to know how reliable the coverage we have there is and how often we run it and report bugs against it. We can discuss IE 11 Win 7 only testing vs another platform using flv sources - maybe one for quick reliable results, and IE11 for full coverage.

#### Addresses Issue(s):

JW8-1089

### DIFF
--- a/src/js/program/media-controller.js
+++ b/src/js/program/media-controller.js
@@ -189,6 +189,10 @@ export default class MediaController extends Eventable {
         if (!this.attached) {
             return false;
         }
+        // A provider without a video tag cannot be backgrounded
+        if (!provider.video) {
+            return false;
+        }
         // A backgrounded provider does not have a parent container, or has one, but without the media tag as a child
         return !container || (container && !container.contains(provider.video));
     }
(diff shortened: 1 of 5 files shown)


src/js/program/media-controller.js
```diff
@@ -231,7 +235,16 @@ export default class MediaController extends Eventable {
 
     set background(shouldBackground) {
         const { container, provider } = this;
-        if (!container || !provider.video) {
+        // A provider without a video tag must use attach and detach
+        if (!provider.video) {
```
johnBartos: Do you think it would be better to handle this higher up, and add this logic to `controller.js`?
https://github.com/jwplayer/jwplayer/blob/master/src/js/controller/controller.js#L632

robwalch: That would bring into question every place that we check `Features.backgroundLoading`. So no, I do not.

johnBartos: Ok

PR #2690, opened 2018-01-19
Prevent duplicate "captionsList" events

robwalch:
### This PR will...

- Only dispatch "captionsList" events when the captions contents don't match the current contents.
- Fix an issue with the ApiQueueDecorator exposed by cucumber tests. (following up with unit tests).
- Keep API queue active until the player is ready so that "beforePlay" can't be fired before ready by calling play() in the "ready" event.

This is needed with native rendering where "subtitleTracks" change events are triggered for both sideloaded tracks and video element 'loaddata' events, which results in an update to "captionsList".

The ApiQueueDecorator fix is very important. The method for resetting decorated methods was flawed in that in only worked if there were commands in the queue. This doesn't matter for other places we use it, because we don't destroy the object while the player is running. Here, on ready it must be turned off and destroyed properly.

#### Addresses Issue(s):

JW8-1101
ADS-871

### DIFF
--- a/src/js/api/api-queue.js
+++ b/src/js/api/api-queue.js
@@ -42,7 +42,7 @@ export default function ApiQueueDecorator(instance, queuedCommands, predicate) {
     };
 
     this.off = function() {
-        commandQueue.forEach(({ command }) => {
+        queuedCommands.forEach((command) => {
             const method = undecoratedMethods[command];
             if (method) {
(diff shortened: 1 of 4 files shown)


robwalch:
@dannyfinks I can't quite figure out why the tests are failing - they seem to not detect playback starting?

PR #2691, opened 2018-01-19
Only pause a tag when actually detached 

johnBartos:
- Only pauses a tag when initializing an ad (as opposed to just creating instream)
- Only remove a tag from the container when backround loading

- DAI and nonlinear ads create an instream adapter, but do not init - and therefore should not pause the video (as this was previous behavior.
- Removing the tag on destroy when not background loading causes ads to pause

No

No

#### Addresses Issue(s):

JW8-1112

### DIFF
--- a/src/js/providers/video-attached-mixin.js
+++ b/src/js/providers/video-attached-mixin.js
@@ -15,10 +15,6 @@ const VideoAttachedMixin = {
         this.stopStallCheck();
         this.eventsOff_();
 
-        if (this.video && !this.video.paused && this.video.pause) {
-            this.video.pause();
-        }
-
         return this.video;
     },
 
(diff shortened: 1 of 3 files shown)


src/js/providers/video-attached-mixin.js
```diff
@@ -15,6 +15,10 @@ const VideoAttachedMixin = {
         this.stopStallCheck();
         this.eventsOff_();
 
+        if (this.video && !this.video.paused && this.video.pause) {
```
robwalch: Why the check for `this.video.pause`?

robwalch: I'd rather not change provider behavior. This should stay scoped to the ads program controller - in setup.

PR #2697, opened 2018-01-23
Read mediaContainer off the playerModel when destroying ad-program 

johnBartos:
The player was failing to rove the BGL tag from the DOM. The ad model does not contain `mediaContainer`

No

No

#### Addresses Issue(s):

JW8-1126


robwalch:
It looks like some captions tests failed (this branch may not be in sync with master) and there was one ad test that failed but only in IE.

PR #2698, opened 2018-01-23
Fix handling of stalling buffering state and QoE in html5 provider

robwalch:
### This PR will...

Replace the `stalledHandler` with a media element "waiting" event handler to change the player state to "buffering" (and QoE track "stalling") when the player under-buffers with the html5 provider. The "waiting" handler also replaces the "waiting" event listeners added and removed while seeking.

The `stalledHandler` stopped working when we began changing the state to "playing" on the media element "playing" event. The state is changed to playing before the first "timeupdate" event so `startStallCheck` is never called.

We still want to set the state to "playing" on the "playing" event when "waiting" hasn't fired. It ensures the buffer wheel goes away quickly on start and after seeking. We don't use that event if waiting fired (`this.stallTime` is set) because Safari and Firefox will fire it before playback has recovered from under-buffering. In that state we rely on "timeupdate" as we did before with an additional check that the media element's `currentTime` no longer equals `this.stallTime`.

The `waiting` event lister is set to `noop` on Android native to keep with the previous practice of not running the stall check on Android native.

Unused methods in video-attached-mixin.js have been removed, and the `atEdgeOfLiveStream` method moved to video-actions-mixin.js.

https://github.com/jwplayer/jwplayer-commercial/pull/4575

#### Addresses Issue(s):

JW8-1125

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -1,7 +1,7 @@
 import { qualityLevel } from 'providers/data-normalizer';
 import { Browser, OS } from 'environment/environment';
 import { isAndroidHls } from 'providers/html5-android-hls';
-import { STATE_IDLE, MEDIA_META, MEDIA_ERROR, MEDIA_VISUAL_QUALITY, MEDIA_TYPE,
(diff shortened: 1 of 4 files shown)


robwalch:
@dannyfinks @johnBartos The test failures here appear to be test/browser specific and not caused by these changes or the changes in https://github.com/jwplayer/jwplayer-commercial/pull/4575

PR #2702, opened 2018-01-25
Change mediaType when switching to/from instream mode

robwalch:
### This PR will...

- Prevent Safari from making html5 provider dispatch "audio" mediaType events while preloading HLS video streams
- Update mediaType on the view model when switching to and from instream mode

This ensures that the view toggles the '.jw-flag-media-audio' CSS class when entering and exiting ads mode. And, that this class does not get added to the player when preloading HLS video in Safari.

#### Addresses Issue(s):

JW8-1135

### DIFF
--- a/src/js/program/ad-program-controller.js
+++ b/src/js/program/ad-program-controller.js
@@ -12,6 +12,8 @@ export default class AdProgramController extends ProgramController {
         this.playerModel = model;
         this.provider = null;
 
+        adModel.mediaModel.attributes.mediaType = 'video';
+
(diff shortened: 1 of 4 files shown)


robwalch:
@dannyfinks @mikesalvia It looks like our automated clicking didn't work in IE here. While "last interaction" did capture a click, there was no "playAttempt" from the player: https://jenkins.jwplayer.com/view/JW7/job/jw7-commercial-cucumber/10192/allure/#suites/4ac97c44220ad8d34e7c613de6efd387/942969430ed397f7/

PR #2711, opened 2018-02-05
Add currentTime property to timeEventObject

DanFerrer:
### This PR will...

Instantiate metadata on the timeEventObject and set the currentTime property on metadata to the videos currentTime.

We need this to be visible when getAlignmentPosition is called so track.source can exist we can return the current time of the video when the stream type is DVR instead of position minus duration. This is so captions are properly displayed on Shaka DVR streams.

https://github.com/jwplayer/jwplayer-commercial/pull/4631

#### Addresses Issue(s):

JW8-868

### DIFF
--- a/src/js/providers/video-listener-mixin.js
+++ b/src/js/providers/video-listener-mixin.js
@@ -64,15 +64,16 @@ const VideoListenerMixin = {
         }
 
         var timeEventObject = {
-            position: position,
-            duration: duration
+            position,
+            duration,
+            metdata: {
+                currentTime: this.video.currentTime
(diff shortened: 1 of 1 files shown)


johnBartos:
@mikesalvia @dannyfinks Some iPad safari issues but the rest looks good

dannyfinks:
@DanFerrer have you pulled from master lately? I think these failures may be due to old test code on your branch.

DanFerrer:
@dannyfinks Says its up to date. Commercial was behind though so i pulled down from master for that one.

PR #2713, opened 2018-02-06
Replace live button contents with localizable text

radium-v:
### This PR will...
Replace the SVG images used for Live and DVR with localizable text.

The SVG icon looked really bad and was practically unreadable in Desktop Safari on external monitors.

This applies to both Live and DVR, and uses the existing `localization.liveBroadcast` string

#### Addresses Issue(s):

JW8-501


src/css/controls/flags/live.less
```diff
@@ -37,3 +31,41 @@
...
+    .rect(auto, @icon-size);
+    align-items: center;
+    border-radius: 1px;
+    color: @inactive-color;
+    display: flex;
+    font-size: 12px;
+    font-weight: bold;
+    margin-right: 10px;
+    padding: 0 1ch;
+    text-rendering: geometricPrecision;
```
johnBartos: There's no support for this prop in IE. Does it look fine without it there?

radium-v: Having this property is to improve consistency in Chrome on MacOS. Firefox treats this value as `optimizeLegibility` (which doesn't have any effect on Arial), Safari doesn't do anything with the property, and IE/Edge use DirectWrite and render the text properly.

PR #2717, opened 2018-02-08
Add canAutoplay test

vseventer:
### This PR will...
Expose a `canAutoplay(mediaPool, options)` function which returns a promise, with options:
- `cancelable`: if this cancelable promise is cancelled, the test will be interrupted and cancelled.
- `muted`: test against a muted `<video>` if `true`.
- `allowMuted`: applicable only if `muted` is false (i.e. the test was performed on an unmuted `<video>` element) - performs a second test to see if we can autoplay a muted `<video>` element.
- `timeout`:  rejects the promise if an individual test takes more than the specified amount (ms).

Promise is rejected with error (no autoplay errors here! - e.g. timeout error), or resolve with one of:
- `AUTOPLAY_ENABLED`: The browser allows (unmuted) autoplay.
- `AUTOPLAY_MUTED`: The browser allows muted autoplay. Note if the `muted` option was `true`, this is the worst-case scenario (i.e. no test for unmuted autoplay was performed, so the browser setting might actually be `AUTOPLAY_ENABLED`). 
- `AUTOPLAY_DISABLED`: The browser does not allow autoplay. Note if the `allowMuted` option was `false`, the browser setting might actually be `AUTOPLAY_MUTED`.

Deal with autoplay issues in Safari / Chrome 65+. Follow-up work to be done (other tickets).

The `muted` and `allowMuted` options allow you to control if one or two tests will be performed. For example, if, as a publisher, you always want to autoplay muted, there is no point in testing if the browser supports unmuted autoplay. The options allow you to control what test(s) is/are actually performed.

#### Addresses Issue(s):
JW8-1157
ADS-937

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -12,6 +12,7 @@ import ViewModel from 'view/view-model';
 import changeStateEvent from 'events/change-state-event';
 import eventsMiddleware from 'controller/events-middleware';
 import Events from 'utils/backbone.events';
+import { canAutoplay } from 'utils/can-autoplay';
 import { OS, Features } from 'environment/environment';
 import { streamType } from 'providers/utils/stream-type';
 import Promise, { resolved } from 'polyfills/promise';
@@ -254,6 +255,17 @@ Object.assign(Controller.prototype, {
                 // this player has been destroyed
                 return;
             }
+
+            // Detect and store browser autoplay setting in the model.
+            const adConfig = _this._model.get('advertising');
+            canAutoplay(mediaPool, {
+                cancelable: checkAutoStartCancelable,
+                muted: _this.getMute(),
+                allowMuted: adConfig ? adConfig.autoplayadsmuted : false
+            })
+                .then(result => _model.set('canAutoplay', result))
+                .catch(function() {});
+
             if (!OS.mobile && _model.get('autostart') === true) {
                 // Autostart immediately if we're not mobile and not waiting for the player to become viewable first
                 _autoStart();
@@ -742,7 +754,7 @@ Object.assign(Controller.prototype, {
         this.getState = _getState;
         this.next = _nextUp;
         this.setConfig = (newConfig) => {
-            setConfig(_this, newConfig); 
+            setConfig(_this, newConfig);
         };
         this.setItemIndex = _setItem;
 
@@ -771,7 +783,7 @@ Object.assign(Controller.prototype, {
             updateProgramSoundSettings();
         };
         this.setPlaybackRate = (playbackRate) => {
-            _model.setPlaybackRate(playbackRate); 
+            _model.setPlaybackRate(playbackRate);
         };
         this.getProvider = () => _model.get('provider');
         this.getWidth = () => _model.get('containerWidth');
--- a/src/js/utils/can-autoplay.js
+++ b/src/js/utils/can-autoplay.js
@@ -23,6 +23,7 @@
 // SOFTWARE.
 
 import createPlayPromise from '../providers/utils/play-promise';
+import Promise from '../polyfills/promise';
 
 /**
  * Small video file with audio.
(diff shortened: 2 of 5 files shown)


src/js/utils/can-autoplay.js
```diff
@@ -0,0 +1,94 @@
...
+    // Set-up.
+    const mediaPool = controller.mediaPool;
+    const element = mediaPool.getPrimedElement();
+    const initialMute = element.muted;
+    const reset = () => {
+        mediaPool.syncMute(initialMute);
+        mediaPool.recycle(element);
+        controller.attached = true; // Re-attach.
+    };
+    controller.attached = false; // Detach.
```
robwalch: I don't think we should detach and attach controllers. That functionality is specific to switching binding of events from the playlist item to something else, but this test should not interfere with that. Calling `attached = true` might do more bad than good.

What we will need to do handle any change in state that might have happened while the test was being performed. Example: user clicks to start playback while test is running. This could be done externally by making the `then` callback use our `cancelable` util.

vseventer: The test does interfere with the model if there is only one `<video>` element - it seems that events on the video element trigger model changes without detaching. Is there a better way to reset the model after the test, or ideally, detach the video event listeners during the test?

Regarding cancellation, yes, we need to handle this, which `then` callback are you referring to? I see trouble ahead when canceling right after `startPlayback` - the `then` will not be executed, but the attributes of the `<video>` element will still be altered in `startPlayback`, which conflicts when the user clicked to start playback. What would help here is if we can attach an `onCancel` handler to the promise somehow, so we can clean things up.

vseventer:
For your first point - wouldn't it make sense to save a property on the model so we can refer to it during `play*` events?

src/js/utils/can-autoplay.js
robwalch: @vseventer In this case we must use a different video tag than one attached to the player. The test would wipe out any preloading performed on iOS or Safari Mac and we don't want that. Maybe we can extend the pool to provide a primed tag for this test that isn't the main "mediaElement".

vseventer:
Single failing "sharing" test in Allure Report above is unrelated to this PR.

src/js/controller/controller.js
```diff
@@ -254,6 +255,17 @@ Object.assign(Controller.prototype, {
                 // this player has been destroyed
                 return;
             }
+
+            // Detect and store browser autoplay setting in the model.
+            const adConfig = _this._model.get('advertising');
+            canAutoplay(mediaPool, {
+                cancelable: checkAutoStartCancelable,
+                muted: _this.getMute(),
+                allowMuted: adConfig ? adConfig.autoplayadsmuted : false
```
robwalch: If we want to test a muted video tag when the unmuted test fails this should go `allowMuted: adConfig ? adConfig.autoplayadsmuted : true` correct?  The player should perform both tests because the current player behavior (on mobile) is to autostart muted if that is the only way autostart works.

vseventer: On mobile, I don't think we ever attempt to autoplay unmuted - `_this.getMute()` will return `true` if `autostartMuted` or already muted. `autostartMuted` is set if we want to autostart on mobile - hence we only do one test (unmuted) on mobile.

src/js/utils/can-autoplay.js
```diff
@@ -0,0 +1,81 @@
...
+        setTimeout(reject, timeout, new Error('Autoplay test timed out'));
+    });
+    return Promise.race([ promise, timer ]);
+}
+
+export const AUTOPLAY_ENABLED = 'autoplayEnabled';
+export const AUTOPLAY_MUTED = 'autoplayMuted';
+export const AUTOPLAY_DISABLED = 'autoplayDisabled';
+
+export function canAutoplay (mediaPool, { cancelable, muted = false, allowMuted = false, timeout = 250 }) {
```
vseventer: It appears the default timeout of 250ms is a bit too short on mobile - the timeout gets hit constantly, resulting in click-to-play while it should autoplay. Should we bump this (optionally on mobile only) to 500ms or 1s? @robwalch

src/js/controller/controller.js
vseventer: Looks like you are right here, if we want to duplicate the mobile behavior (play silently unless `ads && !autoplayAdsMuted`), we should default to `true`.

PR #2718, opened 2018-02-08
Default writingMode to display horizontal text

egreaves:
### This PR will...
Set `'horizontal-tb'` as the default value for `writingMode` and only switch to vertical mode when an enumerated option is supplied.
`writingMode` may be `undefined`. Checking for an empty string led to a false positive that caused text to be vertically oriented in IE/Edge.
No.
Yes - https://github.com/jwplayer/jwplayer-commercial/pull/4670
#### Addresses Issue(s):
JW8-868

### DIFF
--- a/src/js/polyfills/webvtt.js
+++ b/src/js/polyfills/webvtt.js
@@ -418,7 +418,7 @@ function CueStyleBox(window, cue) {
         top: 0,
         bottom: 0,
         display: 'inline',
-        writingMode: writingMode,
+        writingMode,
         unicodeBidi: 'plaintext',
     };
 
@@ -433,7 +433,7 @@ function CueStyleBox(window, cue) {
         whiteSpace: 'pre-line',
         position: 'absolute',
         direction: determineBidi(this.cueDiv),
-        writingMode: writingMode,
+        writingMode,
         unicodeBidi: 'plaintext',
     };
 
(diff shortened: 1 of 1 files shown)


egreaves:
@dannyfinks the failure doesn't seem legit.

dannyfinks:
@egreaves the way the report shows where it fails is deceiving, but it fails when switching to the next playlist item (an hls.js source)

src/js/polyfills/webvtt.js
```diff
@@ -412,7 +418,7 @@ function CueStyleBox(window, cue) {
         top: 0,
         bottom: 0,
         display: 'inline',
-        writingMode: cue.vertical === '' ? 'horizontal-tb' : cue.vertical === 'lr' ? 'vertical-lr' : 'vertical-rl',
+        writingMode: writingMode,
```
johnBartos: You can use the shorthand here

src/js/polyfills/webvtt.js
```diff
@@ -743,7 +749,7 @@ function moveBoxToLinePosition(window, styleBox, containerBox, boxPositions, num
         // need to increase our initial position by the length or width of the
         // video, depending on the writing direction, and reverse our axis directions.
         if (linePos < 0) {
-            position += cue.vertical === '' ? containerBox.height : containerBox.width;
+            position += !cue.vertical ? containerBox.height : containerBox.width;
```
johnBartos: Can you make this a positive conditional? `cue.vertical ? containerBox.width : containerBox.height`

PR #2719, opened 2018-02-09
Prevent idle state over ads while background loading providers

robwalch:
### This PR will...

Prevent the view-model from propagating attribute changes from the player model for attributes present in instream.

When a preroll ad is played in a mixed provider playlist, the player enters an idle state once the new provider is loaded in the background. Changes on the player model should be masked by the instream model. The getters in the view-model already work this way. This extends that masking to "change:" events so that the view only reflects instream state during ad breaks.

#### Addresses Issue(s):

JW8-1205

### DIFF
--- a/src/js/view/view-model.js
+++ b/src/js/view/view-model.js
@@ -3,6 +3,8 @@ import SimpleModel from 'model/simplemodel';
 class SimpleModelExtendable {}
 SimpleModelExtendable.prototype = Object.assign({}, SimpleModel);
 
+const changeEventRegEx = /^change:(.+)$/;
+
 function dispatchDiffChangeEvents(viewModel, newAttributes, oldAttributes) {
     Object.keys(newAttributes).forEach((attr) => {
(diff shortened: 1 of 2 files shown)


src/js/view/view-model.js
```diff
@@ -87,7 +91,20 @@ class PlayerViewModel extends SimpleModelExtendable {
 
 export default class ViewModel extends PlayerViewModel {
     constructor(playerModel) {
-        super(playerModel);
+        super(playerModel, (type) => {
+            // Do not propagate attribute changes from the player model for attributes present in instream
+            const instreamModel = this._instreamModel;
+            if (instreamModel) {
+                const match = changeEventRegEx.exec(type);
```
johnBartos: Want to unit test this?

robwalch: on it.

johnBartos:
Test failure seems like a test issue. Merging

PR #2721, opened 2018-02-12
Add liveTimeout, a configurable duration to kill a stale live stream

johnBartos:
### This PR will...
- Add `liveTimeout` config option with default value `null`
- Constraint `liveTimeout` values between 1 and 30 to 30; allow a value of 0 to indicate that a stream should never be killed
- Leave stream timeout up to provider when `liveTimeout === null`
- Alphabetize config options

Customers do not always want to have a stale stream killed. Instead of a boolean option, we decided on a configurable value with sensible defaults.

Is a minimum value of 30 fair?

https://github.com/jwplayer/jwplayer-commercial/pull/4696

#### Addresses Issue(s):

JW8-1171

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -152,11 +152,11 @@ const Config = function(options, persisted) {
     if (config.liveTimeout !== null) {
         if (_isNaN(liveTimeout) || !_isNumber(liveTimeout)) {
             liveTimeout = null;
-        } else if (config.liveTimeout !== 0) {
+        } else if (liveTimeout !== 0) {
             liveTimeout = Math.max(30, liveTimeout);
         }
+        config.liveTimeout = liveTimeout;
     }
-    config.liveTimeout = liveTimeout;
 
     return config;
 };
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -1,6 +1,7 @@
 import _ from 'utils/underscore';
 import { loadFrom, getScriptPath } from 'utils/playerutils';
 import { serialize } from 'utils/parser';
+import { _isNumber, _isNaN } from 'utils/underscore';
 
 /* global __webpack_public_path__:true */
 /* eslint camelcase: 0 */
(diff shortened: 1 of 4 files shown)


src/js/api/config.js
```diff
@@ -145,6 +147,12 @@ const Config = function(options, persisted) {
     config.qualityLabels = config.qualityLabels || config.hlslabels;
     delete config.duration;
 
+    if (config.liveTimeout !== null) {
+        if (config.liveTimeout !== 0) {
```
robwalch:
```js
const liveTimeout = config.liveTimeout;
if (!isNaN(parseFloat(liveTimeout)) && isFinite(liveTimeout) && liveTimeout > 0) {
    config.liveTimeout = Math.max(30, liveTimeout);
```

src/js/api/config.js
```diff
@@ -145,6 +148,16 @@ const Config = function(options, persisted) {
     config.qualityLabels = config.qualityLabels || config.hlslabels;
     delete config.duration;
 
+    let liveTimeout = config.liveTimeout;
+    if (config.liveTimeout !== null) {
```
robwalch: use `liveTimeout`

src/js/api/config.js
```diff
@@ -145,6 +148,16 @@ const Config = function(options, persisted) {
...
+        if (_isNaN(liveTimeout) || !_isNumber(liveTimeout)) {
+            liveTimeout = null;
+        } else if (config.liveTimeout !== 0) {
```
robwalch: use `liveTimeout`

src/js/api/config.js
```diff
@@ -145,6 +148,16 @@ const Config = function(options, persisted) {
...
+            liveTimeout = Math.max(30, liveTimeout);
+        }
+    }
+    config.liveTimeout = liveTimeout;
```
robwalch: This should go inside the `config.liveTimeout !== null` _if_ block.

PR #2726, opened 2018-02-14
Add mousedown listener w/prevent default to volume icon

karimJWP:
### This PR will...
add a mousedown listener which will call preventDefault on the event
On ie11, The volume icon needed to be clicked 3 times before being toggled
This fix copies the implementation of [`button.js`](https://github.com/jwplayer/jwplayer/blob/master/src/js/view/controls/components/button.js#L24). It is currently present in 3 places (including [related](https://github.com/jwplayer/jwplayer-plugin-related/pull/251/files#diff-fdc2dbda0ff4965c5fbbf0dc10c17cbaR79)). If we set `preventScrolling` to true when constructing the ui element with [`ui.js`](https://github.com/jwplayer/jwplayer/blob/master/src/js/utils/ui.js#L79), it will call `preventDefault` on the event; is there a reason this was excluded in `button.js`?

#### Addresses Issue(s):
JW8-1225

### DIFF
--- a/src/js/view/controls/components/volumetooltip.js
+++ b/src/js/view/controls/components/volumetooltip.js
@@ -23,6 +23,10 @@ export default class VolumeTooltip extends Tooltip {
             .on('over', this.openTooltip, this)
             .on('out', this.closeTooltip, this);
 
+        this.el.addEventListener('mousedown', (e) => {
+            e.preventDefault();
+        });
+
         this._model.on('change:volume', this.onVolume, this);
     }
 


johnBartos:
@karimJWP Can you explain why calling `preventDefault` fixes this?

karimJWP:
@johnBartos I could not tell, but the bug did not reproduce for elements created with button.js so I applied the main difference i noticed, which fixes it

johnBartos:
@karimJWP I think it's best to understand why this code works (just incase it doesn't around an edge case). Also, what's the milestone for this?

karimJWP:
@johnBartos I agree. I tried to understand why it works but was unsuccessful; I figure for IE weirdness it's not worth spinning my wheels. Since this PR applies a pattern already used in the code, i feel comfortable using it here

robwalch:
What you're doing here is prevent an element from getting focus on mouseDown. The reason we don't want to pass in `preventScrolling` is because that has several other performance downgrades which you should be familiar with after looking at https://github.com/jwplayer/jwplayer/pull/2601

PR #2727, opened 2018-02-14
Add new method for DAI to enable time slider

pajong:
### This PR will...
Create new function in InstreamAdapter that does NOT detach the media, but still listen to the time event and remove view click handler.

We want to be able to update the duration and time when DAI ad is playing. Since DAI is a stream with ad already in the stream, we do not want to detach the media.

https://github.com/jwplayer/jwplayer-ads-dai/pull/28

#### Addresses Issue(s):
ADS-656

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -64,6 +64,25 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
 
     this.type = 'instream';
 
+    this.addAdProgramTimeListener = function() {
+        if (_inited || _destroyed) {
+            return;
+        }
+
+        _adProgram.on(MEDIA_TIME, _instreamTime, this);
+
+        // This enters the player into instream mode
+        _model.set('instream', _adProgram);
+
+        // don't trigger api play/pause on display click
+        const clickHandler = _view.clickHandler();
+        if (clickHandler) {
+            clickHandler.setAlternateClickHandlers(() => {}, null);
+        }
+
+        return this;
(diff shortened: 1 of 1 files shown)


src/js/controller/instream-adapter.js
```diff
@@ -64,6 +64,24 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
...
+        }
+
+        _adProgram.on(MEDIA_TIME, _instreamTime, this);
+
+        // This enters the player into instream mode
+        _model.set('instream', _adProgram);
+
+        // don't trigger api play/pause on display click
+        if (_view.clickHandler()) {
+            _view.clickHandler().setAlternateClickHandlers(() => {}, null);
```
robwalch: Does DAI handle ad clickthrough?

src/js/controller/instream-adapter.js
```diff
@@ -64,6 +64,24 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
...
+            return;
+        }
+
+        _adProgram.on(MEDIA_TIME, _instreamTime, this);
+
+        // This enters the player into instream mode
+        _model.set('instream', _adProgram);
+
+        // don't trigger api play/pause on display click
+        if (_view.clickHandler()) {
```
vseventer: Can we save this in a `const`, so we don't execute the function twice?
```
const clickHandler = _view.clickHandler();
if(clickHandler) {
  clickHandler.setAlternateClickHandlers(...);
}
```

PR #2728, opened 2018-02-14
Set item.starttime when seek is called before starting 

johnBartos:
So that we seek and start buffering at the set time if seek has been called before starting (right after `ready`, for example). Setting `starttime` leverages existing code to load at the set time.

No

https://github.com/jwplayer/jwplayer-commercial/pull/4718
https://github.com/jwplayer/jwplayer-commercial/pull/4581

#### Addresses Issue(s):

JW8-1074

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -492,11 +492,10 @@ Object.assign(Controller.prototype, {
             if (_model.get('state') === STATE_ERROR) {
                 return;
             }
+            _programController.position = pos;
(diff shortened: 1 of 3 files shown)


hghazzi:
test this please

hghazzi:
While testing this, I noticed the block of code in the ticket to seek after firstFrame is not seeking.
```
jwplayer().on('firstFrame', function() {
    jwplayer().seek(50);
}
```

PR #2731, opened 2018-02-15
Unmute instream pseudo provider when ending autostart muted

robwalch:

IMA and Freewheel have to receive mute state updates via instream provider.

#### Addresses Issue(s):

JW8-1234


robwalch:
test this please

PR #2733, opened 2018-02-16
Align with the left property instead of paddingLeft

johnBartos:
`paddingLeft` will squish the captions against the edge of the captions container, causing them to display vertically at one line per word. Using `left` allows the captions to break as intended in the VTT file. I believe this is because the container's  (`jw-captions` div) `content-box` box model, which considered `padding` as part of the overall size.

Do we still need paddingLeft in some cases? Maybe not all properties should be changed from `paddingLeft` to `left`

No

#### Addresses Issue(s):

 JW8-1109

### DIFF
--- a/src/js/polyfills/webvtt.js
+++ b/src/js/polyfills/webvtt.js
@@ -415,6 +415,7 @@ function CueStyleBox(window, cue) {
         position: 'relative',
         paddingLeft: 0,
         paddingRight: 0,
+        left: 0,
         top: 0,
         bottom: 0,
         display: 'inline',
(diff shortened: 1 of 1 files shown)


robwalch:
@egreaves @johnBartos looks good to me with setup:
```js
{
    file: 'http://content.jwplatform.com/manifests/s8BPzDe0.m3u8',
    image: "http://qa.jwplayer.com/~todd/bunny.jpg",
    tracks: [{
        "file": "http://qa.jwplayer.com/~todd/corus.vtt",
        "label": "English",
        "kind": "captions",
    }]
}
```

PR #2734, opened 2018-02-16
[JW8-1158 ADS-938] Use autoplay test to determine autostart.

vseventer:
### This PR will...

- Adjust autostart behavior to only autostart if we are allowed to do so by the browser setting. If configured to do so, the behavior will fall back to autostart muted. 
- Simplify the explicit OS / browser checking in the model, as we rely on the `canAutoplay` test instead. 
- Emit an `autostartNotAllowed` event if:
-- The browser forbids autoplay,
-- The browser allows autoplay muted, but we want to autoplay unmuted.

Deal with browser autoplay policy changes.

- I'm not too happy with the changes in `controls.js`. Since the `autostartMuted` flag is available only after the autoplay test, we can't configure the unmuted controls upon view setup. Instead, the updated code waits for the flag to be set, before updating the controls.
- Also, if the test errors (e.g. timeout), we currently fall back to play-to-click, but do not emit the `autostartNotAllowed` event. Do we want to emit this if a test error occurs?
- The `autostartNotAllowed` event currently does not have any properties - do we want to add any information to this event?

jwplayer/jwplayer-commercial#4730
jwplayer/jwplayer-ads-freewheel#104

#### Addresses Issue(s):
JW8-1158
ADS-938

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -272,7 +272,10 @@ Object.assign(Controller.prototype, {
                     if (!_model.get('playOnViewable')) {
                         _autoStart();
                     }
-                }).catch(function() { });
+                }).catch(function() {
+                    // If the test failed, we will never autostart.
+                    _model.set('autostart', false);
+                });
             }
             apiQueue.flush();
         }
@@ -436,6 +439,7 @@ Object.assign(Controller.prototype, {
                 const mode = _model.get('canAutoplay');
 
                 if (mode === AUTOPLAY_DISABLED) {
+                    _model.set('autostart', false);
                     return _this.trigger(AUTOSTART_NOT_ALLOWED);
                 }
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -273,8 +273,11 @@ Object.assign(Controller.prototype, {
                         _autoStart();
                     }
                 }).catch(function() {
-                    // If the test failed, we will never autostart.
+                    // Never autostart if the test fails. Emit event unless test was explicitly canceled.
                     _model.set('autostart', false);
+                    if (!checkAutoStartCancelable.cancelled()) {
+                        _this.trigger(AUTOSTART_NOT_ALLOWED, { reason: 'autoplayTestFailed' });
+                    }
                 });
             }
             apiQueue.flush();
@@ -440,7 +443,7 @@ Object.assign(Controller.prototype, {
 
                 if (mode === AUTOPLAY_DISABLED) {
                     _model.set('autostart', false);
-                    return _this.trigger(AUTOSTART_NOT_ALLOWED);
+                    return _this.trigger(AUTOSTART_NOT_ALLOWED, { reason: mode });
                 }
 
                 // Only apply autostartMuted on un-muted autostart attempt.
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -423,7 +423,7 @@ Object.assign(Controller.prototype, {
             }
 
             // Detect and store browser autoplay setting in the model.
-            const adConfig = _this._model.get('advertising');
+            const adConfig = _model.get('advertising');
             canAutoplay(mediaPool, {
                 cancelable: checkAutoStartCancelable,
                 muted: _this.getMute(),
@@ -442,7 +442,6 @@ Object.assign(Controller.prototype, {
                     }
                     _actionOnAttach = null;
                 });
-
             }).catch(error => {
                 _model.set('canAutoplay', AUTOPLAY_DISABLED);
                 _model.set('autostart', false);
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -12,7 +12,7 @@ import ViewModel from 'view/view-model';
 import changeStateEvent from 'events/change-state-event';
 import eventsMiddleware from 'controller/events-middleware';
 import Events from 'utils/backbone.events';
-import { canAutoplay } from 'utils/can-autoplay';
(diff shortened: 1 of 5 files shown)


robwalch:
@vseventer Can "autostartNotAllowed" contain the details of the test? What would that look like?

johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

robwalch:
@vseventer @egreaves I think we need a "muted" `event.reason`. It's not something we did before on mobile to explain that the player muted sound in order to autostart, but I think it would be helpful now that we could see this behavior in more browsers.

vseventer:
@robwalch We do have `autostartMuted`, which will only be set if the player forcefully muted to allow autostart - wouldn't this be sufficient?

src/js/controller/controller.js
```diff
@@ -256,19 +258,27 @@ Object.assign(Controller.prototype, {
...
+                    _model.set('canAutoplay', result);
 
-            if (!OS.mobile && _model.get('autostart') === true) {
-                // Autostart immediately if we're not mobile and not waiting for the player to become viewable first
-                _autoStart();
+                    // Autostart immediately if we're not waiting for the player to become viewable first.
+                    if (!_model.get('playOnViewable')) {
+                        _autoStart();
+                    }
+                }).catch(function() {
```
robwalch: Can we get the `error` and pass include it in the `AUTOSTART_NOT_ALLOWED` event? Is the only way that this throws is cancelling and timing out?

That means we get `AUTOSTART_NOT_ALLOWED` events under those conditions, and not when the autoplay test fails legitimately?

src/js/controller/controller.js
```diff
@@ -256,19 +258,27 @@ Object.assign(Controller.prototype, {
...
+                    // Never autostart if the test fails. Emit event unless test was explicitly canceled.
+                    _model.set('autostart', false);
```
robwalch: Shouldn't `'canAutoplay'` be set here when we catch too?

src/js/controller/controller.js
```diff
@@ -428,7 +438,20 @@ Object.assign(Controller.prototype, {
 
         function _autoStart() {
             const state = _model.get('state');
-            if (state === STATE_IDLE || state === STATE_PAUSED) {
+            if (!checkAutoStartCancelable.cancelled() && (state === STATE_IDLE || state === STATE_PAUSED)) {
+                const mode = _model.get('canAutoplay');
+
+                if (mode === AUTOPLAY_DISABLED) {
```
robwalch: It's odd that we duplicate this functionality here (Re: my comment above, before I say this I thought we weren't sending it).

src/js/view/controls/controls.js
```diff
@@ -161,21 +161,23 @@ export default class Controls {
...
-            // Set mute state in the controlbar
-            controlbar.renderVolume(true, model.get('volume'));
-            // Hide the controlbar until the autostart flag is removed
-            utils.addClass(this.playerContainer, 'jw-flag-autostart');
-
-            model.on('change:autostartFailed change:autostartMuted change:mute', unmuteCallback, this);
-            this.unmuteCallback = unmuteCallback;
-        }
+        // Unmute Autoplay Button.
+        model.once('change:autostartMuted', (_model, muted) => {
```
robwalch: If "autostartMuted" is true when controls are added, this should be invoked immediately, but it will not be. YOu can use `model.change('autostartMuted'` to help with this, but then you have to manage removal of the "change:" listeners.

Controls can be added and removed from the player with `jwplayer().setControls(bool)`. Each time controls are readded, this code is rerun with the current model state - so you could have missed the one "change:" event.

src/js/controller/controller.js
```diff
@@ -234,9 +234,11 @@ Object.assign(Controller.prototype, {
             eventsReadyQueue = null;
 
             _model.change('viewable', viewableChange);
-            _model.change('viewable', _checkPlayOnViewable);
-            _model.once('change:autostartFailed change:autostartMuted change:mute', function(model) {
-                model.off('change:viewable', _checkPlayOnViewable);
+            _model.once('change:canAutoplay', function(model) {
+                model.once('change:autostartFailed change:autostartMuted change:mute', function() {
```
robwalch: "autostartMuted" is set as a result of "canAutoplay" which could remove the "change:viewable" listener before it had a chance to work.

src/js/controller/controller.js
vseventer: Yes, only on cancel and timeout. Do you want the entire Error, or just error.message?

src/js/controller/controller.js
vseventer: I was thinking to refactor the `canAutoplay`, `autostart`, and `autostartMuted` to `_checkAutostart`, as well as these handlers there. This will make it easier to emit `autostartNotAllowed` in the different scenarios, tidy things up, and leave `_autoStart` untouched.

robwalch: See https://github.com/jwplayer/jwplayer/pull/2743/files

src/js/controller/controller.js
robwalch: https://github.com/jwplayer/jwplayer/pull/2743/files I've made all scenarios that trigger `AUTOSTART_NOT_ALLOWED` throw, and added 'reason' to the errors.

vseventer:
Tests failing because of a bug - we are now autostarting all videos...

PR #2735, opened 2018-02-21
Close volume tooltip after tab

hghazzi:
### This PR will...

check that an event has pointer coords before `document.elementFromPoint`

When focus on the volume button, the volume tooltip should appear. When tabbing out of it, the logic to check that it should appear on hover out should not be called. Otherwise, `document.elementFromPoint` fails:

`Uncaught TypeError: Failed to execute 'elementFromPoint' on 'Document': The provided double value is non-finite.`

#### Addresses Issue(s):

JW8-1253

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -151,6 +151,10 @@ const UI = function (elem, options) {
         }
     }
 
+    function blurHandler(evt) {
+        triggerEvent(OUT, evt);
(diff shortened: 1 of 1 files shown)


robwalch:
@karimJWP Why does the PR title contain [DO NOT MERGE]?
It looks like this fixes a serious TypeError we should merge immediately.

PR #2739, opened 2018-02-22
Return the same autoplay Promise...

robwalch:
### This PR will...

When multiple autostart players are embedded on the page only run one autoplay test for the mute options at once. Once autostart is allowed for that config, always return the same promise.

This prevents us from running the autoplay test unnecessarily. Multiple players all trying to start at the same time can use the same test. New players that want to autostart later can use the successful result of an earlier test.

#### Addresses Issue(s):

JW8-1158 ADS-936


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

vseventer:
@robwalch This is a nice performance improvement! Can you explain why you are resetting the `autoplayPagePromises[key]` to `null` on muted or disabled autoplay? Wouldn't this cause the test to be re-ran for the same config?

The only caveat I see is what happens if the cancelable is cancelled (e.g. user clicks to play while we are doing the test / timeout). Right now, the promise will be stored in `autoplayPagePromises[key]`, and will always reject from there on for that config. Shouldn't we reset the `key` in a catch handler, so we re-run the test the next time?

robwalch:
@vseventer So if the test fails, it should be performed again - that's why the promise is set to null when it fails.

There is a bit of a catch 22 here. The latest request could be the first to work if it was started by a user gesture event. I don't think memoize is right.

- Once an element plays muted, any element can play muted.
- Once an element plays unmuted, any element can play unmuted.
- Calling load() on the test tag while the test is being performed will fail the test (Couldn't happen if there is a user interaction during the test?)
- Attempting to run the test again before it is complete would fail the test (same video tag)
- Attempting to run multiple tests at once (multiple players/video tags) will not work in clients that only allow a single active video tag

With this in mind, how can handle race conditions between players more elegantly? Once playback is active how can we avoid running the test again for new players (Many pubs embed players after page load with `jwplayer().setup()` on user interaction).

robwalch:
@vseventer here's some ideas:

- Put the `timeout` and `Promise.race` in `canAutoplay`. The timeout should be the same whether we're running one test or two for muted. This would allow clearing of `autoplayPagePromises[key]` on timeout.
- Return `null` or the error caught from `startPlayback`. The error could then be included in the result when there is a failure. This is sorely missing now for debugging purposes.

src/js/utils/can-autoplay.js
```diff
@@ -56,11 +56,18 @@ export const AUTOPLAY_ENABLED = 'autoplayEnabled';
 export const AUTOPLAY_MUTED = 'autoplayMuted';
 export const AUTOPLAY_DISABLED = 'autoplayDisabled';
 
+let autoplayPagePromises = {};
+
 export function canAutoplay (mediaPool, { cancelable, muted = false, allowMuted = false, timeout = 250 }) {
     const element = mediaPool.getTestElement();
+    const key = `${muted}-${allowMuted}`;
```
vseventer: We can eliminate one more test case, as there are really one three possibilities:
1. `muted` is `false` - `allowMuted` is `true`,
2. `muted` is `false` - `allowMuted` is `false`,
3. `muted` is `true` - `allowMuted` is irrelevant,

```js
const key = muted ? `${muted}` : `${muted}-${allowMuted}`;
const key = muted ? 'muted' : `${allowMuted}`; // Even shorter if we want to.
```

src/js/utils/can-autoplay.js
```diff
@@ -56,11 +56,18 @@ export const AUTOPLAY_ENABLED = 'autoplayEnabled';
...
+
+    if (autoplayPagePromises[key]) {
```
vseventer: We can tidy up the code a bit by doing:
```js
if (!autoplayPagePromises[key]) {
    autoplayPagePromises[key] = ...
}
return autoplayPagePromises[key];
```

src/js/utils/can-autoplay.js
```diff
@@ -74,8 +81,15 @@ export function canAutoplay (mediaPool, { cancelable, muted = false, allowMuted
...
-            return muted ? AUTOPLAY_MUTED : AUTOPLAY_ENABLED;
+            if (muted) {
+                autoplayPagePromises[key] = null;
+                return AUTOPLAY_MUTED;
+            }
+            return AUTOPLAY_ENABLED;
         }
+        autoplayPagePromises[key] = null;
         return AUTOPLAY_DISABLED;
     });
```
vseventer: We should add a catch handler which resets the promise.

```js
.catch((e) => {
    autoplayPagePromises[key] = null;
    throw e;
});
```

vseventer:
@robwalch Yeah, the timeout probably belongs on `canAutoplay` directly.

src/js/utils/can-autoplay.js
```diff
@@ -35,47 +35,58 @@ import Promise from '../polyfills/promise';
...
-    });
-    return Promise.race([ promise, timer ]);
+    return element.play() || createPlayPromise(element);
 }
 
 export const AUTOPLAY_ENABLED = 'autoplayEnabled';
 export const AUTOPLAY_MUTED = 'autoplayMuted';
 export const AUTOPLAY_DISABLED = 'autoplayDisabled';
 
+let autoplayPagePromises = {};
```
robwalch: const

PR #2742, opened 2018-02-26
Avoid defaulting item.starttime to 0 in the html5 provider

johnBartos:
Seeking in the idle state correct sets the video's currentTime to the seek time; however, since the provider is attached, we don't set item.startttime. When `completeLoad` fires after this idle seek, the default starttime of 0 causes us to begin playback at the start of the video. This PR removes this default so that we don't seek if starttime isn't explicitly set.

This PR only fixes non-preroll seeking. Seeking on idle with a preroll is buggy now and should either be addressed in this PR or in a future PR.

No

No

#### Addresses Issue(s):

JW8-1074


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/program/media-controller.js
```diff
@@ -65,8 +65,8 @@ export default class MediaController extends Eventable {
         }
         // The provider has a video tag, but has not started nor preloaded
         if (this.attached && !this.setup && !this.preloaded) {
-            mediaModel.set('preloaded', true);
             provider.preload(item);
+            mediaModel.set('preloaded', true);
```
robwalch: If `provider.preload` emits events, it makes sense to update the media model first so that callbacks can react and update that state.

robwalch:
Make sure this doesn't impact "repeat".

PR #2743, opened 2018-02-26
Adjust autoplay behavior suggestions

robwalch:
### This PR will...

Additional changes we should consider as part of the autoplay test and behavior changes.

- Change the test timeout to 10 seconds. Sites are slow. There is no rush.
- Revert "playOnViewable" behavior. Run the autostart test on viewable when attempting to autostart so that the test results represent page state at the time the player becomes viewable.
- Any failure in the autoplay test causes the test to catch
- Adds the actual play request to the test promise change, so if the actual play request fails we will still get an "autostartNotAllowed" event with `reason` "playAttemptFailed".
- Clear the timeout when the test ends first - this prevents the timeout callback from running when it isn't needed and stops the debugger from pausing on Promise rejection after the test is over

Builds on top of https://github.com/jwplayer/jwplayer/pull/2734 to provide error details when autostart is cancelled. This also makes the test more reliable by reducing the chance of a timeout and the chance of getting an fail result before a play-on-viewable player comes into view.

#### Addresses Issue(s):

JW8-1158
ADS-938

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -234,13 +234,6 @@ Object.assign(Controller.prototype, {
             eventsReadyQueue = null;
 
             _model.change('viewable', viewableChange);
-            _model.once('change:canAutoplay', function(model) {
-                model.once('change:autostartFailed change:autostartMuted change:mute', function() {
-                    model.off('change:viewable', _checkPlayOnViewable);
-                });
-                model.change('viewable', _checkPlayOnViewable);
-            });
-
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2744, opened 2018-02-26
Added configuration used in setup() call to model attributes

boyntoni:
### This PR will...

Adds a property, "setupConfig", to the model, and exposes it in the jwplayer().getConfig() method. "setupConfig" holds an unmodified copy of the configuration provided in the setup() call.

The addition of the unmodified configuration object assists in debugging and troubleshooting for the TS teams.

At the point in which I make a clone of the configuration, the "id" and "analytics" properties have been appended to the configuration object, regardless of whether these were explicitly defined in the setup() block.

Given that the intended use for this value is to receive the unmodified object, it is somewhat counter-intuitive, but I figured it would be less cumbersome to avoid any logic for removing these additional properties (especially given that analytics could be a user defined key/value pair). 

No.

#### Addresses Issue(s):

JW8-1223

### DIFF
--- a/src/js/api/core-shim.js
+++ b/src/js/api/core-shim.js
@@ -77,7 +77,10 @@ Object.assign(CoreShim.prototype, {
         Object.assign(this.mediaShim, INITIAL_MEDIA_STATE);
 
         // Assigning config properties to the model needs to be synchronous for chained get API methods
-        const configuration = Config(options, persisted);
+        const setupConfig = options;
+        const configuration = Config(Object.assign({}, options), persisted);
+        configuration.id = api.id;
+        configuration.setupConfig = setupConfig;
         Object.assign(model.attributes, configuration, INITIAL_PLAYER_STATE);
         model.getProviders = function() {
             return new Providers(configuration);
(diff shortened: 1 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/api/core-shim.js
```diff
@@ -78,7 +78,8 @@ Object.assign(CoreShim.prototype, {
 
         // Assigning config properties to the model needs to be synchronous for chained get API methods
         const configuration = Config(options, persisted);
-        Object.assign(model.attributes, configuration, INITIAL_PLAYER_STATE);
+        const setupConfig = { setupConfig: Object.assign({}, options) };
```
robwalch: You should do this in "Config" before `options` has properties appended to it. Then there will be no changes to core-shim.js and you can assign  `setupConfig` directly:

```js
// In config.js after let config = Object.assign({}, Defaults, allOptions); (should be const)
config.setupConfig = Object.assign({}, options);
```

boyntoni: @robwalch  Making the changes now.. It will still have the ID appended to it, which occurs in the setup method in api.js, as well as the analytics block, which is added before that.

robwalch:
@boyntoni I just updated your branch. It was out of date according to the status checks on this page.

PR #2745, opened 2018-02-27
Recs Accessibility

hghazzi:
### This PR will...

If the recs shelf or overlay is open when you hit escape it will close the overlay.

If the player is in fullscreen and recs is open, when you hit escape it will exit fullscreen and keep recs open. 

To add accessibility support to the recs shelf, and allow the shelf to be closed via the ESC key when you are focused on any part of the player.

https://github.com/jwplayer/jwplayer-plugin-related/pull/260

#### Addresses Issue(s):

JW8-1068

### DIFF
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -213,12 +213,15 @@ export default class Controls {
         }
 
         function onEscape() {
-            const related = api.getPlugin('related');
             if (model.get('fullscreen')) {
                 api.setFullscreen(false);
                 this.playerContainer.blur();
                 this.userInactive();
-            } else if (related) {
+                return;
+            }
+
+            const related = api.getPlugin('related');
+            if (related) {
                 related.close({ type: 'escape' });
             }
         }
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -212,6 +212,17 @@ export default class Controls {
             api.setVolume(newVol);
         }
 
+        function onEscape() {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/controls.js
```diff
@@ -199,6 +199,17 @@ export default class Controls {
             api.setVolume(newVol);
         }
 
+        function onEscape() {
+            const related = api.getPlugin('related');
+            if (model.get('fullscreen')) {
+                api.setFullscreen(false);
+                this.playerContainer.blur();
+                this.userInactive();
+            } else if (related) {
```
egreaves: I prefer not having to create the `const` if we don't need to go down this code path. @karimJWP had a good suggestion of returning early, in which case we wouldn't need an else.

```
if() {
...
   return;
}
const related = api.getPlugin('related');
if (related) {   
   related.close({ type: 'escape' });
}

```

egreaves: I prefer not having to create the `const` if we don't need to go down this code path. @karimJWP had a good suggestion of returning early, in which case we wouldn't need an else.

```
if() {
   ...
   return;
}
const related = api.getPlugin('related');
if (related) {   
   related.close({ type: 'escape' });
}

```

Issue #2696, opened 2018-01-22
Parse UC browser; add to environment 


johnBartos:
The UC browser comes in as Chrome. We should parse it so that we're able to know usage stats & fix specific bugs

waxidiotic:
@johnBartos I can do this since I recently did the same for Windows Phones.

robwalch:
Leaving this open until `Browsers.UC` is added to environment.

waxidiotic:
Done. Will be in 8.2.0.

PR #2746, opened 2018-02-27
Stopped UCBrowser as being detected as Chrome

waxidiotic:
JW8-1256

### This PR will...

Exclude browsers with user-agent strings that include `UCBrowser` from being interpreted as Chrome in `getEnvironment()`.

This will allow publishers to work around features that are not supported in UC Browser but are in Chrome. It also should improve the chance that returning `true` for Chrome would be correct.

n/a

n/a

#### Addresses Issue(s):

JW8-1256
GitHub - #2696

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -32,7 +32,7 @@ export function isMSIE() {
 }
 
 export function isChrome() {
-    return userAgentMatch(/\s(?:Chrome|CriOS)\//i) && !isEdge();
+    return userAgentMatch(/\s(?:Chrome|CriOS)\//i) && !isEdge() && !userAgentMatch(/UCBrowser/i);
 }
 
 export function isIE() {


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

waxidiotic:
Awaiting on a fix version from Alicia in JIRA, then I'll add it here. Also don't think a test is needed but let me know otherwise.

waxidiotic:
@dannyfinks confirmed the one automated failure is a known issue that @msiddique is fixing.

johnBartos:
@waxidiotic You can add a unit test to this suite: https://github.com/jwplayer/jwplayer/blob/master/test/unit/browser-version-test.js

The test(s) can verify that that we're not returning UC as chrome.

PR #2748, opened 2018-02-28
Improve mouse handling that shows and hides controls

robwalch:
### This PR will...

- Fix mouse out handling not hiding controls when mousing out from the bottom of the player (controlbar)
- Remove dead code for adding logo to controls (`controls.right` is always `null`)
- Use only two event listeners "mousemove" and "mouseout" on the player container
  - Removes "ActiveListeners" code that added event listeners to controlbar, logo and nextup
  - The out handler is only used when the mouse leaves the player container
  - The move handler doesn't create a timeout when the mouse is over controlbar, logo or nextup
  - Don't use `ClickHandler` or `UI` for these listeners - It adds overhead making it difficult to work directly with input events
- Simplify `controls.userActive` interface
  - Just pass in a timeout of `0` to prevent a user active timeout from being created 
- Prevent new timeouts from being started for every call to `controls.userActive` - since this is called on "mousemove", there are a lot of timeouts cleared and set that we can avoid.
- Add initial unit-tests for Controls

Fixes a regression in mouse out handling caused by removing the use of "mouseout". Without the `_playerElement.contains(event.relatedTarget)` filter on that event it fires even when mousing out of child player elements.

Reducing the number of input event handlers (ActiveListeners) helps improve UI responsiveness.

- Cross-browser testing: mobile
- Could we use a `_.debounce` pattern the `controls.userActive` timeout?

#### Addresses Issue(s):

JW8-1182

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -141,7 +141,7 @@ const UI = function (elem, options) {
     }
 
     function pointerOutHandler(evt) {
-        if (evt.pointerType !== 'touch') {
+        if (evt.pointerType !== 'touch' && 'x' in evt) {
             // elementFromPoint to handle an issue where setPointerCapture is causing a pointerout event
             const { x, y } = evt;
             const overElement = document.elementFromPoint(x, y);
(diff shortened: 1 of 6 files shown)


robwalch:
@dannyfinks I'm unable to reproduce this test failure in IE11 on browserstack https://jenkins.jwplayer.com/view/JW7/job/jw7-commercial-cucumber/11072/allure/#suites/b9cb8385c37c0d6570a24dbf08ce345d/a84296d2f9a3c6db/
The only thing that I could think of that would prevent controls from hiding is if you are forcing the controlbar to show, or the mouse is over the controlbar or logo.

PR #2752, opened 2018-03-01
IE11 button and tabbing fixes

robwalch:
### This PR will...

- [JW8-1225] Handle blur events missing relatedTarget property
- [JW8-899] Prevent SVG icons from getting focus in IE11
- Make each UI instance use it's own interactionEnd delegate for document "mouseup" listeners

IE11 does not include the `relatedTarget` property on "blur" events. This caused controls to be hidden the first time the player controls are clicked (JW8-1225: Volume Icon needs 3 clicks to mute the volume).

IE11 required two presses of the TAB key in order to tab from one button to the next. This is because SVG docs (our button icons) receive tab focus by default in IE11. Adding `focusable="false"` to each doc prevents this. 

The UI module may handle mouse events for elements that overlap. Since all mouseup listeners attached to document used the same callback, one instance could remove the listener for another instance. Creating delegates will prevent this, ensuring that all elements receive their interactions.

#### Addresses Issue(s):

JW8-1225
JW8-899


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2753, opened 2018-03-01
Bugfix: Mobile controls toggle regression

robwalch:
### This PR will...

Fix a regressions with toggling controls on tap caused by https://github.com/jwplayer/jwplayer/pull/2748

- Create a single delegate for `userInactiveTimeout` for each Controls instance
- Remove "mousemove" listener when "tap" events are received

Having a single delegate for `userInactiveTimeout` allows `userInactive` to always hide controls immediately when called, like it did before, without interference from debounced timeout logic.

"mousemove" events trigger user activity, which we do not want on devices with tap input. Removing the "mousemove" listener on "tap" allows us to support both modes of controls toggling without performing a user-agent check or using UI.js to forward "move" events.

https://github.com/jwplayer/jwplayer/pull/2752/files

#### Addresses Issue(s):

JW8-1182

### DIFF
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -60,6 +60,15 @@ export default class Controls {
         this.logo = null;
         this.div = null;
         this.dimensions = {};
+        this.userInactiveTimeout = () => {
+            // Rerun at the scheduled time if remaining time is greater than the display refresh rate
+            const remainingTime = this.inactiveTime - now();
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2755, opened 2018-03-02
[JW8-1288] Reset autostart cancelable on pause.

vseventer:
### This PR will...
Reset the autostart cancelable on pause.

~~On mobile, if the element is not viewable (i.e. added to the DOM only after ready), we currently autopause the video, which cancels the autostart check. Once the element becomes viewable (i.e. is added to the DOM), however, we now want to perform the canAutoplay test - but because the autostart was cancelled earlier, this will fail. By resetting the autostart cancelable, the test always uses the latest cancelable, which will work as expected.~~

Resets cancelable before every test instead instead of on pause. In addition, moved the `autostartMuted` change listener to after we set it for the first time. Finally, once a user clicks-to-play, we never autostart/pause on viewable change (similar to click-to-pause behavior).

~~Alternatively, I looked into only calling _pause is the element is not currently idle - however this is a bit more complicated as there are other non-playing states (e.g. buffering / error). This seems to be the cleaner solution.~~

#### Addresses Issue(s):
JW8-1288

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -486,6 +486,7 @@ Object.assign(Controller.prototype, {
         function _pause(meta) {
             _actionOnAttach = null;
             checkAutoStartCancelable.cancel();
+            checkAutoStartCancelable = cancelable(_checkAutoStart);
 
             const pauseReason = _getReason(meta);
             _model.set('pauseReason', pauseReason);
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -235,7 +235,7 @@ Object.assign(Controller.prototype, {
 
             _model.change('viewable', viewableChange);
             _model.change('viewable', _checkPlayOnViewable);
-            _model.once('change:autostartFailed change:autostartMuted change:mute', function(model) {
+            _model.once('change:autostartFailed change:mute', function(model) {
                 model.off('change:viewable', _checkPlayOnViewable);
             });
 
@@ -376,6 +376,10 @@ Object.assign(Controller.prototype, {
 
             const playReason = _getReason(meta);
             _model.set('playReason', playReason);
+            // Stop autoplay behavior if the video is started by the user or an api call
+            if (playReason === 'interaction' || playReason === 'external') {
+                _model.set('playOnViewable', false);
+            }
 
             const adState = _getAdState();
             if (_.isString(adState)) {
@@ -422,6 +426,9 @@ Object.assign(Controller.prototype, {
                 return;
             }
 
+            // Reset cancelable for new autoplay test below.
+            checkAutoStartCancelable = cancelable(_checkAutoStart);
+
             // Detect and store browser autoplay setting in the model.
             const adConfig = _model.get('advertising');
             canAutoplay(mediaPool, {
@@ -435,6 +442,10 @@ Object.assign(Controller.prototype, {
                 if (result === AUTOPLAY_MUTED && !_this.getMute()) {
                     _model.set('autostartMuted', true);
                     updateProgramSoundSettings();
+
+                    _model.once('change:autostartMuted', function(model) {
+                        model.off('change:viewable', _checkPlayOnViewable);
+                    });
                 }
 
                 return _play({ reason: 'autostart' }).catch(() => {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -486,6 +486,7 @@ Object.assign(Controller.prototype, {
         function _pause(meta) {
             _actionOnAttach = null;
             checkAutoStartCancelable.cancel();
+            checkAutoStartCancelable = cancelable(_checkAutoStart);
```
robwalch: @vseventer I notice we're also doing this on `_load()`, but not not when cancelling on `_play()` and `_stop()`.

The reason we reset the cancelable on load is because loading a new playlist results in an autostart check. play, stop and even pause should cancel autostart behavior.

If you pause while the player is attempting to start, it shouldn't start, or at least it should start and then pause. I'm not sure that's what will happen with this change.

vseventer:
@robwalch Fair enough, we can move it to `_checkPlayOnViewable`, but keep in mind we still need to reset the cancelable to allow for playback to pause -> resume -> pause -> resume etc. when scrolled in- and out of view without interaction.

src/js/controller/controller.js
egreaves: @robwalch would it suffice to check the pause reason here to see if it's from a viewability change and not a user interaction or api call to `pause()`? This may be simpler than changing `_checkPlayOnViewable`.
```
if (pauseReason === 'autostart') {
    checkAutoStartCancelable = cancelable(_checkAutoStart);
}
```

vseventer:
@robwalch @egreaves I'm revisiting this ticket altogether - there are some issues with viewable mainly on mobile, mostly caused by the cancelable and autostartMuted change listener which fires prematurely since the autoplay test is now async.

src/js/controller/controller.js
robwalch: I don't think so. We should look at the timing of event listeners. There is not reason to pause on "viewable" until playback has started, so splitting up the "viewable" event handler to one that autostarts, and another that pauses once playback has started is the right move.

robwalch:
@vseventer (+ @egreaves ) Have you already started on this? I think I have a good solution.

The essence of the problem is that pause is called while we're trying to autostart (or test capabilities). 
```
function _playerReadyNotify() {
...
...
        function _checkPlayOnViewable(model, viewable) {
                if (viewable) {
                    _autoStart();
                } else if (OS.mobile) {
                }
            }
        }
```

We always want this pause behavior for mobile autostart so we just need to split this logic out:
function _playerReadyNotify() {
...
...
        function _checkPlayOnViewable(model, viewable) {
                if (viewable) {
                    _autoStart();
            }
        }
```

Then in `_autoStart` after the test if we're `OS.mobile` implement play/pause toggling on viewable, until or unless the state of these has changed:`_model.once('change:autostartFailed change:autostartMuted change:mute', function(model) {`.

**Comment by vseventer** [comment]
Updated PR to reset cancelable before every test instead of on pause. In addition, moved the `autostartMuted` change listener to after we set it for the first time. Finally, once a user clicks-to-play, we never autostart/pause on viewable change (similar to click-to-pause behavior).

File: src/js/controller/controller.js [hunk 2]
```diff
@@ -376,6 +376,10 @@ Object.assign(Controller.prototype, {
 
             const playReason = _getReason(meta);
             _model.set('playReason', playReason);
+            // Stop autoplay behavior if the video is started by the user or an api call
+            if (playReason === 'interaction' || playReason === 'external') {
+                _model.set('playOnViewable', false);
+            }
```
robwalch: This is fine for now, but we need to make a tech-debt note that model properties shouldn't span view-controls code and player-model-controller. Everything "autostart" and "playOnViewable" is split up between controls and controller instead of being encapsulated in one place, like a module that listens to model changes ("change:playReason" where reason === "interaction").

src/js/controller/controller.js
```diff
@@ -435,6 +442,10 @@ Object.assign(Controller.prototype, {
                 if (result === AUTOPLAY_MUTED && !_this.getMute()) {
                     _model.set('autostartMuted', true);
                     updateProgramSoundSettings();
+
+                    _model.once('change:autostartMuted', function(model) {
```
robwalch: Better to do this right after the `set`

PR #2757, opened 2018-03-02
[JW8-1290] Display autostart unmute button only on touch.

vseventer:
### This PR will...
Hide the autostart unmute button on desktop and only show if the player has `jw-flag-touch`.

The autostart unmute button cannot be interacted with on desktop, as hovering it will display the full controlbar instead.

Alternatively, we can move the `OS.mobile` check to inside `setupUnmuteAutoplayButton` in `controls.js` to not render the button in the first place. But, looking at how we general do things, we simply hide the button using CSS.

#### Addresses Issue(s):
JW8-1290

### DIFF
--- a/src/css/controls/flags/user-inactive.less
+++ b/src/css/controls/flags/user-inactive.less
@@ -40,6 +40,6 @@
     }
 }
 
-.jw-state-playing.jw-flag-user-inactive .jw-autostart-mute {
+.jw-state-playing.jw-flag-touch.jw-flag-user-inactive .jw-autostart-mute {
     display: flex;
 }


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

vseventer:
Note - the change in specificity now causes the `autostart-mute` to be rendered on top of the controls for audio-only - working on a fix.

vseventer:
See #2761.

PR #2760, opened 2018-03-03
Disable Airplay for Playlists with Dash

hghazzi:
### This PR will...

Expose the html5 supports methods so it can be accessible by the commercial controller

Since airplay is not supported by shaka, attempts to play a dash stream when airplay is enabled will cause the player to fail and produce video error code 4 `'File could not be played'`. 

`supportsType` is used to check if the playlist has any unsupported items, which will trigger airplay to be disabled.

https://github.com/jwplayer/jwplayer-commercial/pull/4835

#### Addresses Issue(s):

JW8-1282

### DIFF
--- a/src/js/providers/providers-supported.js
+++ b/src/js/providers/providers-supported.js
@@ -25,35 +25,39 @@ const MimeTypes = {
 const SupportsMatrix = [
     {
         name: 'html5',
-        supports: function (source) {
-            if (isAndroidHls(source) === false) {
-                return false;
-            }
-
-            if (!video.canPlayType) {
-                return false;
-            }
-
-            const file = source.file;
-            const type = source.type;
-
-            // Ensure RTMP files are not seen as videos
-            if (isRtmp(file, type)) {
-                return false;
-            }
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/providers/providers-supported.js
```diff
@@ -25,35 +25,39 @@ const MimeTypes = {
...
-                return false;
-            }
+        supports: supportsType
+    }
+];
 
-            if (!video.canPlayType) {
-                return false;
-            }
+export function supportsType(source) {
```
egreaves: Is `supportsType` used anywhere else? Wondering why it's being exported.

hghazzi:
It was a recommendation by to was the best way to verify what can be played
by video tag.

PR #2761, opened 2018-03-05
[JW8-1290] Display autostart unmute button only on touch.

vseventer:
### This PR will...
Hide the autostart unmute button on desktop and only show if the player has `jw-flag-touch`.

The autostart unmute button cannot be interacted with on desktop, as hovering it will display the full controlbar instead.

Alternatively, we can move the `OS.mobile` check to inside `setupUnmuteAutoplayButton` in `controls.js` to not render the button in the first place. But, looking at how we general do things, we simply hide the button using CSS.

#### Addresses Issue(s):
JW8-1290

### DIFF
--- a/src/css/controls/flags/ads.less
+++ b/src/css/controls/flags/ads.less
@@ -5,7 +5,6 @@
     .jw-logo,
     .jw-captions.jw-captions-enabled,
     .jw-nextup-container,
-    .jw-autostart-mute,
     .jw-text-duration,
     .jw-text-elapsed {
         display: none;
--- a/src/css/controls/flags/media-audio.less
+++ b/src/css/controls/flags/media-audio.less
@@ -1,10 +1,3 @@
-.jwplayer.jw-flag-media-audio {
-    // Hide autostart mute button
-    .jw-autostart-mute {
-        display: none;
-    }
-}
-
 .jw-flag-media-audio {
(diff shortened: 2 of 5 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

egreaves:
@vseventer I actually prefer the idea of using JS to setup the mute button only on mobile devices. The CSS solution would be better if it was likely that we'd need to toggle between mobile/desktop views, but that's not the case.

vseventer:
@egreaves Although I agree, I want to make sure this is what we want to do, since nowhere else in the code we seem to conditionally render elements based on OS, it's all done through CSS. /cc @robwalch @DanFerrer

robwalch:
I'm fine with this. It's not UI that pubs should customize. We are creating this element conditionally, so there is no reason to add it if it will never be shown. Pubs can style the controlbar.

PR #2762, opened 2018-03-06
Background load the next playlist item

johnBartos:
### This PR will...
- Background load the next item 2 seconds before the end of a video
- Update unit tests
- Refactors code

To reduce or eliminate the buffering time between subsequent playlist items.

This solution (with the `adPlaying`) flag is necessary to avoid a race condition. The race condition appears when placing background loading item into the foreground; this is an async operation, and if the promise resolves while an ad is playing, that video is placed into the DOM. This flag ensures that if we have entered ad playback during this operation we instead place the background loading video into the `loadedMedia` property, which contains media frozen during ad playback. When the ad is over `restoreBackgroundMedia` is called, which places whatever `loadedMedia` is into the background.

### Is there anything left to do?
- Remove buffering events between item switches when the next item is sufficiently buffered

I had a commit for this but the removal of buffering makes the `IDLE` state appear between items. This is worse.

- Apply race condition resolution to the "regular" case of loading a new media controller

This is also an async operation, and while much harder to reproduce, is theoretically possible. 

Are there any gaps in unit tests which would be worth filling?

https://github.com/jwplayer/jwplayer-commercial/pull/4850

#### Addresses Issue(s):

JW8-30


johnBartos:
@robwalch Addressed

johnBartos:
test this please

src/js/controller/controller.js
```diff
@@ -162,6 +163,14 @@ Object.assign(Controller.prototype, {
                 const type = streamType(duration, minDvrWindow);
                 model.setStreamType(type);
             });
+            if (Features.backgroundLoading) {
+                mediaModel.on('change:position', function (changedMediaModel, position) {
+                    if (!_programController.backgroundLoading
```
robwalch: You could remove the listener `mediaModel.off('change:position ,..` instead of checking `_programController.backgroundLoading`.

src/js/program/background-media.js
```diff
@@ -0,0 +1,42 @@
+/**
+ * A simple data structure for containing both of the background loading objects.
+ * loadedMedia is the currently active item which has been put into the background during ad playback.
+ * loadingMedia is an item which is preloading in the background which may be selected in the future. It is usually the
```
robwalch: Wouldn't `nextItem`, `nextPromise` and `currentMedia` be clearer? I feel like I'm going to be doing a lot of double-takes outside of this file with loaded/loading and Media where it's a mediaController vs a Promise.

src/js/program/background-media.js
```diff
@@ -0,0 +1,42 @@
...
+ * next item in the playlist, or the next up item in a recommendations feed.
+ * @returns {BackgroundMedia}
+ */
+/**
+ * @typedef {Object} BackgroundMedia
+ * @property {MediaController} loadingItem - The mediaController which has been placed into the background during playback.
+ * @property {Object} loadingItem - The playlist item loading in the background.
```
robwalch: {Item} https://github.com/jwplayer/jwplayer/blob/v8.2.0-alpha.1/src/js/playlist/item.js

robwalch:
I'm seeing a period of idle state between playlist item 1 and 2 changing providers http://player-pr-test-jenkins.longtailvideo.com/builds/13495/archive/test/squash/bootstrap.html?primary=html5&edition=ads&feature=playlist/playlists_mixed_providers&suite=Playlist

What's odd is that when profiling it shows that we are loading both providers for items 2 and 3 during this transition:
screen shot 2018-03-07 at 9 37 02 am

That might not be new, but it would be better to only load the next provider if required.

src/js/program/program-controller.js
```diff
@@ -163,8 +171,8 @@ class ProgramController extends Eventable {
      * @returns {void}
      */
     preloadVideo() {
-        const { backgroundMedia, mediaController } = this;
-        const media = mediaController || backgroundMedia;
+        const { background: { currentMedia: backgroundMedia }, mediaController } = this;
+        const media = mediaController || backgroundMedia.activeMedia;
```
robwalch: Can you comment on why we preload the background's  currentMedia's backgroundMedia's activeMedia?

And why wouldn't currentMedia  not === mediaController?

src/js/program/program-controller.js
```diff
@@ -229,18 +234,19 @@ class ProgramController extends Eventable {
      * @returns {void}
      */
     backgroundActiveMedia() {
-        const { backgroundMedia, mediaController } = this;
+        this.adPlaying = true;
```
robwalch: Doesn't this.adPlaying === mediaController.background ?

src/js/program/program-controller.js
```diff
@@ -251,18 +257,47 @@ class ProgramController extends Eventable {
      * @returns {void}
      */
     restoreBackgroundMedia() {
-        const { backgroundMedia, mediaController } = this;
+        this.adPlaying = false;
+        const { background, background: { currentMedia }, mediaController } = this;
```
robwalch: Still confused about the difference between mediaController and currentMedia.

src/js/program/background-media.js
```diff
@@ -0,0 +1,42 @@
...
+
+    return Object.defineProperties(Object.create(null), {
+        nextItem: {
+            get() {
+                return nextMedia ? nextMedia.item : null;
+            }
+        },
+        nextMedia: {
+            get() {
+                return nextMedia ? nextMedia.loadPromise : null;
```
robwalch: Ah this is very confusing. You set `nextMedia` to `{ item, loadPromise }` but get them separately. If you want to only do one set, I would make another setter with a different name ex: `set next`.

src/js/program/program-controller.js
johnBartos: mediaController.background is set too late. This flag needs to be set first thing

src/js/program/program-controller.js
johnBartos: Oops, a bit of a refactoring typo

johnBartos:
test this please

johnBartos:
test this please

src/js/program/program-controller.js
```diff
@@ -321,48 +351,62 @@ class ProgramController extends Eventable {
...
         const { model, providerController } = this;
+        const makeMediaController = ProviderConstructor => new MediaController(
+            new ProviderConstructor(model.get('id'), model.getConfiguration(), this.primedElement),
+            model
+        );
 
         let ProviderConstructor = providerController.choose(source);
         if (ProviderConstructor) {
-            return Promise.resolve(ProviderConstructor);
+            return Promise.resolve(makeMediaController((ProviderConstructor)));
```
robwalch: `import resolved`

src/js/program/program-controller.js
```diff
@@ -321,48 +351,62 @@ class ProgramController extends Eventable {
...
         }
 
         return providerController.loadProviders(model.get('playlist'))
```
robwalch: We should consider changing `loadProviders` to only load the provider needed for `source`.

robwalch: I would be great to change or replace this method with one that accepts `source` and async returns `ProviderConstructor `.

src/js/program/program-controller.js
```diff
@@ -41,55 +43,58 @@ class ProgramController extends Eventable {
...
             const casting = model.get('castActive');
-            // Buffer between item switches, but remain in the initial state (IDLE) while loading the first provider
-            model.set(PLAYER_STATE, STATE_BUFFERING);
             if (casting || this.providerController.canPlay(mediaController.provider, source)) {
-                this.providerPromise = Promise.resolve(mediaController);
-                // We can reuse the current mediaController and do so synchronously
-                // Initialize the provider and mediaModel, sync it with the Model
-                // This sets up the mediaController and allows playback to begin
+                // We can synchronously reuse the current mediaController
+                this.loadPromise = Promise.resolve(mediaController);
```
robwalch: resolved()

src/js/program/program-controller.js
```diff
@@ -119,13 +124,16 @@ class ProgramController extends Eventable {
             // Make the subsequent promise cancelable so that we can avoid playback when no longer wanted
             const thenPlayPromise = cancelable((nextMediaController) => {
                 // Ensure that we haven't switched items while waiting for the provider to load
+                if (nextMediaController.background) {
+                    this._setActiveMedia(nextMediaController);
+                }
```
robwalch: ????

johnBartos:
test this please

PR #2765, opened 2018-03-06
Fix pause event after restoring background loaded media 

johnBartos:
### This PR will...
Set the background mediaModel's mediaState to buffering before restoring it to the foreground

To prevent a pause event from firing when foregrounding. Before the mediaController enters the background, we pause it (and this it's mediaState is `paused`). On restoration we sync the mediaModel's `mediaState` with the playerModel's; this results in a second `pause` event firing. By setting it to buffering we avoid this and also truthfully reflect the player's state (restoring to the foreground is considered buffering). If the restoration is quick enough no buffering wheel will show at all.

No

No

#### Addresses Issue(s):

JW8-1237

### DIFF
--- a/src/js/program/program-controller.js
+++ b/src/js/program/program-controller.js
@@ -259,6 +259,10 @@ class ProgramController extends Eventable {
             return;
         }
         if (backgroundMedia) {
+            // Set the state to buffering before attaching so that we don't resume in the "paused" state
+            // If the mediaController enters the foreground quickly enough (within one animation frame), no buffering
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2772, opened 2018-03-12
Call controller.play in _seek so that it is queued until playback begins

johnBartos:
So that if we call play during the seek call, that play call is queued in the `ApiQueue`. The reason it was not queued before is that only the prototype `this.play` method is captured by the queueing function.

No

No

#### Addresses Issue(s):

JW8-1302 JW8-1303

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -553,7 +553,7 @@ Object.assign(Controller.prototype, {
             }
             _programController.position = pos;
             if (!_model.get('scrubbing') && _model.get('state') !== STATE_PLAYING) {
-                _play(meta);
+                this.play(meta);
             }
         }
 


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -553,7 +553,7 @@ Object.assign(Controller.prototype, {
             }
             _programController.position = pos;
             if (!_model.get('scrubbing') && _model.get('state') !== STATE_PLAYING) {
-                _play(meta);
```
robwalch: Would `_this.play` be safer?

robwalch: Looks like _seek is only referenced once so we're ok.

PR #2775, opened 2018-03-12
Fix background load exceptions crashing playback, background loading of bad nextUp items

johnBartos:
### This PR will...
- Prevent BGL from destroying active playback on error
- Format `nextUp` so that it's a valid `PlaylistItem`

So that background loading does not affect active playback; and that when we're background loading recommendations, we're only working with valid items. The controller has currently been setting `nextUp` to whatever JSON came out of the feed, which is not always valid. This item was validated when loaded so it has caused no problems so far, but we need it to be valid to background load.

No

No

#### Addresses Issue(s):

JW8-1342

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -3,7 +3,7 @@ import { showView } from 'api/core-shim';
 import setConfig from 'api/set-config';
 import ApiQueueDecorator from 'api/api-queue';
 import PlaylistLoader from 'playlist/loader';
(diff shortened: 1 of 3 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2777, opened 2018-03-13
Listen to mediaType change on view.init()

johnBartos:
## Why is this Pull Request needed?
`view.setup` attaches the `change:mediaType` listener before the `mediaModel` is attached to the `viewModel`. This causes the listener to miss the `mediaType` event, and the event stays missed because we do not dispatch `mediaModel` diffs when setting the first `mediaModel`. This change moves the `mediaType` listener to `view.setup`, which attaches after the mediaModel is attached, causing the `change` handler to get (and set) the correct value.

No

No

#### Addresses Issue(s):

JW8-1321 JW8-1322

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -766,7 +766,7 @@ Object.assign(Controller.prototype, {
                 }, _this)
                 .on(MEDIA_ERROR, _this.triggerError, _this)
                 .on(MEDIA_TYPE, (event) => {
-                    // Re-trigger mediaType
+                    // Re-set mediaType so that it triggers change listeners even if queued
                     _model.set('mediaType', event.mediaType);
                 });
         }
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -764,7 +764,11 @@ Object.assign(Controller.prototype, {
                     // Insert a small delay here so that other complete handlers can execute
                     resolved.then(_completeHandler);
                 }, _this)
-                .on(MEDIA_ERROR, _this.triggerError, _this);
+                .on(MEDIA_ERROR, _this.triggerError, _this)
+                .on(MEDIA_TYPE, (event) => {
+                    // Re-set mediaType so that it triggers change listeners even if queued
+                    _model.set('mediaType', event.mediaType);
```
robwalch: This adds "mediaType" back to the model which we moved to mediaModel intentionally.

PR #2778, opened 2018-03-13
Add (is)Object checks to related "nextUp" event and providers.choose

robwalch:
Prevent the player from throwing `Cannot read property 'indexOf' of undefined` on the last related item.

When there are no more recs, related dispatches "nextUp" with an argument of `null`. This change prevents us from setting "nextUp" on the model to an empty playlist item when that happens. It also adds a guard around running `providerSupports` checks on `source` arguments that aren't Objects.

#### Addresses Issue(s):

JW8-1328

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -199,9 +199,12 @@ Object.assign(Controller.prototype, {
             const related = _api.getPlugin('related');
             if (related) {
                 related.on('nextUp', (nextUp) => {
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2780, opened 2018-03-13
Set posterImage on playlistItem change

johnBartos:
So that the posterImage is set for an item loaded via the `load` API. Since we were always loading the posterImage before (even when `autostart: true`), this change does not negatively impact performance.

No

No

#### Addresses Issue(s):
JW8-1322

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -297,7 +297,6 @@ function View(_api, _model) {
         playerViewModel.change('controls', changeControls);
         playerViewModel.change('streamType', _setLiveMode);
         _model.change('mediaType', _onMediaTypeChange);
-        // Set the title attribute of the video tag to display background media information on mobile devices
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/view.js
```diff
@@ -634,7 +618,7 @@ function View(_api, _model) {
 
         // Set the poster image for each audio file encountered in a playlist
         if (isAudioFile) {
-            setPosterImage(model);
```
robwalch: I don't think we need this anymore.

PR #2782, opened 2018-03-14
Avoid background loading DAI items

johnBartos:
DAI items are dynamically replaced before being loaded, so it does not make sense to background load these items. It also fixes a bug where DAI was not finding the video tag it needed, resulting in the ad not playing.

No

No

#### Addresses Issue(s):

 JW8-1333

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -165,10 +165,12 @@ Object.assign(Controller.prototype, {
                 model.setStreamType(type);
             });
 
-            const recsAuto = (model.get('related') || {}).oncomplete === 'autoplay';
             const index = model.get('item') + 1;
+            const recsAuto = (model.get('related') || {}).oncomplete === 'autoplay';
             let item = model.get('playlist')[index];
-            if ((item || recsAuto) && Features.backgroundLoading) {
+            // Do not background load DAI items because that item will be dynamically replaced
+            const isValidItem = (item && !item.daiSetting) || recsAuto;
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -165,10 +165,12 @@ Object.assign(Controller.prototype, {
                 model.setStreamType(type);
             });
 
-            const recsAuto = (model.get('related') || {}).oncomplete === 'autoplay';
             const index = model.get('item') + 1;
+            const recsAuto = (model.get('related') || {}).oncomplete === 'autoplay';
             let item = model.get('playlist')[index];
-            if ((item || recsAuto) && Features.backgroundLoading) {
+            // Do not background load DAI items because that item will be dynamically replaced
+            const isValidItem = (item && !item.daiSetting) || recsAuto;
```
robwalch: Since `item` can be set on position, this should be done in the event handler, in the event that "nextUp" has `daiSetting`.

johnBartos: I assumed recs are not configured to play with DAI

robwalch: They are not in general, but there is no reason why they could not be.

PR #2783, opened 2018-03-14
Prime on beforePlay even if we will not be interrupting for an ad

johnBartos:
So that if we background load the next item when a gesture is required, it will be allowed to play. This code currently does not prime media elements on the first play gesture, which is necessary to background load (and play) the next item. Non-BGL playlists are working without further gesture now because we are reusing the same tag for content (whose play method is called within the user gesture), and the modified code already works for ads.

No

No

#### Addresses Issue(s):

JW8-1337

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -424,12 +424,12 @@ Object.assign(Controller.prototype, {
                     startTime
                 });
                 _beforePlay = false;
+                if (_inInteraction(window.event)) {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2786, opened 2018-03-15
Shelf crashes when controls are disabled

karimJWP:
### This PR will...
trigger a `controls` event in the view instead of the controller, and include the loadPromise in the controls event if controls module has not been loaded.
When the loadPromise fulfills, it returns the model's current controls state.
When the loadPromise is rejected, it returns an error.
When controls are disabled, an exception in the related plugin occurs when the shelf is added. In order to properly add the shelf to the controls, it is not enough to be notified that the model's controls value is changed because the controls module can be loaded asynchronously. By passing the loadPromise, we can handle the change in controls when the promise settles. 
[Related](https://github.com/jwplayer/jwplayer-plugin-related/pull/272)
#### Addresses Issue(s):

JW8-1335


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/view.js
```diff
@@ -320,13 +320,21 @@ function View(_api, _model) {
...
+                return;
             }
+            addControls();
         } else {
             _this.removeControls();
         }
+        _this.trigger(CONTROLS, {
+            controls: enable
+        });
     }
```
robwalch: I'm fine with triggering controls from the view, but instead of changing this method, I would prefer if you triggered the event at the end of the method. It should always be triggered regardless of whether controls are ready. You can add the promise to the event if we're dealing with async behavior.

```js
    function changeControls(model, enable) {
        const controlsEvent = {
            controls: enable
       };
        if (enable) {
            ControlsModule = ControlsLoader.module.controls;
            if (!ControlsModule) {
                controlsEvent.loadPromise = ControlsLoader.load()
                    .then(function (Controls) {
                        ControlsModule = Controls;
                        // Check that controls is still true after the loader promise resolves
                        const enabledState = model.get('controls');
                        if (enabledState) {
                            addControls();
                        }
                        return enabledState;
                    })
                    .catch(function (reason) {
                        _this.trigger(ERROR, {
                            message: 'Controls failed to load',
                            reason: reason
                        });
                        return reason;
                    });
            } else {
                addControls();
            }
        } else {
            _this.removeControls();
        }
        _this.trigger(CONTROLS, controlsEvent);
    }
```

karimJWP:
test this please

karimJWP:
test this please

src/js/view/view.js
```diff
@@ -304,29 +304,37 @@ function View(_api, _model) {
...
                     .then(function (Controls) {
                         ControlsModule = Controls;
                         // Check that controls is still true after the loader promise resolves
-                        if (model.get('controls')) {
+                        const enabledState = model.get('controls');
+                        if (enabledState) {
                             addControls();
                         }
+                        return enabledState;
                     })
```
robwalch: Let's break down the promise so you can catch in related:
```js
controlsEvent.loadPromise = ControlsLoader.load().then(...);
controlsEvent.loadPromise.catch(...);
```

PR #2787, opened 2018-03-15
Add a flag to beforePlay media pool priming so that we do not destroy BGL buffer on subsequent plays

johnBartos:
Fixes a regression introduced by  https://github.com/jwplayer/jwplayer/pull/2783

JW8-1340

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -51,7 +51,7 @@ Object.assign(Controller.prototype, {
         let _interruptPlay;
         let checkAutoStartCancelable = cancelable(_checkAutoStart);
         let updatePlaylistCancelable = cancelable(function() {});
-        let hasPrimedBeforePlay = false;
+        let primeBeforePlay = true;
 
         _this.originalContainer = _this.currentContainer = originalContainer;
         _this._events = eventListeners;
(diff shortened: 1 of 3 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -51,6 +51,7 @@ Object.assign(Controller.prototype, {
         let _interruptPlay;
         let checkAutoStartCancelable = cancelable(_checkAutoStart);
         let updatePlaylistCancelable = cancelable(function() {});
+        let hasPrimedBeforePlay = false;
```
robwalch: What do you think about flipping this so the flag is read as a command (and we don't have to negate it in the check)?
`let primeBeforePlay = true;`

src/js/controller/controller.js
```diff
@@ -428,9 +429,12 @@ Object.assign(Controller.prototype, {
                     startTime
                 });
                 _beforePlay = false;
-                if (_inInteraction(window.event)) {
+
+                if (_inInteraction(window.event) && !hasPrimedBeforePlay) {
                     _programController.primeMediaElements();
+                    hasPrimedBeforePlay = true;
```
robwalch: We should also update the value when playback starts outside of this check, because if play worked without an interaction, we don't want to prime again when toggles playback.

src/js/controller/controller.js
```diff
@@ -428,17 +429,25 @@ Object.assign(Controller.prototype, {
                     startTime
                 });
                 _beforePlay = false;
-                if (_inInteraction(window.event)) {
+
+                if (_inInteraction(window.event) && primeBeforePlay) {
                     _programController.primeMediaElements();
+                    primeBeforePlay = true;
```
robwalch: `primeBeforePlay = false`

PR #2788, opened 2018-03-15
[ADS-1050] Set state to paused after ad.

vseventer:
### This PR will...
Force the model state to `STATE_PAUSED` after an ad completes - this will cause the player to reliably trigger the "play" event after midrolls (previously, it would only trigger this when BGL is enabled, or on VAST). IMA would not fire this event. 

#### Addresses Issue(s):
ADS-1050


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2789, opened 2018-03-15
Bugfix/beforeplay prime flag

johnBartos:
### This PR will...
Force media elements to prime before an ad if background loading is not supported 

If we go to play a preroll (without BGL support), the media element may not be primed. We must force it to prime before handing off the instream (as indicated by `_interruptPlay`). This is a safe operation because `src` will be replaced by the ad plugin anyway.

No

No

#### Addresses Issue(s):

JW8-1340


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2792, opened 2018-03-19
Remove "dvrSeekLimit" from Controlbar and Timeslider

DanFerrer:
### This PR will...

Remove "dvrSeekLimit" from the duration and position in the UI only.

The seek bar would factor in the full duration of the stream so it would appear you were allowed to seek further than you could, but we don't allow you to seek further than the live edge minus the seek limit

https://github.com/jwplayer/jwplayer-commercial/pull/4949

#### Addresses Issue(s):
JW8-1359


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

package-lock.json
```diff
@@ -0,0 +1,14380 @@
+{
```
robwalch: Do not add "package-lock.json" to the project. It was removed intentionally. We use yarn.lock.

src/js/providers/default.js
```diff
@@ -63,6 +63,8 @@ const DefaultProvider = {
     getCurrentAudioTrack: noop,
     setCurrentAudioTrack: noop,
 
+    getSeekRange: noop,
```
robwalch: `getSeekRange` must be implemented in html5.

robwalch: If this method returns nothing, what will happen for code that expects a range returned?

src/js/view/controls/components/timeslider.js
```diff
@@ -157,7 +141,9 @@ class TimeSlider extends Slider {
         var pct = 0;
         if (duration) {
             if (this.streamType === 'DVR') {
-                pct = (duration - position) / duration * 100;
+                var seekRange = this._model.get('seekRange');
```
robwalch: Not all providers set this property.

Use `const` not `var`.

src/js/providers/default.js
robwalch: By default this should return `{ start: 0, end: duration } ` for backwards compatibility.

robwalch: @DanFerrer @johnBartos ^^

DanFerrer:
@robwalch It was removed because its was being factored into the end of the seek range. So we substract from the seekable end of the video to make sure the viewer is behind enough

robwalch:
@DanFerrer subtract what?

DanFerrer:
test this please

src/js/providers/html5.js
```diff
@@ -282,6 +282,24 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
...
+            end: 0
+        };
+
+        const seekable = this.video.seekable;
+
+        if (seekable.length) {
+            const maxSeekableEnd = Math.max(seekable.end(0), seekable.end(seekable.length - 1));
+
+            seekRange.end = maxSeekableEnd;
+            seekRange.start = Math.max(0, seekRange.end - this.video.duration);
```
robwalch: Why not use `seekable.start(0)`?

or 

`Math.min(seekable.start(0), seekable.start(seekable.length - 1));`

src/js/providers/html5.js
```diff
@@ -282,6 +282,24 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
         return duration;
     };
 
+    _this.getSeekRange = function() {
+        const seekRange = {
+            start: 0,
+            end: 0
```
robwalch: default end should be `duration`. You could invoke `Default. getSeekRange()`

robwalch:
test this please

PR #2794, opened 2018-03-19
Set item.starttime if seeking from complete 

johnBartos:
So that the player loads from the correct time following the re-setting and replaing of an item following complete:

https://github.com/jwplayer/jwplayer/blob/master/src/js/controller/controller.js#L412-L414
```
if (_model.get('state') === STATE_COMPLETE) {
    _stop(true);
    _setItem(0);
}
```

This causes the item to be reset and it's position reset to 0; if we seek via setting the media element's `currentTime`, it will be overwritten. Therefore, we must set the item's starttime so it will begin from the correct position when loaded.

No

No

#### Addresses Issue(s):

JW8-1339

### DIFF
--- a/src/js/program/program-controller.js
+++ b/src/js/program/program-controller.js
@@ -6,7 +6,7 @@ import { MediaControllerListener } from 'program/program-listeners';
 import Eventable from 'utils/eventable';
 import BackgroundMedia from 'program/background-media';
 
-import { ERROR, PLAYER_STATE, STATE_BUFFERING, STATE_IDLE } from 'events/events';
+import { ERROR, PLAYER_STATE, STATE_BUFFERING, STATE_IDLE, STATE_COMPLETE } from 'events/events';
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

robwalch:
@johnBartos can you make another branch off v8.2.x with this commit 84443c3f9bee70a0015d5b6c2d0c08137b726855 ? This PR into master will not make it into 8.2.0.

robwalch:
@johnBartos Never mind. We're good - Nothing 8.3 or 8.2.1 has been merged into master yet.

I just moved the v8.2.x branch to this last merge commit https://github.com/jwplayer/jwplayer/commits/v8.2.x

PR #2797, opened 2018-03-20
Feature/live stop

robwalch:
### This PR will...

- Show stop button in control bar during live stream
- Disable "LIVE" button in live mode
- Localize playback button aria-label in stop and pause mode
- Restart playback on live streams on resume (html5 provider)

https://github.com/jwplayer/jwplayer-commercial/compare/feature/live-stop?expand=1

#### Addresses Issue(s):

JW8-335


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/providers/html5.js
```diff
@@ -329,6 +329,10 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
     }
 
     function _play() {
+        const resumeLive = _videotag.duration === Infinity && _videotag.paused && _videotag.played && _videotag.played.length;
```
johnBartos: Do you think we need to account for the case where the videotag is paused because it's stalling? The change would be (`_videotag.paused && _videotag.readyState > 1`)

robwalch: I don't see why. If the stream has played, you pause it, and then resume it, we want to restart the stream. Why do that if the stream never played? Let it finish buffering.

PR #2802, opened 2018-03-23
Set bitrateSelection & bandwidthEstimate on model & preserve to localstorage

johnBartos:
- Saves the bandwidth estimate during playback to localStorage
- Deprecates qualityLabel as a method of preserving quality between sessions
- If manually selected, preserve the bitrate of the currently playing item in localStorage

So that when the player starts a new session, playback quality is propagated from previous session

No

https://github.com/jwplayer/jwplayer-commercial/pull/5006

#### Addresses Issue(s):

JW8-1204

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -36,9 +36,11 @@ const Model = function() {
 
     this.persistQualityLevel = function(quality, levels) {
         const currentLevel = levels[quality] || {};
-        if (currentLevel.bitrate) {
-            this.set('bitrateSelection', currentLevel.bitrate);
+        const bitrate = currentLevel.bitrate;
+        if (!_.isNumber(bitrate) || _.isNaN(bitrate)) {
+            return;
         }
+        this.set('bitrateSelection', bitrate);
     };
 
     this.setActiveItem = function (index) {
@@ -197,9 +199,10 @@ const Model = function() {
     };
 
     this.persistBandwidthEstimate = function (bwEstimate) {
-        if (bwEstimate && this.attributes.bandwidthEstimate !== bwEstimate) {
-            this.set('bandwidthEstimate', bwEstimate);
+        if (!_.isNumber(bwEstimate) || _.isNaN(bwEstimate)) {
+            return;
         }
+        this.set('bandwidthEstimate', bwEstimate);
     };
 };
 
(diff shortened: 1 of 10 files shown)


src/js/controller/model.js
```diff
@@ -35,9 +35,10 @@ const Model = function() {
     };
 
     this.persistQualityLevel = function(quality, levels) {
-        var currentLevel = levels[quality] || {};
-        var label = currentLevel.label;
-        this.set('qualityLabel', label);
+        const currentLevel = levels[quality] || {};
+        if (currentLevel.bitrate) {
```
robwalch: Let's make sure that levels from providers have valid bitrate values or none at all.

src/js/controller/model.js
```diff
@@ -194,6 +195,12 @@ const Model = function() {
         mediaModel.set('currentTime', 0);
         mediaModel.set('duration', duration);
     };
+
+    this.persistBandwidthEstimate = function (bwEstimate) {
+        if (bwEstimate && this.attributes.bandwidthEstimate !== bwEstimate) {
```
robwalch: "change:" only fires if the value is changed.

src/js/program/program-listeners.js
```diff
@@ -90,6 +90,11 @@ export function ProviderListener(mediaController) {
             case 'visualQuality':
                 mediaModel.set('visualQuality', Object.assign({}, data));
                 break;
+            case BANDWIDTH_ESTIMATE:
+                if (data) {
+                    mediaModel.set('bandwidthEstimate', Object.assign({}, data));
+                }
+                break;
```
robwalch: Don't need this on the mediaModel

src/js/providers/bandwidth-monitor.js
```diff
@@ -0,0 +1,26 @@
+import { BANDWIDTH_ESTIMATE } from 'events/events';
+
+export default function BandwidthMonitor(provider, initialEstimate) {
+    let bandwidthMonitorInterval = null;
+    let bandwidthEstimate = initialEstimate;
+    return {
+        start() {
```
robwalch: It would be good to `clearInterval` here.

src/js/utils/underscore.js
```diff
@@ -736,6 +736,7 @@ _.result = function (object, property) {
 };
 
 const _isNumber = _.isNumber;
+const _isValidNumber = (val) => _isNumber(val) && !_isNaN(val);
 
-export { _isNaN, _isNumber };
+export { _isNaN, _isNumber, _isValidNumber };
 export default _;
```
robwalch: If you export the methods without the underscore prefix, we can eventually do that for all methods and remove the default export. Modules that want `_` will `import * as _ from 'underscore'`.

PR #2803, opened 2018-03-26
Prevent state changes for backgrounded mediaControllers 

karimJWP:
### This PR will...
check if the mediaController is attached and not in background before triggering a change in mediaState
When a midRoll begins or playAd is called, a pause event is triggered causing the related shelf to open. Before BGL was introduced, a pause event was not fired before an ad.
should the media model still be updated ?
A prevention is added in [Related](https://github.com/jwplayer/jwplayer-plugin-related/pull/275) to prevent API calls during ad playback from opening related views
#### Addresses Issue(s):
JW8-1362

### DIFF
--- a/src/js/program/program-listeners.js
+++ b/src/js/program/program-listeners.js
@@ -30,7 +30,9 @@ export function ProviderListener(mediaController) {
                 // Always fire change:mediaState to keep player model in sync
                 const previousState = mediaModel.attributes.mediaState;
                 mediaModel.attributes.mediaState = data.newstate;
-                mediaModel.trigger('change:mediaState', mediaModel, data.newstate, previousState);
+                if (mediaController.attached && !mediaController.background) {
+                    mediaModel.trigger('change:mediaState', mediaModel, data.newstate, previousState);
+                }
                 // This "return" is important because
                 //  we are choosing to not propagate model event.
                 //  Instead letting the master controller do so
--- a/src/js/program/program-listeners.js
+++ b/src/js/program/program-listeners.js
@@ -27,13 +27,10 @@ export function ProviderListener(mediaController) {
                     mediaController.thenPlayPromise.cancel();
                     mediaModel.srcReset();
                 }
-
-                if (mediaController.attached && !mediaController.background) {
-                    //fire change:mediaState to keep player model in sync
-                    const previousState = mediaModel.attributes.mediaState;
-                    mediaModel.attributes.mediaState = data.newstate;
-                    mediaModel.trigger('change:mediaState', mediaModel, data.newstate, previousState);
-                }
+                // Always fire change:mediaState to keep player model in sync
+                const previousState = mediaModel.attributes.mediaState;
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

karimJWP:
@johnBartos it's working for me with config 
```
{
    "file": "//content.jwplatform.com/videos/4jFmgjZr-B2XWaEP9.mp4",
    "image": "//cdn.jwplayer.com/thumbs/4jFmgjZr-720.jpg",
    "title": "Agent 327",
    "description": "Operation Barbershop",
    "related": {
        "client": "../bin-debug/related.js",
        "file": "//cdn.jwplayer.com/feed.json?feed_id=9qjYmZBC&related_media_id=Zb9M8K61",
        "displayMode": "shelf"
    },
    "advertising": {
        "debug": true,
        "client": "../../bin-debug/googima.js"
    },
    "analytics": {
        "debug": true
    }
}
```
when paused before playAd, I am seeing the  `mediaController.background` is true

src/js/program/program-listeners.js
```diff
@@ -27,10 +27,13 @@ export function ProviderListener(mediaController) {
                     mediaController.thenPlayPromise.cancel();
                     mediaModel.srcReset();
                 }
-                // Always fire change:mediaState to keep player model in sync
-                const previousState = mediaModel.attributes.mediaState;
-                mediaModel.attributes.mediaState = data.newstate;
-                mediaModel.trigger('change:mediaState', mediaModel, data.newstate, previousState);
+
+                if (mediaController.attached && !mediaController.background) {
```
robwalch: Backgrounded media controllers don't emit events while backgrounded. Is this being queued?

Only the state of the mediaModel after reattaching should come out. I don't think this is the right place to address this. I would need to see the callstack at the point the pause event is being fired to determine the best course of action.

robwalch: Can we try replacing the trigger work-around with `mediaModel.set('mediaState', data.newstate);` ? That should only fire on change. We may need to do a little more to queue or ignore changes while detached.

robwalch: scratch that.

Since this involves instream mode, it should be handled higher up in the controller:
```js
_model.on('change:mediaModel', function(model, mediaModel) {
            model.set('errorEvent', undefined);
            mediaModel.change('mediaState', function (changedMediaModel, state) {
                if (!model.get('errorEvent') && !model.get('instream')) {
                    model.set(PLAYER_STATE, normalizeState(state));
                }
            }, this);
```

PR #2804, opened 2018-03-26
Copy 'androidhls', 'hlsjsdefault' and 'safarihlsjs' into playlist item sources when checking for provider support

robwalch:
### This PR will...

- Copy 'androidhls', 'hlsjsdefault' and 'safarihlsjs' into playlist item sources when checking for provider support
- Don't load html5 provider twice when bundled

These attributes need to be added to sources for provider selection to work properly. Provider selection is part of provider bundling and loaded. When the html5 provider is bundled with jwplayer.core.js, it should not be loaded again.

https://github.com/jwplayer/jwplayer-commercial/pull/5014

#### Addresses Issue(s):

JW8-1319

### DIFF
--- a/src/js/api/core-loader.js
+++ b/src/js/api/core-loader.js
@@ -1,4 +1,5 @@
 import Item from 'playlist/item';
+import { fixSources } from 'playlist/playlist';
 import ProvidersSupported from 'providers/providers-supported';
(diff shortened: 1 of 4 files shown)


johnBartos:
@robwalch Set a real title please

PR #2810, opened 2018-03-28
Disable Related Open & Close API during ads

karimJWP:
### This PR will...
set instreamMode in the model and remove it from the ads model
To allow the Related Plugin and other clients to determine when the player is playing ads via `player.getConfig().instreamMode`
I assume the adModel does not need an instreamMode since it should always be in instreamMode
[related PR](https://github.com/jwplayer/jwplayer-plugin-related/pull/275) 
#### Addresses Issue(s):
JW8-1362

### DIFF
--- a/src/js/model/player-model.js
+++ b/src/js/model/player-model.js
@@ -3,6 +3,7 @@ import { STATE_IDLE } from 'events/events';
 export const INITIAL_PLAYER_STATE = {
     audioMode: false,
     flashBlocked: false,
+    instreamMode: false,
     item: 0,
     itemMeta: {},
     playbackRate: 1,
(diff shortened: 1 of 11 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

robwalch:
"instreamMode" is a view-model only event. It should not be set on the model from the view. instream is managed in the controller, and if you want to add another property to detect it's presence we should discuss it first.

src/js/model/player-model.js
```diff
@@ -3,6 +3,7 @@ import { STATE_IDLE } from 'events/events';
 export const INITIAL_PLAYER_STATE = {
     audioMode: false,
     flashBlocked: false,
+    instreamMode: false,
```
robwalch: This should not be added to the player by default.

PR #2816, opened 2018-03-30
Return currentTime in getAlignmentPosition

DanFerrer:
### This PR will...

Return currentTime from the time event in getAlignmentPosition if metadata source is undefined.

We no longer update position on timeupdate if the stream is DVR, which causes captions to be out of sync. We can return currentTime instead to cover all cases.

#### Addresses Issue(s):

JW8-1111


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2822, opened 2018-04-03
Don't background load for live and DVR streams

robwalch:
### This PR will...

Prevent the next item from being preloaded in the background during a live or DVR stream.

Since we can't use the position and duration to know when a live or DVR stream will end, we shouldn't use them to trigger background loading. Duration either `Infinity` (live) or position is a negative number. There is no reliable signal to trigger preloading of the next item, other than a stream error which is out of scope for background loading.

#### Addresses Issue(s):

JW8-1374


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Issue #2828, opened 2018-04-05
Local custom provider support broken in JWPlayer 8.2.x


tri170391:
### Expected Behavior

Playback providers regstered to `jwplayer` through `jwplayer.api.registerProvider` and `supports` the current `file` should handle the playback. 

### Actual Behavior

This used to work in `8.1.x` and before but since `8.2.x` is released the player failed to load with error:
```
Error loading player: Failed to load media
```

### Steps to reproduce

1. Register a local playback provider through `jwplayer.api.registerProvider`.
2. Play the video.
3. The player failed to load with the aforementioned error.

### Environment

- x86-64 PC
- Windows 10
- Chrome 65.0.3325.181
- JwPlayer 8.2.2

----------------------------------------------------------------------------------------

I have digged into the issue and find that this is the root cause: https://docs.google.com/document/d/1548315Xpz54IkLUGfnOjq8N_NkWom6NTK8BvUYdF1is/edit

I will attempt to fix this and create a pull request when it's ready.

johnBartos:
@tri170391 We're aware of the issue & working on a fix

PR #2830, opened 2018-04-06
Only load provider when it is not loaded

tri170391:
### This PR will...

Fix https://github.com/jwplayer/jwplayer/issues/2828

Restore support for locally registered custom playback providers.

I attempted to write a unit test for this but this involves `global-api.registerProvider` which is not available in the `setup-steps` unit test.

None.

#### Addresses Issue(s):

JW8-2828


robwalch:
Thanks @tri170391 

Changes in master will make it into v8.3.0. I'm talking with the team about cherry picking this for v8.2.3.

PR #2832, opened 2018-04-09
Add guard check for visualQuality Event

DanFerrer:
### This PR will...

Only trigger visualQuality event in media-controller if the eventData argument isn't a falsy value.

An empty visualQuality event is triggered in the media-controller when preload is set to "none" on play attempt for HLS. The eventData argument that would be passed in when it's triggered returns null. We dont know the actual visualQuality until the adaptation event fires

#### Addresses Issue(s):

JW8-1404

### DIFF
--- a/src/js/program/media-controller.js
+++ b/src/js/program/media-controller.js
@@ -138,6 +138,9 @@ export default class MediaController extends Eventable {
                 // Start firing visualQuality once playback has started
                 mediaModel.off(MEDIA_VISUAL_QUALITY, null, this);
                 mediaModel.change(MEDIA_VISUAL_QUALITY, (changedMediaModel, eventData) => {
+                    if (!eventData) {
+                        return;
+                    }
                     this.trigger(MEDIA_VISUAL_QUALITY, eventData);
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/program/media-controller.js
```diff
@@ -138,6 +138,9 @@ export default class MediaController extends Eventable {
                 // Start firing visualQuality once playback has started
                 mediaModel.off(MEDIA_VISUAL_QUALITY, null, this);
                 mediaModel.change(MEDIA_VISUAL_QUALITY, (changedMediaModel, eventData) => {
+                    if (!eventData) {
+                        return;
```
robwalch: Providers should not dispatch visualQuality events without data. The guard should be in the provider before "visualQuality" is set on mediaModel.

PR #2835, opened 2018-04-10
Open Right Click Menu on Long Press

hghazzi:
### This PR will...

- Add a timeout on `touchstart` that will trigger a new `longPress` event if no other touch event occurs within 500ms
- Cancel the timeout whenever a `touchmove`, `touchend`, `touchcancel` event occurs. This prevents a long press to cause the controls be active or pause the player
- Bump up the right click menu 100 pixels up on touch devices to make it more visible
- Android Chrome supports `oncontextmenu`, which allows the right click menu to be displayed. It will prevent the default browser context menu to display for mp4 videos
- For iOS, we need to listen for the `longPress` event to display the right click menu. Added checks to not add `oncontextmenu` for iOS in case it ever becomes supported, which would cause two events to be triggered on longpress

Just like we display a menu with right click on desktop, we want to have the same behaviour when performing a long press. We want to allow users access the same information no matter the device.

We want to stop the long press to be considered as a tap, so we needed to update `ui.js` to treat them differently, even when we are not listening to the `longPress` event for the android chrome.

#### Addresses Issue(s):

JW8-1378

### DIFF
--- a/src/js/view/controls/rightclick.js
+++ b/src/js/view/controls/rightclick.js
@@ -68,8 +68,6 @@ export default class RightClick {
     }
 
     showMenu(evt) {
-        console.log('Show Menu', evt);
-
         // Offset relative to player element
         var off = this.getOffset(evt);
 
--- a/src/js/view/controls/rightclick.js
+++ b/src/js/view/controls/rightclick.js
@@ -60,8 +60,10 @@ export default class RightClick {
         let x = evt.pageX - playerBounds.left;
         let y = evt.pageY - playerBounds.top;
 
+        // move menu up on touch devices
+        // so it is not be blocked by fingers
         if (this.model.get('touchMode')) {
-            y -= 100;
+            y -= 75;
         }
 
         return { x: x, y: y };
--- a/src/js/view/controls/rightclick.js
+++ b/src/js/view/controls/rightclick.js
@@ -138,9 +138,7 @@ export default class RightClick {
 
 
         // Defer the rest of setup until the first click
-        if (!_model.get('touchMode')) {
-            _playerElement.oncontextmenu = this.rightClick.bind(this);
-        }
+        _playerElement.oncontextmenu = this.rightClick.bind(this);
     }
 
     addOffListener(element) {
--- a/src/js/view/controls/rightclick.js
+++ b/src/js/view/controls/rightclick.js
@@ -84,10 +84,7 @@ export default class RightClick {
         return false;
     }
 
-    hideMenu(evt) {
-        if (evt && evt.type === 'mousedown') {
-            return;
-        }
+    hideMenu() {
         this.elementUI.off('out', this.hideMenu, this);
         if (this.mouseOverContext) {
             // If mouse is over the menu, hide the menu when mouse moves out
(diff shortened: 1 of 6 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/rightclick.js
```diff
@@ -77,7 +83,10 @@ export default class RightClick {
         return false;
     }
 
-    hideMenu() {
+    hideMenu(evt) {
+        if (evt && evt.type === 'mousedown') {
+            return;
+        }
```
hghazzi: @robwalch I added this because `mousedown` was being triggered on iOS after the right click became visible causing it to become hidden. Is there a better way to guard against that occurring?

src/js/view/controls/rightclick.js
```diff
@@ -127,8 +137,11 @@ export default class RightClick {
         this.mouseOverContext = false;
         this.layer = layer;
 
-        // Defer the rest of setup until the first click
-        _playerElement.oncontextmenu = this.rightClick.bind(this);
+        // iOS does not support oncontextmenu
+        if (!OS.iOS) {
```
robwalch: There is no harm in setting this on iOS, even though it is not supported. It costs more to import environment and check for iOS than it does to just set `oncontextmenu`.

src/js/view/view.js
```diff
@@ -396,7 +396,14 @@ function View(_api, _model) {
                     }
                 }
             },
-            doubleClick: () => _controls && api.setFullscreen()
+            doubleClick: () => _controls && api.setFullscreen(),
+            longPress: (evt) => {
+                // manually trigger the right click menu on iOS
+                // because context menu is not supported
+                if (OS.iOS && _controls && _controls.rightClickMenu) {
```
robwalch: Instead of checking for iOS here, you should only create a `UI` instance that implements longPress event handling on iOS. This currently implementation makes all browsers run a timeout on mousedown or touchstart for an event that will do nothing.

src/js/view/controls/rightclick.js
```diff
@@ -56,9 +57,15 @@ export default class RightClick {
...
+        let x = evt.pageX - playerBounds.left;
+        let y = evt.pageY - playerBounds.top;
+
+        // move menu up on touch devices
+        // so it is not be blocked by fingers
+        if (this.model.get('touchMode')) {
+            y -= 100;
+        }
 
         return { x: x, y: y };
```
robwalch: `return { x, y };`

src/js/view/controls/rightclick.js
robwalch: `mousedown` or `touchend`? There is no `touchend` event listener here. If you don't want `mousedown` to run `hideMenu` then remove `element.addEventListener('mousedown', this.hideMenuHandler);` for the player element or the document, or both. The issue there is that the menu is supposed to close the moment you begin interacting with something else on the page.

I don't think you want to filter `hideMenu` by event type, I think you need to look specifically at the targets for the situation you are describing. Please include the contents of the `mousedown` event or show me the debugger state in Safari when the menu opens. It makes sense for `mousedown` to fire when a tap starts, not when it ends.

src/js/view/view.js
robwalch: I don't think you should be tagging this on to the `clickHandlerHelper`. This should have it's own UI instance so we can change the element, and only create and register it for environments that need it.

src/js/view/utils/clickhandler.js
```diff
@@ -18,7 +18,8 @@ export default class ClickHandler {
                     return;
                 }
                 this.trigger('doubleClick');
-            }
+            },
+            'longPress': (evt) => this.trigger('longPress', evt)
```
robwalch: Same as below. This shouldn't be tagged onto the video element's clickhandler.

src/js/view/view.js
robwalch: You could also implement the longpress without UI outside of here on player element. This would be better for performance since we only need to register touchstart for iOS and no other listeners.

PR #2837, opened 2018-04-10
Show which rendition is being used when quality is set to "auto"

DanFerrer:
### This PR will...

Show the current rendition when "auto" is selected in the quality menu (ex: **Auto** 720p (4763 kbps))

- An initial span element is created and appended to the first button in the quality submenu (auto) when the quality submenu is created.

- The innerHTML inside the span changes when the visualQuality event is fired. If the mode or reason is "auto" we use the index from the object from the callback to find the label associated with the new quality and set the html to that.  It will be set to an empty string if mode or reason switches from "auto" to "manual"

- Added a class to style the font weight of the span so its independent  from "Auto" being bolded when its selected.

To show viewers the current quality they're on when quality levels are being upswitched/downswitched which may encourage them to make a manual selection.

The DOM manipulation being used in settings-menu.js and submenu-factory.js can be moved somewhere else.

#### Addresses Issue(s):

JW8-233


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/templates/settings/auto-quality.js
```diff
@@ -0,0 +1,5 @@
+export default () => {
+    return (
+        `<span class="jw-reset jw-auto-label"></span>`
```
robwalch: Shouldn't this just be added to the parent template? A module for this span is a bit much. The span can always be present in the menu as long as it's empty when there is no auto label.

src/js/view/utils/submenu-factory.js
```diff
@@ -72,10 +75,16 @@ export function removeAudioTracksSubmenu(settingsMenu) {
...
-        return SettingsContentItem(track.label, track.label, (evt) => {
+        const element = SettingsContentItem(track.label, track.label, (evt) => {
             action(index);
             settingsMenu.close(evt);
         });
+
+        if (index === 0) {
+            const autoQualityElement = createElement(AutoQualityTemplate());
+            element.element().appendChild(autoQualityElement);
+        }
```
robwalch: See above. Just always include the label span in the menu.

src/js/view/controls/settings-menu.js
```diff
@@ -164,6 +164,28 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api)
...
+    // Visual Quality
+    model.on('change:visualQuality', (changedModel, quality) => {
+        const qualitySubMenu = settingsMenu.getSubmenu('quality');
+        const items = qualitySubMenu.getItems();
+        const item = items[0].element().querySelector('.jw-auto-label');
+        const levels = model.get('levels');
+        const { mode, reason, level } = quality;
+
+        if (mode === 'auto' || reason === 'auto') {
+            item.innerHTML = ` ${levels[level.index].label}`;
```
robwalch: Please use `textContent` when setting setting text in an element that does not contain markup.

src/js/view/controls/settings-menu.js
```diff
@@ -164,6 +164,28 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api)
...
+            qualitySubMenu.replaceContent(items);
+        } else {
+            // return if there already isnt a quality label on auto
+            if (!item.innerHTML) {
+                return;
+            }
+
+            item.innerHTML = ``;
```
robwalch: `item.textContent = ''`

src/js/view/controls/settings-menu.js
```diff
@@ -164,6 +164,28 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api)
...
+        const { mode, reason, level } = quality;
+
+        if (mode === 'auto' || reason === 'auto') {
+            item.innerHTML = ` ${levels[level.index].label}`;
+            qualitySubMenu.replaceContent(items);
+        } else {
+            // return if there already isnt a quality label on auto
+            if (!item.innerHTML) {
+                return;
+            }
```
robwalch: This return statement is not necessary - especially if using the more performant `textContent`.

src/js/view/controls/settings-menu.js
```diff
@@ -164,6 +164,28 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api)
...
+
+            item.innerHTML = ``;
+            qualitySubMenu.replaceContent(items);
```
robwalch: Why do we need to call `replaceContent` when setting text directly on one of the items?

src/js/view/controls/settings-menu.js
```diff
@@ -164,6 +164,28 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api)
...
+    model.on('change:visualQuality', (changedModel, quality) => {
+        const qualitySubMenu = settingsMenu.getSubmenu('quality');
+        const items = qualitySubMenu.getItems();
+        const item = items[0].element().querySelector('.jw-auto-label');
+        const levels = model.get('levels');
+        const { mode, reason, level } = quality;
+
+        if (mode === 'auto' || reason === 'auto') {
+            item.innerHTML = ` ${levels[level.index].label}`;
+            qualitySubMenu.replaceContent(items);
```
robwalch: We should not need to replace the elements when setting text directly on them.

src/js/view/utils/submenu-factory.js
```diff
@@ -72,10 +75,16 @@ export function removeAudioTracksSubmenu(settingsMenu) {
 
 export function addQualitiesSubmenu(settingsMenu, qualitiesList, action, initialSelectionIndex, tooltipText) {
     const qualitiesItems = qualitiesList.map((track, index) => {
-        return SettingsContentItem(track.label, track.label, (evt) => {
+        const element = SettingsContentItem(track.label, track.label, (evt) => {
```
robwalch:
```js
let content = track.label;
if (index === 0) {
    content += '&nbsp;<span class="jw-reset jw-auto-label"></span>'
}
return SettingsContentItem(track.label, content, (evt) => {

File: src/js/view/controls/settings-menu.js [hunk 9]
```diff
@@ -164,6 +164,17 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api)
         }
     }, settingsMenu);
 
+    // Visual Quality
+    model.on('change:visualQuality', (changedModel, quality) => {
+        const qualitySubMenu = settingsMenu.getSubmenu('quality');
```
robwalch: `qualitySubMenu`is not going to be defined if we're playing a video with only one level (mp4 file).

This page shows an exception in the console when trying to call `qualitySubMenu.getItems()`
http://player-pr-test-jenkins.longtailvideo.com/builds/14376/archive/test/squash/bootstrap.html?primary=html5&edition=ads&feature=api/api_setup_event_order&suite=undefined&debug=1

Just add a guard and return early.

PR #2839, opened 2018-04-11
Don't hide right click menu on tap

hghazzi:
### This PR will...

- Prevent the right click menu to hide when a touch event occurs
- Remove usage of `ui.js` to avoid unnecessary events being added, which is causing links in the menu to not work

- The right click menu should stay open when a tap on it occurs. This mirrors the behaviour on desktop devices that keeps the menu open when hovering above it. 
- Ensures that links in the menu are working

#### Addresses Issue(s):

JW8-1378

### DIFF
--- a/src/js/view/controls/rightclick.js
+++ b/src/js/view/controls/rightclick.js
@@ -152,12 +152,10 @@ export default class RightClick {
     }
 
     onMouseOver() {
-        console.log('onMouseOver');
         this.mouseOverContext = true;
     }
 
     onMouseOut() {
-        console.log('onMouseOut');
         this.mouseOverContext = false;
     }
 
--- a/src/js/view/controls/rightclick.js
+++ b/src/js/view/controls/rightclick.js
@@ -75,6 +75,7 @@ export default class RightClick {
 
         this.el.style.left = off.x + 'px';
         this.el.style.top = off.y + 'px';
+        this.outCount = 0;
 
         addClass(this.playerElement, 'jw-flag-rightclick-open');
         addClass(this.el, 'jw-open');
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/rightclick.js
```diff
@@ -119,13 +124,10 @@ export default class RightClick {
...
-            .on('over', function() {
-                this.mouseOverContext = true;
-            }, this)
-            .on('out', function() {
-                this.mouseOverContext = false;
-            }, this);
+        this.onMouseOverHandler = this.onMouseOver.bind(this);
+        this.onMouseOutHandler = this.onMouseOut.bind(this);
+        this.el.addEventListener('mouseover', this.onMouseOverHandler);
+        this.el.addEventListener('mouseout', this.onMouseOutHandler);
```
robwalch: Since these are only called on this element to toggle an instance property, and this element is never reused, I would suggest we just add arrow listeners and let GC take care of the cleanup:

```js
this.el.addEventListener('mouseover', () => {
  this.mouseOverContext = true
});
this.el.addEventListener('mouseout', () => {
  this.mouseOverContext = false
});
```

src/js/view/controls/rightclick.js
```diff
@@ -84,13 +83,19 @@ export default class RightClick {
...
-    hideMenu() {
-        this.elementUI.off('out', this.hideMenu, this);
-        if (this.mouseOverContext) {
+    hideMenu(evt) {
+        if (!this.model.get('touchMode')) {
             // If mouse is over the menu, hide the menu when mouse moves out
-            this.elementUI.on('out', this.hideMenu, this);
+            this.el.removeEventListener('mouseout', this.hideMenuHandler);
+            if (this.mouseOverContext) {
+                this.el.addEventListener('mouseout', this.hideMenuHandler);
```
robwalch: Can this be added once in lazySetup instead of on every `hideMenu` call?

hghazzi: Looks like the reason this is here is that after you right click, you're technically inside the menu, so as soon as you move the mouse, the menu will hide.

robwalch: Right. So then what's wrong with always having this listener?

hghazzi: Yes

egreaves: @hghazzi can we close the loop on this? Not sure what's meant by 'yes'.

PR #2841, opened 2018-04-12
Mute button is only set on mobile.

vseventer:
### This PR will...
Fix a `TypeError: null is not an object (evaluating 'this.mute.hide')` on desktop when attempting to unmute from `autostartMuted`.

The mute button is only set on mobile ([here](https://github.com/jwplayer/jwplayer/blob/master/src/js/view/controls/controls.js#L178)), but we try to hide it on all devices. This PR restricts it to mobile.

N/A

N/A

#### Addresses Issue(s):
N/A (found as part of ADS-1074)

### DIFF
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -408,7 +408,7 @@ export default class Controls {
         api.setMute(mute);
         // the model's mute value may not have changed. ensure the controlbar's mute button is in the right state
         this.controlbar.renderVolume(mute, model.get('volume'));
-        if (OS.mobile) {
+        if (this.mute) {
             this.mute.hide();
         }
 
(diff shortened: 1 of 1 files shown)


src/js/view/controls/controls.js
```diff
@@ -408,7 +408,10 @@ export default class Controls {
         api.setMute(mute);
         // the model's mute value may not have changed. ensure the controlbar's mute button is in the right state
         this.controlbar.renderVolume(mute, model.get('volume'));
-        this.mute.hide();
+        if (OS.mobile) {
+            this.mute.hide();
+        }
```
robwalch: Instead we should guard with `if (this.mute) {`. That was it doesn't matter if we change which env we add or don't add the mute button.

PR #2843, opened 2018-04-12
Do not fire programController events during an ad

pajong:
### This PR will...
stop sending player events from programController during an ad playback.

We do not want to fire player events (time event specifically) during DAI ad playback.
We cannot detach the provider during the ad playback when DAI stream is playing, because it will stop the content.
Do not forward the events from programController while ad playback to stop player events from triggering.

No, but I want to check the tests on the build to see if any regressions are caused by this.

#### Addresses Issue(s):
ADS-971

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -771,7 +771,12 @@ Object.assign(Controller.prototype, {
 
         function addProgramControllerListeners() {
             _programController
-                .on('all', _trigger, _this)
+                .on('all', (type, event) => {
+                    if (_this._instreamAdapter) {
+                        return;
+                    }
+                    _trigger(type, event);
+                }, _this)
                 .on('subtitlesTracks', (e) => {
                     _captions.setSubtitlesTracks(e.tracks);
                     const defaultCaptionsIndex = _captions.getCurrentIndex();


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -771,7 +771,12 @@ Object.assign(Controller.prototype, {
 
         function addProgramControllerListeners() {
             _programController
-                .on('all', _trigger, _this)
+                .on('all', (type, event) => {
+                    if (_this._instreamAdapter) {
+                        return;
+                    }
+                    _trigger(type, event);
+                }, _this)
```
robwalch: @pajong did you find any issues today with this change?

@dannyfinks can we get a larger regression test run on this?

One concern I have with DAI is that queueing could cause all of the provider events that fire during instream mode will fire after the adbreak ended.

dannyfinks: @robwalch will do

dannyfinks: @robwalch looking at the test results from the smoke test...it seems these changes is causing time events to not be emitted after a VAST Preroll.

PR #2845, opened 2018-04-13
Remove quality indication label if level clicked has the same quality as auto

DanFerrer:
### This PR will...

- Remove the quality from the "auto" label if we switch to the quality that auto was representing

- Make css changes to the label to make it more readable(discussed with @aliciahurst )

When making a manual change to a quality that was already represented by "auto", the label indicating the quality would stay there and vice versa. This is because there isnt a visualQuality event being triggered when we switch to a level that is already represented on auto

There may be some redundant code now that the "currentLevel" and "visualQuality" events  handle when the label gets changed and the purpose for it in "currentLevel" is just to handle this one case where "visualQuality" wont fire because the qualities are the same

#### Addresses Issue(s):

JW8-233

### DIFF
--- a/src/js/view/controls/settings-menu.js
+++ b/src/js/view/controls/settings-menu.js
@@ -182,7 +182,7 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api) 
     model.on('change:visualQuality', (changedModel, quality) => {
         const qualitySubMenu = settingsMenu.getSubmenu('quality');
         if (qualitySubMenu) {
-            changeAutoLabel(quality, qualitySubMenu, 0);
+            changeAutoLabel(quality, qualitySubMenu, model.get('currentLevel'));
         }
     });
 
--- a/src/js/view/controls/settings-menu.js
+++ b/src/js/view/controls/settings-menu.js
@@ -133,21 +133,22 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api) 
         );
     };
 
+    const changeAutoLabel = function (quality, qualitySubMenu, currentQuality) {
+        const items = qualitySubMenu.getItems();
+        const item = items[0].element().querySelector('.jw-auto-label');
+        const levels = model.get('levels');
+        const { level, mode } = quality;
+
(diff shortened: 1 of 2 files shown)


src/js/view/controls/settings-menu.js
```diff
@@ -168,12 +182,7 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api)
     model.on('change:visualQuality', (changedModel, quality) => {
         const qualitySubMenu = settingsMenu.getSubmenu('quality');
         if (qualitySubMenu) {
-            const items = qualitySubMenu.getItems();
-            const item = items[0].element().querySelector('.jw-auto-label');
-            const levels = model.get('levels');
-            const { mode, level } = quality;
-
-            item.textContent = mode === 'auto' ? `${levels[level.index].label}` : ``;
+            changeAutoLabel(quality, qualitySubMenu, 0);
```
robwalch: Pass the currentLevel here, not `0`.
const currentLevel = model.get('currentLevel');

src/js/view/controls/settings-menu.js
```diff
@@ -133,9 +133,23 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api)
         );
     };
 
+    const changeAutoLabel = function (quality, qualitySubMenu, currentQuality) {
+        const items = qualitySubMenu.getItems();
+        const item = items[0].element().querySelector('.jw-auto-label');
+        const levels = model.get('levels');
+        const { level, mode } = quality;
+
+        item.textContent = currentQuality || mode === 'manual' ? '' : levels[level.index].label;
```
robwalch: When would `currentQuality` be greater than `0` and the player not be in manual mode? The last visual quality event could be a manual change, but that doesn't matter if the user selected a quality, unless we want to continue showing the auto quality label until the quality actually changes.

robwalch: I think we should remove `|| mode === 'manual'`  - that just prevents the label from being added until an ABR switch. If you were in say "240p" and then picked "auto", we should just immediately see "Auto 240p" because we know the visual quality, and it does matter that we got there in manual mode.

PR #2852, opened 2018-04-14
Fix quality label being displayed in html5 provider

DanFerrer:
### This PR will...

Prevent quality label from being created and displayed in html5 provider where we have a player that has multiple sources for a video instead of an adaptive stream.

We currently create the auto label when we iterate over "qualitiesList" in "addQualitiesSubmenu" and hit the first item because "Auto" will always be the first one. That isn't the case where we have multiple sources for a video instead of it being an adaptive stream. The first quality in that case would be whatever source is first in the config

#### Addresses Issue(s):

JW8-1439

### DIFF
--- a/src/js/view/controls/settings-menu.js
+++ b/src/js/view/controls/settings-menu.js
@@ -135,8 +135,7 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api) 
 
     const changeAutoLabel = function (quality, qualitySubMenu, currentQuality) {
         const levels = model.get('levels');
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2856, opened 2018-04-16
Add Padding to Display Icons While Casting

hghazzi:
### This PR will...

Apply the margin and padding used for the display icon container towards all breakpoints instead of just breakpoint 1

We do not only show rewind and next display icons at breakpoint 1. While casting, they are always displayed at larger breakpoints. We need to apply the same margin and padding to the display icon containers so they are spaced out appropriately

#### Addresses Issue(s):

JW8-1307

### DIFF
--- a/src/css/controls/imports/display.less
+++ b/src/css/controls/imports/display.less
@@ -49,6 +49,8 @@ display icons
 
 .jw-display-icon-container {
     display: inline-block;
+    padding: (@mobile-touch-target * 0.125);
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2858, opened 2018-04-17
Add touchstart handlers to rightclick for ios; only add hideMenu handlers when rightclick is open

johnBartos:
So that the rightclick menu closes on mobile, and that handlers are only attached when needed

No

No

#### Addresses Issue(s):

JW8-1379

### DIFF
--- a/src/js/view/controls/rightclick.js
+++ b/src/js/view/controls/rightclick.js
@@ -54,6 +54,7 @@ export default class RightClick {
 
         this.hideMenu();
         this.showMenu(evt);
+        this.addHideMenuHandlers();
 
         return false;
     }
@@ -131,7 +132,6 @@ export default class RightClick {
             this.hideMenu();
             this.infoOverlay.open();
         };
-        this.addListeners();
     }
 
     setup(_model, _playerElement, layer) {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/rightclick.js
```diff
@@ -167,24 +167,24 @@ export default class RightClick {
         clearTimeout(this.longPressTimeout);
     }
 
-    addListeners() {
-        if (!OS.iOS) {
+    addHideMenuHandlers() {
+        this.removeHideMenuHandlers();
```
hghazzi: This is setting `oncontextmenu` to null, and it is never reset. So for Android Chrome, the right-click menu only shows up the first time you long press. Any subsequent long-press will open the default context menu.

johnBartos: thanks

PR #2859, opened 2018-04-18
Populate media notifications with all known video information on Android Chrome

DanFerrer:
### This PR will...

Populate Media Notifications for the player on Android Chrome with available video info (title, image, website)

@hghazzi I create the metadata for the notification when playlistItem is set to try and grab the info as soon as possible. Let me know if you think this should be in a different place.

Also you can specify image fallbacks including sizes, and types and it will try to use the biggest size it can find.

#### Addresses Issue(s):

JW8-812


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -248,6 +248,16 @@ Object.assign(Controller.prototype, {
                         index: _model.get('item'),
                         item: playlistItem
                     });
+                    const { title, image } = playlistItem;
+                    if ('mediaSession' in navigator && title && image) {
+                        navigator.mediaSession.metadata = new window.MediaMetadata({
+                            title: title,
+                            artist: window.location.hostname,
+                            artwork: [
+                                { src: image, sizes: '384x384' }
```
hghazzi: is `sizes` required? If so, where does '384x384' come from?

src/js/controller/controller.js
```diff
@@ -248,6 +248,16 @@ Object.assign(Controller.prototype, {
                         index: _model.get('item'),
                         item: playlistItem
                     });
+                    const { title, image } = playlistItem;
+                    if ('mediaSession' in navigator && title && image) {
```
hghazzi: We only want to do this if we have both `title` and `image`?

robwalch: How about `(title || image)`?

src/js/controller/controller.js
```diff
@@ -248,6 +248,16 @@ Object.assign(Controller.prototype, {
...
+                        navigator.mediaSession.metadata = new window.MediaMetadata({
+                            title: title,
+                            artist: window.location.hostname,
+                            artwork: [
+                                { src: image, sizes: '384x384' }
+                            ]
+                        });
+                    }
```
robwalch: This code should go before triggering the playlist item event `_this.trigger(PLAYLIST_ITEM`

src/js/controller/controller.js
robwalch: The image could be any size. But the default thumbnail from our platform is always 720x406* (720 width, height may vary). Ex:
http://assets-jpcust.jwpsrv.com/thumbs/sYKZyjxB-720.jpg

1,920
1,280
480

DanFerrer: @robwalch @hghazzi works fine without having to use the sizes attribute or having to change the width.   

screenshot_20180419-164555

hghazzi:
test this please

robwalch:
test this please

robwalch:
test this please

src/js/controller/controller.js
```diff
@@ -244,6 +244,16 @@ Object.assign(Controller.prototype, {
 
             _model.change('playlistItem', function(model, playlistItem) {
                 if (playlistItem) {
+                    const { title, image } = playlistItem;
+                    if ('mediaSession' in navigator && (title || image)) {
```
robwalch: This code comes right before we trigger "playlistItem". Any exceptions would break player functionality. As such you need to add better guards here.

Just because "mediaSession" is in window, that does not guarantee that "MediaMetadata" is as well. Add it to the if block.

I would also like to see the following code in a `try catch` block to handle any potential exceptions instantiating `MediaMetadata`. For example if artwork src is not parsed correctly a `TypeError` will be thrown (https://wicg.github.io/mediasession/#convert-artwork-algorithm).

Don't we also need action handlers to control the player from the device's info panel?
 https://developer.mozilla.org/en-US/docs/Web/API/MediaSession

src/js/controller/controller.js
```diff
@@ -244,6 +244,22 @@ Object.assign(Controller.prototype, {
...
+                        try {
+                            navigator.mediaSession.metadata = new window.MediaMetadata({
+                                title,
+                                artist: window.location.hostname,
+                                artwork: [
+                                    { src: image || '' }
+                                ]
+                            });
+                        } catch (error) {
+                            _this.triggerError({
```
robwalch: Catching here does not need to trigger an error. The only failure is setting up a media session. The PLAYLIST_ITEM event will be triggered.

PR #2862, opened 2018-04-19
Close right click menu on tap (Android Chrome)

hghazzi:
### This PR will...

always add a `touchstart` listener to hide the menu, not just for iOS. Otherwise, the right click menu will not hide on a tap on the player or page when its open.

After the right-click menu was opened on Android Chrome, when tapping anywhere outside of the menu should close it. This was only happening on iOS.

#### Addresses Issue(s):

JW8-1464

### DIFF
--- a/src/js/view/controls/rightclick.js
+++ b/src/js/view/controls/rightclick.js
@@ -169,10 +169,11 @@ export default class RightClick {
 
     addHideMenuHandlers() {
         this.removeHideMenuHandlers();
-        if (OS.iOS) {
-            this.playerElement.addEventListener('touchstart', this.hideMenuHandler);
-            document.addEventListener('touchstart', this.hideMenuHandler);
-        } else {
+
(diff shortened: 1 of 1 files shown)


src/js/view/controls/rightclick.js
```diff
@@ -169,10 +169,11 @@ export default class RightClick {
...
         this.removeHideMenuHandlers();
-        if (OS.iOS) {
-            this.playerElement.addEventListener('touchstart', this.hideMenuHandler);
-            document.addEventListener('touchstart', this.hideMenuHandler);
-        } else {
+
+        this.playerElement.addEventListener('touchstart', this.hideMenuHandler);
+        document.addEventListener('touchstart', this.hideMenuHandler);
+
+        if (!OS.iOS) {
```
johnBartos: Should this be restricted to mobile, not just iOS?

PR #2865, opened 2018-04-19
Support requestFilter and xhr options in ajax utils

robwalch:
Allow an `XmlHttpRequest` object to be passed to `ajax` via options, or returned via `options.requestFilter`. This will allow request filters to be implemented in player components that use the ajax util.

A request filter would run before all xhr listeners are set and before `open` is called. Ajax callback and timeout cannot be changed and will still be controlled by the player to prevent bad filters from breaking player functionality.

### DIFF
--- a/src/js/utils/ajax.js
+++ b/src/js/utils/ajax.js
@@ -1,14 +1,11 @@
 import { parseXML } from 'utils/parser';
-import _ from 'utils/underscore';
 
 const noop = function() {};
 
-let useDomParser = false;
-
 // TODO: deprecate (jwplayer-ads-vast uses utils.crossdomain(url)). It's used here for IE9 compatibility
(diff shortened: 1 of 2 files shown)


src/js/utils/ajax.js
```diff
@@ -35,33 +32,45 @@ export function ajax(url, completeCallback, errorCallback, args) {
...
-        xhr.onreadystatechange = _readyStateChangeHandler(options);
+        xhr = options.xhr = options.xhr || new window.XMLHttpRequest();
     } else {
         // browser cannot make xhr requests
         options.onerror('', url);
         return;
     }
-    var requestError = _requestError('Error loading file', options);
+    if (typeof options.requestFilter === 'function') {
+        let result
```
egreaves: ;

PR #2868, opened 2018-04-23
Only display loading ad text at load (VAST)

boyntoni:
### This PR will...

IMDB noticed that there was not parity between Googima and VAST as it relates to when we display "Loading Ad"; this is due to a previous change I made ([here](https://github.com/jwplayer/jwplayer-ads-googima/pull/326)). 

This PR creates parity between the two clients, setting the localized "Loading Ad" text only when load item has been called.

Customer request and parity.

N/A

No.

#### Addresses Issue(s):

ADS-1143

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -115,8 +115,6 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
             clickHandler.setAlternateClickHandlers(() => {}, null);
         }
 
-        this.setText(_model.get('localization').loadingAd);
-
         // We need to know if we're beforeComplete before we reattach, since re-attaching will toggle the beforeComplete flag back if set
         _beforeComplete = _controller.isBeforeComplete() || _model.get('state') === STATE_COMPLETE;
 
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/instream-adapter.js
```diff
@@ -194,6 +192,9 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         if (_destroyed || !_inited) {
             return Promise.reject(new Error('Instream not setup'));
         }
+        
+        this.setText(_model.get('localization').loadingAd);
```
robwalch: `loadItem` is never called for IMA so doesn't break the change for IMA?

robwalch: This change would only set the text once a creative is loading.

The text should be set while a tag is loaded. To ensure this, calls `setText` in vast instead of  in instream.

src/js/controller/instream-adapter.js
```diff
@@ -115,8 +115,6 @@ var InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
             clickHandler.setAlternateClickHandlers(() => {}, null);
         }
 
-        this.setText(_model.get('localization').loadingAd);
```
robwalch: This must remain as is for freewheel and DAI.

PR #2871, opened 2018-04-23
Implement standardized errors for webpack module loading

johnBartos:
So that webpack module load errors are trackable, programmable, all those good things etc. 

No

https://github.com/jwplayer/jwplayer-commercial/pull/5154

#### Addresses Issue(s):

JW8-1444

### DIFF
--- a/src/js/api/errors.js
+++ b/src/js/api/errors.js
@@ -25,6 +25,11 @@ export const SETUP_ERROR_LICENSE_INVALID = 100012;
 export const SETUP_ERROR_LICENSE_EXPIRED = 100013;
 
 /**
+ * @enum {ErrorCode} Setup failed because a core module failed to load.
+ */
+export const SETUP_ERROR_LOADING_CORE_JS = 101000;
+
+/**
  * Class representing the jwplayer() API.
  * Creates an instance of the player.
  * @class PlayerError
--- a/src/js/controller/controls-loader.js
+++ b/src/js/controller/controls-loader.js
@@ -12,7 +12,7 @@ export function loadControls() {
             return ControlsModule;
         }, function() {
             controlsPromise = null;
-            chunkLoadErrorHandler();
+            chunkLoadErrorHandler(130)();
         }, 'jwplayer.controls');
     }
     return controlsPromise;
--- a/test/unit/errors/core-loader-error-tests.js
+++ b/test/unit/errors/core-loader-error-tests.js
@@ -1,0 +1,12 @@
+import sinon from 'sinon';
+import { chunkLoadErrorHandler } from 'api/core-loader';
+
+describe('core-loader errors', function () {
+    it('exports a chunkLoadErrorHandler function which throws a JWError', function () {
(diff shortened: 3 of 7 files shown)


src/js/api/errors.js
```diff
@@ -24,6 +24,11 @@ export const SETUP_ERROR_LICENSE_INVALID = 100012;
  */
 export const SETUP_ERROR_LICENSE_EXPIRED = 100013;
 
+/**
+ * @enum {ErrorCode} Setup failed because a core module failed to load.
+ */
+export const SETUP_ERROR_LOADING_CORE_JS = 101000;
```
karimJWP: @robwalch @johnBartos earlier we agreed that we would export 6 digit codes, but if a base code or a category code were to change we wouldn't want to have to update all the codes that are built off of it. How do you guys feel about
```
const SETUP_ERROR_BASE_CODE = 100000;
export const SETUP_ERROR_LOADING_CORE_JS = SETUP_ERROR_BASE_CODE + 1000;
export const SETUP_ERROR_LICENSE_EXPIRED = SETUP_ERROR_BASE_CODE + 13;
```

src/js/controller/controls-loader.js
```diff
@@ -12,7 +12,7 @@ export function loadControls() {
             return ControlsModule;
         }, function() {
             controlsPromise = null;
-            chunkLoadErrorHandler();
+            chunkLoadErrorHandler(130)();
```
karimJWP: remove extra `()`

test/unit/errors/core-loader-error-tests.js
```diff
@@ -0,0 +1,12 @@
+import sinon from 'sinon';
+import { chunkLoadErrorHandler } from 'api/core-loader';
+
+describe('core-loader errors', function () {
+    it('exports a chunkLoadErrorHandler function which throws a JWError', function () {
+        const handler = chunkLoadErrorHandler(105);
+        expect(handler).to.be.a('function');
+        // The handler itself is a function, which returns a function which throws
+        expect(chunkLoadErrorHandler).to.not.throw();
+        expect(handler).to.throw();
```
karimJWP: this is great. Could we check that it returns an error and that the code is 105 ?

src/js/controller/controls-loader.js
johnBartos: That's not extra, it's invoking the returned function

test/unit/errors/core-loader-error-tests.js
johnBartos: No. It's very hard to make `require.ensure` fail via network conditions in a unit test; webpack injects a script tag into the DOM instead of using `xhr`, so we can't mock it.

src/js/api/errors.js
robwalch: I think this is over doing it, and prevents someone from searching our code base for some of the specific error codes.

We should not update base codes. New errors can have new codes. Old ones should not change.

PR #2875, opened 2018-04-24
[DO NOT MERGE] Set fullscreen when screen.orientation changes on Android Chrome

DanFerrer:
### This PR will...

Set "fullscreen" to true when the viewer switches from portrait to landscape (works turning your phone clockwise and counter clockwise). Sets it to false when going from landscape to portrait 

- Player must be at least 75% visible
- Player exits from fullscreen if video is paused 

#### Addresses Issue(s):

JW8-1380

### DIFF
--- a/src/js/view/utils/views-manager.js
+++ b/src/js/view/utils/views-manager.js
@@ -42,6 +42,9 @@ function scheduleResponsiveRedraw() {
         views.forEach(view => {
             view.checkResized();
         });
+        views.forEach(view => {
+            view.checkOrientation();
+        });
     });
 }
 
--- a/src/js/view/utils/views-manager.js
+++ b/src/js/view/utils/views-manager.js
@@ -42,9 +42,12 @@ function scheduleResponsiveRedraw() {
         views.forEach(view => {
             view.checkResized();
         });
-        views.forEach(view => {
-            view.checkOrientation();
-        });
+    });
+}
+
+function checkOrientation(event) {
+    views.forEach(view => {
+        view.checkOrientation(event.target.type);
     });
 }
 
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

karimJWP:
@DanFerrer a few painful lessons learned after implementing this for the iOS SDKs. 
Initially we created a 'forceFullscreenOnLandscape' bool in our config to allow developers to disable this behavior. Customers then requested a 'forceLandscapeOnFullscreen' bool that would force a landscape display when the viewer tapped the fullscreen button and entered fullscreen. 
These configurations opened a can of worms because in iOS apps, (at a high level) you can configure specific orientation support for different layouts, and those combinations of settings have different effects on orientation detection, and the player's layout was also affected by the parent view's layout. 
We were never able to make our SDKs support all combinations of layout settings and to this day we get bug requests from customers using very specific layouts. It's gotten to the point where fixing bugs in one customer's layout will break support for another customer's layout.
My recommendation before exposing this functionality is to understand if different layouts or page settings can affect mobile browser orientation, orientation change detection and player layout; I have a feeling the experience is uniform and handled at a browser level, but it's worth making sure and avoiding the fullscreen orientation support nightmare we've had in the SDKs.

src/js/view/utils/views-manager.js
```diff
@@ -56,11 +62,19 @@ document.addEventListener('webkitvisibilitychange', onVisibilityChange);
 window.addEventListener('resize', scheduleResponsiveRedraw);
 window.addEventListener('orientationchange', scheduleResponsiveRedraw);
 
+if (screen.orientation) {
+    screen.orientation.addEventListener('change', checkOrientation);
```
robwalch: Don't add event listeners for code that is user-agent guarded. If this functionality is only for Android Chrome, then the `OS.android && Browser.chrome` check should happen here before adding the listener, instead of making all browsers with `screen.orientation` do the same work for nothing.

src/js/view/view.js
```diff
@@ -150,6 +150,23 @@ function View(_api, _model) {
...
+            const isLandscape = type === 'landscape-primary' || type === 'landscape-secondary';
+
+            if (!isLandscape && state === 'paused') {
+                // Set fullscreen to false when going back to portrait while paused and return early
+                _this.model.set('fullscreen', false);
+                return;
+            }
+
+            if (state === 'playing') {
+                _this.model.set('fullscreen', isLandscape);
```
robwalch: Hmm don't you mean `api.setFullscreen(isLandscape)` ? Setting "fullscreen" in the model doesn't actually make the player go fullscreen, going fullscreen changes the model.

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "jwplayer",
-  "version": "8.3.0-rc.1",
```
robwalch: This change shouldn't be showing up. You should rebase or make new branch with just the changes for this feature.

PR #2879, opened 2018-04-25
Set "mediaType" to audio when mimeType is "audio/mp4"

DanFerrer:
### This PR will...

Set "mediaType" to audio when mimeType is "audio/mp4"

Setting the type to "audio/mp4" in the config causes the type to become "mp4" when we trigger a media_type event and because of that, we trigger mediaType as video instead of audio. This is because when we do the check in default.js, the type of the level becomes "mp4" instead of "aac" and the poster image will disappear on playback

Should this be accounted for in source.js instead?

#### Addresses Issue(s):

JW8-1407

### DIFF
--- a/src/js/providers/default.js
+++ b/src/js/providers/default.js
@@ -94,9 +94,9 @@ const DefaultProvider = {
     },
 
     sendMediaType: function(levels) {
-        var level = levels[0];
-        var type = level.type;
-        var isAudioFile = (type === 'oga' || type === 'aac' || type === 'mp3' ||
+        const level = levels[0];
+        const type = level.type;
+        const isAudioFile = (type === 'oga' || type === 'aac' || type === 'mp3' ||
             type === 'mpeg' || type === 'vorbis' || level.mimeType === 'audio/mp4');
 
         this.trigger(MEDIA_TYPE, {
--- a/src/js/providers/default.js
+++ b/src/js/providers/default.js
@@ -94,9 +94,10 @@ const DefaultProvider = {
     },
 
     sendMediaType: function(levels) {
-        var type = levels[0].type;
+        var level = levels[0];
+        var type = level.type;
         var isAudioFile = (type === 'oga' || type === 'aac' || type === 'mp3' ||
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/providers/default.js
```diff
@@ -94,9 +94,10 @@ const DefaultProvider = {
     },
 
     sendMediaType: function(levels) {
-        var type = levels[0].type;
-        var isAudioFile = (type === 'oga' || type === 'aac' || type === 'mp3' ||
-            type === 'mpeg' || type === 'vorbis');
+        const level = levels[0];
+        const type = level.type;
+        const isAudioFile = (type === 'oga' || type === 'aac' || type === 'mp3' ||
+            type === 'mpeg' || type === 'vorbis' || level.mimeType === 'audio/mp4');
```
robwalch: Do `level.mimeType.indexOf('audio/') === 0` to support all mime-types starting with 'audio/' .

robwalch: Re source.js:  We could consider adding `mediaType` as a property of source in a future release, but should not change how `source.type` is set, because it could impact existing integrations with our API.

src/js/providers/default.js
```diff
@@ -94,9 +94,10 @@ const DefaultProvider = {
     },
 
     sendMediaType: function(levels) {
-        var type = levels[0].type;
-        var isAudioFile = (type === 'oga' || type === 'aac' || type === 'mp3' ||
-            type === 'mpeg' || type === 'vorbis');
+        const level = levels[0];
+        const { type, mimeType } = level;
```
robwalch: level is not used. just destructure levels[0]

`const { type, mimeType } = levels[0];`

PR #2881, opened 2018-04-26
Ajax Error Codes

robwalch:
### This PR will...

- Implement request (1-6), network (400-599) and format (600+) error codes for ajax utils.
- Add a `PlayerError` parameter to the `onerror` ajax callback
- Move `crossdomain(uri)` function into helpers.js since we only keep it around to maintain the `utils` interface.
- Update ajax unit tests, maintaining existing coverage through promises, and adding coverage for error codes, and new errors
 
Provide standardized error objects for errors returned by the ajax util.

Some errors are not tested because they require modification to browser functionality that I don't want to tamper with or cannot mock in our tests. If there are specific tests or assertions we want to make, let's discuss.
 
No

#### Addresses Issue(s):

JW8-1445

### DIFF
--- a/src/js/utils/ajax.js
+++ b/src/js/utils/ajax.js
@@ -1,19 +1,23 @@
 import { parseXML } from 'utils/parser';
+import { PlayerError } from 'api/errors';
+
+// XHR Request Errors
+const ERROR_TIMEOUT = 1;
+const ERROR_XHR_UNDEFINED = 2;
+const ERROR_XHR_OPEN = 3;
+const ERROR_XHR_SEND = 4;
+const ERROR_XHR_FILTER = 5;
+const ERROR_XHR_UNKNOWN = 6;
+
+// Network Responses with http status 400-599
+// will produce an error with the http status code
+
+// Format Errors
+const ERROR_DOM_PARSER = 601;
+const ERROR_NO_XML = 602;
+const ERROR_JSON_PARSE = 611;
 
 const noop = function() {};
-
-// TODO: deprecate (jwplayer-ads-vast uses utils.crossdomain(url)). It's used here for IE9 compatibility
-export function crossdomain(uri) {
-    const a = document.createElement('a');
-    const b = document.createElement('a');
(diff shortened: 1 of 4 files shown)


src/js/utils/ajax.js
```diff
@@ -152,7 +165,8 @@ function _readyStateChangeHandler(options) {
                 } else {
                     message = '' + xhr.status + '(' + xhr.statusText + ')';
                 }
-                return options.onerror(message, options.url, xhr);
+                _error(options, message, (xhr.status < 600) ? xhr.status : ERROR_XHR_UNKNOWN);
```
karimJWP: This error handling seems odd, we only handle the case where the code is 200 or a value greater than 400. Is that because the state change handler gets called several times with a ready state of 4 until status is 200 or > 400? If not, I could rewrite in a separate PR as:
```
if (xhr.readyState === 4) {
            clearTimeout(options.timeoutId);
            if (xhr.status === 200) {
                return _ajaxComplete(options)(e);
            }
            let message;
            if (xhr.status >= 400) {
                if (xhr.status === 404) {
                    message = 'File not found';
                } else {
                    message = '' + xhr.status + '(' + xhr.statusText + ')';
                }
            }
            _error(options, message, (xhr.status < 600) ? xhr.status : ERROR_XHR_UNKNOWN);
}
```
that way we can catch a broader status range, i.e. xhr.status is undefined

robwalch: > This error handling seems odd, we only handle the case where the code is 200 or a value greater than 400. Is that because the state change handler gets called several times with a ready state of 4 until status is 200 or > 400?

`if (xhr.readyState === 4) {` handles waiting for the request to be complete.

When it is, code `200` signifies that the response was OK, and we can use it. Anything else that is not an error would be handled by XHR (redirect), or by triggering the onerror callback.

`400-599` are client-server errors, and we want to know when these happen for a file because it helps us tell why the client couldn't retrieve the file. Other status codes like I mentioned would have been handled already so not helpful.

PR #2892, opened 2018-05-04
Implement standard errors for provider load failures

johnBartos:
### This PR will...
- Add codes to errors thrown by `providers.load` found in `program-controller` and during `setup-steps`

So that we can track provider load failures

Did I miss any codes?

No

#### Addresses Issue(s):

JW8-1508

### DIFF
--- a/src/js/api/errors.js
+++ b/src/js/api/errors.js
@@ -30,11 +30,6 @@ export const SETUP_ERROR_LICENSE_EXPIRED = 100013;
 export const SETUP_ERROR_LOADING_CORE_JS = 101000;
 
 /**
- * @enum {ErrorCode} Setup failed because the playlist failed to load
- */
-export const SETUP_ERROR_LOADING_PLAYLIST = 102000;
-
-/**
  * @enum {ErrorCode} Setup failed because the initial provider failed to load
  */
 export const SETUP_ERROR_LOADING_PROVIDER = 104000;
--- a/src/js/api/errors.js
+++ b/src/js/api/errors.js
@@ -30,6 +30,21 @@ export const SETUP_ERROR_LICENSE_EXPIRED = 100013;
 export const SETUP_ERROR_LOADING_CORE_JS = 101000;
 
 /**
+ * @enum {ErrorCode} Setup failed because the playlist failed to load
+ */
+export const SETUP_ERROR_LOADING_PLAYLIST = 102000;
+
+/**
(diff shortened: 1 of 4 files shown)


karimJWP:
Also @johnBartos  it seems to me that `ERROR_LOADING_PROVIDER` can get caught in https://github.com/jwplayer/jwplayer/blob/master/src/js/controller/controller.js#L357, where it would be summed to `ERROR_LOADING_PLAYLIST` resulting in code 406,000. This would happen in `controller.js` 's `_load` in the case where item is a number (`loadPromise = _setItem(item);`)

src/js/api/errors.js
```diff
@@ -39,6 +39,21 @@ export const SETUP_ERROR_LOADING_PLAYLIST = 102000;
...
 
+/**
+ * @enum {ErrorCode} Setup failed because the initial provider failed to load
+ */
+export const SETUP_ERROR_LOADING_PROVIDER = 104000;
+
+/**
+ * @enum {ErrorCode} Between playlist items, the required provider could not be loaded be
+ */
+export const PLAYLIST_ERROR_LOADING_PROVIDER = 203000;
```
karimJWP: The `203,000` category is for Playlist items, so we should rename this to `ERROR_LOADING_PLAYLIST_ITEM`. The category covers more than provider errors, i.e. we can have code 203,640 which is the `No media` error.

karimJWP: Also let's add periods to the descriptions, it seems to be the pattern in api.js

src/js/controller/controller.js
```diff
@@ -585,7 +586,11 @@ Object.assign(Controller.prototype, {
 
         function _item(index, meta) {
             _stop(true);
-            _setItem(index);
+            _setItem(index)
+                .catch(e => {
```
karimJWP: this looks good but my concern is you can switch items by calling `load(1)` which would switch to item index 2 which would directly call `_setItem`. I feel like loading an index should call `_item`

robwalch: `load()` should not `stop` and `play` before and after changing the item. It doesn't with `load(feed)` or `load(object)` so the same applies to `load(Number)`.

karimJWP: @robwalch why do we have a different implementation for `load(Number)` and `next()`? Shouldn't `load(currentIndex+1)` be the same as calling `next()`?

PR #2893, opened 2018-05-06
Only allow one plugin loader per api/model instance

robwalch:
### This PR will...

Prevent multiple plugin instances of the same plugin from being created for one player instance when setup is called several times during setup.

The plugin loader instance was not bound by id rather than a specific player instance or model, so async calls started for one instance could resolve to another if the first one was replaced.

Test page https://github.com/jwplayer/jwplayer-commercial/pull/5208

#### Addresses Issue(s):

JW8-###

### DIFF
--- a/src/js/api/Setup.js
+++ b/src/js/api/Setup.js
@@ -37,6 +37,7 @@ const Setup = function(_model) {
 
     this.destroy = function() {
         clearTimeout(_setupFailureTimeout);
+        _model.set('_destroyed', true);
         _model = null;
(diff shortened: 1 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2895, opened 2018-05-07
 Trigger "visualQuality" from html5 provider with audio streams

robwalch:
### This PR will...

 Trigger "visualQuality" from html5 provider with audio streams.

When the html5 provider is used to play an audio-only HLS stream in Safari, or mp4/AAC file in Chrome, the player should still fire a "visualQuality" event once playback has begun.

These changes make sure that we always do fire a "visualQuality" event where before we did not, and is careful not to fire one with a width and height of `0` for HLS video streams still loading in Safari.

#### Addresses Issue(s):

JW8-1544


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2896, opened 2018-05-08
Fix for IE/Edge play promise polyfill

robwalch:
### This PR will...

Ignore "abort" events fired before the "play" in in the IE/Edge play promise polyfill.

IE and Edge do not return a Promise from `mediaElement.play()`. In this case, we create a promise that resolves when playback begins with the "playing" event, or is rejected if playback is interrupted by an "abort", "error" or "pause" event - the play promise polyfill.

The problem is that IE and Edge respond to, and notify us, of changes on the media element asynchronously. Even though we call `mediaElement.play()` and create this promise after changing the `mediaElement.src`, the "abort" event fires after the promise is created, making it look like the play attempt failed. Fortunately "play" is also always fired asynchronously after `mediaElement.play()`, with these other events in order. So as long as we wait for the "play" event to fire first, the play promise polyfill works correctly.

We should verify supported mobile platforms that are missing play promise support (Android browser, and iOS 8.4?). If calling `mediaElement.play()` fires the "play" event synchronously on those platforms, this change could result in the promise not resolving.

#### Addresses Issue(s):

JW8-1483

### DIFF
--- a/src/js/providers/utils/play-promise.js
+++ b/src/js/providers/utils/play-promise.js
@@ -6,10 +6,17 @@ export default function createPlayPromise(video) {
             return reject(new Error('Play refused.'));
         }
         const removeEventListeners = function() {
+            video.removeEventListener('play', playListener);
             video.removeEventListener('playing', listener);
             video.removeEventListener('pause', listener);
             video.removeEventListener('abort', listener);
             video.removeEventListener('error', listener);
+        };
+        const playListener = function() {
+            video.addEventListener('playing', listener);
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2897, opened 2018-05-08
Set fullscreen on change event for screen.orientation

DanFerrer:
### This PR will...

Attach an event listener to screen.orientation to set fullscreen using the api when the viewer switches from portrait to landscape (works turning your phone clockwise and counter clockwise). Sets it to false when going from landscape to portrait 

- Player must be at least 75% visible
- Player exits from fullscreen if video is paused 

#### Addresses Issue(s):

JW8-1380

### DIFF
--- a/src/js/view/utils/views-manager.js
+++ b/src/js/view/utils/views-manager.js
@@ -1,9 +1,9 @@
 import activeTab from 'utils/active-tab';
 import { requestAnimationFrame, cancelAnimationFrame } from 'utils/request-animation-frame';
-import { Browser, OS } from 'environment/environment';
 
 const views = [];
 const observed = {};
+const hasOrientation = 'screen' in window && 'orientation' in window.screen;
 
 let intersectionObserver;
 let responsiveRepaintRequestId = -1;
@@ -57,9 +57,7 @@ function onOrientationChange() {
                 // Set fullscreen to false when going back to portrait while paused and return early
                 view.api.setFullscreen(false);
                 return;
-            }
-
-            if (state === 'playing') {
+            } else if (state === 'playing') {
                 view.api.setFullscreen(isLandscape);
             }
         }
@@ -77,7 +75,7 @@ document.addEventListener('webkitvisibilitychange', onVisibilityChange);
 window.addEventListener('resize', scheduleResponsiveRedraw);
 window.addEventListener('orientationchange', scheduleResponsiveRedraw);
 
-if (OS.android && Browser.chrome) {
+if (hasOrientation) {
     window.screen.orientation.addEventListener('change', onOrientationChange);
 }
 
@@ -87,7 +85,7 @@ window.addEventListener('beforeunload', () => {
     window.removeEventListener('resize', scheduleResponsiveRedraw);
     window.removeEventListener('orientationchange', scheduleResponsiveRedraw);
 
-    if (OS.android && Browser.chrome) {
+    if (hasOrientation) {
         window.screen.orientation.removeEventListener('change', onOrientationChange);
     }
 });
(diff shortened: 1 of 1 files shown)


src/js/view/utils/views-manager.js
```diff
@@ -56,11 +77,19 @@ document.addEventListener('webkitvisibilitychange', onVisibilityChange);
 window.addEventListener('resize', scheduleResponsiveRedraw);
 window.addEventListener('orientationchange', scheduleResponsiveRedraw);
 
+if (OS.android && Browser.chrome) {
+    window.screen.orientation.addEventListener('change', onOrientationChange);
```
robwalch: `OS.android && Browser.chrome` Does not guarantee that `window.screen.orientation` is defined. 

`'screen' in window && 'orientation' in window.screen`

src/js/view/utils/views-manager.js
```diff
@@ -56,11 +77,19 @@ document.addEventListener('webkitvisibilitychange', onVisibilityChange);
...
+}
+
 window.addEventListener('beforeunload', () => {
     document.removeEventListener('visibilitychange', onVisibilityChange);
     document.removeEventListener('webkitvisibilitychange', onVisibilityChange);
     window.removeEventListener('resize', scheduleResponsiveRedraw);
     window.removeEventListener('orientationchange', scheduleResponsiveRedraw);
+
+    if (OS.android && Browser.chrome) {
```
robwalch: At this point you don't need to check the user-agent. Just make sure `window.screen.orientation` is defined (see above).

src/js/view/utils/views-manager.js
```diff
@@ -45,6 +46,26 @@ function scheduleResponsiveRedraw() {
...
+    views.forEach(view => {
+        if (view.model.get('visibility') >= 0.75) {
+            const state = view.model.get('state');
+            const orientation = window.screen.orientation.type;
+            const isLandscape = orientation === 'landscape-primary' || orientation === 'landscape-secondary';
+
+            if (!isLandscape && state === 'paused') {
+                // Set fullscreen to false when going back to portrait while paused and return early
+                view.api.setFullscreen(false);
+                return;
```
robwalch: Just `} else if (state === 'playing') {`

src/js/view/utils/views-manager.js
robwalch: I didn't say remove the Android Chrome check. We still need to gate this feature. The point is that you still need to guard from writing JS that can throw exceptions.

src/js/view/utils/views-manager.js
```diff
@@ -56,11 +75,19 @@ document.addEventListener('webkitvisibilitychange', onVisibilityChange);
 window.addEventListener('resize', scheduleResponsiveRedraw);
 window.addEventListener('orientationchange', scheduleResponsiveRedraw);
 
+if (hasOrientation) {
```
robwalch: This feature is still only supposed to be for Android Chrome

src/js/view/utils/views-manager.js
```diff
@@ -45,6 +48,24 @@ function scheduleResponsiveRedraw() {
...
+function onOrientationChange() {
+    views.forEach(view => {
+        if (view.model.get('visibility') >= 0.75) {
+            const state = view.model.get('state');
+            const orientation = window.screen.orientation.type;
+            const isLandscape = orientation === 'landscape-primary' || orientation === 'landscape-secondary';
+
+            if (!isLandscape && state === 'paused') {
+                // Set fullscreen to false when going back to portrait while paused and return early
+                view.api.setFullscreen(false);
```
robwalch: Shouldn't you check if this api is in fullscreen mode first? What if it isn't but another visible one is?

src/js/view/utils/views-manager.js
```diff
@@ -45,6 +48,24 @@ function scheduleResponsiveRedraw() {
...
+                return;
+            } else if (state === 'playing') {
+                view.api.setFullscreen(isLandscape);
```
robwalch: Since only one player can be fullscreen at a time, you should return early here. It would not be good to call `setFullscreen(true)` on two api instances at once.

PR #2898, opened 2018-05-08
Set tabindex to -1 on video element

DanFerrer:
### This PR will...

Explicitly set the tabindex to -1 on video element to make sure it isnt reachable when tabbing through the player

Firefox assumes the next element should be focusable sequentially even though tabindex isn't specified while chrome leaves it alone. Chrome has the desired behavior so to keep it consistent, we should explicitly set it with the value -1.

#### Addresses Issue(s):

JW8-935

### DIFF
--- a/src/js/program/media-element-pool.js
+++ b/src/js/program/media-element-pool.js
@@ -92,6 +92,7 @@ function createMediaElement() {
     const mediaElement = document.createElement('video');
 
     mediaElement.className = 'jw-video jw-reset';
+    mediaElement.setAttribute('tabindex', '-1');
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2900, opened 2018-05-09
Use a single matrix transform to resize video in IE/Edge

robwalch:
### This PR will...

Scale the video without repositioning it so that our object-fit workaround always renders correctly.

Changing left and top to 50% and then translating x and y resulted in video being rendered incorrectly intermittently.

### Other notes

Setting object-fit to the default 'contain' ensures this code path works regardless of whether or not object-fit is supported in video tags - so you can test it in Chrome by setting `fitVideoUsingTransforms` to `true`, and if Microsoft gets around to adding support, we won't over or under scale the video.

https://github.com/jwplayer/jwplayer-commercial/pull/5225

#### Addresses Issue(s):

JW8-1475


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2903, opened 2018-05-11
Play pause video on Android to prime tags already preloading

robwalch:
### This PR will...
Change our video tag priming workaround for Android so that it does not empty the video tag.

This PR build should be used to verify if this fix addresses issues we've found on Android Chrome with preroll content, and MSE players and/or DRM content.

So that preloading is not thrown away, and internal players do not have to setup twice on Android.

Pausing asynchronously when the play promise resolves could be problematic, if the promise resolves late, like once the video is actually expected to play. We might want to listen for additional calls to play and ignore this handler if that happens.

#### Addresses Issue(s):

JW8-1584 JW8-1594

### DIFF
--- a/src/js/program/media-element-pool.js
+++ b/src/js/program/media-element-pool.js
@@ -73,17 +73,17 @@ function primeMediaElementForPlayback(mediaElement) {
         mediaElement.load();
     } else if (OS.android && !mediaElement.parentNode) {
         // If the player sets up without a gesture and preloads, the background tag may not be primed for playback.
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2904, opened 2018-05-13
Volume & Seek Slider Accessibility

hghazzi:
### This PR will...

- Add the `slider` role to the time slider so that VoiceOver will read out the `aria-valuetext` attribute whenever it is in focus and updated
- Add the `status` role to the volume button so that VoiceOver will read out the `aria-label` attribute when it is updated, even when not in focus
- Make the time slider be focusable
- Remove the time slider focus ring
- create a `setAttribute` util method in `dom.js`

To improve player accessibility, we want users who have VoiceOver enabled to be aware of the current volume when interacting with the volume slider, and the current position when interacting with the seek slider. 

- For VoiceOver will only read out the `aria-valuetext` attribute when the element is focused

#### Addresses Issue(s):

JW8-1382 & JW8-1381

### DIFF
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -1,5 +1,8 @@
 import { throttle, each } from 'utils/underscore';
-import utils from 'utils/helpers';
+import { between } from 'utils/helpers';
+import { style } from 'utils/css';
+import { timeFormat } from 'utils/parser';
+import { addClass, removeClass, setAttribute, bounds } from 'utils/dom';
 import UI, { getPointerType } from 'utils/ui';
 import Slider from 'view/controls/components/slider';
 import Tooltip from 'view/controls/components/tooltip';
@@ -25,8 +28,8 @@ class TimeTip extends Tooltip {
         this.addContent(wrapper);
     }
 
-    image(style) {
-        utils.style(this.img, style);
+    image(styles) {
+        style(this.img, styles);
     }
 
     update(txt) {
@@ -53,7 +56,7 @@ class TimeTip extends Tooltip {
             return;
         }
 
-        this.containerWidth = utils.bounds(this.container).width + tolerance;
+        this.containerWidth = bounds(this.container).width + tolerance;
     }
 
     resetWidth () {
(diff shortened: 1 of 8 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/controlbar.js
```diff
@@ -343,6 +346,7 @@ export default class Controlbar {
             this.elements.volumetooltip.volumeSlider.render(muted ? 0 : vol);
             utils.toggleClass(this.elements.volumetooltip.element(), 'jw-off', muted);
             utils.toggleClass(this.elements.volumetooltip.element(), 'jw-full', vol >= 75 && !muted);
+            this.elements.volumetooltip.element().setAttribute('aria-label', `volume ${muted ? 0 : vol}%`);
```
robwalch: If you keep accessing the same property more than three times and these are lengthy strings that won't minify, it's a good rule of thumb to put them in a var, and replace all instances of them in the function:
```js
const mute = this.elements.mute;
const volumetooltip = this.elements.volumetooltip;
if (mute) {
utils.toggleClass(mute, 'jw-off', muted);
...
if (volumetooltip) {
  volumeSlider.render(muted ? 0 : vol);
...

File: src/js/view/controls/components/timeslider.js [hunk 2]
```diff
@@ -96,12 +96,18 @@ class TimeSlider extends Slider {
             .change('buffer', this.onBuffer, this)
             .change('streamType', this.onStreamType, this);
 
+
+        this.el.setAttribute('tabindex', '0');
+        this.el.setAttribute('role', 'slider');
+        this.el.setAttribute('aria-label', 'Time Slider');
```
robwalch: Seems like we could make things a lot of space by adding a static `setAttribute` helper .

hghazzi: Could you clarify

robwalch: The three lines above will not be minified at all. This will:
```js
function setAttribute(element, attribute, value) {
    element.setAttribute(attribute, value);
}
....
        const element = this.el;
        setAttribute(element, 'tabindex', '0');
        setAttribute(element, 'role', 'slider');
        setAttribute(element, 'aria-label', 'Time Slider');
```

To something more like
```js
function a(b, c, d) {
    b.setAttribute(c, d);
}
....
        var x = this.el;
        a(x, 'tabindex', '0');
        a(x, 'role', 'slider');
        a(x, 'aria-label', 'Time Slider');
```

hghazzi:
test this please

PR #2913, opened 2018-05-15
Only preload video on Android when safe

robwalch:
### This PR will...

Only preload video on Android, if the media element does not need to be primed for playback. The autostart test is used to determine if priming is required.

On Android in Chrome media elements do not always play without `video.load()` being called on a user-gesture event. It's completely unclear how this decision to require an interaction is made between versions of Chrome and versions of Android (Chrome 32 always, Chrome 66 media score... who knows).

In a situation where Chrome will block playback, we cannot preload media (set `video.src` with blob and prepare MSE) until the user has clicked to play. We can still detect whether or not Chrome will allow playback, and if it does, preload normally.

#### Addresses Issue(s):

JW8-1609


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2915, opened 2018-05-17
Set player element role to "application"

hghazzi:
### This PR will...

set the jwplayer element's role to "application" so that any keystrokes that occur within the player are not intercepted by screen readers.

Screen readers such as JAWS will intercept the ENTER and SPACE key, stopping the player from pausing or playing. By adding `role="application"`, it tells the screen reader not to do so. 

Using `role="application"` should be used sparingly and wisely. According to the spec, it seems to make sense that the player is considered an application.

https://www.w3.org/TR/aria-in-html/#using-application

```
You only want to use role=application if the content you’re providing consists of only focusable, 
interactive controls, and of those, mostly advanced widgets that emulate a real desktop application. 
```

#### Addresses Issue(s):

JW8-1614

### DIFF
--- a/src/templates/player.js
+++ b/src/templates/player.js
@@ -1,6 +1,6 @@
 export default (id, ariaLabel = '') => {
     return (
-        `<div id="${id}" class="jwplayer jw-reset jw-state-setup" tabindex="0" aria-label="${ariaLabel}">` +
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2921, opened 2018-05-21
Maintain cursor after setControls(false), except in fullscreen mode

boyntoni:
### This PR will...

Result in parity between setting ```controls: false``` in the player configuration object, and using the ```setControls(false)``` API method (with the exception of Fullscreen mode). 

Parity in UX

No

No

#### Addresses Issue(s):

JW8-1626


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/css/controls/flags/controls-hidden.less
```diff
@@ -10,6 +10,14 @@
         top: 0;
     }
 
+    .jw-media {
+        cursor: none;
+    }
+
+    &:not(.jw-flag-fullscreen) .jw-media {
+        cursor: default;
+    }
```
robwalch: Have you looked in "src/css/jwplayer/flags/controls-hidden.less"?

What isn't this handled there once and for all? Not specific enough? Looks like it should always be default.

boyntoni: @robwalch  This is in that file -- 

The reason for the two rules is that, per your feedback on the last PR, my understanding was that we do not want a pointer in Fullscreen (this mirrors the behavior of setting ```controls: false``` in the configuration). So I wanted to ensure the cursor was visible in all instances, but fullscreen.

robwalch: "src/css/controls/flags/controls-hidden.less" does not equal
"src/css/jwplayer/flags/controls-hidden.less"

radium-v: @boyntoni Following up, the reason for the split in files is that `controls` is loaded only if `controls` are enabled. The code in `jwplayer` is always loaded as critical CSS. The fix will likely need to be addressed in both locations, or removed from `controls` completely since the styles may be cascading over each other.

boyntoni:
@robwalch In speaking w/ @egreaves about this, added an additional rule to the user-inactive file to prevent ```cursor: none``` from taking effect when the ```jw-flag-controls-hidden``` flag is set.

I also went to confirm the fullscreen behavior when using ```controls: false``` in the configuration. We do maintain the cursor in fullscreen in this use case, so maintained the same behavior for when customer uses ```setControls(false)```.

PR #2922, opened 2018-05-22
Read out time and volume once per focus 

hghazzi:
### This PR will...

- Trigger the `aria-valuetext` to be updated after a seek occurs or when the time slider has focus
- The volume will now be read out only when the volume button is in focus, making it more consistent to the time slider

When using a screen reader, if the `aria-valuetext` attribute is updated while an element is in focus, the text will be read out loud. We do not want it to read out load every second, but only when it becomes in focus or after a seek occurs.

#### Addresses Issue(s):

JW8-1382

### DIFF
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -108,9 +108,17 @@ class TimeSlider extends Slider {
 
         // Show the tooltip on while dragging (touch) moving(mouse), or moving over(mouse)
         this.elementUI = new UI(this.el, { useHover: true, useMove: true })
-            .on('drag move over', this.showTimeTooltip.bind(this), this)
-            .on('dragEnd out', this.hideTimeTooltip.bind(this), this)
+            .on('drag move over', this.showTimeTooltip, this)
+            .on('out', this.hideTimeTooltip, this)
             .on('click', () => this.el.focus());
+
+        this.el.addEventListener('focus', () => this.updateAriaText());
+        this._model.on('seeked', () => {
+            if (this._model.get('scrubbing')) {
+                return;
+            }
+            this.updateAriaText();
+        });
     }
 
     update(percent) {
@@ -156,11 +164,9 @@ class TimeSlider extends Slider {
                 const diff = duration + dvrSeekLimit;
                 const pos = position + dvrSeekLimit;
                 pct = (diff - pos) / diff * 100;
-                setAttribute(this.el, 'aria-valuetext', timeFormat(pos, true));
             } else if (this.streamType === 'VOD' || !this.streamType) {
                 // Default to VOD behavior if streamType isn't set
                 pct = position / duration * 100;
-                setAttribute(this.el, 'aria-valuetext', `${timeFormat(position)} of ${timeFormat(duration)}`);
             }
         }
         this.render(pct);
@@ -269,11 +275,8 @@ class TimeSlider extends Slider {
         style(timeTip.el, { left: safePct + '%' });
     }
 
-    hideTimeTooltip(evt) {
+    hideTimeTooltip() {
         removeClass(this.timeTip.el, 'jw-open');
-        if (evt.type === 'dragEnd') {
-            this.el.focus();
-        }
     }
 
     addCues(model, cues) {
@@ -284,6 +287,19 @@ class TimeSlider extends Slider {
             });
             this.drawCues();
         }
+    }
+
+    updateAriaText() {
+        const position = this._model.get('position');
+        const duration = this._model.get('duration');
+        let ariaText;
+
+        if (this.streamType === 'DVR') {
+            ariaText = timeFormat(position);
+        } else {
+            ariaText = `${timeFormat(position)} of ${timeFormat(duration)}`;
+        }
+        setAttribute(this.el, 'aria-valuetext', ariaText);
     }
 
     reset() {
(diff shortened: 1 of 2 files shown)


src/js/view/controls/controlbar.js
```diff
@@ -352,7 +352,7 @@ export default class Controlbar {
             volumeTooltip.volumeSlider.render(muted ? 0 : vol);
             toggleClass(volumeTooltipEl, 'jw-off', muted);
             toggleClass(volumeTooltipEl, 'jw-full', vol >= 75 && !muted);
-            setAttribute(volumeTooltipEl, 'aria-label', `volume ${muted ? 0 : vol}%`);
+            setAttribute(volumeTooltipEl, 'aria-label', `Volume ${muted ? 0 : vol}%`);
```
hghazzi: To fix automated tests, the `V` needs to be capitalized.

src/js/view/controls/components/timeslider.js
```diff
@@ -286,6 +294,27 @@ class TimeSlider extends Slider {
...
+
+    updateAriaText(evt) {
+        if (evt && evt.type === 'keydown' &&
+            evt.key !== 'ArrowLeft' &&
+            evt.key !== 'ArrowRight') {
+            return;
+        }
+
+        // wait for seek to complete before updating aria text
+        setTimeout(() => {
```
robwalch: If you add a timeout you have to cancel it when appropriate.

This doesn't make sense. A timeout doesn't not mean a seek completed. If you need to update a value based on change, use the model.

src/js/view/controls/components/timeslider.js
```diff
@@ -109,8 +109,16 @@ class TimeSlider extends Slider {
         // Show the tooltip on while dragging (touch) moving(mouse), or moving over(mouse)
         this.elementUI = new UI(this.el, { useHover: true, useMove: true })
             .on('drag move over', this.showTimeTooltip.bind(this), this)
-            .on('dragEnd out', this.hideTimeTooltip.bind(this), this)
+            .on('out', this.hideTimeTooltip.bind(this), this)
```
robwalch: You shouldn't need `bind(this)` when passing `this` as the listener context (third param) for `on`:

`.on('event', fn.bind(this))` is the same as `.on('event', fn, this)` and it's best to avoid `bind` since it can be replaced on `Function.prototype`.

src/js/view/controls/components/timeslider.js
```diff
@@ -108,9 +108,17 @@ class TimeSlider extends Slider {
...
         // Show the tooltip on while dragging (touch) moving(mouse), or moving over(mouse)
         this.elementUI = new UI(this.el, { useHover: true, useMove: true })
-            .on('drag move over', this.showTimeTooltip.bind(this), this)
-            .on('dragEnd out', this.hideTimeTooltip.bind(this), this)
+            .on('drag move over', this.showTimeTooltip, this)
+            .on('out', this.hideTimeTooltip, this)
             .on('click', () => this.el.focus());
+
+        this.el.addEventListener('focus', (e) => this.updateAriaText(e));
+        this._model.on('seeked', () => {
```
robwalch: Use the model argument, instead of `this._model` inside this callback:

```js
this._model.on('seeked', (model) => {
    if (model.get('scrubbing')) {
```

src/js/view/controls/components/timeslider.js
```diff
@@ -108,9 +108,17 @@ class TimeSlider extends Slider {
 
         // Show the tooltip on while dragging (touch) moving(mouse), or moving over(mouse)
         this.elementUI = new UI(this.el, { useHover: true, useMove: true })
-            .on('drag move over', this.showTimeTooltip.bind(this), this)
-            .on('dragEnd out', this.hideTimeTooltip.bind(this), this)
+            .on('drag move over', this.showTimeTooltip, this)
+            .on('out', this.hideTimeTooltip, this)
             .on('click', () => this.el.focus());
+
+        this.el.addEventListener('focus', (e) => this.updateAriaText(e));
```
robwalch: The event argument `e` is not being used in `updateAriaText` and should be removed here.

src/js/view/controls/components/timeslider.js
```diff
@@ -286,6 +289,18 @@ class TimeSlider extends Slider {
...
+    updateAriaText() {
+        const position = this._model.get('position');
+        const duration = this._model.get('duration');
+
+        if (this.streamType === 'DVR') {
+            this.ariaText = timeFormat(position);
+        } else {
+            this.ariaText = `${timeFormat(position)} of ${timeFormat(duration)}`;
+        }
+        setAttribute(this.el, 'aria-valuetext', this.ariaText);
```
robwalch: Why is `ariaText` a property of `this`? Can we make this a `let`?

src/js/view/controls/components/timeslider.js
hghazzi: @robwalch this is a trigger, not a change, so the `{type: "seeked"}` is what is being passed into callback, not the model

hghazzi:
test this please

src/js/view/controls/components/timeslider.js
robwalch: Gotcha. Then just chain this listener with the other model listeners above.

```js
this._model
             .on('change:duration', this.onDuration, this)
	    .on(...
```

PR #2923, opened 2018-05-22
Create network error parser module & test

johnBartos:
### This PR will...
- Break out the `jwplayer-errors` repo into a `network-error-parser` module
- Avoid passing `message` via `super()` call so that we can define an enumerable `message` property on `PlayerError`

- So that we can keep all standardized error codes in OS and Commercial. 
- It's a small module & easier to maintain as part of this project
- So that we preserve the prototype chain of events; for errors cases this is important (the `message` property is lost if copied via `Object.assign`).

Does the new event creation method introduce any breaking changes into our API?

https://github.com/jwplayer/jwplayer-commercial/pull/5268

#### Addresses Issue(s):

 JW8-1554

### DIFF
--- a/src/js/api/errors.js
+++ b/src/js/api/errors.js
@@ -78,8 +78,10 @@ export const FLASH_MEDIA_ERROR = 214000;
  * @param {code} [ErrorCode] - The error code.
  * @param {sourceError} [Error] - The lower level error, caught by the player, which resulted in this error.
  */
-export class PlayerError {
+export class PlayerError extends Error {
     constructor(message, code, sourceError = null) {
+        // Avoid passing message via super so that we can define an enumerable message property on PlayerError
+        super();
         this.code = isValidNumber(code) ? code : null;
         this.message = message;
         this.sourceError = sourceError;


src/js/providers/utils/network-error-parser.js
```diff
@@ -0,0 +1,33 @@
+export default function parseNetworkError(baseCode, statusCode, url = '', responseText) {
+    let message;
+    let code = baseCode + 1000;
+
+    if (statusCode > 0) {
+        message = badStatusMessage(statusCode, responseText);
+        code += statusCode;
```
karimJWP: if `statusCode` is out of the 400-600 range we want to assign it a `010` code. I know it's a very unlikely scenario but I want to ensure our codes never conflict.

johnBartos: ok

src/js/api/errors.js
```diff
@@ -78,10 +78,10 @@ export const FLASH_MEDIA_ERROR = 214000;
  * @param {code} [ErrorCode] - The error code.
  * @param {sourceError} [Error] - The lower level error, caught by the player, which resulted in this error.
  */
-export class PlayerError extends Error {
+export class PlayerError {
```
karimJWP: What is the reasoning behind this change ? We want inheritance to leverage existing error features like stack trace, and we want a consistent interface to make our errors compatible with any technologies that handle javascript errors.

johnBartos: It's because of the way `program-listeners.js` handles provider events. `error.message` isn't enumerable and is lost when the event is created via `Object.assign`; we don't want to change this (to use `Object.create`, for example) because we will potentially change our API.

johnBartos: Disregard. If we avoid passing message via `super` we're allowed to define an enumerable `message` property on our `PlayerError`.

src/js/api/errors.js
```diff
@@ -80,8 +80,10 @@ export const FLASH_MEDIA_ERROR = 214000;
  */
 export class PlayerError extends Error {
     constructor(message, code, sourceError = null) {
-        super(message);
+        // Avoid passing message via super so that we can define an enumerable message property on PlayerError
+        super();
```
robwalch: PlayerError extends Error

PR #2931, opened 2018-05-29
Implement html5 standard errors

johnBartos:
### This PR will...
- Implement `live stream down or ended` error
- Implement media element errors

So that we can standardize & track html5 errors

No

https://github.com/jwplayer/jwplayer-commercial/pull/5289

#### Addresses Issue(s):

JW8-1553

### DIFF
--- a/src/js/providers/video-listener-mixin.js
+++ b/src/js/providers/video-listener-mixin.js
@@ -164,17 +164,17 @@ const VideoListenerMixin = {
         }
     },
     error() {
-        const code = (this.video.error && this.video.error.code) || -1;
-        const message = ({
-            1: 'Unknown operation aborted',
-            2: 'Unknown network error',
-            3: 'Unknown decode error',
-            4: 'File could not be played'
-        }[code] || 'Unknown');
+        const errorCode = (this.video.error && this.video.error.code) || -1;
+        const [message, code] = ({
+            1: ['Unknown operation aborted', 1],
+            2: ['Unknown network error', 0],
+            3: ['Unknown decode error', 2],
+            4: ['File could not be played', 3]
+        }[errorCode] || ['Unknown', 0]);
 
         this.trigger(
             MEDIA_ERROR,
-            new PlayerError(`Error loading media ${message}`, HTML5_BASE_MEDIA_ERROR + Math.max(code, 0))
+            new PlayerError(`Error loading media ${message}`, HTML5_BASE_MEDIA_ERROR + code, this.video.error)
         );
     }
 };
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/providers/video-listener-mixin.js
```diff
@@ -170,11 +171,18 @@ const VideoListenerMixin = {
...
             4: 'File could not be played'
         }[code] || 'Unknown');
-        this.trigger(MEDIA_ERROR, {
-            code: code,
-            message: 'Error loading media: ' + message
-        });
+
+        this.trigger(
+            MEDIA_ERROR,
+            new PlayerError(`Error loading media ${message}`, HTML5_BASE_MEDIA_ERROR + Math.max(code, 0))
```
karimJWP: @johnBartos the old codes don't map 1:1 to the new ones; `'Unknown network error'` is `221,000`, reason being it falls in the network category. The other codes are also different

karimJWP: also `this.video.error` should be passed in as sourceError

src/js/providers/video-listener-mixin.js
```diff
@@ -163,18 +164,25 @@ const VideoListenerMixin = {
...
-            4: 'File could not be played'
-        }[code] || 'Unknown');
-        this.trigger(MEDIA_ERROR, {
-            code: code,
-            message: 'Error loading media: ' + message
-        });
+        const errorCode = (this.video.error && this.video.error.code) || -1;
+        const [message, code] = ({
+            1: ['Unknown operation aborted', 1],
+            2: ['Unknown network error', 0],
```
karimJWP: the resulting error for `'Unknown network error'` should be `221,000`; so it's not in the 224,XXX range. Since we know it's a network error, we group it in that range

robwalch: A `video.error.code` of `2` maps to `221000` (Unknown network error).

`video.error.code` 1, 3, 4 and other are correct `224xxx` (media errors).

src/js/providers/video-listener-mixin.js
```diff
@@ -163,18 +164,34 @@ const VideoListenerMixin = {
...
-            1: 'Unknown operation aborted',
-            2: 'Unknown network error',
-            3: 'Unknown decode error',
-            4: 'File could not be played'
-        }[code] || 'Unknown');
-        this.trigger(MEDIA_ERROR, {
-            code: code,
-            message: 'Error loading media: ' + message
-        });
+        const errorCode = (this.video.error && this.video.error.code) || -1;
```
robwalch: `|| 0` since we're pulling from an Array now.

I think a little switch call to a static function that triggers the error would  be a lot simpler and easier to read.

PR #2935, opened 2018-05-30
Implement standard error logging 

johnBartos:
### This PR will...
- Log `errorEvent` changes in core-shim, if the error is a Player error
- Implement function to generate a log message from a Player error

So that developers reading the console can be directed to documentation about any thrown errors

Should I add more guard checks to `get logMessage`?

No

#### Addresses Issue(s):

JW8-1542

### DIFF
--- a/src/js/api/errors.js
+++ b/src/js/api/errors.js
@@ -78,13 +78,21 @@ export const FLASH_MEDIA_ERROR = 214000;
  * @param {code} [ErrorCode] - The error code.
  * @param {sourceError} [Error] - The lower level error, caught by the player, which resulted in this error.
  */
-export class PlayerError extends Error {
+export class PlayerError {
     constructor(message, code, sourceError = null) {
-        // Avoid passing message via super so that we can define an enumerable message property on PlayerError
-        super();
-        this.code = isValidNumber(code) ? code : null;
+        this.code = isValidNumber(code) ? code : 0;
         this.message = message;
         this.sourceError = sourceError;
+    }
+
+    get logMessage() {
+        let codeStr = this.code.toString();
+        const prefix = codeStr.slice(0, 3);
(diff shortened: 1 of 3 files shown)


src/js/api/errors.js
```diff
@@ -78,15 +78,23 @@ export const FLASH_MEDIA_ERROR = 214000;
...
+        this.code = isValidNumber(code) ? code : 0;
         this.message = message;
         this.sourceError = sourceError;
     }
 
+    get logMessage() {
+        let codeStr = this.code.toString();
+        const prefix = codeStr.slice(0, 3);
+        const suffix = codeStr.slice(-3);
+        if (suffix >= 400 && suffix < 600) {
```
robwalch: We should avoid using gt/lt operators to compare strings to numbers. 

mod should do the trick:
```js
> 623404 % 1000
404
> 623404 - 404
623000
```

johnBartos:
test this please

src/js/api/errors.js
```diff
@@ -78,15 +78,25 @@ export const FLASH_MEDIA_ERROR = 214000;
  * @param {code} [ErrorCode] - The error code.
  * @param {sourceError} [Error] - The lower level error, caught by the player, which resulted in this error.
  */
-export class PlayerError extends Error {
+export class PlayerError {
     constructor(message, code, sourceError = null) {
-        // Avoid passing message via super so that we can define an enumerable message property on PlayerError
-        super();
-        this.code = isValidNumber(code) ? code : null;
+        this.code = isValidNumber(code) ? code : 0;
```
robwalch: @johnBartos Note that this would result in errors with code `0` being pinged back by our analytics plugin. Since we should not use `PlayerError` without specifying a code, I'm OK with this, since the data would indicate exactly that issue.

PR #2941, opened 2018-05-31
[ADS-1002] Emit preload event for next playlistItem.

vseventer:
### This PR will...
Emit a "preload" event when background loading a playlistItem.

Signal to the ads plugins to preload ads for the playlistItem.

@robwalch Do we want the event to have a more specific name? 

N/A

#### Addresses Issue(s):
ADS-1002


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -188,6 +188,7 @@ Object.assign(Controller.prototype, {
                     if (allowPreload && position && duration > 0 && position >= duration - BACKGROUND_LOAD_OFFSET) {
                         mediaModel.off('change:position', onPosition, this);
                         _programController.backgroundLoad(item);
+                        _this.trigger(PRELOAD, { index, item });
```
robwalch: This matches the "playlistItem" event format correct?

robwalch: Maybe this should be triggered in `_programController.backgroundLoad`. I believe this controller forwards all program events. This way we'll always get a "preload" event when `backgroundLoad` is called successfully.

robwalch:
> @robwalch Do we want the event to have a more specific name?

@vseventer suggestions? I like "preload" but I think it should also apply to `_programController.preloadVideo`, not just `_programController.backgroundLoad`.

vseventer:
Benching this idea - see JW8-1660.

PR #2946, opened 2018-06-01
Output friendly error messages

egreaves:
### This PR will...
- Add friendly error messages to all instances of `PlayerError`
- Localize friendly error messages, making them customizable
To reduce the number of variantes of error messages that get displayed to the user and allow publishers to localize the error messages.
Seems Chai doesn't handle custom error assertions well, so I made [this](https://github.com/jwplayer/jwplayer/pull/2946/files#diff-b3ee4f02a5ebb4014e652dd0fea88581R187) generic since the returned value is a `PlayerError` that contains a key property and not a message. Chai allows message or strict (===) equality comparisons. Open to suggestions for improvement.

Yes: https://github.com/jwplayer/jwplayer-commercial/pull/5310
#### Addresses Issue(s):
JW8-1366
JW8-1367


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/playlist/loader.js
```diff
@@ -53,7 +53,7 @@ const PlaylistLoader = function() {
                         throw Error('Playlist is not an array');
                     }
                 } catch (e) {
-                    throw new PlayerError('Not a valid RSS/JSON feed', 621, e);
+                    throw new PlayerError(MSG_CANT_LOAD_PLAYER, 621, e);
```
karimJWP: I think `cantPlayVideo ` would be a better message because the problem is the media files

src/js/playlist/loader.js
```diff
@@ -65,9 +65,8 @@ const PlaylistLoader = function() {
 
     function playlistError(error) {
         if (!error.code) {
-            error = new PlayerError(error ? error : 'Error loading file', 0);
+            error = new PlayerError(MSG_CANT_LOAD_PLAYER, 0);
```
karimJWP: I think `cantPlayVideo` would be a better message because the problem is the media files

src/js/providers/video-listener-mixin.js
```diff
@@ -177,7 +177,7 @@ const VideoListenerMixin = {
 
         this.trigger(
             MEDIA_ERROR,
-            new PlayerError(`Error loading media ${errorData[0]}`, code, this.video.error)
+            new PlayerError(MSG_TECHNICAL_ERROR, code, this.video.error)
```
karimJWP: I think each error code should have a specific message; 
'File could not be played' -> `cantPlayVideo `
'Unknown decode error' -> `cantPlayVideo `
the rest sounds like `MSG_TECHNICAL_ERROR `

src/js/utils/ajax.js
```diff
@@ -150,7 +150,7 @@ function _requestError(message, options) {
 }
 
 function _error(options, message, code, error) {
-    options.onerror(message, options.url, options.xhr, new PlayerError(message, code, error));
+    options.onerror(message, options.url, options.xhr, new PlayerError(MSG_CANT_LOAD_PLAYER, code, error));
```
karimJWP: We can pass a specific key depending on the scenario; i.e. 'File not found' is really a `cantPlayVideo `

src/js/api/config.js
```diff
@@ -19,35 +19,45 @@ const Defaults = {
...
+        playbackRates: 'Playback Rates',
+        playlist: 'Playlist',
+        prev: 'Previous',
         related: 'More Videos',
-        close: 'Close',
+        replay: 'Replay',
+        rewind: 'Rewind 10 Seconds',
+        stop: 'Stop',
+        volume: 'Volume',
         settings: 'Settings',
```
karimJWP: @egreaves since we're alphabetizing:
settings, stop, unmute, videoInfo, volume

src/js/providers/video-listener-mixin.js
```diff
@@ -177,7 +177,7 @@ const VideoListenerMixin = {
 
         this.trigger(
             MEDIA_ERROR,
-            new PlayerError(`Error loading media ${errorData[0]}`, code, this.video.error)
+            new PlayerError(MSG_CANT_PLAY_VIDEO, code, this.video.error)
```
karimJWP: we are no longer using the strings in `errorData`, so we could get rid of them. 'Unknown network error' i think maps best with `badConnection`

karimJWP: we might want to rewrite this. Basically we map the HTML5 `video.error.code` to our own codes:

0->  224000
1->  224001
2-> 221000 (notice this is a `1000` level to indicate it is network specific, vs `4000` indicating it is media specific)
3 -> 224002
4 -> 224003
(> 4) -> 224000 (fallback to base code because the video element's error code is unknown)

egreaves: @karimJWP so 'File could not be played' is the only one that should be mapped to `MSG_CANT_PLAY_VIDEO ` now?

src/js/controller/controller.js
```diff
@@ -1028,11 +1028,14 @@ Object.assign(Controller.prototype, {
...
     },
     triggerError(evt) {
-        this._model.set('errorEvent', evt);
-        this._model.set('state', STATE_ERROR);
-        this._model.once('change:state', function() {
-            this._model.set('errorEvent', undefined);
-        }, this);
+        const model = this._model;
+        evt.message = model.get('localization').errors[evt.key];
+        delete evt.key;
```
johnBartos: `delete` causes a deoptimization in v8. I think we should set it to null if acceptable; or create a new object

johnBartos: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers

>An object will go into hash table mode for example when you add too many properties dynamically (outside constructor), delete properties, use properties that cannot be valid identifiers and so on. 

>Objects that are in "hash table mode" (aka "normalized objects", "dictionary mode" - objects who have a hash table as a backing data structure) are not simple enumerables**

src/js/api/core-shim.js
```diff
@@ -244,18 +244,22 @@ Object.assign(CoreShim.prototype, {
...
+        const model = core._model || core.modelShim;
+
         if (!isValidNumber(error.code)) {
             // Transform any unhandled error into a PlayerError so emitted events adhere to a uniform structure
-            errorEvent = new PlayerError(error.message, SETUP_ERROR_UNKNOWN, error);
+            errorEvent = new PlayerError(MSG_TECHNICAL_ERROR, SETUP_ERROR_UNKNOWN, error);
         }
+        // The message may have already been created (eg. multiple players on a page where a plugin fails to load)
+        errorEvent.message = errorEvent.message || model.get('localization').errors[errorEvent.key];
+        errorEvent.key = null;
```
robwalch: Let's no output undocumented properties with null values in our API events. @egreaves @johnBartos better to deop on error than degrade the developer experience.

PR #2947, opened 2018-06-01
Fix initial playback rate menu selection

robwalch:
### This PR will...

Remove the default parameter value of `0` from submenu `activateItem(itemOrdinal)`, and correct the value we pass in for this argument in the playback rate menu to be an index and not a playback rate.

Removing unnecessary default parameters keeps the build smaller.

The wrong playback rate could be selected by default in the rates menu if the second option does not match "1" or the default rate.


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2953, opened 2018-06-05
Cleanup default provider methods

robwalch:
### This PR will...
Fixes some of the defaults applied to all providers registered with the player:
- `setContainer` has no return value
- `supportsFullscreen` is required and should return `false` by default
- `setControls` should be deprecated with Flash. Clarify this with a comment. The methods following it are being used.
- 'oga' and 'vorbis' are not specific to audio and do not need to be included in our `sendMediaType` (which really belongs in a mixin, not here, but save that for another time)

Providers registered with the player should implement these methods and if they do not, these are the defaults.

### DIFF
--- a/src/js/providers/default.js
+++ b/src/js/providers/default.js
@@ -44,13 +44,14 @@ const DefaultProvider = {
     setVisibility: noop,
     setFullscreen: noop,
     getFullscreen: returnFalse,
+    supportsFullscreen: returnFalse,
 
     // If setContainer has been set, this returns the element.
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #2958, opened 2018-06-08
Standard errors overlay

johnBartos:
### This PR will...
- Implement a new standard overlay, which contains a scrollable container

To ensure that any length error message is readable at any player size.

Any further CSS cleanup?

No

#### Addresses Issue(s):

JW8-1395

### DIFF
--- a/src/css/jwplayer/states.less
+++ b/src/css/jwplayer/states.less
@@ -11,6 +11,10 @@
     }
 }
 
+span.jw-break {
+    display: block;
+}
+
 // errors
 body .jw-error,
 body .jwplayer.jw-state-error {
@@ -53,10 +57,6 @@ body .jwplayer.jw-state-error {
             }
         }
 
-        .jw-error-text {
-            align-self: center;
-        }
-
         .jw-info-container {
             margin: 0;
             padding: 0;
@@ -69,9 +69,30 @@ body .jwplayer.jw-state-error {
             .jw-error-msg {
                 flex-direction: column;
 
+                .jw-error-text {
+                    text-align: center;
+                }
+
                 .jw-icon {
+                    flex: 0.5 0 auto;
                     margin-right: 0;
                     margin-bottom: 20px;
+                }
+            }
+        }
+    }
+}
+
+.jwplayer.jw-state-error {
+    &.jw-flag-audio-player,
+    &.jw-flag-small-player,
+    &.jw-breakpoint-2 {
+        .jw-error-msg {
+            .jw-break {
+                display: inline;
+
+                &:before {
+                    content: " ";
                 }
             }
         }
@@ -87,8 +108,9 @@ body .jwplayer.jw-state-error {
         padding: 4px 16px;
         z-index: 1;
 
-        .jw-info-container {
-            justify-content: center;
+        &.jw-info-overlay {
+            max-width: none;
+            max-height: none;
         }
     }
 }
--- a/src/css/jwplayer/states.less
+++ b/src/css/jwplayer/states.less
@@ -60,7 +60,6 @@ body .jwplayer.jw-state-error {
         .jw-info-container {
             margin: 0;
             padding: 0;
-            justify-content: center;
         }
     }
 
@@ -71,6 +70,7 @@ body .jwplayer.jw-state-error {
                 flex-direction: column;
 
                 .jw-icon {
+                    margin-right: 0;
                     margin-bottom: 20px;
                 }
             }
@@ -86,6 +86,10 @@ body .jwplayer.jw-state-error {
         transform: none;
         padding: 4px 16px;
         z-index: 1;
+
+        .jw-info-container {
+            justify-content: center;
+        }
     }
 }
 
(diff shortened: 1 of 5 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

johnBartos:
test this please

src/templates/error.js
```diff
@@ -1,22 +1,20 @@
...
-                `[id="${id}"] .jw-error-msg{top:50%;left:50%;position:absolute;align-items:center;display:flex;transform:translate(-50%,-50%)}` +
-                `[id="${id}"] .jw-title{color:#FFF;position:static}` +
-                `[id="${id}"] .jw-title-primary,` +
-                `[id="${id}"] .jw-title-secondary{font:600 14px/1.35 Arial,Helvetica,sans-serif}` +
-                `[id="${id}"] .jw-title-secondary{font-weight:400}` +
+                `[id="${id}"].jw-error{background:#000;overflow:hidden;position:relative}` +
+                `[id="${id}"] .jw-error-msg{top:50%;left:50%;position:absolute;align-items:center;transform:translate(-50%,-50%)}` +
+                `[id="${id}"] .jw-error-container{display:flex;flex-direction:column;align-items:center}` +
+                `[id="${id}"] .jw-error-text{color:#FFF;position:static;font:600 14px/1.35 Arial,Helvetica,sans-serif}` +
+                `[id="${id}"] .jw-error-text{font-weight:400}` +
```
robwalch: Combine `.jw-error-text` rules into the first line `font:400 14px/1.35 Arial` (or just delete the font weight value: `font:14px/1.35 Arial`)

src/templates/error.js
```diff
@@ -1,22 +1,20 @@
 export default (id, message, errorCode, code) => {
-    const detail = code ? `(${errorCode}: ${code})` : '';
+    message += code ? ` (${errorCode}: ${code})` : '';
```
robwalch: Can you create a assign a new constant the new message value, instead of mutating `message`?

src/css/jwplayer/states.less
```diff
@@ -90,11 +95,6 @@ body .jwplayer.jw-state-error,
 body .jw-error,
 .jw-state-idle,
 .jwplayer.jw-state-complete:not(.jw-flag-casting):not(.jw-flag-audio-player):not(.jw-flag-overlay-open-related) {
-    // Show the title
-    .jw-title {
-        display: block;
-    }
```
robwalch: This is modifying the idle and complete states, not just error. Is that OK?

src/css/jwplayer/states.less
```diff
@@ -36,39 +35,45 @@ body .jwplayer.jw-state-error {
...
+            justify-content: center;
+        }
+    }
 
-            &-primary {
-                font-weight: 600;
-                white-space: normal;
+    &:not(.jw-flag-audio-player) {
+        &.jw-breakpoint-0,
+        &.jw-breakpoint-1,
```
robwalch: Can we `small-player` these?

src/css/jwplayer/states.less
```diff
@@ -20,7 +20,6 @@ body .jwplayer.jw-state-error {
     font-size: @default-em-size;
     display: table;
     opacity: 1;
-    overflow: hidden;
```
robwalch: Is `overflow: hidden` already applied elsewhere? A developer might be resizing the player in the error state specifically to hide the error from users.

src/templates/error.js
johnBartos: Will do

src/templates/error.js
johnBartos: Sure

src/css/jwplayer/states.less
johnBartos: Yes, in the `info-overlay` div. It doesn't make sense to have `overflow:hidden` on the outermost container because nothing inside can overflow.

robwalch: OK. We should have that on the player contain already anyway.

src/templates/error.js
```diff
@@ -1,22 +1,19 @@
...
+                `[id="${id}"] .jw-error-msg{top:50%;left:50%;position:absolute;align-items:center;transform:translate(-50%,-50%)}` +
+                `[id="${id}"] .jw-error-container{display:flex;flex-direction:column;align-items:center}` +
+                `[id="${id}"] .jw-error-text{color:#FFF;position:static;font:14px/1.35 Arial,Helvetica,sans-serif}` +
                 `</style>` +
                 `<div class="jw-icon jw-reset"></div>` +
-                `<div class="jw-title jw-reset">` +
-                    `<div class="jw-title-primary jw-reset">${message || ''}</div>` +
-                    `<div class="jw-title-secondary jw-reset">${detail}</div>` +
+                `<div class="jw-info-container jw-reset">` +
+                    `<div class="jw-error-text jw-reset">${(message || '') + detail}</div>` +
```
robwalch: If there were two `div.jw-error-text` elements, one for the message and one for detail we could keep the line break between the two, at least on larger players.

PR #2966, opened 2018-06-14
Handle dom exceptions in setup and between items

robwalch:
### This PR will...

Handle `DOMException` instances in places where we catch errors. 

Two new methods are exported from "api/errors""
- `convertToPlayerError` takes an error and if it is not a `PlayerError` converts it to one with the default error type (key) and code provided.
- `composePlayerError` replaces the static `PlayerError.compose` call. It uses `convertToPlayerError` to make sure the error code is writeable before attempting to change it.

The error handling in controller and program-controller has been cleaned up a little:
- The catch on `_item` has been moved to inside `_item` on `setActiveItem`. There is no point in throwing from `_item` since methods that call it do not pass on it's promise.
- `setActiveItem` is where all provider and item related errors are caught. This handles errors in all places `_item` is called in the controller. These errors are intercepted here before `_load`. 
- `_load` only catches playlist errors.

`DOMException` object's `message` and `code` properties are read-only. If a promise chain errors as the result of one of these, we need to convert it to a `PlayerError`.

Static class methods are only useful for public API. Exporting function work the same and their names can be minified.

No.

#### Addresses Issue(s):

JW8-1692

### DIFF
--- a/src/js/api/errors.js
+++ b/src/js/api/errors.js
@@ -45,7 +45,7 @@ export const ERROR_LOADING_PLAYLIST = 202000;
 export const SETUP_ERROR_LOADING_PROVIDER = 104000;
 
 /**
- * @enum {ErrorCode} An error occurred setting up the current playlist item.
+ * @enum {ErrorCode} An error occurred when switching playlist items.
  */
 export const ERROR_LOADING_PLAYLIST_ITEM = 203000;
 
--- a/src/js/api/errors.js
+++ b/src/js/api/errors.js
@@ -45,12 +45,17 @@ export const ERROR_LOADING_PLAYLIST = 202000;
 export const SETUP_ERROR_LOADING_PROVIDER = 104000;
 
 /**
- * @enum {ErrorCode} Between playlist items, the required provider could not be loaded be.
+ * @enum {ErrorCode} An error occurred setting up the current playlist item.
  */
 export const ERROR_LOADING_PLAYLIST_ITEM = 203000;
 
 /**
- * @enum {ErrorCode} Using the load API, the required provider could not be loaded.
+ * @enum {ErrorCode} The current playlist item has no source media.
+ */
+export const ERROR_PLAYLIST_ITEM_MISSING_SOURCE = 203640;
+
+/**
+ * @enum {ErrorCode} Between playlist items, the required provider could not be loaded.
  */
 export const ERROR_LOADING_PROVIDER = 204000;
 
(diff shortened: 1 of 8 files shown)


src/js/api/errors.js
```diff
@@ -45,12 +45,17 @@ export const ERROR_LOADING_PLAYLIST = 202000;
 export const SETUP_ERROR_LOADING_PROVIDER = 104000;
 
 /**
- * @enum {ErrorCode} Between playlist items, the required provider could not be loaded be.
+ * @enum {ErrorCode} An error occurred setting up the current playlist item.
```
karimJWP: How does 'An error occurred **when switching** playlist items' sound ?

src/js/api/errors.js
```diff
@@ -45,12 +45,17 @@ export const ERROR_LOADING_PLAYLIST = 202000;
...
 export const ERROR_LOADING_PLAYLIST_ITEM = 203000;
 
 /**
- * @enum {ErrorCode} Using the load API, the required provider could not be loaded.
+ * @enum {ErrorCode} The current playlist item has no source media.
+ */
+export const ERROR_PLAYLIST_ITEM_MISSING_SOURCE = 203640;
+
+/**
+ * @enum {ErrorCode} Between playlist items, the required provider could not be loaded.
```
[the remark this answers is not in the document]
karimJWP: sounds good to me, I'll update the docs then. no more 203,150 - 203160 errors then

src/js/controller/controller.js
```diff
@@ -366,16 +368,12 @@ Object.assign(Controller.prototype, {
                 default:
                     return;
             }
-            loadPromise.catch(e => {
-                if (e.code >= 151 && e.code <= 162) {
```
robwalch: @karimJWP The provider errors never reach this point because they are caught in `_setItem`.

robwalch: @karimJWP if we go your route we should:
1. rename`ERROR_LOADING_PROVIDER` to `ERROR_SETTING_ITEM`
2. rename `ERROR_LOADING_PLAYLIST` to `ERROR_CALLING_LOAD`
3. pass either `ERROR_SETTING_ITEM` or `ERROR_CALLING_LOAD` to `_setItem` to establish the correct base code.

Calling load can load a playlist, change the playlist, replace the playlist with a new playlist item, or just set the playlist index, so "ERROR_LOADING_PLAYLIST" isn't really valid base code.

PR #2967, opened 2018-06-14
Validate volume and mute values

robwalch:
### This PR will...

- Clamp volume between 0 and 100
- Convert mute to a boolean
- Ignore invalid volume input

Prevent `DOMException` errors from being thrown by setting `video.volume` to a non-normal value.

#### Addresses Issue(s):

JW8-1696

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -113,8 +113,9 @@ const Config = function(options, persisted) {
     __webpack_public_path__ = config.base;
     config.width = _normalizeSize(config.width);
     config.height = _normalizeSize(config.height);
-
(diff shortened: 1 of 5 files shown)


src/js/api/config.js
```diff
@@ -113,8 +113,9 @@ const Config = function(options, persisted) {
     __webpack_public_path__ = config.base;
     config.width = _normalizeSize(config.width);
     config.height = _normalizeSize(config.height);
-
     config.aspectratio = _evaluateAspectRatio(config.aspectratio, config.width);
+    config.volume = parseFloat(config.volume) || Defaults.volume;
```
robwalch: This is bad if volume is set to `0` it'll fall through to the default.

PR #2969, opened 2018-06-15
Lazy instantiate video blob and guard against URL being overwritten

robwalch:
### This PR will...

- Lazy instantiat video blob
- Guard against URL being overwritten

We've found setup errors caused by sites overwritting the `URL` global. This results in a TypeError that fails setup with code `100000` on Android where we perform a video play test before preloading.

```js
JW Player Error 100000. For more information see https://developer.jwplayer.com/jw-player/docs/developer-guide/api/errors-reference#100000

TypeError: URL.createObjectURL is not a function
at Je (https://ssl.p.jwpcdn.com/player/v/8.4.0-beta.2/jwplayer.core.controls.html5.js:10:63924)
```

`URL` is critical to this test and to MSE usage in hls.js and shaka. Catching it here won't help video playback, but it prevents setup from failing. mp4s will play fine, and DASH and HLS engines should surface the issue.

#### Addresses Issue(s):

JW8-1363


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/utils/can-autoplay.js
```diff
@@ -25,21 +25,42 @@
...
+    }
+    return URL.createObjectURL(videoBlob);
+}
+
+// Catch errors caused by sites replacing the `URL` global with another value
+function getObjectURL() {
+    try {
+        return createObjectURL(window.URL);
+    } catch (e) {
+        console.error(window.URL, e);
```
vseventer: Do we want to log this?

src/js/utils/can-autoplay.js
```diff
@@ -25,21 +25,42 @@
...
+    }
+    return URL.createObjectURL(videoBlob);
```
vseventer: Instead of all the `try catch` below, what if we check if the method exists, rather than recovering from failure?
```js
const method = (window.URL && window.URL.createObjectURL) ||
    (window.webkitURL && window.webkitURL.createObjectURL) // etc.
```

At the end of the day, third-party scripts might override not only `URL`, but also `Blob`, etc.

src/js/utils/can-autoplay.js
robwalch: See the description above. I believe we do since site developers are making a big mistake. It would be even better for browsers to make `window.URL` read-only.

PR #2970, opened 2018-06-15
[ADS-1233] Set itemReady when ad is playing.

vseventer:
### This PR will...
Revert part of 7a07c90, where `itemReady` was no longer set to `true` when an ad is playing. This causes the controls (play, pause, audio track, captions, quality, and fullscreen) to be [enqueued](https://github.com/jwplayer/jwplayer/blob/master/src/js/controller/controller.js#L1003) rather then executed on subsequent playlist items.

Fix ad controls.

Setting `itemReady` to `true` when playing ads causes issues with VR videos, but that's already a current issue (see JW8-1578). This PR does not break JW8-1332.

N/A

#### Addresses Issue(s):
ADS-1233

### DIFF
--- a/src/js/program/program-controller.js
+++ b/src/js/program/program-controller.js
@@ -445,7 +445,7 @@ class ProgramController extends Eventable {
      * @memberOf ProgramController
      */
     _activateBackgroundMedia() {
-        const { background, background: { nextLoadPromise } } = this;
+        const { background, background: { nextLoadPromise }, model } = this;
         // Activating this item means that any media already loaded in the background will no longer be needed
         this._destroyMediaController(background.currentMedia);
         background.currentMedia = null;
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

vseventer:
@robwalch Do we want to get this in as is, or using your suggestion? We'd still have the issue that `itemReady` is never set (b/c `this._setActiveMedia(nextMediaController);` is never called in this scenario.

robwalch:
@vseventer Please reopen against v8

robwalch:
@vseventer use /test/public/events/ to make sure this change doesn't result in player event firing during ad breaks in the circumstances surrounding this bug.

vseventer:
@robwalch Yep, events look good!

Issue #2981, opened 2018-06-27
Multiple loading of plugins and skin.url


artur-michalak:
### Expected Behavior
Plugins and skin should load only once.

### Actual Behavior
When new player is created then the same JW Player related scripts are added to HEAD.

![image](https://user-images.githubusercontent.com/2443904/41971488-e328801a-7a0e-11e8-969a-961098743d52.png)
### Steps to reproduce

1. Create html
2. Add this code
```javascript
    var player = null,
        count = 0,
        intervalId;

    function createPlayer() {
        player = jwplayer("player1").setup({
            "playlist": [{
                "file": "http://content.jwplatform.com/videos/bkaovAYt-640.mp4"
            }],
            skin: {
                url: '/css/player.css',
                name: 'svp'
            },
            autostart: false,
            "key": "KEY",
        });
    }

    intervalId = setInterval(function () {
        count++;
        // cleanup previous
        if(player) {
            player.remove();
        }

        createPlayer();

        if (count === 4) {
            clearInterval(intervalId);
            player.remove();
        }
    }, 500);
```
3.

### Environment

- *hardware | device* PC
- *operating system | version* Windows 10
- *browser | version* Chrome 67
- *JW Player version | url* 8.4.0

----------------------------------------------------------------------------------------

This template is provided as a guide to filing bugs and feature requests. Please replace 
all text *in italics* with details describing your issue. When submitting a feature request, 
"Steps to reproduce" can be substituted with a user story or list of acceptance criteria.

robwalch:
Hi @artur-michalak,

Thank you for the bug report. Is this happening only when removing a player before setting up a new one (If you call setup on an existing player instance, the current instance will be replaced/removed for you)?

artur-michalak:
It has the same bug when setup is called on player instance, however it works well when .load() is used

robwalch:
Hi @artur-michalak,

I'm going to make sure we only load plugin urls's once on a page, but for CSS, I think it's on you to change your setup. skin loading is like a courtesy. Ideally you should embed it on the page yourself for best performance. If you are setting up multiple players, specify the url on the first player, and then only include "name" on subsequent setups. That would ensure that you only load the CSS once. The player can't know that that your CSS URL returns static content and should only load it once.

robwalch:
Hi @artur-michalak,

[...]

PR #2983, opened 2018-06-28
Plugins with the same name should only load and be registered once

robwalch:
### This PR will...
Ensure that individual plugins are only loaded and registered once.
Remove registered plugins that fail to load or instantiate (allowing new player setups to retry).

If the player is setup multiple times it should not reload javascript for a plugin that is already been registered.

### How do jwplayer plugins work?

There are two ways plugins can be registered in the player.
1. inline by calling `jwplayer.api.registerPlugin(name, minimumVersion, pluginClass)`
2. async by through `jwplayer().setup({ plugins: { 'path/to/plugin.js': pluginInstanceOptions  }})`

The commercial player automatically converts specific configuration blocks like `advertising` to a `plugins` entry:
```js
advertising: { client: 'vast' }
// becomes
plugins: { 'path/to/vast.js': { /* advertising block */ } }
```
Plugins already in the registry can be addressed by name:
```js
jwplayer.api.registerPlugin('myPlugin', '8.0', function(playerApi, options, div) { /* plugin implementation*/ });
// in jwplayer.setup({})
plugins: { 'myPlugin': { /* options */ } }
```

During setup, the player looks at all entries in the `plugins` block, extracts the plugin name (path/**name**.js) from the url key and adds the plugin to the library's plugin registry.

All players on the page share this registry. **These changes ensure that we do not reload plugins already in the registry, and if loading or instantiation fails, remove the plugin from the registry.**

There can be multiple plugins in the registry, but only those found in the player setup options will be instantiated for the player being setup. Simply registering a plugin does not bind it to all subsequent player setups.

Each plugin that is not found in the registry is then loaded. It is expected that plugin.js files will execute `registerPlugin` once loaded, adding their class function `plugin.js` to the plugin object in the registry.

Finally each plugin class is instantiated for the new player being setup. The plugin class's constructor receives three arguments:
1. The player instance (`api`)
2. The plugin options (`config`)
3. The overlay container (`div`) which can be used to render DOM elements between the player's media and controls elements.

```js
    // plugins/plugin.js
    const PluginClass = this.js;
    const pluginInstance = new PluginClass(api, config, div);
```

#### Addresses Issue(s):
JW8-1721 #2981

### DIFF
--- a/src/js/plugins/model.js
+++ b/src/js/plugins/model.js
@@ -4,27 +4,46 @@ import { log } from 'utils/log';
 const pluginsRegistered = {};
 
 // Extract a plugin name from a string
-const getPluginName = function (url) {
+const getPluginName = function(url) {
     // Regex locates the characters after the last slash, until it encounters a dash.
     return url.replace(/^(.*\/)?([^-]*)-?.*\.(js)$/, '$2');
 };
 
-const PluginModel = function () {
-    this.addPlugin = function (url, fromLoader) {
-        const pluginName = getPluginName(url);
-        let plugin = pluginsRegistered[pluginName];
-        if (!plugin) {
-            plugin = new Plugin(url);
-            pluginsRegistered[pluginName] = plugin;
-        } else if (fromLoader && plugin.url !== url) {
-            log(`JW Plugin "${pluginName}" already loaded from "${plugin.url}". Ignoring "${url}."`);
-        }
-        return plugin;
-    };
+const PluginModel = function() {};
+const prototype = PluginModel.prototype;
 
-    this.getPlugins = function () {
-        return pluginsRegistered;
-    };
+prototype.setupPlugin = function(url) {
+    const registeredPlugin = this.getPlugin(url);
+    if (registeredPlugin) {
+        if (registeredPlugin.url !== url) {
+            log(`JW Plugin "${getPluginName(url)}" already loaded from "${registeredPlugin.url}". Ignoring "${url}."`);
+        }
+        return registeredPlugin.promise;
+    }
+    const plugin = this.addPlugin(url);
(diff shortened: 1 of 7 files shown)


src/js/plugins/model.js
```diff
@@ -4,27 +4,46 @@ import { log } from 'utils/log';
...
+
+prototype.setupPlugin = function(url) {
+    const registeredPlugin = this.getPlugin(url);
+    if (registeredPlugin) {
+        if (registeredPlugin.url !== url) {
+            log(`JW Plugin "${getPluginName(url)}" already loaded from "${registeredPlugin.url}". Ignoring "${url}."`);
         }
-        return plugin;
-    };
+        return registeredPlugin.promise;
```
vseventer: See comment below - would we be able to `return Promise.resolve(registeredPlugin)`?

src/js/plugins/plugin.js
```diff
@@ -37,6 +37,9 @@ function getJSPath(url) {
 
 const Plugin = function(url) {
     this.url = url;
+    this.promise = new Promise((resolve) => {
```
vseventer: Can we go without this - seems a bit of an anti-pattern.

robwalch: I don't like the circular reference, but this was the most elegant way to ensure all plugins have a promise that resolves when `this.js` is available.

src/js/plugins/model.js
robwalch: No because the this could be a plugin whose request is still active. Resolving while the request is still in flight would cause `configurePlugin` to fail because the plugin class `plugin.js` is not yet available.

The test case for this is you setup a player, and while plugin requests are in flight, setup another player or cancel and resetup the same player.

PR #2994, opened 2018-07-12
Handle non-plugin JavaScript loading via plugins block

robwalch:
### This PR will...

Resolve plugin promises on load, in case the loaded file does not call `registerPlugin` first resolving the plugin class (`plugin.js`).

The `plugins` setup options block is being used to load JavaScript files that do not call `registerPlugin`. In these cases we need to resolve the plugin promise shortly after loading so that setup is not blocked.

This will also prevent valid plugins that throw exceptions before `registerPlugin` is executed or complete from blocking setup.

#### Addresses Issue(s):
JW8-1750 #2981


egreaves:
@robwalch @dannyfinks gonna merge this, then fix my 8.4.x pr so it includes these changes.

PR #2996, opened 2018-07-13
Prime video tags on first user-gesture

robwalch:
### This PR will...

Cleanup video tag priming and add event listeners on the player container's window or document to prime videos on first gesture (click or keydown).

These changes allow playback to start programmatically (not in a gesture event) provided the user triggered at least one gesture event since the `jwplayer().setup()` was called.

This will help in cases where playback should start when a player comes into view by scrolling down the page after the user has interacted with the page, or after dismissing a dialog or ad, and those interactions did not directly result in `video.play()` being called.

With these changes, browsers with autoplay restrictions that require a gesture to call `video.play()` or `video.load()` to enable unmuted playback should produce less "playAttemptFailed" events - provided there is an interaction with the page after player setup (iOS, Safari, and Android Chrome (prior to v65)). 

### Notes?

This line in the controller prevents uncaught promise exceptions that result from calling `jwplayer().play()` on "playlistItem"
`this.play = (meta) => _play(meta).catch(noop);`

#### Testing
1. This setup will attempt to play one second after setup, but result in a "playAttemptFailed" even on mobile devices with audio unmuted. As long as you tap anywhere on the page within that second, playback will start.
```js
{
    events: {
        playlistItem: function() {
            setTimeout(this.play, 1000);
        }
    },
    playlist: "//cdn.jwplayer.com/v2/playlists/ThjcbXPb&sources=mp4&poster_width=320"
}
```
2. See test/squash/bootstrap.html?primary=html5&edition=ads&feature=autostart_viewable/autostart_when_viewable&suite=Autostart

These changes depend on updates to UI.js made in https://github.com/jwplayer/jwplayer/pull/2991. Once merged this branch should be rebased against master.

#### Addresses Issue(s):

JW8-1584

### DIFF
--- a/src/js/api/core-shim.js
+++ b/src/js/api/core-shim.js
@@ -12,6 +12,7 @@ import { resolved } from 'polyfills/promise';
 import ErrorContainer from 'view/error-container';
 import MediaElementPool from 'program/media-element-pool';
 import SharedMediaPool from 'program/shared-media-pool';
+import UI, { getElementWindow } from 'utils/ui';
 import { PlayerError, composePlayerError, convertToPlayerError,
     SETUP_ERROR_LOADING_PLAYLIST, SETUP_ERROR_UNKNOWN, MSG_TECHNICAL_ERROR } from 'api/errors';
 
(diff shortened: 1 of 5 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/api/core-shim.js
```diff
@@ -95,7 +96,12 @@ Object.assign(CoreShim.prototype, {
         if (!model.get('backgroundLoading')) {
             mediaPool = SharedMediaPool(mediaPool.getPrimedElement(), mediaPool);
         }
-        mediaPool.prime();
```
robwalch: Note that all video tags are "primed" as they are created. So this call is no longer necessary after creating the pool.

PR #2997, opened 2018-07-16
Prevent styling control bar with background transparent

boyntoni:
### This PR will...

Prevent an issue in which, when setting ```controlbar.background: 'transparent'```, the submenu is negatively impacted (losing its transparency).

This PR addresses this issue by preventing writing the CSS rule when ```controlbar.background: transparent``` is detected.

When setting ```controlbar.background: transparent```, the submenu loses its transparency.

In speaking with @radium-v , it's my understanding that the better solution would be to decouple the submenu from the controlbar so that this issue would not exist in the first place.

That being said, this is a short term patch. I am unsure if this fully covers the scenario, as I do not test hex values, etc. I do not believe any method exists in the css utils that would allow this test (getRbga does not seem to cover this). 

No

#### Addresses Issue(s):

JW8-1663

### DIFF
--- a/src/js/view/utils/skin.js
+++ b/src/js/view/utils/skin.js
@@ -279,18 +279,18 @@ export function handleColorOverrides(playerId, skin) {
 }
 
 function isTransparent(color) {
-if (color.toLowerCase() === 'transparent') {
-    return true;
-}
+    if (color.toLowerCase() === 'transparent') {
+        return true;
+    }
 
-let channels;
-if (color.indexOf('(') > -1) {
-    channels = color.match(/(0\.\d+)|\d+/g);
-}
+    let channels;
+    if (color.indexOf('(') > -1) {
+        channels = color.match(/(0\.\d+)|\d+/g);
+    }
 
-if (color.indexOf('#') > -1) {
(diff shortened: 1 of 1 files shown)


johnBartos:
@boyntoni I think you need to look for `rgba(x, x, x, 0)` (note the 4th field is the alpha channel); `rgb(0, 0, 0)` is just black.

johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/utils/skin.js
```diff
@@ -274,3 +277,20 @@ export function handleColorOverrides(playerId, skin) {
         }
     }
 }
+
+function isTransparent(color) {
```
robwalch: This method is overkill for a workaround. Just check the default @radium-v uses and 'transparent'.

PR #3001, opened 2018-07-17
Prevent captions renderer from setting shadowDOMFontSize in Safari iOS

waxidiotic:
JW8-1700

### This PR will...

Prevent the captions renderer from setting the `shadowDOMFontSize` in Safari for iOS.

The captions were appearing too small when bringing the player full screen in Safari for iOS.

n/a

n/a

#### Addresses Issue(s):

JW8-1700

### DIFF
--- a/src/js/view/captionsrenderer.js
+++ b/src/js/view/captionsrenderer.js
@@ -1,4 +1,4 @@
-import { Browser } from 'environment/environment';
+import { Browser, OS } from 'environment/environment';
 import { chunkLoadErrorHandler } from '../api/core-loader';
 import Events from 'utils/backbone.events';
 import { ERROR } from 'events/events';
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/captionsrenderer.js
```diff
@@ -213,7 +213,7 @@ const CaptionsRenderer = function (viewModel) {
 
         const fontSize = Math.round(height * _fontScale);
 
-        if (_model.get('renderCaptionsNatively')) {
```
robwalch: So we only call `_setShadowDOMFontSize` on Safari Desktop? Do we even need to?

egreaves: The idea behind doing this was to add styles to ensure native/player rendered captions are sized similarly. If it's not working as intended, we should just get rid of it.

PR #3003, opened 2018-07-18
Trigger 'remove' when player is reset for plugins

robwalch:
### This PR will...
Trigger 'remove' when player is reset so that plugins are notifies that the api instance is no longer in use.

Previously this event was only fired when calling `api.remove()`. Now it will also be invoked when `api.setup()` is called as well, since all internals and listeners of the instance have been reset.

This will also notify the analytics plugin that the current player session has ended.

These changes are needed for related to cleanup external elements when the player is removed or reset.

https://github.com/jwplayer/jwplayer-plugin-related/pull/310

#### Addresses Issue(s):

JW8-1414

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -46,6 +46,7 @@ function resetPlayer(api, core) {
     Object.keys(plugins).forEach(key => {
         delete plugins[key];
     });
+    api.trigger('remove');
     api.off();
     core.playerDestroy();
     core.getContainer().removeAttribute('data-jwplayer-id');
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

robwalch:
Replaced by https://github.com/jwplayer/jwplayer/pull/3004

PR #3006, opened 2018-07-19
Handle Error Completing Setup

robwalch:
### This PR will...
Handle exceptions in `playerReadyNotify` as a new player error rather than a setup error.

When the player is setup with `jwplayer().setup()`, the expectation is that the setup process will complete with either a "ready" event or a "setupError" event. If an exception is thrown as setup completes, during the "ready", "playlist" or "playlistItem" events, in debug mode (set `jwplayer.debug = true` before setup), then the player should handle the exception gracefully with an "error" event, rather than a "setupError" event.

*Steps to reproduce:*
```js
// enable debug to not catch exceptions thrown in event handlers
jwplayer.debug = true;

// then
jwplayer().setup({ file: '//video' }).on('playlistItem', function() {
   throw new Error('This should not result in a setup error.');
});

// or
jwplayer().setup({
  file: '//video',
  events: {
    playlistItem: function() { throw new Error('This should not result in a setup error.'); }
  }
});
```

*Actual:*
Player fires "ready", "playlist", then "setupError" with code 100,000.

*Expected:*
Player fires "ready", "playlist", then "error" with new error code 200,001.

*Note* When `jwplayer.debug` is not set or set to false, the exception is caught and logged to the console correctly.

#### Addresses Issue(s):
JW8-1774

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -191,15 +191,17 @@ Object.assign(Controller.prototype, {
             // Fire 'ready' once the view has resized so that player width and height are available
             // (requires the container to be in the DOM)
             _view.once(RESIZE, () => {
-                resolved.then(_playerReadyNotify).catch((error) => {
+                try {
+                    playerReadyNotify();
+                } catch (error) {
                     _this.triggerError(convertToPlayerError(MSG_TECHNICAL_ERROR, ERROR_COMPLETING_SETUP, error));
-                });
+                }
             });
 
             _view.init();
         };
 
-        function _playerReadyNotify() {
+        function playerReadyNotify() {
             _model.change('visibility', _updateViewable);
             eventsReadyQueue.off();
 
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -190,7 +190,11 @@ Object.assign(Controller.prototype, {
 
             // Fire 'ready' once the view has resized so that player width and height are available
             // (requires the container to be in the DOM)
-            _view.once(RESIZE, _playerReadyNotify);
+            _view.once(RESIZE, () => {
+                resolved.then(_playerReadyNotify).catch((error) => {
```
johnBartos: Would a try/catch be better to keep the code synchronous?

PR #3008, opened 2018-07-19
Only trigger the "remove" event on api instances that have been setup

robwalch:

This is a follow up to https://github.com/jwplayer/jwplayer/pull/3004 which adds tests, and changes the behavior for the "remove" event to only fire when resetting an api that has been setup.

This means that on first setup we will continue to receive events starting with "ready", "playlist"... and will not get a "remove" event before "ready" until the after the second time that setup is called on an api instance.

#### Addresses Issue(s):
JW8-1414

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -46,7 +46,9 @@ function resetPlayer(api, core) {
     Object.keys(plugins).forEach(key => {
         delete plugins[key];
     });
-    api.trigger('remove');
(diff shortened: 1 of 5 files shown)


robwalch:
^ https://github.com/jwplayer/jwplayer/pull/3008#issuecomment-406317919 merging v8.4.x into master and rebasing to make these tests more stable and performant.

PR #3011, opened 2018-07-20
Create setPlaylistItem method in api

DanFerrer:
### This PR will...

- Creates new method in the public API called ```setPlaylistItem```
- Change double quotes to single (breaks linting, but build still passes)

To allow publishers to modify the properties of a playlist item (ex: ```item.starttime```)

- JSDoc
-  Use of ```updatePlaylist``` in the controller
- Return the API instance or the new playlist item/playlist?

#### Addresses Issue(s):

JW8-1688


src/js/api/api.js
```diff
@@ -648,6 +648,29 @@ export default function Api(element) {
...
+        setPlaylistItem(index, item) {
+            const state = this.getState();
+
+            if (state === 'idle' || state === 'complete') {
+                let playlist = this.getPlaylist();
+
+                if (playlist) {
+                    playlist = playlist.slice();
+                    playlist[index] = item;
+                    core.updatePlaylist(playlist, core.get('feedData'));
```
robwalch: Why are you copying the playlist and callling `updatePlaylist`?

I can see how it `filterPlaylist` and `_setItem` would be useful if you were to replace an item with something that could not be played or update the current item, however, there some side effects. `stop()` is called and both a "playlist" and "playlistItem" API event will be dispatched. This has not been discussed and is not functionality we want introduced without going through the impact this could have.

Ideally calling `setPlaylistItem` should update an item without changed to the playlist, or interruption to playback, unless the current item is replaces while not idle or complete. The array should remain the same array, unless the item is invalid (in that case we could make the call fail instead of filtering), and the only part of `filterPlaylist` we need for a single item is the `forEach` function being called on items in the list.

src/js/api/api.js
```diff
@@ -648,6 +648,27 @@ export default function Api(element) {
...
+                }
+            }
+
+            return this;
```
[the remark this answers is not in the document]
robwalch: chainable is good - `getPlaylistItem` can always be used to get the updated item.

`setPlaylistItem` should be added to this list for coverage: https://github.com/jwplayer/jwplayer/blob/master/test/data/api-methods-chainable.js

src/js/controller/controller.js
```diff
@@ -918,6 +918,21 @@ Object.assign(Controller.prototype, {
             return _setItem(_model.get('item'));
         };
 
+        this.setPlaylistem = function (index, item) {
+            const state = _model.get('state');
+            const selectedItemPlaying = index === _model.get('item') && state === 'playing';
```
robwalch: Just checking 'playing' isn't enough. The item is active when the player state is paused and buffering as well.

But this doesn't mean we shouldn't replace the item. It just means we can't reload it;

- Always replace the item
- When idle, if the item being replaced is the current item, call `setItem` to update the player state so that when playback starts the new item is played. This should also kick off preloading at the correct `item.startTime`.

src/js/controller/controller.js
```diff
@@ -918,6 +918,19 @@ Object.assign(Controller.prototype, {
             return _setItem(_model.get('item'));
         };
 
+        this.setPlaylistItem = function (index, item) {
+            item = normalizePlaylistItem(_model, item, {});
```
robwalch: Make sure to pass in existing `item.feedData` if present:

`item = normalizePlaylistItem(_model, item, item.feedData || {});`

src/js/playlist/playlist.js
```diff
@@ -12,32 +12,13 @@ const Playlist = function(playlist) {
...
 
     playlist.forEach(function(item) {
-        item = Object.assign({}, item);
-
-        item.preload = getPreload(item.preload, preload);
-
-        item.allSources = formatSources(item, model);
-
-        item.sources = filterSources(item.allSources, providers);
+        item = normalizePlaylistItem(model, item, feedData);
```
robwalch: We should still copy `itemFeedData` once (here), then pass it in to each item (so we're not making a new copy of each item):

```js
const itemFeedData = Object.assign({}, feedData);	
delete itemFeedData.playlist;	

playlist.forEach(function(item) {
    item = normalizePlaylistItem(model, item, itemFeedData);
```

robwalch: Just an idea for you to thing about - Consider whether map/filter would be an improvement over defining an array and pushing into it when item is defined:

`return playlist.map((item) => normalizePlaylistItem(model, item, itemFeedData)).filter((item) => !!item)`

src/js/playlist/playlist.js
```diff
@@ -49,6 +30,35 @@ export function validatePlaylist(playlist) {
         throw new PlayerError(MSG_CANT_PLAY_VIDEO, 630);
     }
 }
+
+export function normalizePlaylistItem(model, item, feedData) {
+    const providers = model.getProviders();
+    const preload = model.get('preload');
+    const itemFeedData = Object.assign({}, feedData);
+    delete itemFeedData.playlist;
```
robwalch: Don't copy or delete `itemFeedData` here.

src/js/playlist/playlist.js
```diff
@@ -49,6 +30,35 @@ export function validatePlaylist(playlist) {
...
+
+    item = Object.assign({}, item);
```
robwalch: Can we define a new const for the copied item (no-params-reassign).

`const playlistItem = Object.assign({}, item);`

src/js/view/controls/nextuptooltip.js
```diff
@@ -140,7 +140,9 @@ export default class NextUpTooltip {
     onNextUp(model, nextUp) {
         this.reset();
         if (!nextUp) {
-            return;
+            nextUp = {
+                showNextUp: false
+            };
```
robwalch: How did this change show up?

src/js/controller/controller.js
```diff
@@ -918,6 +918,19 @@ Object.assign(Controller.prototype, {
...
 
+        this.setPlaylistItem = function (index, item) {
+            item = normalizePlaylistItem(_model, item, item.feedData || {});
+
+            if (item) {
+                const playlist = _model.get('playlist');
+                playlist[index] = item;
+
+                if (index === _model.get('item') && _model.get('state') === 'idle') {
+                    _setItem(index);
```
robwalch: @DanFerrer @vseventer @msiddique  Let's make sure we understand the impact on ads this has. Updating "playlistItem" on idle will kick off preloading of the item, and bids/schedules/ads. The "playlistItem" event is necessary. Making sure any previous loading is cleaned up, or exceptions are made for when the change to the item do not require new requests is follow up you will need to do for start on seek.

PR #3014, opened 2018-07-23
Set reason as 'File Not Found' when ajax request returns 404

DanFerrer:
### This PR will...

Change reason on error event from 'cantPlayVideo' to 'File Not Found' when ajax request for captions file returns 404

The wrong reason was being shown for not fatal captions events?

#### Addresses Issue(s):

JW8-1690

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -37,7 +37,7 @@ const Captions = function(_model) {
                     _addTrack(track);
                     loadFile(track, (vttCues) => {
                         _addVTTCuesToTrack(track, vttCues);
-                    }, (key, file, url, error) => {
+                    }, (key, url, xhr, error) => {
                         if (error.code === 404) {
                             this.trigger(ERROR, {
                                 message: 'Captions failed to load',
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -38,12 +38,10 @@ const Captions = function(_model) {
                     loadFile(track, (vttCues) => {
                         _addVTTCuesToTrack(track, vttCues);
                     }, (key, url, xhr, error) => {
-                        if (error.code === 404) {
-                            this.trigger(ERROR, {
-                                message: 'Captions failed to load',
-                                reason: 'File not found'
-                            });
-                        }
+                        this.trigger(ERROR, {
+                            message: 'Captions failed to load',
+                            reason: error.code === 404 ? 'File not found' : key
+                        });
                     });
                 }
             }
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -37,13 +37,11 @@ const Captions = function(_model) {
                     _addTrack(track);
                     loadFile(track, (vttCues) => {
                         _addVTTCuesToTrack(track, vttCues);
-                    }, (key, file, url, error) => {
-                        if (error.code === 404) {
-                            this.trigger(ERROR, {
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/captions.js
```diff
@@ -37,11 +37,13 @@ const Captions = function(_model) {
                     _addTrack(track);
                     loadFile(track, (vttCues) => {
                         _addVTTCuesToTrack(track, vttCues);
-                    }, (error) => {
-                        this.trigger(ERROR, {
-                            message: 'Captions failed to load',
-                            reason: error
-                        });
+                    }, (key, file, url, error) => {
```
johnBartos: I think the signature goes `key, url, xhr, error`.

```
function _error(options, key, code, error) {
    options.onerror(key, options.url, options.xhr, new PlayerError(key, code, error));
}
```

src/js/providers/tracks-mixin.js
```diff
@@ -459,11 +459,13 @@ function addTextTracks(tracksArray) {
                 (vttCues) => {
                     this.addVTTCuesToTrack(textTrackAny, vttCues);
                 },
-                (error) => {
-                    this.trigger(ERROR, {
-                        message: 'Captions failed to load',
-                        reason: error
-                    });
+                (key, file, url, error) => {
+                    if (error.code === 404) {
```
karimJWP: if `error.code` is 404, we should set the reason to `'File not found'`, but if the error code is anything else we still want to trigger the error event

src/js/controller/captions.js
```diff
@@ -37,11 +37,13 @@ const Captions = function(_model) {
...
+                        if (error.code === 404) {
```
karimJWP: if error code is not 404 we still want to trigger an error. What happens if the code is 403?

src/js/providers/tracks-mixin.js
```diff
@@ -459,11 +459,13 @@ function addTextTracks(tracksArray) {
...
-                (error) => {
-                    this.trigger(ERROR, {
-                        message: 'Captions failed to load',
-                        reason: error
-                    });
+                (key, url, xhr, error) => {
+                    if (error.code === 404) {
+                        this.trigger(ERROR, {
+                            message: 'Captions failed to load',
+                            reason: 'File not found'
```
robwalch: Why is reason supposed to be a string?

src/js/controller/captions.js
```diff
@@ -37,10 +37,10 @@ const Captions = function(_model) {
                     _addTrack(track);
                     loadFile(track, (vttCues) => {
                         _addVTTCuesToTrack(track, vttCues);
-                    }, (error) => {
+                    }, (key, url, xhr, error) => {
                         this.trigger(ERROR, {
                             message: 'Captions failed to load',
-                            reason: error
+                            reason: error.code === 404 ? 'File not found' : key
```
robwalch: This event is supposed to be changed to a "warning". We shouldn't even be supporting "reason". The `error` object should be much more useful and is what was intended, but instead the key was being passed in because of a miss-understanding of the callback arguments. Make reason the error or remove it.

PR #3018, opened 2018-08-03
Set intersection ratio to the view instead of the model

karimJWP:
### This PR will...
- differentiate widgets from views in the views-manager
- view-manager commands the view to set the intersection ratio to the model.
As we add new views for the related plugin, it's preferable to allow the view-manager to set the intersection on a view instead of reading it from a shared model.
I couldn't come up with a better name than `matchIntersection`
#### Addresses Issue(s):
JW8-1418

### DIFF
--- a/src/js/view/utils/views-manager.js
+++ b/src/js/view/utils/views-manager.js
@@ -3,6 +3,7 @@ import { requestAnimationFrame, cancelAnimationFrame } from 'utils/request-anima
 import { Browser, OS } from 'environment/environment';
 
 const views = [];
+const widgets = [];
 const observed = {};
 const hasOrientation = 'screen' in window && 'orientation' in window.screen;
 const isAndroidChrome = OS.android && Browser.chrome;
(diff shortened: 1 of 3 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/utils/views-manager.js
```diff
@@ -21,7 +21,7 @@ function lazyInitIntersectionObserver() {
                     for (let j = views.length; j--;) {
                         let view = views[j];
                         if (entry.target === view.getContainer()) {
-                            view.model.set('intersectionRatio', entry.intersectionRatio);
+                            view.setIntersectionRatio(entry.intersectionRatio);
```
robwalch: I would like us to add widgets to the views-manager in a different array. SO we can have:

```js
const views = [];
const widgets = [];
```

and for adding and removing:
```js
 export default {	 export default {
     add: function(view) {
         views.push(view);
     },
     remove: function(view) {
         const index = views.indexOf(view);
         if (index !== -1) {
             views.splice(index, 1);
         }
     },
     addWidget: function(widget) {
         widgets.push(widget);
     },
     removeWidget: function(widget) {
         const index = widgets.indexOf(widget);
         if (index !== -1) {
             widgets.splice(index, 1);
         }
     },
```

PR #3022, opened 2018-08-07
Only modify item.starttime when attached equals false

DanFerrer:
### This PR will...

- Move the logic for setting `starttime` on an item outside of `mediaController.detach()` and set it after we call the detach function 

- ```detatch()``` should not be responsible for setting the starttime in this case
- Prevents starttime on the new active item from being set on the item being detached in a playlist

#### Addresses Issue(s):

JW8-1688

### DIFF
--- a/src/js/program/media-controller.js
+++ b/src/js/program/media-controller.js
@@ -101,7 +101,7 @@ export default class MediaController extends Eventable {
     }
 
     detach() {
-        const { item, mediaModel, provider } = this;
+        const { provider } = this;
         this.thenPlayPromise.cancel();
         provider.detachMedia();
         this.attached = false;
--- a/src/js/program/media-controller.js
+++ b/src/js/program/media-controller.js
@@ -105,13 +105,6 @@ export default class MediaController extends Eventable {
         this.thenPlayPromise.cancel();
         provider.detachMedia();
         this.attached = false;
-
-        // If detaching to play a midroll (pos > 0), ensure that the player resumes at the detached time by setting starttime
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/program/media-controller.js
```diff
@@ -101,17 +101,10 @@ export default class MediaController extends Eventable {
...
     detach() {
-        const { item, mediaModel, provider } = this;
+        const { provider } = this;
         this.thenPlayPromise.cancel();
         provider.detachMedia();
         this.attached = false;
-
-        // If detaching to play a midroll (pos > 0), ensure that the player resumes at the detached time by setting starttime
-        // We don't need to do this for prerolls because the player re-attaches at time 0 by default
-        const pos = mediaModel.get('position');
```
karimJWP: detach gets called in [`set background`](https://github.com/jwplayer/jwplayer/blob/72b1b9ced8ecf6124e37ed2a55097cff7cb208df/src/js/program/media-controller.js#L252). Is it ok if `item.starttime = pos;` is not being called when an item is being backgrounded ?

robwalch: Yes. because backgrounded items are paused and do not need `starttime` updated to resume at the correct position.

PR #3023, opened 2018-08-08
Fix orientation detection on mobile devices for captions font resizing

egreaves:
### This PR will...
Detect orientation correctly on mobile devices that support the `screen.orientation` API.
`screen.availWidth` and `screen.availHeight` do not get updated when an iOS device is rotated. However, they do on Android devices. Detecting if the `screen.orientation` API is supported gives us confidence that the device's height and width will change when the orientation changes.
No, but we've discovered that there's a font scaling bug in iOS 11.4. The code in this PR does not introduce the bug, nor is it a regression. I'll be filing an issue with webkit.
No.
#### Addresses Issue(s):
 JW8-1794

### DIFF
--- a/src/js/view/captionsrenderer.js
+++ b/src/js/view/captionsrenderer.js
@@ -237,11 +237,18 @@ const CaptionsRenderer = function (viewModel) {
 
             // Use screen dimensions when in fullscreen on mobile devices
             if (_model.get('fullscreen') && OS.mobile) {
-                // device is in portrait mode when window.orientation = 0 || 180
-                const portraitMode = !(window.orientation % 180);
                 const { screen } = window;
-                containerHeight = portraitMode ? screen.availHeight : screen.availWidth;
-                containerWidth = portraitMode ? screen.availWidth : screen.availHeight;
+                if (screen.orientation) {
+                    containerHeight = screen.availHeight;
+                    containerWidth = screen.availWidth;
+                } else {
+                    // availHeight and availWidth don't change in iOS when the orientation changes
+                    // iOS device is in portrait mode when window.orientation = 0 || 180
+                    const portraitMode = !(window.orientation % 180);
+                    containerHeight = portraitMode ? screen.availHeight : screen.availWidth;
+                    containerWidth = portraitMode ? screen.availWidth : screen.availHeight;
+                }
+
             }
 
             if (containerWidth && containerHeight && videoWidth && videoHeight) {


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/captionsrenderer.js
```diff
@@ -237,11 +237,18 @@ const CaptionsRenderer = function (viewModel) {
...
                 const { screen } = window;
-                containerHeight = portraitMode ? screen.availHeight : screen.availWidth;
-                containerWidth = portraitMode ? screen.availWidth : screen.availHeight;
+                if (screen.orientation) {
+                    containerHeight = screen.availHeight;
+                    containerWidth = screen.availWidth;
+                } else {
+                    // availHeight and availWidth don't change in iOS when the orientation changes
+                    // iOS device is in portrait mode when window.orientation = 0 || 180
+                    const portraitMode = !(window.orientation % 180);
```
johnBartos: Did you mean to use both `screen` and `window`?

src/js/view/captionsrenderer.js
```diff
@@ -237,11 +237,18 @@ const CaptionsRenderer = function (viewModel) {
 
             // Use screen dimensions when in fullscreen on mobile devices
             if (_model.get('fullscreen') && OS.mobile) {
-                // device is in portrait mode when window.orientation = 0 || 180
-                const portraitMode = !(window.orientation % 180);
                 const { screen } = window;
-                containerHeight = portraitMode ? screen.availHeight : screen.availWidth;
-                containerWidth = portraitMode ? screen.availWidth : screen.availHeight;
+                if (screen.orientation) {
```
johnBartos: What if the orientation is 0? Seems like it'd be better to check if `orientation` is a number.

src/js/view/captionsrenderer.js
egreaves: Yep. `window.orientation` is the older standard which will be dropped in the future (see: https://developer.mozilla.org/en-US/docs/Web/API/Window/orientation)

`window.screen.orientation` isn't yet supported on iOS, so we're using `window.orientation` as the fallback.

src/js/view/captionsrenderer.js
egreaves: `screen.orientation` is an object, not a number.

PR #3024, opened 2018-08-08
Default network error message key is "This video file cannot be played."

robwalch:
### This PR will...
Always return 'cantPlayVideo' as the default `key` from `parseNetworkError`.

`parseNetworkError` should always return a message key. It prevents us from having to import `MSG_CANT_PLAY_VIDEO` everywhere that `parseNetworkError` is used (which is only in commercial).

We should move these modules to jwplayer-commercial because they are not used anywhere in this project. I don't want to do that now since merging module changes will cause other open PRs for the upcoming release to fail if they are out of sync.

https://github.com/jwplayer/jwplayer-commercial/pull/5551

#### Addresses Issue(s):

JW8-1786

### DIFF
--- a/src/js/providers/utils/network-error-parser.js
+++ b/src/js/providers/utils/network-error-parser.js
@@ -1,18 +1,18 @@
 import { MSG_PROTECTED_CONTENT, MSG_CANT_PLAY_VIDEO } from 'api/errors';
 
-export default function parseNetworkError(baseCode, statusCode, url = '') {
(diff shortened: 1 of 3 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3026, opened 2018-08-09
Create displayPlaybackLabel config option

DanFerrer:
### This PR will...

- Adds new option called ```displayPlaybackLabel``` to config (default is ```false```)
- Creates CTA underneath the play button when ```displayPlaybackLabel === true```
- Changes CTA text given the value of ```localization.playback```
- Replace "Start Playback" with "Play" as the default text for ```localization.playback```
- Add unit tests for new config option
- Rename less file to ```playback-label.less```
- Remove old unit tests

Configure the idle state to maximize plays.

- The CSS is ```playback-label.less``` will only be used when ```displayPlaybackLabel === true```
- Using long strings for ```localization.playback``` to ensure the CTA doesn't get truncated/printed on a new line.

#### Addresses Issue(s):

JW8-1796


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/css/controls/imports/playback-label.less
```diff
@@ -1,8 +1,12 @@
 @import "../../shared-imports/vars.less";
 
+.jw-idle-icon-text {
+    display: none;
```
vseventer: Can we move the styling (except `display: block`) from lines 21–26 here? This will make it easier to re-use the class when the player is not idle (`.jw-state-idle`). We can even remove the `display: block`, and control this on `.jw-idle-label` instead.

src/css/controls/imports/playback-label.less
```diff
@@ -1,8 +1,12 @@
...
+}
+
 .jw-state-idle {
     .jw-icon-display {
-        &.jw-ab-idle-label {
+        &.jw-idle-label {
```
vseventer: Same here - if we move the default styling for `.jw-idle-label` to outside `.jw-state-idle` its easier to re-use for other player states.

test/unit/play-display-icon-test.js
```diff
@@ -25,84 +25,47 @@ describe('PlayDisplayIcon', function() {
...
+        it('should set playback label to the value of localization.playback', function() {
+            model.set('displayPlaybackLabel', true);
+            localization.playback = 'Jugar';
             displayIcon = new PlayDisplayIcon(model, {}, element);
 
-            expect(displayIcon.icon.className).to.include('jw-ab-idle-label');
+            expect(displayIcon.icon.className).to.include('jw-idle-label');
             expect(displayIcon.icon.lastChild.className).to.include('jw-idle-icon-text');
-            expect(displayIcon.icon.lastChild.textContent).to.equal('Play');
+            expect(displayIcon.icon.textContent).to.not.equal('Play');
```
karimJWP: not a big deal but it would be better to have a variable storing the default localization value instead of hard coding; that way if we change the default value again, we don't need to update this test
```
const defaultPlaybackLocalization = localization.playback;
localization.playback = 'Jugar';
...
expect(displayIcon.icon.textContent).to.not.equal(defaultPlaybackLocalization);
```

test/unit/play-display-icon-test.js
```diff
@@ -25,84 +25,47 @@ describe('PlayDisplayIcon', function() {
         icon.appendChild(svg);
         element = document.createElement('div');
         element.appendChild(icon);
+
```
karimJWP: remove blank

src/js/view/controls/play-display-icon.js
```diff
@@ -42,34 +41,20 @@ export default class PlayDisplayIcon {
...
         });
 
-        this.toggleIdleClass(_model.get('state'), idleButtonText);
+        if (_model.get('displayPlaybackLabel')) {
+            let iconText = this.icon.getElementsByClassName('jw-idle-icon-text')[0];
+            if (!iconText) {
+                iconText = createElement(`<div class="jw-idle-icon-text"></div>`);
+                this.icon.appendChild(iconText);
+                addClass(this.icon, 'jw-idle-label');
+                iconText.textContent = localization.playback;
```
robwalch: You should always `appendChild` after making changes to the element, not before in the same function.

PR #2781, opened 2018-03-13
When play is triggered by seek, add startTime to meta event to notify plugins

pajong:
### This PR will...
Add startTime to meta event when play is triggered by seek API.

### Why is this Pull Request needed?
A new ad rule needs to be added to follow up with the seek API before the start of the content, so that the viewer is not overloaded with too many ads.
This PR will notify the ad plugins on beforePlay event that the play is due to seek API, so that the ad plugins can correctly avoid playing the preroll.

### Are there any Pull Requests open in other repos which need to be merged with this?
https://github.com/jwplayer/jwplayer-ads-vast/pull/346

#### Addresses Issue(s):

JW8-1327

PR #3030, opened 2018-08-14
Set meta.starttime to playlistitem.starttime

DanFerrer:
### This PR will...

- Set ```meta.startTime``` to ```playlistItem.starttime``` when attempting playback

We currently don't set ```meta.startTime``` in the ```_play``` method and use that property for "startOnSeek" ad rules (pre and none are currently ignored because of this)

Original PR: https://github.com/jwplayer/jwplayer/pull/2781

#### Addresses Issue(s):

JW8-1688

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -430,7 +430,7 @@ Object.assign(Controller.prototype, {
                 _beforePlay = true;
                 _this.trigger(MEDIA_BEFOREPLAY, {
                     playReason,
-                    startTime: meta && meta.startTime ? meta.startTime : _model.get('playlistItem').starttime
+                    startTime: meta && meta.startTime ? (meta.startTime || 0) : _model.get('playlistItem').starttime
                 });
                 _beforePlay = false;
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -430,7 +430,7 @@ Object.assign(Controller.prototype, {
                 _beforePlay = true;
                 _this.trigger(MEDIA_BEFOREPLAY, {
                     playReason,
-                    startTime: meta && meta.startTime ? (meta.startTime || 0) : _model.get('playlistItem').starttime
+                    startTime: meta && meta.startTime ? meta.startTime : _model.get('playlistItem').starttime
                 });
                 _beforePlay = false;
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -18,7 +18,7 @@ import { OS } from 'environment/environment';
 import { streamType } from 'providers/utils/stream-type';
 import Promise, { resolved } from 'polyfills/promise';
 import cancelable from 'utils/cancelable';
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -408,7 +408,9 @@ Object.assign(Controller.prototype, {
             }
 
             const playReason = _getReason(meta);
-            const startTime = meta ? meta.startTime : null;
+
+            const startTime = meta.startTime = isValidNumber(meta.startTime) ? meta.startTime : _model.get('playlistItem').starttime;
```
egreaves: This changes the assumption about whether `meta` is defined or not. If it's possible for meta to be not an object, there needs to be a check before trying to assign a value to `meta.startTime`.

src/js/controller/controller.js
```diff
@@ -408,7 +408,14 @@ Object.assign(Controller.prototype, {
             }
 
             const playReason = _getReason(meta);
-            const startTime = meta ? meta.startTime : null;
+            let startTime;
+            
+            if (meta && meta === Object(meta)) {
```
robwalch: When is `meta` truthy but not an object?

If `item.startime` is set, why would we only set `meta.startTime ` when `meta` is passed into `play` as an object?

DanFerrer: @robwalch Theres an integration test where we call ```play()``` from the api with ```true``` as an argument in the last unstable build but I didnt see any case where we pass anything but an object in:

```unknown error: Cannot create property 'startTime' on boolean 'true' ```

File: src/js/controller/controller.js [hunk 3]
```diff
@@ -408,7 +408,14 @@ Object.assign(Controller.prototype, {
...
+                startTime = meta.startTime = isValidNumber(meta.startTime) ? meta.startTime : _model.get('playlistItem').starttime;
+            } else {
```
robwalch: This probably is not the right place to read `playlistItem.starttime`.  Why don't we always add `playlistItem.starttime` to the `MEDIA_BEFOREPLAY` when it exists on the playlistItem (unless overwritten by meta)?

src/js/controller/controller.js
```diff
@@ -431,7 +430,7 @@ Object.assign(Controller.prototype, {
                 _beforePlay = true;
                 _this.trigger(MEDIA_BEFOREPLAY, {
                     playReason,
-                    startTime
+                    startTime: meta && meta.startTime ? (meta.startTime || 0) : _model.get('playlistItem').starttime
```
johnBartos: Oops, I misunderstood what you were doing before. You can remove the `|| 0`

PR #3032, opened 2018-08-20
Only forward and queue mute changes

robwalch:
When playback autostarts muted with a preroll, the background video fires a "volumechange" event asynchronously after being muted and after being backgrounded. The `ProviderListener`'s call to `trigger` is then queued until after the ad break. If audio is unmuted during the ad playback the reverse "volumechange" event is not fired (this may be a bug in Chrome since it is fired only when dev tools are open) so we never get a second unmute `trigger` added to the event queue. As a result a mute event is flushed after the break telling the player the video has been muted but has since been unmuted resulting in the UI and media being out of sync.

This change prevents us from forwarding or queuing mute events that match the player's global mute state so that we avoid this situation entirely. Since mute true events will not be forwarded when autostartMuted is true, we no longer need to check for it in the `MediaControllerListener` handler.

#### Addresses Issue(s):
JW8-1806

### DIFF
--- a/src/js/program/program-listeners.js
+++ b/src/js/program/program-listeners.js
@@ -22,6 +22,12 @@ export function ProviderListener(mediaController) {
             case MEDIA_VISUAL_QUALITY:
                 mediaModel.set(MEDIA_VISUAL_QUALITY, Object.assign({}, data));
                 return;
+            case MEDIA_MUTE:
+                // Only forward and queue mute changes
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3034, opened 2018-08-21
Defer play command when the item is changed before the first playlistItem event

robwalch:
### This PR will...
Queue the play command when the item is changed.

So that "playlist" and "playlistItem" events are dispatched before the "beforePlay" and "playAttempt" events which need to be attributed to the correct item.

#### Addresses Issue(s):
JW8-1691

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -262,6 +262,12 @@ Object.assign(Controller.prototype, {
             // Run _checkAutoStart() last
             // 'viewable' changes can result in preload() being called on the initial provider instance
             _checkAutoStart();
+
+            _model.on('change:itemReady', (changeModel, itemReady) => {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

robwalch:
@dannyfinks I've addressed the mixed providers issue. It turns out we ignore play and pause commands completely in that state. We also ignore pause when not buffering which is pretty limiting (you can't pause on playAttempt).

dannyfinks:
@robwalch @egreaves the failing test requires updating based on these changes. Everything else looks good.

PR #3038, opened 2018-08-27
Trigger share menu on right click

DanFerrer:
### This PR will...

- Create a new config option for the ```sharing``` block called ```shareOnRightClick```
- Create a new ```localization ``` called ```share```
- Create share option on rightclick menu if ```sharing.shareOnRightClick``` is ```true```

### Why is this Pull Request needed

New feature for A/B test

https://github.com/jwplayer/borg/pull/817
https://github.com/jwplayer/jwplayer-commercial/pull/5673

#### Addresses Issue(s):

JW8-2160


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/rightclick.js
```diff
@@ -16,8 +16,9 @@ function createDomElement(html) {
 }
 
 export default class RightClick {
-    constructor(infoOverlay) {
```
johnBartos: I don't see any usages for `this.controlbar`. Did you forget or is it not needed anymore?

DanFerrer: @rob Addressed your changes from the ticket in commercial . Everything is now in open source and instead of passing in the controlbar, i grab the ```sharing``` pliugin instance from the api. ```api.getPlugin('sharing')``` isnt available when we call ```new RightClick()``` so i pass the whole api object in. @johnBartos to your point earlier we could use a function here instead of passing the api here.

```javascript
new RightClick(this.infoOverlay, () => {
  api.getPlugin('sharing').open();
});
```

robwalch: I like the idea of passing in a `openSharing` callback handler. Ideally each item in right click could be organized with label, callback etc... and added in `rightclick.setup` not the constructor.

src/js/view/controls/controls.js
```diff
@@ -86,6 +86,7 @@ export default class Controls {
 
         const touchMode = model.get('touchMode');
 
+
```
robwalch: Remove extra line.

src/js/view/controls/rightclick.js
```diff
@@ -16,8 +16,9 @@ function createDomElement(html) {
...
+    constructor(infoOverlay, api) {
```
robwalch: Pass in a callback on `setup`. Let's keep the constructor clean and not add properties we don't need to the instance.

src/js/view/controls/rightclick.js
```diff
@@ -131,13 +132,21 @@ export default class RightClick {
...
+            };
+        }
     }
 
     setup(_model, _playerElement, layer) {
         this.playerElement = _playerElement;
         this.model = _model;
         this.mouseOverContext = false;
         this.layer = layer;
+        this.shareOnRightClick = (this.model.get('sharing') && this.model.get('sharing').shareOnRightClick);
```
robwalch: We don't really need a `shareOnRightClick` property. Just define `shareHandler` here, and use that and/or the presence of `'.jw-share-item'` to determine if sharing is enabled in this object.

src/js/view/controls/rightclick.js
```diff
@@ -156,6 +165,9 @@ export default class RightClick {
             this.el.addEventListener('mouseout', this.outHandler);
         }
         this.el.querySelector('.jw-info-overlay-item').addEventListener('click', this.infoOverlayHandler);
+        if (this.shareOnRightClick) {
```
robwalch: We don't need `this.shareOnRightClick`.

You could do this, but you should **use `UI` for tap support**.
```js
const shareItemElement = this.el.querySelector('.jw-share-item');
if (shareItemElement) {
   shareItemElement.addEventListener('click', this.shareHandler);
}
```

src/js/view/controls/rightclick.js
```diff
@@ -166,6 +178,9 @@ export default class RightClick {
             this.el.querySelector('.jw-info-overlay-item').removeEventListener('click', this.infoOverlayHandler);
             this.el.removeEventListener('mouseover', this.overHandler);
             this.el.removeEventListener('mouseout', this.outHandler);
+            if (this.shareOnRightClick) {
+                this.el.querySelector('.jw-share-item').removeEventListener('click', this.shareHandler);
+            }
```
robwalch: Use `UI` for tap support, and remove `shareOnRightClick`.

src/js/view/controls/controls.js
```diff
@@ -104,7 +105,7 @@ export default class Controls {
         this.infoOverlay = new InfoOverlay(element, model, api, visible => {
             toggleClass(this.div, 'jw-info-open', visible);
         });
-        this.rightClickMenu = new RightClick(this.infoOverlay);
+        this.rightClickMenu = new RightClick(this.infoOverlay, api);
```
robwalch: Don't change the constructor or add `api` (see other comments).

src/js/view/controls/rightclick.js
```diff
@@ -98,7 +99,7 @@ export default class RightClick {
     }
 
     lazySetup() {
-        const html = rightclickTemplate(this.buildArray(), this.model.get('localization'));
+        const html = rightclickTemplate(this.buildArray(), this.model.get('localization'), this.shareOnRightClick);
```
robwalch: Use `buildArray()` to add items. `<li class="jw-reset jw-rightclick-item">` is repetitive and now that there are three types of items with different content, just add a type to each item in the menu array and handle the three types of items in the template `rightClickItem` method.

For example:
```js
// Handle all items with this method:
const menuItems = items.map(item => rightClickItem(item, localization))


const rightClickItem = (item, localization) => {
    const { link = '', title = '', featured, showLogo, type } = item;
    const item.logo = showLogo ? `<span class="jw-rightclick-logo jw-reset"></span>` : '';
    return `<li class="jw-reset jw-rightclick-item ${featured ? 'jw-featured' : ''}">${itemContentTypes[type](item, localization)}</li>`;
};

const itemContentTypes = {
    link({ link = '', title = '', logo }) => `<a href="${link}" class="jw-rightclick-link jw-reset" target="_blank">${logo}${title}</a>`,
    info(item, localization) => `<button type="button" class="jw-reset jw-rightclick-link jw-info-overlay-item">${localization.videoInfo}</button>`;
    share(item, localization) => /* you get the gist*/
};


```

src/js/api/config.js
```diff
@@ -56,6 +56,7 @@ const Defaults = {
         rewind: 'Rewind 10 Seconds',
         settings: 'Settings',
         stop: 'Stop',
+        share: 'Share',
```
karimJWP: We might want to make `sharing` a block within `localization` (similar to `errors`) since there are plans to add more fields and customizations to sharing

DanFerrer:
@robwalch Addressed feedback and made the following changes:

- Removed the check for sharing from ```rightclick.js``` to ```controls.js```
- If ```sharing``` and ```sharing.shareOnRightClick``` exist we create an ```openShareMenu``` function which will call ```open``` from the plugin instance of the share menu and we pass that function into the ```setup``` function on ```rightclick```
- We then use that function in ```rightclick.js``` to determine if we should add/remove event listeners to the right click template for sharing.
- Made ```buildArray()``` responsible for determining what template to create so all the menu items now have a ```type``` property ```(share, info, link)``` and we use that type to determine what template to create like in your example before.

If thats okay to do, I also need to put in a PR in for commercial to add the ```type``` property for customLinks.

@karimJWP As discussed earlier, if we wanna make ```share``` point to an object, lets sync up on that next sprint.

DanFerrer:
Issue was caused by my commercial pr that I closed that had the same branch name. Going to delete the branch and rebuild the PR

src/js/view/controls/templates/rightclick.js
```diff
@@ -13,21 +9,16 @@ export default (menu, localization) => {
...
 };
 
-const infoOverlayItem = (videoInfoLocalization) => {
-    return (
-        `<li class="jw-reset jw-rightclick-item">` +
-            `<button type="button" class="jw-reset jw-rightclick-link jw-info-overlay-item">${videoInfoLocalization}</button>` +
-        `</li>`
-    );
+const itemContentTypes = {
+    link: ({ link = '', title = '', logo }) => `<a href="${link}" class="jw-rightclick-link jw-reset" target="_blank">${logo}${title}</a>`,
```
karimJWP: defaults come at a cost. I would suggest using fallbacks instead:
```
link: ({ link, title, logo }) => `<a href="${link || ''}" class="jw-rightclick-link jw-reset" target="_blank">${logo}${title || ''}</a>`,
```

PR #3040, opened 2018-08-28
[JW8-1843] Outstream player

vseventer:
### This PR will...
Set-up the player to support outstream functionality.

Add outstream functionality without any changes in plugins.

`playlist` and `playlistItem` events are emitted with the following playlist item: `{ sources: [{}] }`. This is required as analytics attempts to read `sources[0].type`.

Line 445 in `controller.js` is only necessary for outstream players without a valid tag - right now the player is always idle. This allows `playAd` on outstream players. If we don't want to support this, we can update this line to throw an error instead. 

jwplayer/jwplayer-commercial#5635

#### Addresses Issue(s):
JW8-1843

### DIFF
--- a/src/js/api/setup-steps.js
+++ b/src/js/api/setup-steps.js
@@ -37,7 +37,8 @@ function setPlaylistAttributes(model, playlist, feedData) {
 
 function loadProvider(_model) {
     return loadPlaylist(_model).then(() => {
-        if (destroyed(_model)) {
+        const adConfig = _model.get('advertising');
+        if (destroyed(_model) || (adConfig && adConfig.outstream)) {
             return;
         }
 
--- a/src/js/api/setup-steps.js
+++ b/src/js/api/setup-steps.js
@@ -37,8 +37,7 @@ function setPlaylistAttributes(model, playlist, feedData) {
 
 function loadProvider(_model) {
     return loadPlaylist(_model).then(() => {
-        const adConfig = _model.get('advertising');
-        if (destroyed(_model) || (adConfig && adConfig.outstream)) {
+        if (destroyed(_model)) {
             return;
         }
 
@@ -83,7 +82,7 @@ function isSkinLoaded(skinPath) {
     return false;
 }
 
-function loadSkin(_model) {
+export function loadSkin(_model) {
     const skinUrl = _model.get('skin') ? _model.get('skin').url : undefined;
     if (typeof skinUrl === 'string' && !isSkinLoaded(skinUrl)) {
         const isStylesheet = true;
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -441,7 +441,7 @@ Object.assign(Controller.prototype, {
                     mediaPool.prime();
                 }
 
-                if (_interruptPlay) {
+                if (_interruptPlay || (adConfig && adConfig.outstream)) {
                     // Force tags to prime if we're about to play an ad
                     // Resetting the source in order to prime is OK since we'll be switching it anyway
                     if (_inInteraction(window.event) && !_backgroundLoading) {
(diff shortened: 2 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

robwalch:
> jwplayer-commercial browserstack tests FAILURE

Please rebase to resolve the issue with commercial tests.

src/js/api/setup-steps.js
```diff
@@ -38,7 +38,8 @@ function setPlaylistAttributes(model, playlist, feedData) {
 
 function loadProvider(_model) {
     return loadPlaylist(_model).then(() => {
-        if (destroyed(_model)) {
+        const adConfig = _model.get('advertising');
+        if (destroyed(_model) || (adConfig && adConfig.outstream)) {
```
robwalch: Do we even need the loadProvider() in the setup promises for outsream? Could we remove it here to get around this check?
https://github.com/jwplayer/jwplayer-commercial/blob/v8.5.2/src/js/api/setup-steps.js#L49

src/js/controller/controller.js
```diff
@@ -730,6 +734,10 @@ Object.assign(Controller.prototype, {
 
         function _attachMedia() {
             // Called after instream ends
+            const adConfig = _model.get('advertising');
+            if (adConfig && adConfig.outstream) {
+                _programController.trigger(MEDIA_COMPLETE, {});
+            }
```
robwalch: For these case it probably makes sense to add these methods to the instance here and extend in https://github.com/jwplayer/jwplayer-commercial/blob/v8.5.2/src/js/controller/controller.js

src/js/controller/instream-adapter.js
```diff
@@ -442,7 +447,7 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         // when instream was inited and the player was not destroyed\
         _controller.attachMedia();
 
-        if (this.noResume) {
+        if (this.noResume || (adConfig && adConfig.outstream)) {
```
robwalch: Set `noResume` to true in commercial controller by extending `createInstream`.

src/js/controller/controller.js
```diff
@@ -421,7 +421,11 @@ Object.assign(Controller.prototype, {
                 return resolved;
             }
 
-            if (_model.get('state') === STATE_COMPLETE) {
+            const adConfig = _model.get('advertising');
+            if (adConfig && adConfig.outstream) {
+                _stop(true);
+                _model.setActiveItem(0);
+            } else if (_model.get('state') === STATE_COMPLETE) {
```
vseventer: @robwalch How can we move this block to commercial?

src/js/controller/controller.js
vseventer: Extending `attachMedia` works, but there's no reference to `_programController`. Any idea how to trigger `MEDIA_COMPLETE` from commercial?

vseventer:
Test this please.

src/js/controller/controller.js
```diff
@@ -422,7 +422,7 @@ Object.assign(Controller.prototype, {
 
             if (_model.get('state') === STATE_COMPLETE) {
                 _stop(true);
-                _setItem(0);
+                _this.setItemIndex(0);
```
robwalch: There are five calls to `_setItem` in the controller. Should we not make all calls go through the instance `_this.setItemIndex`?

vseventer: They are the same, as line 810 assigns `this.setItemIndex = _setItem`. For clarity, we can probably always use the `this` variant.

PR #3041, opened 2018-08-28
Feature/html5 video src emptied error

robwalch:
### This PR will...
Handle issues caused by 3rd party ads libraries clearing the video.src after an ad break or error.

If the src is cleared during a play attempt, the player will attempt to recover within the play promise chain.

If the src is cleared after playback has started a new error code - 224005 - will be included in errors to differentiate this type of failure from ones that result from MediaElement media not supported errors - 224003 (video.error.code 4).

This will add some resilience and tracking for errors we've identified on iPhone with IMA.

Changes to the media controller's play promise chain adds support for retries

Returning the promise chain from `mediaController._playAttempt` rather than just the provider's play promise from `mediaController._loadAndPlay` may impact the order of events for things that use it > `programController.playVideo()` > `controller _play` > `controller _autostart` | `controller|api.play()`.

We may want to consider just implementing the new error code and tracking the impact of this issue before extending the media controller's play promise to support retries.

#### Addresses Issue(s):
JW8-2158

### DIFF
--- a/src/js/program/media-controller.js
+++ b/src/js/program/media-controller.js
@@ -42,7 +42,7 @@ export default class MediaController extends Eventable {
             mediaModel.set('setup', true);
             playPromise = this._loadAndPlay(item, provider);
             if (!mediaModel.get('started')) {
-                this._playAttempt(playPromise, playReason);
+                playPromise = this._playAttempt(playPromise, playReason);
             }
         }
         return playPromise;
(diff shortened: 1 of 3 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3042, opened 2018-08-28
Prevent phone from going to fullscreen on Orientation change on audio mode and chromless players

DanFerrer:
### This PR will...

- Prevent ```setFullScreen``` from being called the player is audio only or chromeless.
- Destructure  ```model``` and ```api```from ```view``` argument

### Why is this Pull Request needed

- Player goes to a large black screen when we call ```setFullScreen``` on audio players
- Improving UX on chromeless players since viewer cant click on fullscreen button without controls.

#### Addresses Issue(s):

JW8-1851


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/utils/views-manager.js
```diff
@@ -56,17 +56,19 @@ function scheduleResponsiveRedraw() {
 
 function onOrientationChange() {
     views.forEach(view => {
-        if (view.model.get('visibility') >= 0.75) {
-            const state = view.model.get('state');
+        const { model } = view;
```
johnBartos: I wouldn't recommend destructuring when only one argument is extracted; it incurs a small transpiler cost

src/js/view/utils/views-manager.js
```diff
@@ -56,7 +56,7 @@ function scheduleResponsiveRedraw() {
 
 function onOrientationChange() {
     views.forEach(view => {
-        if (view.model.get('visibility') >= 0.75) {
+        if (!view.model.get('audioMode') && view.model.get('controls') && view.model.get('visibility') >= 0.75) {
```
johnBartos: Probably worth assigning `view.model` to a variable since it's used 4 times. It might also look neater to return early and reduce nesting by 1 indentation

robwalch:
@karimJWP @egreaves @johnBartos I'm fine merging as-is, but would like other opinions on gating fullscreen on and/or off vs listening for orientation change, and whether we'd need to handle controls or audio mode changing while in fullscreen: https://github.com/jwplayer/jwplayer/pull/3042#pullrequestreview-152121996

src/js/view/utils/views-manager.js
```diff
@@ -56,8 +56,9 @@ function scheduleResponsiveRedraw() {
 
 function onOrientationChange() {
     views.forEach(view => {
-        if (view.model.get('visibility') >= 0.75) {
-            const state = view.model.get('state');
+        const model = view.model;
+        if (!model.get('audioMode') && model.get('controls') && model.get('visibility') >= 0.75) {
```
karimJWP: I would suggest an early return over nested if statements.
```
if (model.get('audioMode') || !model.get('controls') || model.get('visibility') < 0.75) {
    return;
}
```

karimJWP: Also we don't need the returns in the in the following if statements
```
if (!isLandscape && state === 'paused' && view.api.getFullscreen()) {
                 // Set fullscreen to false when going back to portrait while paused and return early
                 view.api.setFullscreen(false);
} else if (state === 'playing') {
                 view.api.setFullscreen(isLandscape);
}
```

PR #3043, opened 2018-08-29
[DNM] [Experiment] Add wrapper div around contents of the player

radium-v:
### This PR will...
Add a div called `.jw-wrapper` that acts as an intermediate parent for manipulating the contents of the player without affecting the player container itself.

I keep running into similar issues while building experiments. Adding a wrapping div works in multiple cases.

There are likely some other elements that I'm not aware of which should be aware of this new div.

#### Related Issue:
JW8-1863

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -194,7 +194,7 @@ function View(_api, _model) {
         _captionsRenderer.setup(_playerElement.id, _model.get('captions'));
 
         // captions should be placed behind controls, and not hidden when controls are hidden
-        _playerElement.insertBefore(_captionsRenderer.element(), _title.element());
+        _title.element().parentNode.insertBefore(_captionsRenderer.element(), _title.element());
 
         // Display Click and Double Click Handling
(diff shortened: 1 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/view.js
```diff
@@ -194,7 +194,7 @@ function View(_api, _model) {
         _captionsRenderer.setup(_playerElement.id, _model.get('captions'));
 
         // captions should be placed behind controls, and not hidden when controls are hidden
-        _playerElement.insertBefore(_captionsRenderer.element(), _title.element());
```
robwalch: You should put `wrapperElement` in a var and use it in these places.

vseventer:
See #3092.

PR #3044, opened 2018-08-29
Return Promise.all result in Setup.start for compatibility with zone.js

robwalch:
### This PR will...
- Return `Promise.all` result in `Setup.start` for compatibility with [zone.js](https://github.com/angular/zone.js)
- Add setup error code 100,002 for falsey `Setup.start` result 

In 8.5.0 we introduced this change https://github.com/jwplayer/jwplayer/commit/c1c9b77b70dbf52310005b7e2ed36b5a6c1705ca#diff-49ba114a4fff54f99f3a7f73e2c67df0L15 which works when the Promise implementation is consistent.

Adding zone.js to the page after our library results in a mix of native Promise and ZoneAwarePromise instances being used. In this case results in the call to Promise.race([setup, timeout]) thinking that the timeout resolved first and returning that result.

Any players on pages that add zone.js after our library would fail setup with error code 100,000.
"//unpkg.com/zone.js@0.8.26/dist/zone.js"

Adding zone.js to the page before jwplayer.js would work because then we would use reference to ZoneAwarePromise consistently. We specifically avoid that and try to only use the reference to window.Promise on embed to avoid conflicts with third party libraries.

Despite us keeping a reference to window.Promise, not all of the frameworks that we use do. Webpack's chunk loader for example which is used in our setup promise chain accesses it globally and as a result introduces zone.js Promises into the chain. 

#### Addresses Issue(s):
JW8-2185

### DIFF
--- a/src/js/api/Setup.js
+++ b/src/js/api/Setup.js
@@ -12,9 +12,10 @@ const Setup = function(_model) {
 
     this.start = function (api) {
 
-        const setup = startSetup(_model, api, [
+        const setup = Promise.all([
             loadCoreBundle(_model),
-            loadPlugins(_model, api)
+            loadPlugins(_model, api),
+            startSetup(_model, api, [])
         ]);
 
         const timeout = new Promise((resolve, reject) => {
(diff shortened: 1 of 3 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

robwalch:
test this please

PR #3050, opened 2018-09-05
Use native font scaling when in iOS fullscreen

johnBartos:
### This PR will...
- Set the fontSize to initial when in iOS fullscreen instead of using the `getScaledFontSize` method

iOS uses extra magic to get the proper font scale that we haven't figured out. Instead of spending the effort to do it we can just remove our style and inherit iOS'

- This means that viewers loose custom styling in iOS fullscreen. Is that OK?
- Removing `px` from the `setFontSize` method causes all font sizes to be rounded up. Is this OK? 

No

#### Addresses Issue(s):

JW8-1814

### DIFF
--- a/src/js/view/captionsrenderer.js
+++ b/src/js/view/captionsrenderer.js
@@ -206,26 +206,27 @@ const CaptionsRenderer = function (viewModel) {
 
     function _setFontSize() {
         const height = _model.get('containerHeight');
+
         if (!height) {
             return;
         }
 
+        let fontSize;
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3055, opened 2018-09-06
Delete captions submenu & button on playlistItem change

johnBartos:
To mirror the silent captions list clearing on the `playlistItem` event done here:
https://github.com/jwplayer/jwplayer/blob/master/src/js/controller/captions.js#L19

We don't dispatch an empty `captionsList` event (because captions may be received later in a stream so we cannot definitively say captions are empty); however, our own internal events rely on an empty list to clear the submenu. This method clears out the menu on the same event which clears the captions.

Is there a better way to do this? Ideally we'd let the events propagate through the player but not out of it; however, this seemed pretty uncertain because I'm not 100% sure that we should never dispatch a captions list consisting of just `off`.

No

#### Addresses Issue(s):

JW8-1174

### DIFF
--- a/src/js/view/controls/settings-menu.js
+++ b/src/js/view/controls/settings-menu.js
@@ -163,6 +163,12 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api) 
     }, settingsMenu);
 
     // Captions
+    model.on('change:playlistItem', () => {
+        // captions.js silently clears captions when the playlist item changes. The reason it silently clear captions
+        // instead of dispatching an event is because we don't want to emit 'captionsList' if the new list is empty.
+        removeCaptionsSubmenu(settingsMenu);
+        controlbar.elements.captionsButton.hide();
+    });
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/settings-menu.js
```diff
@@ -171,6 +171,12 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api)
             controlbar.toggleCaptionsButtonState(!!index);
         }
     }, settingsMenu);
+    model.on('change:playlistItem', () => {
```
egreaves: This shouldn't affect the behavior, but let's put this handler above the `captionsList` handler to be safe since the order of events should be playlistItem > captionsList > captionsIndex.

PR #3057, opened 2018-09-07
[JW8-1840] Add autoPause.viewability.

vseventer:
### This PR will...
Add `autoPause` for autostarting videos, so they pause when transitioning from viewable to not viewable, and resume when viewable again. Config: `autoPause: { viewability: true }`.

Mainly for outstream, but applicable to all autostarting players.

* Feature only applicable to autostarting videos.
* User gestures (i.e. manual pause) cancel out `autoPause`.
* The above means if the video cannot autostart (i.e. browser restrictions), setting does not apply since a user gesture is needed before playback starts.
* `pauseReason` is `autostart`.

N/A

#### Addresses Issue(s):
JW8-1840

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -1,4 +1,3 @@
-import { OS } from 'environment/environment';
 import SimpleModel from 'model/simplemodel';
 import { INITIAL_PLAYER_STATE, INITIAL_MEDIA_STATE } from 'model/player-model';
 import { STATE_IDLE } from 'events/events';
@@ -186,9 +185,7 @@ const Model = function() {
         if (autoStart !== undefined) {
             this.set('autostart', autoStart);
         }
-
-        const autoStartOnMobile = OS.mobile && this.get('autostart');
-        this.set('playOnViewable', autoStartOnMobile || this.get('autostart') === 'viewable');
+        this.set('playOnViewable', this.get('autostart'));
     };
 
     this.resetItem = function (item) {
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -277,7 +277,7 @@ Object.assign(Controller.prototype, {
             }
 
             // Autostart immediately if we're not waiting for the player to become viewable first.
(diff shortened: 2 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

vseventer:
Test this please.

src/js/controller/model.js
```diff
@@ -186,9 +185,7 @@ const Model = function() {
         if (autoStart !== undefined) {
             this.set('autostart', autoStart);
         }
-
-        const autoStartOnMobile = OS.mobile && this.get('autostart');
-        this.set('playOnViewable', autoStartOnMobile || this.get('autostart') === 'viewable');
```
robwalch: Why is this logic being removed?

"autoPause" should only result in playback resuming when the player becomes visible, if it was auto paused. 

I wouldn't overload "playOnViewable" here. I would actually get rid of it and base all logic on "autostart", "autoPause" and user interaction which disables "playOnViewable" ("autostartViewable"). This logic doesn't belong here in model.js - keep it all in controller.js even if you want to keep the "playOnViewable" property.

vseventer: The logic is replaced by a looser definition of `playOnViewable`. This way, the `_checkPlayOnViewable` handler in `controller.js` is executed for `autostart: true` players as well. This handler then pauses the video if not viewable.

robwalch: This turns on viewable autostart for desktop players when `autostart` is set to `true` - Which changes behavior:

- if autostart is not allowed then view-ability changes
- when autostart is allowed ... ?

And what about this statement in controller `_checkAutoStart`?

```js
            if (_model.get('autostart') === true && !_model.get('playOnViewable')) {
                _autoStart();
            }
```

PR #3060, opened 2018-09-10
Get player language

karimJWP:
### This PR will...
- introduce `getLanguage` util
- add unit tests for `getLanguage`
- the first step to automatic translation is to detect the player's language. We first attempt to detect the html's language attribute. If the player is in an iFrame which does not have a lang attribute, we get it from the top window. If we do not have a lang attribute, we check the navigator's language properties. 
n/a
n/a
#### Addresses Issue(s):
JW8-1347

### DIFF
--- a/src/js/api/core-shim.js
+++ b/src/js/api/core-shim.js
@@ -13,7 +13,6 @@ import ErrorContainer from 'view/error-container';
 import MediaElementPool from 'program/media-element-pool';
 import SharedMediaPool from 'program/shared-media-pool';
 import UI, { getElementWindow } from 'utils/ui';
-import { getLanguage } from 'utils/language';
 import { PlayerError, composePlayerError, convertToPlayerError,
     SETUP_ERROR_LOADING_PLAYLIST, SETUP_ERROR_PROMISE_API_CONFLICT, SETUP_ERROR_UNKNOWN,
     MSG_TECHNICAL_ERROR } from 'api/errors';
@@ -88,7 +87,6 @@ Object.assign(CoreShim.prototype, {
         const setupConfig = options;
         const configuration = Config(Object.assign({}, options), persisted);
         configuration.id = api.id;
-        model.language = getLanguage();
         configuration.setupConfig = setupConfig;
         Object.assign(model.attributes, configuration, INITIAL_PLAYER_STATE);
         model.getProviders = function() {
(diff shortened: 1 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/utils/language.js
```diff
@@ -35,3 +36,16 @@ export function getLabel(language) {
 export function getCode(language) {
     return langToCode[language] || '';
 }
+
+export function getPlayerLanguage() {
+    let language = extractLanguage(document);
+    if (!language && isIframe) {
+        language = extractLanguage(window.top.document);
+    }
+    return language || navigator.language || navigator.browserLanguage || navigator.userLanguage || navigator.systemLanguage;
```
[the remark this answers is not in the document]
karimJWP: that's a good catch. I'll update the tickets to make the codes lower case when it comes to matching them. We want this function to return the code as written by the lang attribute or the browser properties to facilitate comparison.

src/js/utils/language.js
```diff
@@ -35,3 +36,16 @@ export function getLabel(language) {
 export function getCode(language) {
     return langToCode[language] || '';
 }
+
+export function getPlayerLanguage() {
```
robwalch: Wouldn't `getPageLanguage` be a more appropriate name. We will make this the language that the player uses, but we're getting it from the page.

test/data/model-properties.js
```diff
@@ -107,5 +108,6 @@ export default {
     setupConfig: {},
     bandwidthEstimate: null,
     bitrateSelection: null,
-    backgroundLoading: Features.backgroundLoading
+    backgroundLoading: Features.backgroundLoading,
+    language: getPlayerLanguage(),
```
robwalch: It's not optional if it's always added by the player. Don't we default it to 'en'?

src/js/api/core-shim.js
```diff
@@ -87,6 +88,7 @@ Object.assign(CoreShim.prototype, {
         const setupConfig = options;
         const configuration = Config(Object.assign({}, options), persisted);
         configuration.id = api.id;
+        configuration.language = getPlayerLanguage();
```
robwalch: Why set this here on the config? Does something else need to read this from anything other than the model? If not set this in the model, not the config.

karimJWP:
test this please

karimJWP:
test this please

PR #3071, opened 2018-09-19
Wait for plugins to load before handling setup errors

robwalch:
### This PR will...

- Wait for plugins to load before handling setup errors.
- Export each setup step so that each step can be extended.

This will allow plugins that listen to "setupError" to load before we trigger that event so that they can execute callbacks.

Exporting each step allows us to extend steps instead of blocking all promises with additional promises that only needs to block a single step.

https://github.com/jwplayer/jwplayer-commercial/pull/5724

#### Addresses Issue(s):

JW8-1830


egreaves:
I believe these failures are the same ones @dannyfinks mentioned need test updates.

PR #3077, opened 2018-09-24
Update translation files with "playback" property

DanFerrer:
### This PR will...

Add ```playback``` key to translation files with the same value as ```play```

```playback``` is the property being used for the click to play CTA when the player is idle. Its mapped to ```play```  because it performed the best during the a/b test

#### Addresses Issue(s):

JW8-2919

### CODE REVIEW

**Comment by johnBartos** [comment]
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

File: src/assets/translations/fr.json [hunk 1]
```diff
@@ -34,6 +34,7 @@
     "off": "Désactiver",
     "pause": "Pause",
     "play": "Play",
+    "playback": "Play",
```
robwalch: Playback is Play even in English?
You should explain in the PR description why we need another field which seems to echo play and isn't used in these changes.

DanFerrer: @robwalch Changed the PR description.

PR #3078, opened 2018-09-24
Wrap top language extraction in try block

karimJWP:
### This PR will...
wrap `window.top.document` in a `try` block.
When we fail to extract the language from an iFrame, we try to extract it from `window.top.document`. If the origins are different, attempting to access `window.top.document` will result in an exception; by wrapping it in a try block we prevent the exception and fallback to the browser's api. 
n/a
n/a
#### Addresses Issue(s):

JW8-2262

### DIFF
--- a/src/js/utils/language.js
+++ b/src/js/utils/language.js
@@ -50,7 +50,10 @@ function extractLanguage(doc) {
 export function getLanguage() {
     let language = extractLanguage(document);
     if (!language && isIframe()) {
-        language = extractLanguage(window.top.document);
+        try {
+            // Exception is thrown if iFrame is on a different domain name.
+            language = extractLanguage(window.top.document);
+        } catch (e) {/* ignore */}
     }
     return language || navigator.language || navigator.browserLanguage || navigator.userLanguage || navigator.systemLanguage || 'en';
 }


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/utils/language.js
```diff
@@ -50,7 +50,10 @@ function extractLanguage(doc) {
 export function getLanguage() {
     let language = extractLanguage(document);
     if (!language && isIframe()) {
-        language = extractLanguage(window.top.document);
+        try {
+            // Exception is thrown if iFrame is on a different domain name.
+            language = extractLanguage(window.top.document);
+        } catch (e) {/* ignore */}
     }
     return language || navigator.language || navigator.browserLanguage || navigator.userLanguage || navigator.systemLanguage || 'en';
```
robwalch: Why are we falling back to `navigator.language` or `navigator.browserLanguage`? These represent the preferred language of the user which I though we did not want to use.

robwalch: IE 11 supports `navigator.language`. If we want the user's preferred language, that is all we need to check. `navigator.browserLanguage || navigator.userLanguage || navigator.systemLanguage` are not needed.

PR #3081, opened 2018-09-26
[JW8-1838; DO NOT MERGE] floatOnScroll

pajong:
### This PR will...
Introduce a new setting, `floatOnScroll`, which will transform the player to a floating player if it is not visible.

* This implementation requires an extra wrapper around the player element, to allow for proper positioning.

jwplayer/jwplayer-commercial#5769

#### Addresses Issue(s):
JW8-1838


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/templates/floatingPlayer.js
```diff
@@ -0,0 +1,15 @@
+export default (id, ariaLabel = '') => {
```
vseventer: Instead of defining a new template, re-use the player one and add a third argument, `hasFloatOnScroll` and conditionally wrap the template in a `<div>` container.

src/css/jwplayer/flags/floatingplayer.less
```diff
@@ -0,0 +1,10 @@
+.jw-container {
+    margin-top: 2em;
+    background-size: cover;
+}
+
+.jw-flag-floating {
+    bottom: 2rem;
+    right: 2rem;
+    position: fixed;
```
vseventer: `z-index: 1` so we position ourselves on top of other positioned elements.

src/js/view/view.js
```diff
@@ -49,8 +51,25 @@ function View(_api, _model) {
...
     const _videoLayer = _playerElement.querySelector('.jw-media');
+    const _image = _model.get('image');
+    if (floatOnScroll) {
+        if (typeof _image === 'string') {
+            const backgroundImage = 'url("' + _image + '")';
+            style(_containerElement, {
+                backgroundImage: backgroundImage
+            });
+        } else {
+            style(_containerElement, {
```
vseventer: This can be added as `background-color` to the CSS of `jw-container`.

src/js/view/view.js
```diff
@@ -810,9 +840,30 @@ function View(_api, _model) {
         _model.set('intersectionRatio', entry.intersectionRatio);
     };
 
+    this.setFloatingIntersection = function (entry) {
+        console.log(entry.intersectionRatio);
```
vseventer: This method needs some improvement - apart from the `console.log`, I feel we can optimize the listeners below.

src/js/view/view.js
```diff
@@ -181,6 +195,14 @@ function View(_api, _model) {
         }
     }
 
+    this.stopFloating = function () {
+        if (hasClass(_playerElement, 'jw-flag-floating')) {
```
vseventer: Once we implement max. 1 floating player per page, we can improve this check to not be dependent on the class.

src/css/jwplayer/flags/floatingplayer.less
```diff
@@ -0,0 +1,36 @@
...
+    background-color: #000;
+    background-size: cover;
+
+    &.jw-flag-header {
+        margin-top: 2em;
+    }
+}
+
+.jw-float-close {
+    display: none;
```
vseventer: We can add the default styles here.

src/js/view/view.js
```diff
@@ -807,12 +842,41 @@ function View(_api, _model) {
...
+            floatingPlayer = _containerElement;
+
+            const width = _containerElement.offsetWidth;
+            const height = _containerElement.offsetHeight;
+
+            addClass(_playerElement, 'jw-flag-floating');
+            _this.trigger(FLOAT, { floating: true });
+
+            _this.resize(320, 320 * height / width);
+            style(_containerElement, { width, height }); // Keep original dimensions.
```
radium-v: This creates problems when using percentages. `Width` and `height` of the wrapper and the player need to match. Setting `width: 50%` causes the wrapper to be 50% of its container, and also causes the player to be 50% of the wrapper. Rather than having a div wrap around the player, try using one that wraps around the contents _inside_ the player div: https://github.com/jwplayer/jwplayer/pull/3043

PR #3082, opened 2018-09-26
Treat media errors during preload as warnings

johnBartos:
### This PR will...
- Re-emit MEDIA_ERROR as a warning on preload (i.e. when mediaModel `setup`is false)
- Reset video src on HTML5 provider error 

So that players which encountered an error preloading 1) don't show an error screen until interacted with and 2) are given a chance to retry.

Should the preload check (`mediaModel.get('setup')`) be changed to `mediaModel.get('preload') && mediaModel.get('setup')`? It seems that`setup` is  a good proxy because it's false before preload and true after.
 
https://github.com/jwplayer/jwplayer-commercial/pull/5759

#### Addresses Issue(s):

JW8-1816


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/providers/html5.js
```diff
@@ -159,6 +159,11 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
         webkitendfullscreen(e) {
             _fullscreenState = false;
             _sendFullscreen(e);
+        },
+
+        error() {
+            _clearVideotagSource();
```
robwalch: If you call `_clearVideotagSource` before `VideoEvents.error` you will break this error check:
```js
        } else if (errorCode === 3 || errorCode === 4) {
            code += errorCode - 1;
            if (errorCode === 4 && video.src === location.href) {
                code = HTML5_SRC_RESET;
            }
```

We should remove `error` from the mixin, since it fires HTML5 specific codes and we don't want those fired in hls or shaka.

In this work in progress, I'm overwriting the listener https://github.com/jwplayer/jwplayer-commercial/compare/master...feature/provider-error-handling#diff-9eff7f6edcda38b4ffe8c1b9c5559955R380

johnBartos:
test this please

PR #3084, opened 2018-09-28
Backwards support for localization options

DanFerrer:
### This PR will...

- Provide backwards support for options that were customizable outside of ```config.localization``` (```config.sharing```. ```config.advertising```, ```config.related```,  ```config.abouttext```)

We want to use the ```localization``` block for customizable fields in those plugin blocks moving forward without messing with the functionality of customers who still customize it the old way.

Making the function DRYer

https://github.com/jwplayer/jwplayer-plugin-related/pull/351
https://github.com/jwplayer/jwplayer-commercial/pull/5771

#### Addresses Issue(s):

JW8-1349

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -63,37 +63,26 @@ function _adjustDefaultBwEstimate(estimate) {
 
 function _copyToLocalization(allOptions) {
     const { advertising, related, sharing } = allOptions;
-    const localization = Object.assign({}, Defaults.localization, allOptions.localization);
+    const localization = Object.assign({}, allOptions.localization);
 
     if (advertising) {
-        localization.advertising = Object.assign({}, localization.advertising, {
-            admessage: advertising.admessage,
-            cuetext: advertising.cuetext,
-            loadingAd: localization.loadingAd,
-            podmessage: advertising.podmessage,
-            skipmessage: advertising.skipmessage,
-            skiptext: advertising.skiptext
-        });
+        localization.advertising = Object.assign({}, localization.advertising, advertising);
     }
 
     if (related) {
+        localization.related = localization.related || {};
         if (typeof related === 'object') {
-            localization.related = Object.assign({}, localization.related, {
-                autoplaymessage: related.autoplaymessage,
-                heading: related.heading,
-            });
+            localization.related.heading = localization.related.heading || related.heading;
+            localization.related.autoplaymessage = localization.related.autoplaymessage || related.autoplaymessage;
         } else {
-            localization.related = Object.assign({}, Defaults.localization.related, {
-                heading: related.heading
-            });
+            localization.related.heading = localization.related.heading || localization.related;
         }
     }
 
     if (sharing) {
-        localization.sharing = Object.assign({}, localization.sharing, {
-            copied: localization.copied,
-            heading: sharing.heading
-        });
+        localization.sharing = localization.sharing || {};
+        localization.sharing.heading = localization.sharing.heading || sharing.heading;
+        localization.sharing.copied = localization.sharing.copied || localization.copied;
     }
 
     localization.abouttext = localization.abouttext || allOptions.abouttext;
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -80,11 +80,11 @@ function _copyToLocalization(allOptions) {
         if (typeof related === 'object') {
             localization.related = Object.assign({}, localization.related, {
                 autoplaymessage: related.autoplaymessage,
-                heading: localization.nextUp || related.heading,
+                heading: related.heading,
             });
         } else {
             localization.related = Object.assign({}, Defaults.localization.related, {
-                heading: localization.related
+                heading: related.heading
             });
         }
     }
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -77,10 +77,16 @@ function _copyToLocalization(allOptions) {
     }
 
     if (related) {
-        localization.related = Object.assign({}, localization.related, {
-            autoplaymessage: related.autoplaymessage,
-            heading: localization.nextUp || related.heading,
-        });
+        if (typeof localization.related === 'object') {
+            localization.related = Object.assign({}, localization.related, {
+                autoplaymessage: related.autoplaymessage,
+                heading: localization.nextUp || related.heading,
+            });
+        } else {
+            localization.related = Object.assign({}, Defaults.localization.related, {
+                heading: localization.related
+            });
+        }
     }
 
     if (sharing) {
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -86,8 +86,7 @@ function _copyToLocalization(allOptions) {
     if (sharing) {
         localization.sharing = Object.assign({}, localization.sharing, {
             copied: localization.copied,
-            heading: sharing.heading,
-            link: sharing.link,
+            heading: sharing.heading
         });
     }
 
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -77,7 +77,7 @@ function _copyToLocalization(allOptions) {
     }
 
     if (related) {
-        if (typeof localization.related === 'object') {
+        if (typeof related === 'object') {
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/api/config.js
```diff
@@ -61,9 +61,50 @@ function _adjustDefaultBwEstimate(estimate) {
     return Defaults.bandwidthEstimate;
 }
 
+function _copyToLocalization(allOptions) {
+    const { advertising, related, sharing } = allOptions;
+    const localization = Object.assign({}, Defaults.localization, allOptions.localization);
+
+    if (advertising) {
+        localization.advertising = Object.assign({}, localization.advertising, {
```
karimJWP: since these map to exactly the same fields we can just do 
`localization.advertising = Object.assign({}, localization.advertising, advertising);`

src/js/api/config.js
```diff
@@ -61,9 +61,50 @@ function _adjustDefaultBwEstimate(estimate) {
...
+            skipmessage: advertising.skipmessage,
+            skiptext: advertising.skiptext
+        });
+    }
+
+    if (related) {
+        if (typeof localization.related === 'object') {
+            localization.related = Object.assign({}, localization.related, {
+                autoplaymessage: related.autoplaymessage,
+                heading: localization.nextUp || related.heading,
```
karimJWP: `related.heading` does not appear in the documentation; i say we don't need to backwards support it

src/js/api/config.js
```diff
@@ -61,9 +61,50 @@ function _adjustDefaultBwEstimate(estimate) {
     return Defaults.bandwidthEstimate;
 }
 
+function _copyToLocalization(allOptions) {
+    const { advertising, related, sharing } = allOptions;
+    const localization = Object.assign({}, Defaults.localization, allOptions.localization);
```
karimJWP: We shouldn't apply the localizations to the `Defaults.localization`, it would break the translation process.

src/js/api/config.js
```diff
@@ -61,9 +61,50 @@ function _adjustDefaultBwEstimate(estimate) {
...
+            cuetext: advertising.cuetext,
+            loadingAd: localization.loadingAd,
+            podmessage: advertising.podmessage,
+            skipmessage: advertising.skipmessage,
+            skiptext: advertising.skiptext
+        });
+    }
+
+    if (related) {
+        if (typeof localization.related === 'object') {
```
karimJWP:
```
if (typeof localization.related === 'string') {
    localization.related = {
                heading: localization.related
    }
}
```

src/js/api/config.js
karimJWP: why would we use `localization.nextUp`?

src/js/api/config.js
karimJWP: followed by
```
localization.related. autoplaymessage = localization.related. autoplaymessage || related.autoplaymessage;
```

src/js/api/config.js
```diff
@@ -61,9 +61,50 @@ function _adjustDefaultBwEstimate(estimate) {
...
+            });
+        } else {
+            localization.related = Object.assign({}, Defaults.localization.related, {
+                heading: localization.related
+            });
+        }
+    }
+
+    if (sharing) {
+        localization.sharing = Object.assign({}, localization.sharing, {
```
karimJWP: the problem here is that if the developer specified both in the localization block and in the sharing block, the localization block should take precedence;
instead i would write
```
localization.sharing.heading = localization.sharing.heading || sharing.heading;
localization.sharing. copied = localization.sharing. copied || localization. copied;
```

src/js/api/config.js
```diff
@@ -61,9 +61,39 @@ function _adjustDefaultBwEstimate(estimate) {
...
+    const localization = Object.assign({}, allOptions.localization);
+
+    if (advertising) {
+        localization.advertising = Object.assign({}, localization.advertising, advertising);
+    }
+
+    if (related) {
+        localization.related = localization.related || {};
+        if (typeof related === 'object') {
+            localization.related.heading = localization.related.heading || related.heading;
```
karimJWP: there is no `related.heading` in the documentation, why do we fallback to it ? https://developer.jwplayer.com/jw-player/docs/developer-guide/customization/configuration-reference/

src/js/api/config.js
```diff
@@ -61,9 +61,39 @@ function _adjustDefaultBwEstimate(estimate) {
...
+            localization.related.autoplaymessage = localization.related.autoplaymessage || related.autoplaymessage;
+        } else {
+            localization.related.heading = localization.related.heading || localization.related;
```
karimJWP: I think you might have overlooked one of my previous comments https://github.com/jwplayer/jwplayer/pull/3084#discussion_r221345406

karimJWP: this would break if `localization.related` is a string

src/js/api/config.js
```diff
@@ -61,9 +61,39 @@ function _adjustDefaultBwEstimate(estimate) {
...
+        }
+    }
+
+    if (sharing) {
+        localization.sharing = localization.sharing || {};
+        localization.sharing.heading = localization.sharing.heading || sharing.heading;
+        localization.sharing.copied = localization.sharing.copied || localization.copied;
+    }
+
+    localization.abouttext = localization.abouttext || allOptions.abouttext;
```
karimJWP: `const { advertising, related, sharing, abouttext } = allOptions;`
...
`localization.abouttext = localization.abouttext || abouttext;`

src/js/api/config.js
```diff
@@ -61,9 +61,53 @@ function _adjustDefaultBwEstimate(estimate) {
...
+        _mergeProperty(localization.advertising, advertising);
+    }
+
+    if (related) {
+        if (typeof localization.related === 'string') {
+            localization.related = {
+                heading: localization.related
+            };
+
+            _mergeProperty(localization.related, related, 'autoplaymessage');
```
karimJWP: this should be outside of the if block, since it is not related to `localization.related` being a string

src/js/api/config.js
```diff
@@ -61,9 +61,53 @@ function _adjustDefaultBwEstimate(estimate) {
...
+        localizationObj = Object.assign({}, localizationObj, allOptionsObj);
+    }
+}
+
+function _copyToLocalization(allOptions) {
+    const { advertising, related, sharing, abouttext } = allOptions;
+    const localization = Object.assign({}, allOptions.localization);
+
+    if (advertising) {
+        _mergeProperty(localization.advertising, advertising);
```
karimJWP: let's use object.assign here and remove the `else` block from `_mergeProperty` since that function should be property specific.
Also, I made a mistake when I told you to use the entire advertising block in `Object.assign({}, localization.advertising, advertising)` instead of your own object containing the specific advertising options. We should revert it to what you had before since the advertising block will contain other fields we don't want

src/js/api/config.js
```diff
@@ -61,9 +61,61 @@ function _adjustDefaultBwEstimate(estimate) {
...
+        _mergeProperty(localization.advertising, advertising, 'admessage');
+        _mergeProperty(localization.advertising, advertising, 'cuetext');
+        _mergeProperty(localization.advertising, advertising, 'loadingAd');
+        _mergeProperty(localization.advertising, advertising, 'podmessage');
+        _mergeProperty(localization.advertising, advertising, 'skipmessage');
+        _mergeProperty(localization.advertising, advertising, 'skiptext');
+    }
+
+    if (related) {
+        if (typeof localization.related === 'string') {
```
karimJWP: These if statements shouldn't be nested, because you can have `localization.related === 'string'` without a related block
```
        if (typeof localization.related === 'string') {
            localization.related = {
                heading: localization.related
            };
        } else {
            localization.related = localization.related || {};
        }

if (related) {
        _mergeProperty(localization.related, related, 'autoplaymessage');
}
```

PR #3089, opened 2018-09-30
Deprecate unused localization options

DanFerrer:
### This PR will...

- Remove ```nextUpClose``` from localization options.
- Remove ```more``` from localization options.

Deprecate options that no longer needed for localization 

#### Addresses Issue(s):

JW8-2218

### DIFF
--- a/src/assets/translations/en.js
+++ b/src/assets/translations/en.js
@@ -31,7 +31,6 @@ export default {
     more: 'More',
     next: 'Next',
     nextUp: 'Next Up',
-    nextUpClose: 'Next Up Close',
     notLive: 'Not Live',
     off: 'Off',
     pause: 'Pause',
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/assets/translations/en.js
```diff
@@ -28,10 +28,8 @@ export default {
     hd: 'Quality',
     liveBroadcast: 'Live',
     logo: 'Logo',
-    more: 'More',
     next: 'Next',
     nextUp: 'Next Up',
-    nextUpClose: 'Next Up Close',
```
karimJWP: It's not enough to remove this. It should be a fallback for `localization.close`. So in the backwards compatibility mapping:
`localization.close = localization.close || localization. nextUpClose;`

karimJWP: Let's address once (or in) https://github.com/jwplayer/jwplayer/pull/3084 is

PR #3097, opened 2018-10-03
Fix having black background when it should be transparent

pajong:
### This PR will...
Put transparent background to the newly-added .jw-wrapper class

jw-wrapper caused a regression of having black background, when it should be transparent.

#### Addresses Issue(s):

JW8-###

### DIFF
--- a/src/css/jwplayer/imports/jwplayerlayout.less
+++ b/src/css/jwplayer/imports/jwplayerlayout.less
@@ -30,6 +30,10 @@
         }
     }
 
+    .jw-wrapper {
+        background-color: transparent;
+    }
+
     .jw-aspect {
         display: none;
     }
--- a/src/css/jwplayer/imports/jwplayerlayout.less
+++ b/src/css/jwplayer/imports/jwplayerlayout.less
@@ -92,5 +92,6 @@
 
 .jw-wrapper {
     &:extend(.jwplayer);
+    background-color: transparent;
     height: 100%;
 }
--- a/src/css/jwplayer/imports/jwplayerlayout.less
+++ b/src/css/jwplayer/imports/jwplayerlayout.less
@@ -30,10 +30,6 @@
         }
     }
 
-    .jw-wrapper {
-        background-color: transparent;
-    }
-
     .jw-aspect {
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/css/jwplayer/imports/jwplayerlayout.less
```diff
@@ -30,6 +30,10 @@
         }
     }
 
+    .jw-wrapper {
```
vseventer: Remove background-color from `.jw-player` above, and move it to line 99 below.

src/css/jwplayer/states.less
```diff
@@ -4,6 +4,10 @@
 .jw-state-setup {
     background-color: transparent;
 
+    .jw-wrapper {
+        background-color: transparent;
```
robwalch: Should it always be transparent? (not just on setup). Please compare the current Views page in this build (http://player-pr-test-jenkins.longtailvideo.com/builds/17782/archive/test/view/) with an earlier version before the wrapper was added.

pajong: The wrapper is used for floating player, and it needs the black background for floating player.
Without the background, there is no black space when there needs to be.

pajong:
@robwalch Compared all scenario between this build (http://player-pr-test-jenkins.longtailvideo.com/builds/17793/archive/test/view/) and the build that you provided, and confirmed that all of them are matching.

src/css/jwplayer/states.less
robwalch: But it is also present when we're not in floating mode, so the black background is applied everywhere except for setup. 

`.jw-wrapper { background-color: transparent; } ` should be a default (in jwplayerlayout not here in state-setup) setting and only make it black when floating.

PR #3102, opened 2018-10-03
Float at the center if small player

pajong:
#### Addresses Issue(s):
JW8-1838

### DIFF
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -17,6 +17,13 @@
         z-index: 1;
     }
 
+    &.jw-flag-small-player {
+        .jw-wrapper {
+            right: 50%;
+            margin-right: -160px;
+        }
+    }
+
     .jw-float-close {
         display: block;
     }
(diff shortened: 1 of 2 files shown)


src/js/view/view.js
```diff
@@ -855,7 +863,7 @@ function View(_api, _model) {
         if (floatingPlayer === _playerElement) {
             floatingPlayer = null;
 
-            removeClass(_playerElement, 'jw-flag-floating');
+            removeClass(_playerElement, 'jw-flag-floating jw-flag-small-player');
```
pajong: Do we want to do this? Or do we want to do
`removeClass(_wrapperElement, 'jw-flag-small-player')`?

pajong:
@vseventer I can't approve it because I am the creator of the PR. I think you can't either since you contributed

src/css/jwplayer/flags/floatingplayer.less
```diff
@@ -10,6 +10,13 @@
         height: auto;
         margin-top: 2em;
         position: fixed;
+
+        &.jw-flag-small-player {
```
robwalch: Couldn't we just make this a media query to match the AC "Small screen portrait"?

PR #3100, opened 2018-10-03
Expose wrapperElement as a view api

pajong:
### This PR will...
Expose wrapperElement as a view api

### Why is this Pull Request needed?
We want to use `wrapperElement` in commercial project.

### Are there any Pull Requests open in other repos which need to be merged with this?
https://github.com/jwplayer/jwplayer-commercial/pull/5798

#### Addresses Issue(s):

JW8-1838

PR #3104, opened 2018-10-03
[JW8-1838] Misc. floating player bugfixes

vseventer:
### This PR will...
- Fix a lint error - 3d0675c89a6bb8d30ff28d98b1477c2c202aa2a9
- Add a `.jw-aspect` as direct child to `.jw-player`. This way we can maintain aspect ratio in the placeholder if the player is floating. - e0dd787824f87cbe69ec681c33f827cce880843a
- Re-apply `z-index: 1` to the wrapper to avoid our own shelf being positioned on top of the player. There still might be other (non-JW) elements on the page overlapping, but I don't think we want to `z-index: MAX_INTEGER`. /cc @radium-v - 735f46fd25344e1896c3bf8ee52dd80289d1ab3a
- Update how bounds are determined: use `.jw-player`, and only if the player is floating use the wrapper element. This fixes a bug with animating the height in appearing outstream players. - 45f86d89d2ba61b809a20ba865a1badf141fc0cf
- Update the background-image when we start floating, adding support for images in multiple playlist items. - 220545006cfd2e51521125b3719d7506a9ad46e2 c5eb00ba1abdc048d942ab4609e88f4e9179da43

Fix outstanding bugs with float-on-scroll.

N/A

PRs below can be merged independently, but fix other bugs with float-on-scroll:
#3097
#3100
#3102 	
jwplayer/jwplayer-commercial#5798
jwplayer/jwplayer-commercial#5799

#### Addresses Issue(s):
JW8-1838


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/templates/player.js
```diff
@@ -1,6 +1,7 @@
 export default (id, ariaLabel = '') => {
     return (
         `<div id="${id}" class="jwplayer jw-reset jw-state-setup" tabindex="0" aria-label="${ariaLabel}" role="application">` +
+            `<div class="jw-aspect jw-reset"></div>` +
```
pajong: On a non-outstream floating player, we will not show the poster image on the floating player if we do this. I think it is fine as it is though.

vseventer: Looks like you're right. Not sure why though?

pajong: We use `jw-aspect` to display the poster, so if we have this outside of the `jw-wrapper`, then it won't show on the floating player.

radium-v:
> Re-apply `z-index: 1` to the wrapper to avoid our own shelf being positioned on top of the player. There still might be other (non-JW) elements on the page overlapping, but I don't think we want to `z-index: MAX_INTEGER`. /cc @radium-v - 735f46f

A value of 1 is plenty since we just need to address when our own elements might be in the way (eg. shelf widget). Since every environment is unique, developers who use float-on-scroll should adjust the `z-index` to their liking via CSS:

```css
.jw-flag-floating .jw-wrapper {
    z-index: /* just big enough, but don't go overboard :) */;
}
```

PR #3113, opened 2018-10-05
Fire a warning on Translation load failures

karimJWP:
### This PR will...
- add warning codes for Translation load failures
To warn developers when translations fail to load. We do not need individual codes for each language, since the language code is in the model and will be pinged in PP-40
- I updated [the confluence list of errors ](https://jwplayer.atlassian.net/wiki/spaces/ES/pages/231637123/Documented+Errors+list), do I have to open a documentation PR as well ?
- Ajax's `onerror` gets called with arguments that we no longer use (url, message, xhr) can we get rid of these ? We are only using the error (4th argument). 
In fact, Shaka might be erroneously using Ajax, at https://github.com/jwplayer/jwplayer-commercial/blob/master/src/js/providers/shaka.js#L507 
```
ajax(..., (error) => reject(error));
```
where the error returned by Ajax is the key, not the Ajax error
n/a
#### Addresses Issue(s):
JW8-2244

### DIFF
--- a/src/js/utils/language.js
+++ b/src/js/utils/language.js
@@ -99,7 +99,8 @@ export function loadJsonTranslation(base, languageCode) {
     if (!translationLoad) {
         const url = `${base}translations/${normalizeLanguageCode(languageCode)}.json`;
         translationPromises[languageCode] = translationLoad = new Promise((oncomplete, reject) => {
-            ajax({ url, oncomplete, onerror: (message, file, _url, error) => reject(error), responseType: 'json' });
+            const onerror = (message, file, _url, error) => reject(error);
+            ajax({ url, oncomplete, onerror, responseType: 'json' });
         });
     }
     return translationLoad;
--- a/src/js/utils/language.js
+++ b/src/js/utils/language.js
@@ -98,8 +98,8 @@ export function loadJsonTranslation(base, languageCode) {
     let translationLoad = translationPromises[languageCode];
     if (!translationLoad) {
         const url = `${base}translations/${normalizeLanguageCode(languageCode)}.json`;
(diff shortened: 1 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/utils/language.js
```diff
@@ -98,7 +98,8 @@ export function loadJsonTranslation(base, languageCode) {
     let translationLoad = translationPromises[languageCode];
     if (!translationLoad) {
         const url = `${base}translations/${normalizeLanguageCode(languageCode)}.json`;
-        translationPromises[languageCode] = translationLoad = new Promise((oncomplete, onerror) => {
+        translationPromises[languageCode] = translationLoad = new Promise((oncomplete, reject) => {
+            const onerror = (message, file, _url, error) => reject(error);
```
robwalch: When we do encounter an error, do we want to leave a rejected promise in `translationPromises[languageCode]`? That will automatically trigger the same error for the next player that sets up on a page instead of reattempting the request.

karimJWP: True. how likely is it that the error will be a one off error vs a persistent error ? If it's persistent then we don't want to re try for every player.  Also would it be better to have consistency across players or is it better to have some translated and some not ? @aliciahurst any thoughts on the last question ?

karimJWP: @robwalch never mind, addressed

PR #3114, opened 2018-10-08
[JW8-2301] Round intersectionRatio to two decimals.

vseventer:
### This PR will...
Round the `intersectionRatio` to two decimals.

The `IntersectionObserver` polyfill (used in IE11) uses math to return the intersection ratio, which sometimes results in 1.000000123456 or 0.9999999123456 rather than 1.

N/A

N/A

#### Addresses Issue(s):
JW8-2301

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -292,7 +292,6 @@ function View(_api, _model) {
         _model.on('change:fullscreen', _fullscreen);
         _model.on('change:activeTab', updateVisibility);
         _model.on('change:fullscreen', updateVisibility);
-        _model.on('change:floating', updateVisibility);
         _model.on('change:intersectionRatio', updateVisibility);
         _model.on('change:visibility', redraw);
         _model.on('instreamMode', (instreamMode) => {
@@ -302,6 +301,10 @@ function View(_api, _model) {
                 destroyInstream();
             }
         });
+
+        if (_floatOnScroll) {
+            _model.on('change:intersectionRatio', _updateFloating);
+        }
 
         updateVisibility();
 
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/view.js
```diff
@@ -837,8 +837,11 @@ function View(_api, _model) {
     }
 
     function _setFloatingIntersection(entry) {
-        // Entirely invisible and no floating player already in the DOM
-        const isVisible = entry.intersectionRatio === 1;
+        // Round as the IntersectionObserver polyfill sometimes returns ±0.00XXX.
+        const intersectionRatio = Math.round(entry.intersectionRatio * 100) / 100;
```
robwalch: Can we do 1/20ths? We'll still get two decimals, but the second will always be 0 or 5 and we'll shave off two chars.

src/js/view/view.js
```diff
@@ -826,19 +829,18 @@ function View(_api, _model) {
...
     };
 
     function _getCurrentElement() {
         return _model.get('floating') ? _wrapperElement : _playerElement;
     }
 
-    function _setFloatingIntersection(entry) {
-        // Entirely invisible and no floating player already in the DOM
-        const isVisible = entry.intersectionRatio === 1;
+    function _updateFloating() {
```
robwalch: If this is a model "change:" handler, please use the listener args to avoid the extra call to `_model.get`:
```js
function _updateFloating(model, intersectionRatio) {
```

src/js/view/view.js
```diff
@@ -303,6 +302,10 @@ function View(_api, _model) {
             }
         });
 
+        if (_floatOnScroll) {
+            _model.on('change:intersectionRatio', _updateFloating);
```
robwalch: It's a bit odd to create a listener for this when you also set it in the same script. In general we use listeners for changes that happen somewhere else in the application, That is why `_setFloatingIntersection` was called directly before these changes.

vseventer: You're saying to call `_updateFloating` directly inside `this.setIntersection`?

robwalch: If that is the only place `model` "intersectionRatio" is set, then yes. Using event listeners to execute functions in the same context is inefficient.

vseventer: Yep, it is.

PR #3116, opened 2018-10-09
[JW8-2321] Set height on wrapper.

vseventer:
### This PR will...
Set height on the wrapper:
- 100% if absolute (px) height.
- auto if relative (%) height.
- 100% if fullscreen.

Previously, height on the wrapper was always 100%, causing incorrect player sizing.

N/A

Alternative approach in #3122.

#### Addresses Issue(s):
JW8-2321

### DIFF
--- a/src/css/jwplayer/flags/fullscreen.less
+++ b/src/css/jwplayer/flags/fullscreen.less
@@ -1,8 +1,4 @@
 .jwplayer.jw-flag-fullscreen {
-    /* !important required to ensure that fullscreen state is respected */
-    /* stylelint-disable declaration-no-important */
-    width: 100% !important;
-    height: 100% !important;
     /* stylelint-enable declaration-no-important */
     top: 0;
     right: 0;
@@ -11,4 +7,11 @@
     z-index: 1000;
     margin: 0;
     position: fixed;
+
+    .jw-wrapper {
+        /* !important required to ensure that fullscreen state is respected */
+        /* stylelint-disable declaration-no-important */
+        width: 100% !important;
+        height: 100% !important;
+    }
 }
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/css/jwplayer/flags/fullscreen.less
```diff
@@ -1,14 +1,17 @@
 .jwplayer.jw-flag-fullscreen {
-    /* !important required to ensure that fullscreen state is respected */
```
vseventer: Hmm, I'm wondering whether the width and height should be duplicated here, rather then moved? /cc @radium-v

robwalch: Yeah. Which element is fullscreened now? If it's still the player container couldn't it clip the wrapper? (overflow hidden)

src/css/jwplayer/imports/jwplayerlayout.less
```diff
@@ -19,12 +19,12 @@
     }
 
     // Aspect Ratio styles
-    &.jw-flag-aspect-mode {
+    &.jw-flag-aspect-mode .jw-wrapper {
         /* Height auto required for displaying aspect ratio correctly */
         /* stylelint-disable-next-line declaration-no-important */
         height: auto !important;
```
robwalch: This change also removes height settings from the player container which could result in regressions.

vseventer:
See #3122.

PR #3117, opened 2018-10-09
Add check for provider before calling a function

pajong:
### This PR will...
Check if `provider` is defined before calling a function within it

For outstream FreeWheel, `provider` is not loaded and therefore not defined, causing a console error.

https://github.com/jwplayer/jwplayer-ads-freewheel/pull/131

#### Addresses Issue(s):
JW8-2318


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/view.js
```diff
@@ -528,7 +528,9 @@ function View(_api, _model) {
         } else {
             // Request media element fullscreen (iOS)
             const provider = model.getVideo();
-            provider.setFullscreen(state);
+            if (provider) {
+                provider.setFullscreen(state);
+            }
```
robwalch: You're supposed to set a provide with instream.

robwalch: Link the FW PR please

pajong: We don't have fullscreen for IMA for iOS < 12.
For iOS 12 (ipad), the fullscreen works because `fullscreenHelpers.supportsDomFullscreen()` is true.
The `applyProviderListeners` function calls `_setProvider` in `ad-program-controller.js`, but that function doesn't set anything to the provider being passed in: https://github.com/jwplayer/jwplayer/blob/master/src/js/program/ad-program-controller.js#L97
To call `model.setProvider` with the instreamProvider, we need to implement several dummy functions (`getName`, `getPlaybackRate`, etc). Even after that, it causes different problems.

src/js/view/view.js
```diff
@@ -527,8 +527,11 @@ function View(_api, _model) {
             _toggleDOMFullscreen(_playerElement, state);
         } else {
             // Request media element fullscreen (iOS)
-            const provider = model.getVideo();
-            provider.setFullscreen(state);
+            const instreamProvider = model.get('instream') ? model.get('instream').provider : null;
```
vseventer: I think `model.getVideo` already contains this logic.

pajong: It just returns the `_provider`
https://github.com/jwplayer/jwplayer/blob/master/src/js/controller/model.js#L76

robwalch: Can you just do one `get`?

src/js/view/view.js
```diff
@@ -530,8 +530,12 @@ function View(_api, _model) {
             _toggleDOMFullscreen(_playerElement, state);
         } else {
             // Request media element fullscreen (iOS)
-            const provider = model.getVideo();
-            provider.setFullscreen(state);
+            const instream = model.get('instream');
+            const instreamProvider = instream ? instream.provider : null;
+            const provider = model.getVideo() || instreamProvider;
+            if (provider && typeof provider.setFullscreen === 'function') {
```
robwalch: `if (provider && provider.setFullscreen) {` is enough.

PR #3122, opened 2018-10-10
[JW8-2321] Absolutely position wrapper.

vseventer:
### This PR will...
Absolutely position the `jw-wrapper`, so that we get a more reliable behavior when it comes to:
- fullscreen: the container has 100% width/height, so wrapper will be fullscreen too. 
- aspectratio: both `.jw-aspect`s are either both displayed, or both are not.
- height: does not need to be set to 100%, auto, or initial since it relies on the positioning instead. For floating players, height is set inline, or uses aspectratio if relative.

Make things easier, plus reduce our CSS.

N/A

Alternative approach in #3116.

#### Addresses Issue(s):
JW8-2321

### DIFF
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -3,16 +3,13 @@
     overflow: visible;
     background-color: #000;
 
-    &.jw-flag-aspect-mode .jw-aspect {
-        display: block;
-    }
-
     .jw-wrapper {
         overflow: visible;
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

Issue #3126, opened 2018-10-11
Normalize playlist item duration


karimMourra:
### Expected Behavior

in playlist.js `normalizePlaylistItem`, normalize the duration

### Actual Behavior

in the related plugin we often call `timeFormat(seconds(duration))`. If duration is a strings it gets deconstructed into a number to then be reconstructed into the same string. By normalizing the duration we can execute less code and improve performance a bit. 

*Note* we would have to normalize the duration from related feed items as well. 

----------------------------------------------------------------------------------------

This template is provided as a guide to filing bugs and feature requests. Please replace 
all text *in italics* with details describing your issue. When submitting a feature request, 
"Steps to reproduce" can be substituted with a user story or list of acceptance criteria.

PR #3127, opened 2018-10-11
Format duration on nextup

karimJWP:
### This PR will...
- remove unnecessary conversion of strings to elements 
- convert duration to hh:mm:ss format when a number
duration should be displayed as hh:mm:ss
- anything wrong with removing `createElement` ? 
- It would be beneficial to normalize duration on playlist items, see https://github.com/jwplayer/jwplayer/issues/3126
n/a
#### Addresses Issue(s):

JW8-2345

### DIFF
--- a/src/js/view/controls/nextuptooltip.js
+++ b/src/js/view/controls/nextuptooltip.js
@@ -6,6 +6,7 @@ import Events from 'utils/backbone.events';
 import { cloneIcon } from 'view/controls/icons';
 import { seconds } from 'utils/strings';
 import { genId, FEED_SHOWN_ID_LENGTH } from 'utils/random-id-generator';
+import { timeFormat } from 'utils/parser';
 
 export default class NextUpTooltip {
     constructor(_model, _api, playerElement) {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/nextuptooltip.js
```diff
@@ -124,15 +125,15 @@ export default class NextUpTooltip {
...
             this.title.innerText = title ? createElement(title).textContent : '';
 
             // Set duration
-            if (nextUpItem.duration) {
+            const duration = nextUpItem.duration;
+            if (duration) {
                 this.duration = this.content.querySelector('.jw-nextup-duration');
-                const duration = nextUpItem.duration;
-                this.duration.innerText = duration ? createElement(duration).textContent : '';
+                this.duration.innerText = typeof duration === 'number' ? timeFormat(duration) : duration;
```
robwalch: What if `duration` is `undefined`? Have you tested this?

karimJWP: the line would never run because we're in an `if (duration)` block

PR #3128, opened 2018-10-11
Add 'forceLocalizationDefaults' config option

DanFerrer:
### This PR will...

- Add new config option ```forceLocalizationDefaults```
- Add unit test for config option:
  -  Check that page language doesn't change.
  -  Check that we don't try to load translation files if config option is ```true```
  -  Check we don't use ```intl``` block if given.
  -  Check localization options in config block still work.

New a/b test

https://github.com/jwplayer/borg/pull/1056
https://github.com/jwplayer/jwplayer-commercial/pull/5859

#### Addresses Issue(s):

JW8-2260

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -119,7 +119,7 @@ const Config = function(options, persisted) {
     _copyToLocalization(allOptions);
     _deserialize(allOptions);
 
-    const language = getLanguage();
+    const language = allOptions.forceLocalizationDefaults ? Defaults.language : getLanguage();
     const { localization, intl } = allOptions;
     allOptions.localization = applyTranslation(en, getCustomLocalization(localization, intl || {}, language));
 
--- a/test/unit/language-test.js
+++ b/test/unit/language-test.js
@@ -1,4 +1,4 @@
-import * as Language from 'utils/language';
+import { getLabel, getCode, getLanguage, translatedLanguageCodes, isTranslationAvailable, loadJsonTranslation, getCustomLocalization, isLocalizationComplete } from 'utils/language';
 import * as Browser from 'utils/browser';
 import Config from 'api/config';
 import en from 'assets/translations/en';
@@ -6,16 +6,6 @@ import sinon from 'sinon';
 
 describe('languageUtils', function() {
     const sandbox = sinon.sandbox.create();
-    const {
-        getLabel,
-        getCode,
-        getLanguage,
-        translatedLanguageCodes,
-        isTranslationAvailable,
-        loadJsonTranslation,
-        getCustomLocalization,
-        isLocalizationComplete
-    } = Language;
 
     describe('getLabel from unsupported codes', function() {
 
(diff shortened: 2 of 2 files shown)


test/unit/language-test.js
```diff
@@ -337,6 +341,30 @@ describe('languageUtils', function() {
...
+                intl
+            });
+            const configLocalization = config.localization;
+            expect(getLanguage()).to.equal('fr');
+            expect(configLocalization.play).to.equal(localizationPlay);
+            expect(configLocalization.pause).to.equal(localizationPause);
+            expect(configLocalization.stop).to.equal(localizationStop);
+            
+            Object.keys(en).forEach(key => {
+                if (key === 'play' || key === 'pause' || key === 'stop') {
```
karimJWP: we don't need this if block right ? Since what we care is that all localization values match the default values.
I assume you had this because those values are the same in our french translation, but that doesn't matter.

test/unit/language-test.js
```diff
@@ -337,6 +341,30 @@ describe('languageUtils', function() {
...
+        it('should only use custom localization block if "forceLocalizationDefaults" is true', function() {
+            stubHtmlLanguage(document, 'fr');
+            const config = new Config({
+                forceLocalizationDefaults: true,
+                localization,
+                intl
+            });
+            const configLocalization = config.localization;
+            expect(getLanguage()).to.equal('fr');
+            expect(configLocalization.play).to.equal(localizationPlay);
```
karimJWP: I would add a check that config.language === 'en'

test/unit/language-test.js
```diff
@@ -337,6 +341,30 @@ describe('languageUtils', function() {
             expect(customLocalization.pause).to.equal(localizationPause);
             expect(customLocalization.stop).to.equal(localizationStop);
         });
+
+        it('should only use custom localization block if "forceLocalizationDefaults" is true', function() {
+            stubHtmlLanguage(document, 'fr');
```
karimJWP: No need to stub the language on the hmtl, we can stub `getLanguage` to return 'fr'
import the entire language module `import * as Language from 'utils/language';`
`sandbox.stub(Language, 'getLanguage').returns('fr');`

test/unit/language-test.js
```diff
@@ -309,6 +309,10 @@ describe('languageUtils', function() {
             };
         });
 
+        after(function () {
```
karimJWP: I would move your unit test to it's own `describe` block. Make one for `forceLocalizationDefaults`

test/unit/language-test.js
```diff
@@ -337,6 +351,21 @@ describe('languageUtils', function() {
...
+
+        it('should only use custom localization block if "forceLocalizationDefaults" is true', function() {
+            sandbox.stub(Language, 'getLanguage').returns('fr');
+            const config = new Config({
+                forceLocalizationDefaults: true,
+                localization,
+                intl
+            });
+            const configLocalization = config.localization;
+            expect(Language.getLanguage()).to.equal('fr');
```
karimJWP: we don't need to test this; it would mean we are testing the stub

src/js/api/config.js
```diff
@@ -119,7 +119,7 @@ const Config = function(options, persisted) {
     _copyToLocalization(allOptions);
     _deserialize(allOptions);
 
-    const language = getLanguage();
+    const language = allOptions.forceLocalizationDefaults ? Defaults.language : getLanguage();
```
robwalch: What is the value of `Defaults.language`? Why don't we just set a language in the config, and base the test on `setupConfig.language`? I much rather we use existing config options and support the setting of `language` in the config, than add a new option that no one need like `forceLocalizationDefaults`.

karimJWP: `Defaults.language` is 'en'. I have mixed feelings about providing a config option to override language because we want to support google translate, which overrides the language tag on the page. If the player is setup after google translate then we read the updated language tag.
 if the language is being overriden via the config by the developer, then we have to count on them to update it when the page has been translated, which i doubt they would do.

test/unit/language-test.js
```diff
@@ -353,18 +343,23 @@ describe('languageUtils', function() {
         });
 
         it('should only use custom localization block if "forceLocalizationDefaults" is true', function() {
-            sandbox.stub(Language, 'getLanguage').returns('fr');
```
karimJWP: @DanFerrer why did you remove the stubbing ? If we don't stub then how are we verifying that we override the language when `forceLocalizationDefaults` is true ?

karimJWP: I think you might have misunderstood our conversation from friday. What i meant to say was that `expect(Language.getLanguage()).to.equal('fr');` is unnecessary since get language is being stubbed and we don't need to unit test the stubbing. But in order to properly test that `forceLocalizationDefaults ` overrides the language, we need to stub `getLanguage`, makes sense ?

PR #3131, opened 2018-10-12
Clear captionsrenderer display when resuming a live stream 

johnBartos:

So that the currently displaying cues do not re-appear when resuming the stream at a future point. An indeterminate amount of time may pass between pauses - the segment at which the current set of cues should have appeared could be long past.

Is it OK to always clear the renderer on live play? The `livePaused` flag is a bit more complicated in the providers.

This code could also use a bit of cleanup. I think it makes sense to take these states out of the `default` case.

https://github.com/jwplayer/jwplayer-commercial/pull/5875
#### Addresses Issue(s):

JW8-2313

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -716,8 +716,15 @@ function View(_api, _model) {
             default:
                 if (_captionsRenderer) {
                     _captionsRenderer.show();
-                    if (state === STATE_PAUSED && _controls && !_controls.showing) {
-                        _captionsRenderer.renderCues(true);
+                    if (state === STATE_PLAYING) {
+                        if (_model.get('streamType') === 'LIVE') {
+                            _captionsRenderer.clear();
+                        }
+                    }
+                    if (state === STATE_PAUSED) {
+                        if (_controls && !_controls.showing) {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/view.js
```diff
@@ -716,8 +716,15 @@ function View(_api, _model) {
...
-                        _captionsRenderer.renderCues(true);
+                    if (state === STATE_PLAYING) {
+                        if (_model.get('streamType') === 'LIVE') {
+                            _captionsRenderer.clear();
+                        }
+                    }
+                    if (state === STATE_PAUSED) {
+                        if (_controls && !_controls.showing) {
+                            _captionsRenderer.renderCues(true);
+                        }
```
robwalch: I don't think the issue we're trying to fix is related to the captions renderer updating in relation to the player state. We should be able to address this issue entirely in the provider where we are seeing the issue.

PR #3133, opened 2018-10-14
Do not rely on compute visibility if detect that in an iFrame

boyntoni:
### This PR will...

Does not rely on compute visibility if we detect we are within an iFrame.

Regardless of viewability, we are returning that the player is in view when it is in an iFrame. This is due to the method for viewability detection within computeVisibility. The method using the IntersectionObserver is reliable.

N/A

No

#### Addresses Issue(s):

JW8-2248

### DIFF
--- a/src/js/view/utils/visibility.js
+++ b/src/js/view/utils/visibility.js
@@ -27,7 +27,7 @@ export default function getVisibility(model, element) {
     }
     return intersectionRatio;
 }
- 
+
 function computeVisibility(target) {
     const html = document.documentElement;
     const body = document.body;
--- a/src/js/view/utils/visibility.js
+++ b/src/js/view/utils/visibility.js
@@ -17,7 +17,8 @@ export default function getVisibility(model, element) {
     // Otherwise, set it to the intersection ratio reported from the intersection observer
     let intersectionRatio = model.get('intersectionRatio');
 
-    if (intersectionRatio === undefined) {
+    // Compute visibility does not return accurate value for iFrame player
+    if (intersectionRatio === undefined && !(window.top !== window.self)) {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/utils/visibility.js
```diff
@@ -17,14 +17,15 @@ export default function getVisibility(model, element) {
...
+    // Compute visibility does not return accurate value for iFrame player
+    if (intersectionRatio === undefined && !(window.top !== window.self)) {
         // Get intersectionRatio through brute force
         intersectionRatio = computeVisibility(element);
     }
 
     return intersectionRatio;
 }
-
+
```
radium-v: whitespace

src/js/view/utils/visibility.js
```diff
@@ -17,14 +17,15 @@ export default function getVisibility(model, element) {
     // Otherwise, set it to the intersection ratio reported from the intersection observer
     let intersectionRatio = model.get('intersectionRatio');
 
-    if (intersectionRatio === undefined) {
+    // Compute visibility does not return accurate value for iFrame player
+    if (intersectionRatio === undefined && !(window.top !== window.self)) {
```
robwalch: Get the value of `computeVisibility(element);`. Then if it's > 0 and we're in an iFrame, throw it away. But if it comes back as 0 in an iFrame, we can trust that the player is not visible in the iFrame and therefor impossible to view in the page.

PR #3140, opened 2018-10-17
[JW8-2366] Use api rather than rect.

vseventer:
### This PR will...
Rely on the API `getWidth` and `getHeight` methods to return the player dimensions. Since these will only get called when the player is *not* floating, the container dimensions will be returned.

For outstream, using `rect` will return a height of `1`, causing a tiny floating player. Instead, we should use the expected height to calculate the floating player size.

N/A

jwplayer/jwplayer-commercial#5893

#### Addresses Issue(s):
JW8-2366

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -859,7 +859,8 @@ function View(_api, _model) {
         if (!isVisible && _model.get('state') !== STATE_IDLE && floatingPlayer === null) {
             floatingPlayer = _playerElement;
 
-            const rect = bounds(_playerElement);
+            const width = _this.api.getWidth();
+            const height = _this.api.getHeight();
 
             // Copy background from preview element, fallback to image config.
             style(_playerElement, {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

vseventer:
@robwalch I think I can actually use `getSafeRegion` here instead of calling the API - this does not involve any direct calls to API unless it is outstream (as `view-outstream.js` patches `getSafeRegion`).

robwalch:
@vseventer `getSafeRegion` is just `_lastWidth` and `_lastHeight` (aka model `containerWidth` and `containerHeight`).

vseventer:
@robwalch `getSafeRegion` does not always return `_lastHeight` for [outstream](https://github.com/jwplayer/jwplayer-commercial/blob/master/src/js/view/view-outstream.js#L16).

robwalch:
@vseventer Ah ha! Thanks.

PR #3142, opened 2018-10-18
[JW8-2368] Fix double call to _autoStart on viewable

vseventer:
### This PR will...
Fix a bug introduced with autoPause (3c1ff92) where `_autoStart` was called twice with `autostart: true` and viewable when setting up. Instead, autoPause is set-up only after calling `_autoStart`. This works fine since autoPause only works when autostarting (viewable or otherwise).

There is no need to call `_autoStart` twice, plus this was leading to duplicate events (e.g. `beforePlay`).

N/A

N/A

#### Addresses Issue(s):
JW8-2368

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -283,7 +283,7 @@ Object.assign(Controller.prototype, {
             }
 
             // Autostart immediately if we're not waiting for the player to become viewable first.
-            if (_model.get('autostart') === true) {
+            if (_model.get('autostart') === true && !_model.get('playOnViewable')) {
                 _autoStart();
             }
             apiQueue.flush();
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -501,6 +501,11 @@ Object.assign(Controller.prototype, {
                     });
                 }
 
+                // Enable autoPause behavior.
+                if (_model.get('autoPause') && _model.get('autoPause').viewability) {
+                    _model.set('playOnViewable', true);
+                }
+
                 return _play({ reason: 'autostart' }).catch(() => {
                     if (!_this._instreamAdapter) {
                         _model.set('autostartFailed', true);
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -502,7 +502,8 @@ Object.assign(Controller.prototype, {
                 }
 
                 // Enable autoPause behavior.
-                if (_model.get('autoPause') && _model.get('autoPause').viewability) {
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -501,6 +501,11 @@ Object.assign(Controller.prototype, {
                     });
                 }
 
+                // Enable autoPause behavior.
+                if (_model.get('autoPause') && _model.get('autoPause').viewability) {
```
karimJWP:
```
const autoPause = _model.get('autoPause');
if (autoPause && autoPause.viewability) {
```

src/js/controller/controller.js
```diff
@@ -283,7 +283,7 @@ Object.assign(Controller.prototype, {
             }
 
             // Autostart immediately if we're not waiting for the player to become viewable first.
-            if (_model.get('autostart') === true) {
+            if (_model.get('autostart') === true && !_model.get('playOnViewable')) {
```
karimJWP: would it make more sense to check for `!_model.get('playOnViewable')` before even calling `_checkAutoStart` ?

vseventer: @karimJWP They're both related to autostart, so imo it make sense to have this logic in a function called `_check*`.

src/js/controller/model.js
```diff
@@ -185,7 +186,9 @@ const Model = function() {
         if (autoStart !== undefined) {
             this.set('autostart', autoStart);
         }
-        this.set('playOnViewable', this.get('autostart'));
+
+        const autoStartOnMobile = OS.mobile && this.get('autostart');
```
karimJWP:
```
autoStart = autoStart === undefined ? this.get('autostart') : autoStart;
this.set('playOnViewable', autoStartOnMobile || autoStart === 'viewable');
```

karimJWP: or rather
```
if (autoStart === undefined) {
       autoStart = this.get('autostart');
} else {
}
const autoStartOnMobile = OS.mobile && autoStart;
this.set('playOnViewable', autoStartOnMobile || autoStart === 'viewable');
```

robwalch: What does this do?

(`undefined !== 'viewable'`)

PR #3144, opened 2018-10-18
Include sourceError in html5 provider errors

robwalch:
### This PR will...
Add the media element's `MediaError` to the "error" event's `sourceError` property.

## Why is this Pull Request needed?
`video.error` and thus `sourceError` is being set to `null` when `_clearVideotagSource()` is called.

These changes capture the error in a const before that so that it is passed to the `PlayerError` constructor.  

It's important that we provide the original `MediaError` so that it's `message` property can be examined for more information about decode errors in the browser.

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -176,7 +176,8 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
 
         error() {
             const { video } = _this;
-            const errorCode = (video.error && video.error.code) || -1;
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3145, opened 2018-10-19
[JW8-2234] Retrieve mediaController after setting the active item.

vseventer:
### This PR will...
Retrieve the `mediaController` only after the active item is set in `model.setActiveItem`. The latter updates the `playlistItem`, which cause the ad plugins to destroy the current ad break (if any) and re-attach the original media. Therefore, we should get the `mediaController` only after this completes, to ensure we got the latest one.

With ad breaks, upon updating the `playlistItem`, two video tags (one hidden which played, one visible which was idle) appeared on the page as the original video tag was never destroyed (because `mediaController` would be `null` as the ad was still playing).

N/A

N/A

#### Addresses Issue(s):
JW8-2234

### DIFF
--- a/src/js/program/program-controller.js
+++ b/src/js/program/program-controller.js
@@ -42,7 +42,7 @@ class ProgramController extends Eventable {
      * @memberOf ProgramController
      */
     setActiveItem(index) {
-        const { background, mediaController, model } = this;
+        const { model } = this;
         const item = model.get('playlist')[index];
 
         model.attributes.itemReady = false;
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3149, opened 2018-10-22
Safari 12 workaround: array.reverse is cached

karimMourra:
### This PR will...
Call `slice` on the axis array to make a copy before calling `reverse`.
Safari 12 introduced a bug where calling reverse on an array would then cache the reverse value (https://bugs.webkit.org/show_bug.cgi?id=188794). This went undetected in our testing because webVTT is not used in safari, since on safari we render captions natively. The iOS SDK however uses the webVTT polyfill because media is rendered in the native AVPlayer instead of the webView, therefore we cannot render captions natively in the webView.
To reproduce the bug run the following:
```
for (var i = 0; i <= 10; i ++) {
	var array1 = ['one', 'two'];
	console.log('array1: ', array1);
	var reversed = array1.reverse(); 
}
```
notice that even though the array is being instantiated, the value is reversed before `reverse` is even called.
The following screenshot shows the bug in action:
wtf
By reversing a copy of the array, we can work around the bug 
Upstream fix: https://github.com/mozilla/vtt.js/pull/377

`array.prototype.reverse` is not used anywhere else.
#### Addresses Issue(s):

iOS

### DIFF
--- a/src/js/polyfills/webvtt.js
+++ b/src/js/polyfills/webvtt.js
@@ -764,7 +764,7 @@ function moveBoxToLinePosition(window, styleBox, containerBox, boxPositions, num
             // We call reverse on a copy of the 'axis' array, instead of calling it directly on the original var because
             // Safari 12 introduced a bug where the reverse order of an array is cached after calling
             // array.prototype.reverse (https://bugs.webkit.org/show_bug.cgi?id=188794).
-            axis = axis.slice(0).reverse();
+            axis = axis.slice().reverse();
         }
 
         // Shift the position of the captions up to prevent minor overlaps as the text is laid out in IE11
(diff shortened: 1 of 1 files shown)


karimJWP:
@robwalch PR is up for upstream https://github.com/mozilla/vtt.js/pull/376. The bug is in Safari 12.0. The bug is fixed in a tech preview, but not sure when that will be released. On mobile devices, safari can only be updated when updating iOS.
My favorite comment from a [Y combinator discussion](https://news.ycombinator.com/item?id=18021835): "Finally we've discovered why people are asked in interviews to reverse an array."

johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

karimJWP:
Marking as 'Do not merge' because minified code still uses the hard-coded strings instead of variable encapsulation.

PR #3150, opened 2018-10-24
[JW8-2234] Do not resume ads when stopped

vseventer:
### This PR will...
Not resume playback when there was a change in playlist item during ad playback.

Skipping to the a different playlist item during an ad will cause the instream-adapter to call `programController.playVideo` for the old playlist item. We should avoid this and only play the new playlist item.

For Shaka particularly, loading two playlist items concurrently makes it throw an [7000](https://shaka-player-demo.appspot.com/docs/api/shaka.util.Error.html#value:7000) error.

`_getAdState()` implies `instreamAdapter` is set, do we need the extra `if (instream)` check?

#3145 

#### Addresses Issue(s):
JW8-2234


vseventer:
@robwalch Too much copy-paste? ^^

PR #3157, opened 2018-10-25
Clear tracks when resuming live playback in html5

robwalch:
### This PR will...
- Clear tracks when resuming live playback in html5
- Use model and item's `minDvrWindow`

Live streams reload when playback is resumed. With native text track rendering, tracks need to be cleared, otherwise they will be duplicated.

We should not be using the old `MIN_DVR_DURATION = 120` constant in html5 because it conflicts with the logic in the rest of the player. `minDvrWindow` can be set in the config per playlist item. 

#### Addresses Issue(s):
JW8-2289


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3159, opened 2018-10-26
Add resize-listener module

robwalch:
### This PR will...
Move resize listener to jwplayer/src/js/view/utils/ from jwplayer-plugin-related

This is a view util that we can use in other parts of the player where we need to respond to size change in the DOM where a window resize or orientation change would not be fired (ex: container style width is changed directly).

https://github.com/jwplayer/jwplayer-plugin-related/pull/361

#### Addresses Issue(s):
JW8-1798


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3167, opened 2018-11-02
Implement resize-listener css in js module

robwalch:
### This PR will...
- Implement resize-listener css in js module
- Add support for setting "content" in pseudo element class rules

So that additional CSS is not added to all players using related, only those using `ResizeListener`.

https://github.com/jwplayer/jwplayer-plugin-related/pull/362

#### Addresses Issue(s):
JW8-2398

### DIFF
--- a/src/js/view/utils/resize-listener.js
+++ b/src/js/view/utils/resize-listener.js
@@ -1,24 +1,51 @@
 import { requestAnimationFrame, cancelAnimationFrame } from 'os/utils/request-animation-frame';
 import { createElement } from 'os/utils/dom';
+import { css, style } from 'os/utils/css';
 
 export default class ResizeListener {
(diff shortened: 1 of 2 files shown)


src/js/view/utils/resize-listener.js
```diff
@@ -1,24 +1,51 @@
...
+            top: 0,
+            left: 0
+        };
+
+        const stretch = {
+            width: '100%',
+            height: '100%'
+        };
+
+        css('.jw-contract-trigger::before', Object.assign({
```
radium-v: Since `.jw-contract-trigger` and these other styles would theoretically only exist in this form of CSS-in-JS, any static analysis tools used for the CSS output can be rendered inaccurate.

PR #3169, opened 2018-11-05
[JW8-2057 JW8-2292] Handle playOnViewable for ad pauseReason.

vseventer:
### This PR will...
Update `playOnViewable` depending on the `pauseReason` passed by the ad plugins.

`autostart` depends on the `pauseReason` and `playOnViewable` properties, and currently does not work properly for VPAIDs.

N/A

* jwplayer/jwplayer-ads-googima#440
* jwplayer/jwplayer-ads-freewheel#144
* jwplayer/jwplayer-ads-vast#466

#### Addresses Issue(s):
JW8-2057
JW8-2292

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -207,14 +207,8 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         adModel.set('state', newstate);
 
         if (newstate === STATE_PLAYING) {
-            if (event.playReason) {
-                _model.set('playReason', event.playReason);
-            }
             _controller.trigger(AD_PLAY, event);
         } else if (newstate === STATE_PAUSED) {
-            if (event.pauseReason) {
-                _model.set('pauseReason', event.pauseReason);
-            }
             _controller.trigger(AD_PAUSE, event);
         }
     };
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -207,17 +207,14 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         adModel.set('state', newstate);
 
         if (newstate === STATE_PLAYING) {
+            if (event.playReason) {
+                _model.set('playReason', event.playReason);
+            }
             _controller.trigger(AD_PLAY, event);
         } else if (newstate === STATE_PAUSED) {
-            const pauseReason = event.pauseReason;
-            _model.set('pauseReason', pauseReason);
-
-            // Stop autoplay behavior if the ad was paused by the user or an api call.
-            if (pauseReason === 'interaction' || pauseReason === 'external' ||
-                 pauseReason === 'clickthrough' || pauseReason === 'vpaid') {
-                _model.set('playOnViewable', false);
+            if (event.pauseReason) {
+                _model.set('pauseReason', event.pauseReason);
             }
-
             _controller.trigger(AD_PAUSE, event);
         }
     };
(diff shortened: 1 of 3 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

vseventer:
Test this please.

vseventer:
Test this please.

src/js/controller/instream-adapter.js
```diff
@@ -207,8 +207,14 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
...
 
         if (newstate === STATE_PLAYING) {
+            if (event.playReason) {
+                _model.set('playReason', event.playReason);
+            }
             _controller.trigger(AD_PLAY, event);
         } else if (newstate === STATE_PAUSED) {
+            if (event.pauseReason) {
+                _model.set('pauseReason', event.pauseReason);
+            }
```
robwalch: instream should not be set player model state. These should be adModel play and pause reasons if we wanted this state in the adModel. We do not want to attribute the reason for content playback to the reason for ad playback after an ad.

These two models are used to track state. Not simply as event dispatchers. If you want to call `changeReason` in the controller when something happens here in instream, then you should trigger events from instream with handlers in the controller. The good thing is we already trigger events on the controller:

With that in mind, we can listen for `AD_PLAY` and `AD_PAUSE` in the controller instead of adding changes to the player model. In the controller this would look like:

```js
// this can be one line
_model.on('change:pauseReason change:playReason', changeReason);
// listen for play and pause reasons from instream
_this.on(AD_PLAY, (event) => { changeReason(null, event.playReason); });
_this.on(AD_PAUSE, (event) => { changeReason(null, event.pauseReason); });
```

src/js/controller/instream-adapter.js
```diff
@@ -192,10 +195,33 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         }
     }
 
+    function _triggerAdPlayPause(event) {
+        const { newstate } = event;
+        const adEvent = Object.assign({
+            oldstate: _adProgram.model.get('state')
+        }, _data, event);
+
+        if (adEvent.oldstate === newstate) {
```
robwalch: I would prefer that this check be performed before creating `adEvent`.
```js
const { newstate } = event;
const oldstate = _adProgram.model.get('state');
if (oldstate === newstate) {
  return;
...
const adEvent = Object.assign({ oldstate }, _data, event);
...
```

src/js/controller/instream-adapter.js
```diff
@@ -192,10 +195,33 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
...
+            _controller.trigger(AD_PLAY, adEvent);
+            _data = { }; // Reset.
+        } else if (newstate === STATE_PAUSED) {
+            _controller.trigger(AD_PAUSE, adEvent);
+            _data = { }; // Reset.
+        }
+    }
+
+    this.setEventData = function(data) {
+        _data = data;
```
robwalch: This should still be settable (or reset) via `instream.play(meta)`, `instream.pause(meta)` and potentially `instream.loadItem`. Resetting on `init` and `destroy` would be good practice as well.

And what about `_instreamItemNext`? Shouldn't this also reset the ad event data (each ad pod should have a new event data)?

src/js/controller/instream-adapter.js
vseventer: The `oldstate` might come be set on [`event`](https://github.com/jwplayer/jwplayer/pull/3169/files#diff-4b5d9c794240aa20cb620353eb027d56R114) though.

PR #3170, opened 2018-11-06
Use fastSeek when scrubbing

karimMourra:
### This PR will...
- call `fastSeek` when seek is called while scrubbing and fastSeek is available.
- certain providers support an imprecise seek (a.k.a fastSeek) which allows a frame to appear sooner. We want to use `fastSeek` while scrubbing because we prefer a quicker frame rendition over frame precision.
In my opinion an if statement would read better, but the ternary operator minifies better so i chose the latter. Any thoughts ?
n/a
#### Addresses Issue(s):

JW8-2379

### DIFF
--- a/src/js/program/media-controller.js
+++ b/src/js/program/media-controller.js
@@ -304,9 +304,11 @@ export default class MediaController extends Eventable {
 
     set position(pos) {
         const { provider } = this;
-        const { fastSeek } = provider;
-        const seek = this.model.get('scrubbing') && fastSeek ? fastSeek : provider.seek;
-        seek(pos);
+        if (this.model.get('scrubbing') && provider.fastSeek) {
+            provider.fastSeek(pos);
+        } else {
+            provider.seek(pos);
+        }
     }
 
     set quality(index) {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/program/media-controller.js
```diff
@@ -303,7 +303,10 @@ export default class MediaController extends Eventable {
     }
 
     set position(pos) {
-        this.provider.seek(pos);
+        const { provider } = this;
+        const { fastSeek } = provider;
+        const seek = this.model.get('scrubbing') && fastSeek ? fastSeek : provider.seek;
+        seek(pos);
```
robwalch: We have to call `provider.seek` for `this` to refer to the provider. With these change it will refer to `window`.

Let's make it simpler:

```js
    const { provider } = this;
    if (this.model.get('scrubbing') && provider.fastSeek) {
        provider.fastSeek(pos);
    } else {
        provider.seek(pos);
    }
```

PR #3176, opened 2018-11-09
Fallback to model's state when video element is absent

karimMourra:
### This PR will...
- check the model's state when the video element is absent.
- The Mobile SDKs render video natively, therefore the webView in which jwplayer.js runs does not have a video tag. Since the video tag is absent, the buffer state is never set on the model; as a result, the state stays at `playing`, so the change required to activate the `playHandler` in adPlayer.js never gets triggered, preventing adImpressions from firing in adPods. 
By falling back to the model's state, we can ensure a consistent experience across all platforms.
#### Addresses Issue(s):

SDK Channel 5

### DIFF
--- a/src/js/program/media-controller.js
+++ b/src/js/program/media-controller.js
@@ -121,7 +121,7 @@ export default class MediaController extends Eventable {
             playReason
         });
         // Immediately set player state to buffering if these conditions are met
-        if (video && !video.paused) {
+        if (video ? !video.paused : model.get(PLAYER_STATE) !== STATE_PAUSED) {
             model.set(PLAYER_STATE, STATE_BUFFERING);
         }
 
--- a/src/js/program/media-controller.js
+++ b/src/js/program/media-controller.js
@@ -7,7 +7,7 @@ import { MediaModel } from 'controller/model';
 import { seconds } from 'utils/strings';
 import {
     MEDIA_PLAY_ATTEMPT, MEDIA_PLAY_ATTEMPT_FAILED, MEDIA_COMPLETE,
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

karimMourra:
@johnBartos  I confirmed the changes do not break the fix. I used the test configs in the ticket but could not map to http://arrow.wikia.com/wiki/Oliver_Queen.  
the changes for the line I am touching were a just a clean up; since rob introduced a video var, he replaced `provider.video` with `video`. The actual bug fix lies in assigning the play attempt to the playPromise.

src/js/program/media-controller.js
```diff
@@ -121,7 +121,7 @@ export default class MediaController extends Eventable {
             playReason
         });
         // Immediately set player state to buffering if these conditions are met
-        if (video && !video.paused) {
+        if (video ? !video.paused : model.get(PLAYER_STATE) !== STATE_PAUSED) {
```
robwalch: It would be safer to check the for the one state that prevents impressions from firing.

You said in the description it's the play state, or is it the idle state? We should not accept both cases.

`video.paused` is a flag, not part of a state enum, so checking that the player is not paused is not the same as checking if video is expected to play after this point.

PR #3177, opened 2018-11-12
RTL support for concatenation

karimMourra:
### This PR will...
- introduce the `isRtl` util, which will return true when the language is RTL and:
-- the language is defined in the intl block
-- or the translation json is successfully downloaded
- set the 'dir' tag to `auto` for UI elements containing concatenated strings
- Some strings are concatenated with punctuation clearly indicating a sense of direction (i.e. a colon). For these concatenations, it is sufficient to set the dir tag to 'auto' and let the browser handle the component ordering. For the remaining cases, we must rely on the `isRtl` util.  
#### Addresses Issue(s):

JW8-2356

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -3,7 +3,7 @@ import { serialize } from 'utils/parser';
 import { isValidNumber, isNumber, pick, isBoolean } from 'utils/underscore';
 import { Features } from 'environment/environment';
 import en from 'assets/translations/en.js';
-import { getLanguage, getCustomLocalization, applyTranslation } from 'utils/language';
+import { getLanguage, isRtl, getCustomLocalization, applyTranslation, normalizeIntl, isInIntl } from 'utils/language';
 
 /* global __webpack_public_path__:true */
 /* eslint camelcase: 0 */
@@ -22,6 +22,7 @@ const Defaults = {
     height: 360,
     intl: {},
     language: 'en',
+    languageDir: 'ltr',
     liveTimeout: null,
     localization: en,
     mute: false,
@@ -120,7 +121,9 @@ const Config = function(options, persisted) {
     _deserialize(allOptions);
 
     const language = allOptions.forceLocalizationDefaults ? Defaults.language : getLanguage();
-    const { localization, intl } = allOptions;
+    const { localization } = allOptions;
+    const intl = normalizeIntl(allOptions.intl);
+
     allOptions.localization = applyTranslation(en, getCustomLocalization(localization, intl || {}, language));
 
     let config = Object.assign({}, Defaults, allOptions);
@@ -135,6 +138,10 @@ const Config = function(options, persisted) {
     config.volume = isValidNumber(config.volume) ? Math.min(Math.max(0, config.volume), 100) : Defaults.volume;
     config.mute = !!config.mute;
     config.language = language;
+    config.intl = intl;
+    if (isInIntl(intl, language) && isRtl(language)) {
+        config.languageDir= 'rtl';
+    }
 
     let rateControls = config.playbackRateControls;
 
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -3,7 +3,7 @@ import { serialize } from 'utils/parser';
 import { isValidNumber, isNumber, pick, isBoolean } from 'utils/underscore';
 import { Features } from 'environment/environment';
 import en from 'assets/translations/en.js';
-import { getLanguage, isRtl, getCustomLocalization, applyTranslation, normalizeIntl, isInIntl } from 'utils/language';
+import { getLanguage, getCustomLocalization, applyTranslation, normalizeIntl } from 'utils/language';
 
 /* global __webpack_public_path__:true */
 /* eslint camelcase: 0 */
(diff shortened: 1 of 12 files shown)


src/js/api/config.js
```diff
@@ -135,6 +138,10 @@ const Config = function(options, persisted) {
     config.volume = isValidNumber(config.volume) ? Math.min(Math.max(0, config.volume), 100) : Defaults.volume;
     config.mute = !!config.mute;
     config.language = language;
+    config.intl = intl;
+    if (isInIntl(intl, language) && isRtl(language)) {
+        config.languageDir= 'rtl';
```
pajong: small comment - space before `=`

src/templates/error.js
```diff
@@ -6,11 +6,11 @@ export default (id, message, errorCode, code) => {
                 `<style>` +
                 `[id="${id}"].jw-error{background:#000;overflow:hidden;position:relative}` +
                 `[id="${id}"] .jw-error-msg{top:50%;left:50%;position:absolute;transform:translate(-50%,-50%)}` +
-                `[id="${id}"] .jw-error-text{color:#FFF;font:14px/1.35 Arial,Helvetica,sans-serif}` +
+                `[id="${id}"] .jw-error-text{text-align:start;color:#FFF;font:14px/1.35 Arial,Helvetica,sans-serif}` +
                 `</style>` +
                 `<div class="jw-icon jw-reset"></div>` +
                 `<div class="jw-info-container jw-reset">` +
-                    `<div class="jw-error-text jw-reset">${(message || '')}<span class="jw-break jw-reset"></span>${detail}</div>` +
+                    `<div class="jw-error-text jw-reset-text" dir="auto">${(message || '')}<span class="jw-break jw-reset"></span>${detail}</div>` +
```
pajong: `class ="jw-error-text jw-reset-text" dir="auto"` is correct?

src/js/view/controls/controlbar.js
```diff
@@ -216,6 +217,7 @@ export default class Controlbar {
         }, () => {
             feedShownId = '';
         });
+        setAttribute(nextElement, 'dir', 'auto');
```
pajong: What does this do?

karimMourra: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/dir
"The dir global attribute is an enumerated attribute indicates the directionality of the element's text"
"dir=auto lets the user agent decide. It uses a basic algorithm as it parses the characters inside the element until it finds a character with a strong directionality, then applies that directionality to the whole element."

src/templates/error.js
karimMourra: good catch, that was meant to be in #3179

src/js/utils/language.js
```diff
@@ -76,17 +94,24 @@ export function getLanguage() {
     return language || navigator.language || 'en';
 }
 
+export function isRtl(lang) {
+    return rtlLangs.indexOf(normalizeLanguageCode(lang)) > -1;
```
vseventer: On line `104`, we use `>= 0` - would be nice to keep it consistent.

src/js/view/controls/rightclick.js
```diff
@@ -23,21 +23,24 @@ export default class RightClick {
...
+        const poweredBy = model.get('localization').poweredBy;
+        const versionSpan = `<span class="jw-reset">JW Player ${majorMinorPatchPre}</span>`;
 
         const menu = {
             items: [{
                 type: 'info'
             },
             {
-                title: `${this.model.get('localization').poweredBy} <span class="jw-reset">JW Player ${majorMinorPatchPre}</span>`,
+                title: model.get('languageDir') === 'rtl' ? `${versionSpan} ${poweredBy}` : `${poweredBy} ${versionSpan}`,
```
robwalch: I don't think it makes sense to add a config property for this one if statement. Where else and what else is it used for?

Why don't you just call `isRtl(config.language)` here.

robwalch:
I'm not sure if "translation json is successfully downloaded" should be included in the is RTL util. The test shouldn't file while a language is still being loaded or hasn't started loading yet.

test/unit/setup-steps-test.js
```diff
@@ -43,4 +44,16 @@ describe('Load Translations', function () {
...
+
+    it('should be RTL when RTL translation loads successfully', function () {
+        const rtlLang = 'ar';
+        const rtlModel = Object.assign({ language: rtlLang, intl: {} }, model);
+        rtlModel.attributes.language = rtlLang;
+        sandbox.stub(Utils, 'ajax').callsFake(({ oncomplete }) => {
+            oncomplete({ response: {} });
+        });
+        loadTranslations(rtlModel).then(() => {
+            expect(Language.isRtl(rtlModel.language, rtlModel.intl)).to.be.true;
```
robwalch: If you really want this to return false before the translation is loaded you should test that as well (you can add the `expect.to.be.false` before loading in this test case.

src/js/utils/language.js
```diff
@@ -78,12 +81,22 @@ export function getLanguage() {
 
 export const translatedLanguageCodes = ['ar', 'da', 'de', 'es', 'fr', 'he', 'it', 'ja', 'nl', 'no', 'pt', 'ro', 'sv', 'tr', 'zh'];
 
+export function isRtl(message) {
+    // RTL regex can be improved with ranges from:
```
robwalch: We should return `false` early `if (!message)` to be safe.

PR #3178, opened 2018-11-13
Fix starttime regression

pajong:
### This PR will...
Fix iOS regression with seek before the start of the content using `starttime` from the following commit: https://github.com/jwplayer/jwplayer/commit/9e40fccb989592939dd7aa32858615b3a9b26c9b

`getSeekRange` returns `end: _videotag.duration` when `_videotag.seekable` does not have a length.
This results `end` of `seekRange` to be defined as the duration of the video when there should be no `seekable` range, and therefore tries to seek instead of delaying the seek attempt until the video is ready.

#### Addresses Issue(s):

JW8-2423

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -613,7 +613,7 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
             seekPos += seekRange.start + seekRange.end;
         }
         if (!_canSeek) {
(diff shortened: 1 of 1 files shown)


dannyfinks:
@pajong verified that the recs feed failing here was a flake and that the IMA IE11 failure is due to the third party IMA issue I am going to filter that test out

PR #3183, opened 2018-11-20
Add border when focused with keyboard

pajong:
### This PR will...
Add border when focused with keyboard

It is not apparent which element is focused. We want the viewer to know what element is focused when using keyboard.

https://github.com/jwplayer/jwplayer-plugin-related/pull/365

#### Addresses Issue(s):

JW8-2330

### DIFF
--- a/src/js/view/controls/components/button.js
+++ b/src/js/view/controls/components/button.js
@@ -1,5 +1,6 @@
 import UI from 'utils/ui';
 import svgParse from 'utils/svgParser';
+import { addFocusHandler } from 'utils/dom';
 
 export default function (icon, apiAction, ariaText, svgIcons) {
     const element = document.createElement('div');
@@ -21,9 +22,7 @@ export default function (icon, apiAction, ariaText, svgIcons) {
 
     // Prevent button from being focused on mousedown so that the tooltips don't remain visible until
     // the user interacts with another element on the page
-    element.addEventListener('mousedown', (e) => {
-        e.preventDefault();
-    });
+    addFocusHandler(element);
 
     if (svgIcons) {
         Array.prototype.forEach.call(svgIcons, svgIcon => {
--- a/src/js/view/controls/components/button.js
+++ b/src/js/view/controls/components/button.js
@@ -20,8 +20,7 @@ export default function (icon, apiAction, ariaText, svgIcons) {
         });
     }
 
-    // Prevent button from being focused on mousedown so that the tooltips don't remain visible until
-    // the user interacts with another element on the page
+    // prevent button from having a border when focused through click
     addFocusHandler(element);
 
     if (svgIcons) {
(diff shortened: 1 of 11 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/components/button.js
```diff
@@ -21,9 +22,7 @@ export default function (icon, apiAction, ariaText, svgIcons) {
 
     // Prevent button from being focused on mousedown so that the tooltips don't remain visible until
     // the user interacts with another element on the page
```
pajong: How do we want to handle the tooltip?

vseventer:
Tricky problem - one thing I noticed is that the ad skip button does not have an outline when navigating with keyboard.

karimMourra:
@pajong I notice that the tooltips have the focus ring, is that expected behavior ? @aliciahurst do you have a preference? The AC says "Tooltips and other text decorations should also appear, but cannot occlude the visibility of the focus ring" which i do not interpret as requesting a focus ring around the tooltip. 
I also notice that the seek bar and menu items do not have focus rings, is that desired behavior ? @aliciahurst please take http://player-pr-test-jenkins.longtailvideo.com/builds/18697/archive/test/squash/bootstrap.html?primary=html5&edition=ads&feature=player_ui/player_ui_with_the_kitchen_sink&suite=Player%20UI%20and%20Customization for a spin

karimMourra:
@pajong also there is no focus rings around the related items in the overlay

src/js/view/controls/components/button.js
```diff
@@ -19,11 +20,8 @@ export default function (icon, apiAction, ariaText, svgIcons) {
         });
     }
 
-    // Prevent button from being focused on mousedown so that the tooltips don't remain visible until
```
karimMourra: I'm confused by the tooltip part of this comment. Are we sure the behavior is being respected ?

pajong:
> Tricky problem - one thing I noticed is that the ad skip button does not have an outline when navigating with keyboard.

@vseventer Do you have the tag? I see border around skip on my end

pajong:
> @pajong also there is no focus rings around the related items in the overlay

@karimMourra It already has a white border when it is focused. Do we want to have an extra browser border on top of that?

aliciahurst:
Agree with Karim's points:
* Outline should not include tooltips
* Outline missing in menu items
* Outline should be on overlays even when there is an existing border (for consistency I prefer it)

pajong:
@karimMourra Still working on removing the tooltip border, but the rest is in place. Let me know if you find anything else. Thanks!

PR #3194, opened 2018-12-02
[JW8-2343] Volume Button/Slider Accessibility Improvements

waxidiotic:
### This PR will...

Improve the accessibility of the volume button and slider by:
* Adding a tooltip for when the volume button is focused on via tabbing on the keyboard.
* Localizes the tooltip
* Tooltip changes based on `mute` state
* Changes `aria-orientation` for volume slider to `vertical`

There were deficiencies in our accessibility around the volume button and slider.

- [ ] The AC of the ticket states that the volume slider in controlbar-only mode should be read as a left/right, or horizontal slider. This is not the case as it was later clarified that adding a visible volume slider for controlbar-only mode will be handled in a later ticket.

n/a

#### Addresses Issue(s):

JW8-2343

### DIFF
--- a/src/assets/translations/en.js
+++ b/src/assets/translations/en.js
@@ -57,10 +57,10 @@ export default {
         "heading": "Share",
         "link": "Link"
     },
-    "slider": "Seek",
-    "stop": "Stop",
-    "unmute": "Unmute",
-    "videoInfo": "About This Video",
-    "volume": "Volume",
-    "volumeSlider": "Volume"
+    slider: 'Seek Slider',
+    stop: 'Stop',
+    unmute: 'Unmute',
+    videoInfo: 'About This Video',
+    volume: 'Volume',
+    volumeSlider: 'Volume Slider'
 };
--- a/src/assets/translations/en.js
+++ b/src/assets/translations/en.js
@@ -57,10 +57,10 @@ export default {
         "heading": "Share",
         "link": "Link"
     },
-    slider: 'Seek Slider',
-    stop: 'Stop',
-    unmute: 'Unmute',
-    videoInfo: 'About This Video',
-    volume: 'Volume',
-    volumeSlider: 'Volume Slider'
+    "slider": "Seek Slider",
+    "stop": "Stop",
+    "unmute": "Unmute",
+    "videoInfo": "About This Video",
+    "volume": "Volume",
+    "volumeSlider": "Volume Slider"
 };
--- a/src/assets/translations/en.js
+++ b/src/assets/translations/en.js
@@ -57,10 +57,10 @@ export default {
         "heading": "Share",
         "link": "Link"
     },
-    "slider": "Seek Slider",
+    "slider": "Seek",
     "stop": "Stop",
     "unmute": "Unmute",
     "videoInfo": "About This Video",
     "volume": "Volume",
-    "volumeSlider": "Volume Slider"
+    "volumeSlider": "Volume"
 };
--- a/src/css/controls/imports/controlbar.less
+++ b/src/css/controls/imports/controlbar.less
@@ -112,14 +112,6 @@
     display: none;
 }
 
-.jw-icon-volume {
-    &:not(.jw-open) {
-        .jw-overlay {
-            visibility: visible;
-        }
-    }
-}
-
 .jwplayer .jw-text-countdown {
     display: none;
 }
--- a/src/css/controls/imports/controlbar.less
+++ b/src/css/controls/imports/controlbar.less
@@ -112,6 +112,23 @@
     display: none;
 }
 
+.jw-icon-volume {
+    &:not(.jw-open) {
+        div {
+            height: 0;
+        }
+
+        > .jw-overlay {
+            visibility: visible;
+            min-height: 0;
+        }
+    }
+
+    > .jw-overlay:focus {
+        outline: 0;
+    }
+}
+
 .jwplayer .jw-text-countdown {
     display: none;
 }
--- a/src/css/controls/imports/controlbar.less
+++ b/src/css/controls/imports/controlbar.less
@@ -114,21 +114,12 @@
 
 .jw-icon-volume {
     &:not(.jw-open) {
-        div&:not(.jw-tooltip) {
-            height: 0;
-        }
-
-        .jw-tooltip {
-            bottom: (@mobile-touch-target / 2);
-        }
-
-        > .jw-overlay {
+        .jw-overlay {
             visibility: visible;
-            min-height: 0;
         }
     }
 
-    > .jw-overlay:focus {
+    .jw-overlay:focus {
         outline: 0;
     }
 }
--- a/src/css/controls/imports/controlbar.less
+++ b/src/css/controls/imports/controlbar.less
@@ -114,8 +114,12 @@
 
 .jw-icon-volume {
     &:not(.jw-open) {
-        div {
+        div&:not(.jw-tooltip) {
             height: 0;
+        }
+
+        .jw-tooltip {
+            bottom: (@mobile-touch-target / 2);
         }
 
         > .jw-overlay {
--- a/src/css/controls/imports/controlbar.less
+++ b/src/css/controls/imports/controlbar.less
@@ -118,10 +118,6 @@
             visibility: visible;
         }
     }
-
-    .jw-overlay:focus {
-        outline: 0;
-    }
 }
 
 .jwplayer .jw-text-countdown {
--- a/src/js/view/controls/components/volumetooltip.js
+++ b/src/js/view/controls/components/volumetooltip.js
@@ -26,16 +26,10 @@ export default class VolumeTooltip extends Tooltip {
             this.el.focus();
         }, this);
 
-        const closeTooltipHelper = () => {
-            this.closeTooltip();
-        };
-
-        const openTooltipHelper = () => {
-            this.openTooltip();
-        };
-
-        this.container.addEventListener('blur', closeTooltipHelper);
-        this.container.addEventListener('focus', openTooltipHelper);
+        this.closeTooltipHandler = this.closeTooltip.bind(this);
+        this.openTooltipHandler = this.openTooltip.bind(this);
+        this.container.addEventListener('blur', this.closeTooltipHandler);
+        this.container.addEventListener('focus', this.openTooltipHandler);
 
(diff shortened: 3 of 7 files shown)


waxidiotic:
test this please

src/js/view/controls/controlbar.js
```diff
@@ -221,6 +222,7 @@ export default class Controlbar {
         SimpleTooltip(elements.rewind.element(), 'rewind', localization.rewind);
         SimpleTooltip(elements.settingsButton.element(), 'settings', localization.settings);
         const fullscreenTip = SimpleTooltip(elements.fullscreen.element(), 'fullscreen', localization.fullscreen);
+        const muteTip = SimpleTooltip(elements.volumetooltip.element(), 'volumetooltip', localization.mute);
```
karimMourra: should this be called `mutetooltip` instead ? we already have a volume tooltip

src/js/view/controls/components/volumetooltip.js
```diff
@@ -7,25 +7,38 @@ export default class VolumeTooltip extends Tooltip {
...
 
-        this.addContent(this.volumeSlider.element());
+        this.addContent(volumeSliderElement);
 
         this.volumeSlider.on('update', function (evt) {
             this.trigger('update', evt);
             this.el.focus();
         }, this);
 
+        this.volumeSlider.el.parentElement.addEventListener('blur', () => {
```
karimMourra: I believe the parent element in this case is `this.container`.

src/assets/translations/en.js
```diff
@@ -57,10 +57,10 @@ export default {
         "heading": "Share",
         "link": "Link"
     },
-    "slider": "Seek",
```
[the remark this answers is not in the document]
waxidiotic: I don’t remember adding anything else to this file other than the mute/unmute translations, which I guess I wasn’t supposed to add anyway. I’ll revert the changes when I’m in the office tomorrow.

src/js/view/controls/controlbar.js
waxidiotic: I didn’t want to make any fundamental changes to the existing components. I’ll raise this tomorrow.

src/css/controls/imports/controlbar.less
```diff
@@ -112,6 +112,14 @@
     display: none;
 }
 
+.jw-icon-volume {
+    &:not(.jw-open) {
```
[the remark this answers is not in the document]
waxidiotic: @karimMourra Yes, because all other `.jw-overlay`'s are `visible: hidden` but we wanted to make this one visible in order to have it be focusable.

src/js/view/controls/controlbar.js
waxidiotic: @karimMourra `volumetooltip` is actually the slider that appears above the volume icon when you hover over it. The `muteTip` the simple text tooltip that is going to appear when focused on the volume icon. The component (`view/controls/components/volumetooltip.js`) is called `volumetooltip` so I didn't want to change that. Thoughts?

src/css/controls/imports/controlbar.less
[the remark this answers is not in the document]
karimMourra: right

src/js/view/controls/controlbar.js
```diff
@@ -270,7 +275,12 @@ export default class Controlbar {
 
         // Listen for model changes
         _model.change('volume', this.onVolume, this);
-        _model.change('mute', this.onMute, this);
+        _model.change('mute', (model, muted) => {
+            this.renderVolume(muted, model.get('volume'));
+            const muteText = model.get('mute') ? localization.unmute : localization.mute;
```
vseventer: Instead of reading `model.get('mute')` again, you can re-use the `muted` parameter.

src/js/view/controls/components/volumetooltip.js
```diff
@@ -7,25 +7,38 @@ export default class VolumeTooltip extends Tooltip {
...
 
-        this.addContent(this.volumeSlider.element());
+        this.addContent(volumeSliderElement);
 
         this.volumeSlider.on('update', function (evt) {
             this.trigger('update', evt);
             this.el.focus();
         }, this);
 
+        this.container.addEventListener('blur', () => {
```
karimMourra: these event listeners should be removed, at the very least when the volumeTooltip is destroyed.

src/js/view/controls/controlbar.js
```diff
@@ -222,6 +223,10 @@ export default class Controlbar {
         SimpleTooltip(elements.settingsButton.element(), 'settings', localization.settings);
         const fullscreenTip = SimpleTooltip(elements.fullscreen.element(), 'fullscreen', localization.fullscreen);
 
+        const volumeTooltipEl = elements.volumetooltip.element();
+        const muteTip = SimpleTooltip(volumeTooltipEl, 'volumetooltip', localization.mute);
```
karimMourra: remember to replace `'volumetooltip'` with mute tooltip

src/js/view/controls/components/volumetooltip.js
```diff
@@ -46,4 +46,9 @@ export default class VolumeTooltip extends Tooltip {
     toggleValue() {
         this.trigger('toggleValue');
     }
+
+    destroy() {
+        this.off();
```
vseventer: Not sure if you need this - where are the event listeners added? You should also remove the container listeners added in the constructor, plus call `this.ui.destroy()` if you want to clean-up everything nicely.

waxidiotic: @karimMourra requested i did this. Also, the volume tooltip is no longer constructed with `UI`

vseventer: What about the UI on line 37?

waxidiotic: Sorry, I mean the actual containing element needs the event listeners. The `UI` handles the actual slider but the `blur` and `focus` listeners are on the overlay container.

karimMourra: no no, I asked you to remove the event listener that you are adding on the container.

src/js/view/controls/components/volumetooltip.js
karimMourra: this has not been resolved. I think it's up to the person commenting to resolve a comment

src/js/view/controls/components/volumetooltip.js
karimMourra: You should be removing the listener mentioned in this comment https://github.com/jwplayer/jwplayer/pull/3194/files#r238739450

src/js/view/controls/components/volumetooltip.js
karimMourra: actually, no need to use an arrow function to call a function
```
this.container.addEventListener('blur', this.closeTooltip);
this.container.addEventListener('focus', this. openTooltip);
.
.
.
destroy() {
        this.container.removeEventListener('blur', this.closeTooltip);
        this.container.removeEventListener('focus', this.openTooltip);
    }
```

waxidiotic: or do you....

src/js/view/controls/components/volumetooltip.js
```diff
@@ -7,30 +7,52 @@ export default class VolumeTooltip extends Tooltip {
...
 
         this._model.on('change:volume', this.onVolume, this);
     }
 
     toggleValue() {
         this.trigger('toggleValue');
     }
+
+    destroy() {
+        this.container.removeEventListener('blur', () => {
```
johnBartos: This doesn't remove the listener - you need to reference the original function. See https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener

src/js/view/controls/controlbar.js
```diff
@@ -270,7 +275,12 @@ export default class Controlbar {
 
         // Listen for model changes
         _model.change('volume', this.onVolume, this);
-        _model.change('mute', this.onMute, this);
+        _model.change('mute', (model, muted) => {
+            this.renderVolume(muted, model.get('volume'));
+            const muteText = muted ? localization.unmute : localization.mute;
+            muteTip.setText(muteText);
+            setAttribute(this.elements.volumetooltip.element(), 'aria-label', muteText);
```
johnBartos: You can just use the `volumeTooltopEl` variable you created on L226

```suggestion
             setAttribute(volumeTooltipEl, 'aria-label', muteText); 
```

src/js/view/controls/controls.js
```diff
@@ -361,6 +361,7 @@ export default class Controls {
 
     disable(model) {
         const { nextUpToolTip, settingsMenu, infoOverlay } = this;
+        const volumeTooltip = this.controlbar.elements.volumeTooltip;
```
karimMourra: this is always undefined because in `controlbar.js`, volumetooltip is not camel cased. Let's make this `volumetooltip` for now.
 I suggest we call destroy on the controlbar instead of calling it directly on the volumetooltip. THe controls should not know about the internals of the controlbar, it should just interact with its interface. in other words:
```
const { nextUpToolTip, settingsMenu, infoOverlay,  controlbar } = this;
.
.
.
if (controlbar) {
         controlbar.destroy();
}
```
and then add a destroy function in controlbar that calls destroy on the all lower case volumetooltip

src/js/view/controls/components/volumetooltip.js
```diff
@@ -7,30 +7,51 @@ export default class VolumeTooltip extends Tooltip {
...
 
         this._model.on('change:volume', this.onVolume, this);
     }
 
     toggleValue() {
         this.trigger('toggleValue');
     }
+
+    destroy() {
+        this.container.removeEventListener('blur', this.closeTooltipHelper);
```
karimMourra: `this.closeTooltipHelper` would be undefined in this case, it's just a local var in the constructor. Same goes for `this.openTooltipHelper`

johnBartos: >You should instead define them as functions outside of the constructor's scope

Na, they should be a `this` object. Otherwise those handlers will be shared by all instances of the class.

src/js/view/controls/controls.js
johnBartos: But disabling doesn't destroy the controlbar?

PR #3197, opened 2018-12-04
Fix initial "buffer" event.reason

robwalch:
Initial "buffer" `event.reason` should equal "loading" not "idle". Setting the mediaModel's `mediaState` immediately updates the player model state and results in the correct reason being set in `ChangeStateEvent` https://github.com/jwplayer/jwplayer/blob/0ac0e36ae281cc50ec8d7e32c5832ead1e194832/src/js/events/change-state-event.js#L12-L24

Why would we need `mediaModel.mediaState` to remain "idle" in `_playAttempt()`? If there is no reason for it to remain unchanged, then these changes are fine.

#### Addresses Issue(s):
JW8-2483

### DIFF
--- a/src/js/program/media-controller.js
+++ b/src/js/program/media-controller.js
@@ -7,7 +7,7 @@ import { MediaModel } from 'controller/model';
 import { seconds } from 'utils/strings';
 import {
     MEDIA_PLAY_ATTEMPT, MEDIA_PLAY_ATTEMPT_FAILED, MEDIA_COMPLETE,
-    PLAYER_STATE, STATE_PAUSED, STATE_PLAYING, STATE_BUFFERING, STATE_COMPLETE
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/program/media-controller.js
```diff
@@ -121,7 +121,7 @@ export default class MediaController extends Eventable {
         });
         // Immediately set player state to buffering if these conditions are met
```
vseventer: Comment might be confusing since STATE_BUFFERING no longer appears - maybe clarify to "Propagate player state to buffering by setting mediaState to loading if these conditions are met" or sth.

PR #3199, opened 2018-12-05
[JW8-2488] enableAdsMode: set initial state to loading.

vseventer:
### This PR will...
Upon enabling ads mode, set the model state to `loading` rather than `playing`.

In #3169 `adPlay` events are only emitted if `oldstate !== newstate`. No `adPlay` event is emitted for DAI since the `oldstate` already is `playing`. Initializing the model state to `loading` instead will trigger an `adPlay`.

Note - currently, when resuming from ad pause, `oldstate` already yields `loading`. Therefore, it make sense to also use `loading` when starting an ad.

N/A

N/A

#### Addresses Issue(s):
JW8-2488

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -1,5 +1,5 @@
-import { STATE_BUFFERING, STATE_COMPLETE, STATE_LOADING, STATE_PLAYING,
-    STATE_PAUSED, PLAYER_STATE,
+import { STATE_BUFFERING, STATE_COMPLETE, STATE_PLAYING, STATE_PAUSED,
+    PLAYER_STATE,
     MEDIA_META, MEDIA_PLAY_ATTEMPT_FAILED, MEDIA_TIME, MEDIA_COMPLETE,
     PLAYLIST_ITEM, PLAYLIST_COMPLETE,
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/instream-adapter.js
```diff
@@ -137,7 +137,7 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
             }
         });
         _model.set('instream', _adProgram);
-        _adProgram.model.set('state', STATE_PLAYING);
+        _adProgram.model.set('state', STATE_LOADING);
```
robwalch: If anything `_adProgram.model` 'state' should be set to equal the player state, and then you can force an adPlay if that is the expectation when an ad is already playing.

vseventer:
@robwalch While the model state reflects the state of the stream, does it make sense for the adPlay event when an ad first starts playing to have `oldstate: "playing"`?

PR #3207, opened 2018-12-07
Scope all logic for volume tooltip to desktop-only

boyntoni:
### This PR will...

Scope all logic pertaining to the volume tooltip to desktop devices, to prevent exceptions being thrown in mobile environments. In mobile environments, the volume tooltip is not instantiated, however is later referenced. 

Also ensures the volumeTooltip is referenced prior to to the controls being disabled within the disable() method, required as the previous implementation was causing the following error: `null is not an object (evaluating 'this.controlbar.elements')`. 

In mobile environments, player was throwing a 100000 error, as volumeTooltip was not defined in this environment. 

N/A

No

#### Addresses Issue(s):

JW8-2343

### DIFF
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -601,6 +601,7 @@ export default class Controlbar {
                 this.elements[elementName].destroy();
             }
         });
+        this._model.off(null, null, this);
     }
 }
 
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -11,7 +11,6 @@ import Events from 'utils/backbone.events';
 import { prependChild, setAttribute, toggleClass } from 'utils/dom';
 import { timeFormat } from 'utils/parser';
 import UI from 'utils/ui';
-import { each } from 'utils/underscore';
 import { genId, FEED_SHOWN_ID_LENGTH } from 'utils/random-id-generator';
 
 function text(name, role) {
@@ -119,8 +118,8 @@ export default class Controlbar {
         this._volumeAnnouncer = _accessibilityContainer.querySelector('.jw-volume-update');
         const localization = _model.get('localization');
         const timeSlider = new TimeSlider(_model, _api, _accessibilityContainer.querySelector('.jw-time-update'));
+        const menus = [];
         let volumeTooltip;
-        let volumeTooltipEl;
         let muteTip;
         let muteButton;
         let feedShownId = '';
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/controls.js
```diff
@@ -362,6 +362,13 @@ export default class Controls {
 
     disable(model) {
         const { nextUpToolTip, settingsMenu, infoOverlay } = this;
+        const volumeTooltip = this.controlbar.elements.volumetooltip;
+
+        
+        if (volumeTooltip) {
+            volumeTooltip.destroy();
+        }
```
robwalch: Can we create a `destroy` method for `controlbar` which destroys it's elements? This should go above the `rightClickMenu.destroy` block.

src/js/view/controls/controlbar.js
```diff
@@ -120,6 +120,8 @@ export default class Controlbar {
         const localization = _model.get('localization');
         const timeSlider = new TimeSlider(_model, _api, _accessibilityContainer.querySelector('.jw-time-update'));
         let volumeTooltip;
+        let volumeTooltipEl;
+        let muteTip;
```
robwalch: If `volumeTooltipEl` are only used in the `if (!this._isMobile)` block below they should be defined as constants in that block.

src/js/view/controls/controlbar.js
```diff
@@ -139,11 +141,18 @@ export default class Controlbar {
...
             setAttribute(volumeTooltipEl, 'aria-valuemax', 100);
             setAttribute(volumeTooltipEl, 'aria-orientation', 'vertical');
             setAttribute(volumeTooltipEl, 'role', 'slider');
+            muteTip = SimpleTooltip(volumeTooltipEl, 'mutetooltip', localization.mute);
+            volumeTooltipEl.removeEventListener('mouseover', muteTip.open);
+            _model.change('mute', (model, muted) => {
+                const muteText = muted ? localization.unmute : localization.mute;
+                muteTip.setText(muteText);
+                setAttribute(volumeTooltipEl, 'aria-label', muteText);
+            }, this);
```
robwalch: Could we just push `volumeTooltip` into `menus` here instead of that silly filter method below?

src/js/view/controls/controlbar.js
```diff
@@ -597,5 +594,13 @@ export default class Controlbar {
 
         toggleClass(captionsButton.element(), 'jw-off', !active);
     }
+
+    destroy() {
+        Object.keys(this.elements).forEach((elementName) => {
```
johnBartos: Also call `_model.off(null, null, this)` to detach all controlbar listeners

PR #3208, opened 2018-12-07
Resolved Issue with Volume Slider

waxidiotic:
JW8-2343

### This PR will...

* Resolve issue with dragging the knob in the volume slider.

* Dragging the knob on the volume slider would move focus back to the volume icon, thus making it impossible to continue dragging the volume slider.

n/a

n/a

#### Addresses Issue(s):

JW8-2343

### DIFF
--- a/src/js/view/controls/components/volumetooltip.js
+++ b/src/js/view/controls/components/volumetooltip.js
@@ -23,7 +23,6 @@ export default class VolumeTooltip extends Tooltip {
 
         this.volumeSlider.on('update', function (evt) {
             this.trigger('update', evt);
-            this.el.focus();
         }, this);
 
         this.closeTooltipHandler = this.closeTooltip.bind(this);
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/components/volumetooltip.js
```diff
@@ -23,7 +23,6 @@ export default class VolumeTooltip extends Tooltip {
 
         this.volumeSlider.on('update', function (evt) {
             this.trigger('update', evt);
-            this.el.focus();
```
johnBartos: Why can this be safely removed?

waxidiotic: Before, when the slider was interacted with, it would push focus back onto the volume icon. Before the changes for 8.7, the slider would still be visible. Now, it's not so we don't want to push focus back on to the volume icon - we want to keep it on the slider.

PR #3209, opened 2018-12-07
Prevent duplicate volume and seek slider updates

karimMourra:
### This PR will...
Update the hidden announcements only when the slider is not in focus. 
update the time slider's `aria-valuenow` to work around a JAWS bug where the time is announced twice if the `aria-valuetext` is updated without an update to `aria-valuenow`.

We introduced the `jw-hidden-annoucements` element to announce volume and seek changes executed with the arrow keys when the sliders are not in focus. This resulted in the JAWS screen reader making duplicate announcements when the sliders are in focus.
I chose only trigger the hidden announcements when the slider is not in focus  instead of removing the `aria-valuenow` and `aria-valuetext` from the sliders, because per [MDN](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/ARIA_Techniques/Using_the_aria-valuenow_attribute) , `"aria-valuenow is a required attribute of role slider, scrollbar and spinbutton."`, therefore removing the aria values would negatively affect our score in an accessibility audit.
i noticed we never use document.activeElement to check if the element is in focus; any known issues with this api ?
#### Addresses Issue(s):

JW8-2507

### DIFF
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -110,6 +110,7 @@ class TimeSlider extends Slider {
         setAttribute(this.el, 'tabindex', '0');
         setAttribute(this.el, 'role', 'slider');
         setAttribute(this.el, 'aria-label', this._model.get('localization').slider);
+        setAttribute(this.el, 'aria-live', 'off');
         this.el.removeAttribute('aria-hidden');
         this.elementRail.appendChild(this.timeTip.element());
 
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -106,21 +106,20 @@ class TimeSlider extends Slider {
             .change('buffer', this.onBuffer, this)
             .change('streamType', this.onStreamType, this);
 
-
-        setAttribute(this.el, 'tabindex', '0');
-        setAttribute(this.el, 'role', 'slider');
-        setAttribute(this.el, 'aria-label', this._model.get('localization').slider);
-        setAttribute(this.el, 'aria-live', 'off');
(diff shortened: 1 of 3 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

karimMourra:
test this please

pajong:
@karimMourra I think that when `aria-live` is set to `null`, the behavior is the same as setting to `off`. The bug doesn't happen in Mac/Chrome in the first place, so please check in IE11 using JAWS

src/js/view/controls/components/timeslider.js
```diff
@@ -106,20 +106,20 @@ class TimeSlider extends Slider {
...
+        sliderElement.removeAttribute('aria-hidden');
         this.elementRail.appendChild(this.timeTip.element());
 
         // Show the tooltip on while dragging (touch) moving(mouse), or moving over(mouse)
-        this.ui = (this.ui || new UI(this.el))
+        this.ui = (this.ui || new UI(sliderElement))
             .on('move', this.showTimeTooltip, this)
             .on('dragEnd out', this.hideTimeTooltip, this)
-            .on('click', () => this.el.focus());
+            .on('click', () => sliderElement.focus());
```
robwalch: Use ui to handle focus:
`on('focus', () => this.updateAriaText());`

(and then remove the `removeEventListener` for it)

PR #3210, opened 2018-12-09
Do not show focus decoration on mouse interaction

waxidiotic:
JW8-2343

### This PR will...

* Make sure the focus decoration (blue outline) does not appear unless the volume slider is focused via a keyboard interaction.
* Make sure the focus decoration does not appear when the CSS opacity transition for the volume slider takes place.

* After https://github.com/jwplayer/jwplayer/pull/3208, the focus decoration for the volume slider will still appear if the volume slider is hovered on after it has been focused on.
* The opacity transition for the volume slider was causing the focus decoration to briefly appear when mousing over or out of the slider.

n/a

n/a

#### Addresses Issue(s):

JW8-2343

### DIFF
--- a/src/css/controls/imports/tooltip.less
+++ b/src/css/controls/imports/tooltip.less
@@ -4,10 +4,6 @@
     opacity: 1;
     transition-delay: 0s;
     visibility: visible;
-
-    &:hover {
-        outline: 0;
-    }
 }
 
 .jw-slider-time .jw-overlay:before {
@@ -76,6 +72,12 @@
 
     .jw-icon-volume .jw-overlay {
         visibility: visible;
+        outline: none;
+        transition: none;
+
+        &:focus:not(:hover) {
+            outline: @ui-focus;
+        }
     }
 
     .jw-option {


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/css/controls/imports/tooltip.less
```diff
@@ -4,10 +4,6 @@
     opacity: 1;
     transition-delay: 0s;
     visibility: visible;
-
-    &:hover {
-        outline: 0;
```
karimMourra: why are we removing this ? tooltips should not have an outiline when opened

waxidiotic: This is for the volume slider. The text (simple) tooltips don't have the `.jw-overlay` class. This line was actually added by me in a previous PR.

src/css/controls/imports/tooltip.less
```diff
@@ -76,6 +72,12 @@
 
     .jw-icon-volume .jw-overlay {
         visibility: visible;
+        outline: none;
+        transition: none;
```
karimMourra: @waxidiotic why the `transition: none;` ? Do we see a flash without it ?

waxidiotic: yes

PR #3211, opened 2018-12-09
Only close menus if there is no event

waxidiotic:
JW8-2343

### This PR will...

Only close menus when in the controlbar if there is no event

Before, the `!evt` check was inside of the `this.menus.forEach` loop which would cause an exception when tabbing through the controlbar if there are no existing menus.

Tested in Chrome 71, iPad/iPhone on iOS 12.1.1, IE 11 on Windows 7 & 10, Edge on Windows 10

#### Addresses Issue(s):

JW8-###

### DIFF
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -441,7 +441,7 @@ export default class Controlbar {
 
     // Close menus if it has no event.
     closeMenus(evt) {
-        if (!this.menus) {
+        if (!evt || !this.menus) {
             return;
         }
 
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -439,9 +439,9 @@ export default class Controlbar {
         this.elements.alt.textContent = altText;
     }
 
-    // Close menus if it has no event.  Otherwise close all but the event's target.
+    // Close menus if it has no event.
     closeMenus(evt) {
-        if (evt) {
+        if (!this.menus) {
             return;
         }
 
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -441,7 +441,7 @@ export default class Controlbar {
 
     // Close menus if it has no event.
     closeMenus(evt) {
-        if (!evt || !this.menus) {
+        if (evt || !this.menus) {
             return;
         }
 
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -441,15 +441,15 @@ export default class Controlbar {
 
     // Close menus if it has no event.  Otherwise close all but the event's target.
     closeMenus(evt) {
-        if (!evt) {
-            this.menus.forEach(ele => {
-                if (evt.target !== ele.el) {
-                    ele.closeTooltip(evt);
-                } else {
-                    return;
-                }
-            });
+        if (evt) {
+            return;
         }
+
+        this.menus.forEach(ele => {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/controlbar.js
```diff
@@ -441,11 +441,15 @@ export default class Controlbar {
 
     // Close menus if it has no event.  Otherwise close all but the event's target.
     closeMenus(evt) {
-        this.menus.forEach(ele => {
-            if (!evt || evt.target !== ele.el) {
-                ele.closeTooltip(evt);
-            }
-        });
+        if (!evt) {
```
karimMourra: let's reduce the nesting with an early return statement. I find early returns easier to read and nest if statements tend to get tangly. Also you don't need that `else` statement.
```
if (evt) {
        return;
}
this.menus.forEach(ele => {
            if (evt.target !== ele.el) {
                    ele.closeTooltip(evt);
            }
});
```

src/js/view/controls/controlbar.js
```diff
@@ -441,8 +441,12 @@ export default class Controlbar {
...
     // Close menus if it has no event.  Otherwise close all but the event's target.
     closeMenus(evt) {
+        if (evt) {
+            return;
+        }
+
         this.menus.forEach(ele => {
-            if (!evt || evt.target !== ele.el) {
+            if (evt.target !== ele.el) {
                 ele.closeTooltip(evt);
```
karimMourra: @waxidiotic is there ever a case where ele is undefined ? Do we ever add undefined or null elements to the menu array ?

src/js/view/controls/controlbar.js
```diff
@@ -439,10 +439,14 @@ export default class Controlbar {
...
-    // Close menus if it has no event.  Otherwise close all but the event's target.
+    // Close menus if it has no event.
     closeMenus(evt) {
+        if (evt || !this.menus) {
+            return;
+        }
+
         this.menus.forEach(ele => {
-            if (!evt || evt.target !== ele.el) {
+            if (evt.target !== ele.el) {
```
johnBartos:
```suggestion
             if (!evt || evt.target !== ele.el) { 
```

src/js/view/controls/controlbar.js
```diff
@@ -439,8 +439,12 @@ export default class Controlbar {
         this.elements.alt.textContent = altText;
     }
 
-    // Close menus if it has no event.  Otherwise close all but the event's target.
+    // Close menus if it has no event.
     closeMenus(evt) {
+        if (!this.menus) {
+            return;
```
robwalch: No event or no menus?

robwalch: @waxidiotic @boyntoni,

@pajong noticed that `this.menus` is never being set because or recent changes. It always should be set so adding this guard should not have been added because it doesn't address the regression and instead adds code which causes us all to ignore it.

waxidiotic: @robwalch I'll remove it in another PR.

waxidiotic: Actually, now that I think of it - maybe @pajong should remove it in #3215.

robwalch: It is done.

PR #3212, opened 2018-12-10
Add focus to button on mousedown for firefox

pajong:
### This PR will...
- Add focus to settings button on mousedown 
- This is a workaround for Firefox not having correct focus
- Other browsers work fine, and are not affected by this change

- On Firefox, there is an issue with the settings buttons not being focused on click.
- Since we removed the calls to `preventDefault` on button clicks, we should get focus on buttons.
- Having no focus results the buttons to not remove the `jw-no-focus` class added on mousedown, resulting no focus outline when they are back in focus with keyboard for the first time after clicking.

#### Addresses Issue(s):

JW8-2501

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -38,7 +38,7 @@ export default class UI extends Eventable {
         this.pointerId = null;
         this.startX = 0;
         this.startY = 0;
-        this.noFocusHelper = flagNoFocus(element);
+        this.noFocusHelper = flagNoFocus(element, options.focusOnMouseDown);
     }
 
     on(name, callback, context) {
--- a/src/js/view/controls/components/settings/content-item.js
+++ b/src/js/view/controls/components/settings/content-item.js
@@ -5,7 +5,7 @@ import UI from 'utils/ui';
 export default function SettingsContentItem(name, content, action) {
     let active;
     const contentItemElement = createElement(ContentItemTemplate(content));
-    const contentItemUI = new UI(contentItemElement).on('click tap enter', (evt) => {
+    const contentItemUI = new UI(contentItemElement, { focusOnMouseDown: true }).on('click tap enter', (evt) => {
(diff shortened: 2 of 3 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/utils/ui.js
```diff
@@ -38,7 +38,7 @@ export default class UI extends Eventable {
         this.pointerId = null;
         this.startX = 0;
         this.startY = 0;
-        this.noFocusHelper = flagNoFocus(element);
+        this.noFocusHelper = flagNoFocus(element, options.focusOnMouseDown);
```
robwalch: @pajong Why is this it's own module and why is it being added in the constructor? `ui` was refactored recently with the design principal that it should only add listeners when needed.

This is not something that should be added in the constructor. The focus logic should only apply to ui instances handling mouse (and maybe touch) input in `eventRegisters`.

src/js/view/controls/components/settings/content-item.js
```diff
@@ -5,7 +5,7 @@ import UI from 'utils/ui';
 export default function SettingsContentItem(name, content, action) {
     let active;
     const contentItemElement = createElement(ContentItemTemplate(content));
-    const contentItemUI = new UI(contentItemElement).on('click tap enter', (evt) => {
+    const contentItemUI = new UI(contentItemElement, { focusOnMouseDown: true }).on('click tap enter', (evt) => {
```
robwalch: Why is this the only element that does not get focus on FireFox?

PR #3214, opened 2018-12-10
Handle no-focus flag in ui.js

robwalch:
### This PR will...
- Handle no-focus flag in ui.js
- Remove "flag-no-focus" module
- Use ui.js to handle volume slider focus
- Cleanup more ui.js instances

These changes remove duplicate code used to prevent display of the focus ring when mouse/pointer/touch input is used on UI elements in the player.

The number of listeners added to each element is reduced by one. Listeners are removed when ui is destroyed or `off` is called (still missing in some ui modules).

#### Addresses Issue(s):
JW8-2501
JW8-2343

For functionality described in:
JW8-2477
JW8-2476


src/js/view/controls/components/volumetooltip.js
```diff
@@ -25,10 +25,11 @@ export default class VolumeTooltip extends Tooltip {
             this.trigger('update', evt);
         }, this);
 
-        this.closeTooltipHandler = this.closeTooltip.bind(this);
-        this.openTooltipHandler = this.openTooltip.bind(this);
-        this.container.addEventListener('blur', this.closeTooltipHandler);
-        this.container.addEventListener('focus', this.openTooltipHandler);
+
+        this.uiOver = new UI(this.container)
+            .on('click', function() {}, this)
```
karimMourra: why are we adding the click listener ?

waxidiotic: I believe it is because we don't want the tooltip to toggle state when clicked on. `this.uiOver` is on the `.jw-overlay`, but it's a child of the volume tooltip that `this.ui` is created on so adding an empty action when `this.uiOver` is clicked resets it because otherwise, the `this.ui` click handler would toggle the tooltip's open state.

robwalch: Add a comment to explain why the listener is added. The focus helper logic is in ui.js `interactStartHandler`.

PR #3219, opened 2018-12-11
Make volume focus and hover behavior consistent

robwalch:
### This PR will...
Make volume focus and hover behavior consistent. Focus and hover show the slider and no mute tip.

In order to tab into the volume slider it needs to be visible. The simplest way to do this is to open the slider when the volume/mute button has focus - same as when it's hovered over.

The alternative would be to change the volume-icon, overlay and open classes to never set the visibility of the slider to none and instead hide it with transforms or maybe clipping. That could however interfere with animations and other elements sharing some of these classes. This seems like the simplest fix.

#### Addresses Issue(s):
JW8-2343

### DIFF
--- a/src/css/controls/imports/tooltip.less
+++ b/src/css/controls/imports/tooltip.less
@@ -70,15 +70,6 @@
         }
     }
 
-    .jw-icon-volume .jw-overlay {
-        visibility: visible;
-        transition: none;
-
-        &:hover {
-            outline: none;
-        }
-    }
-
     .jw-option {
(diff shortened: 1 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3220, opened 2018-12-12
Add focus ring to volume slider in Firefox

waxidiotic:
JW8-2513

### This PR will...

Make sure the focus ring for the volume slider is the right decoration in Firefox.

In Firefox, the focus ring would just be the default browser style of a dotted outline. All other browsers were fine.

n/a

n/a

#### Addresses Issue(s):

JW8-2513

### DIFF
--- a/src/css/controls/imports/tooltip.less
+++ b/src/css/controls/imports/tooltip.less
@@ -4,6 +4,10 @@
     opacity: 1;
     transition-delay: 0s;
     visibility: visible;
+
+    &:focus:not(.jw-not-focus) {
+        outline: @ui-focus;
+    }
 }
 
 .jw-slider-time .jw-overlay:before {
--- a/src/css/controls/imports/tooltip.less
+++ b/src/css/controls/imports/tooltip.less
@@ -5,7 +5,7 @@
     transition-delay: 0s;
     visibility: visible;
 
-    &:focus:not(.jw-not-focus) {
+    &:focus {
         outline: @ui-focus;
     }
 }


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/css/controls/imports/tooltip.less
```diff
@@ -4,6 +4,10 @@
     opacity: 1;
     transition-delay: 0s;
     visibility: visible;
+
+    &:focus:not(.jw-not-focus) {
+        outline: @ui-focus;
+    }
```
robwalch: 1. Search for all instances of `outline: @ui-focus`
2. Why is there an additional ampersand in skipad (or not one here):
```less
    &:focus&:not(.jw-no-focus) {
        outline: @ui-focus;
    }
```

The combination of `:not(.jw-no-focus)` and `:focus` is inconsistent. Can you make sure these are correct and necessary?

waxidiotic: @robwalch - I just doubled-checked with John K. and it looks like the way I have it is the best way. Functionally, there is no difference between `&:focus&:not(.jw-no-focus)` and `&:focus:not(.jw-no-focus)` except for the fact that if the parent selector is complex, it adds a lot of additional code and complexity to the transpiled result.

But, in the end, I'm going to try to remove the `not:(.jw-not-focus)` to see if this still works without it because the `.jw-overlay` element should never have it.

PR #3223, opened 2018-12-13
Fix display and skip icon focus issues

robwalch:
### This PR will...
- Replace ".jw-no-focus" with ".jw-tab-focus" and only show focus outline when focus changes following a keyboard event
- Revert changes to listen for interactions on display-icon-container instead of display-icon
- Always use display icon for focus and interaction
- Remove `.jw-display-icon-container: pointer-events: all` styling which causes the display icon container to receive focus instead of only the display icons
- Remove ads `pointer-events: all` styling which causes icons to receive focus
- Remove ads `jw-nextup-container` styles (we never show nextup over ads)
- Add "jw-icon" class to mock icon element in play-display-icon-test and make `displayIcon` a local const
- Use `border` to style jw-icon elements. This fixes both Firefox outline issues surrounding child elements with absolute positioning, and Safari which sometimes doesn't paint outlines applied to svg elements

These changes prevent elements from getting focus that should not on IE and mobile. UI listeners should be attached to the elements assigned a tab index, not their parents.

#### Addresses Issue(s):
JW8-2519

### DIFF
--- a/src/css/controls/flags/ads.less
+++ b/src/css/controls/flags/ads.less
@@ -87,15 +87,6 @@
             }
         }
     }
-
-    .jw-nextup-container {
-        bottom: @mobile-touch-target;
-    }
-
-    .jw-icon,
-    .jw-slider-horizontal {
-        pointer-events: all;
-    }
 }
 
 .jwplayer.jw-flag-ads-vpaid,
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -18,6 +18,7 @@ const DOUBLE_CLICK_DELAY = 300;
 const LONG_PRESS_DELAY = 500;
 
 let longPressTimeout;
+let lastInteractionListener;
 
 export default class UI extends Eventable {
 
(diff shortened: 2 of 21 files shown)


robwalch:
@waxidiotic these changes are specific to the display and skip buttons.

waxidiotic:
Makes sense. I was just looking at the ticket.

robwalch:
@waxidiotic Addressed the issue of focus ring being applied to element that receive focus after the previously active element is removed from the DOM or "jw-no-focus" was not applied.

src/css/controls/flags/ads.less
```diff
@@ -87,15 +87,6 @@
             }
         }
     }
-
-    .jw-nextup-container {
```
karimMourra: why would next up show during ads ? Is this for banner ads ?

src/js/utils/ui.js
```diff
@@ -318,11 +325,25 @@ const eventRegisters = {
             }
         });
     },
+    keydown(ui) {
+        const keydown = 'keydown';
+        addEventListener(ui, keydown, 'keydown', (e) => {
```
karimMourra: does `'keydown'` have to be a string ?
`addEventListener(ui, keydown, keydown, (e) => {`

src/css/controls/flags/ads.less
robwalch: If it was `mobile-touch-target` isn't tall enough

PR #3224, opened 2018-12-13
Insert cues in time order in IE/Edge 

johnBartos:
Because IE/Edge will throw an exception if cues are not inserted in order

Did I mess up my discrete math?
```
!(Browser.ie && renderNatively) || !window.TextTrackCue

is equal to:

!Browser.ie || !renderNatively || !window.TextTrackCue

and when negated becomes:

Browser.ie && renderNatively && window.TextTrackCue
```

No

#### Addresses Issue(s):

JW8-2521

### DIFF
--- a/src/js/providers/tracks-mixin.js
+++ b/src/js/providers/tracks-mixin.js
@@ -517,15 +517,35 @@ function addVTTCuesToTrack(track, vttCues) {
 // ////////////////////
 
 function _addCueToTrack(renderNatively, track, vttCue) {
-    if (!(Browser.ie && renderNatively) || !window.TextTrackCue) {
-        track.addCue(vttCue);
-        return;
+    let cue = vttCue;
+    if (Browser.ie && renderNatively && window.TextTrackCue) {
+        // There's no support for the VTTCue interface in IE/Edge.
+        // We need to convert VTTCue to TextTrackCue before adding them to the TextTrack
+        // This unfortunately removes positioning properties from the cues
+        cue = new window.TextTrackCue(vttCue.startTime, vttCue.endTime, vttCue.text);
     }
-    // There's no support for the VTTCue interface in IE/Edge.
-    // We need to convert VTTCue to TextTrackCue before adding them to the TextTrack
-    // This unfortunately removes positioning properties from the cues
-    const textTrackCue = new window.TextTrackCue(vttCue.startTime, vttCue.endTime, vttCue.text);
-    track.addCue(textTrackCue);
+
+    // IE/Edge will throw an exception if cues are not inserted in time order: https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/13183203/
+    if (Browser.ie) {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

vseventer:
@johnBartos Wouldn't the third step become `!(x && y && z)`?

src/js/providers/tracks-mixin.js
```diff
@@ -517,15 +517,35 @@ function addVTTCuesToTrack(track, vttCues) {
 // ////////////////////
 
 function _addCueToTrack(renderNatively, track, vttCue) {
-    if (!(Browser.ie && renderNatively) || !window.TextTrackCue) {
-        track.addCue(vttCue);
-        return;
+    let cue = vttCue;
+    if (Browser.ie && renderNatively && window.TextTrackCue) {
```
egreaves: We returned early in the non-ie use case because that code path will get hit in majority of cases (chrome, safari, ff).

The `!window.TextTrackCue` check was there to prevent this code from blowing up in environments where `window.TextTrackCue` doesn't exist and we use the vttCue polyfill.

Every browser - ie, edge, ff, safari, chrome - supports `window.TextTrackCue`, so if we were to keep this condition, you could simplify it to `if (Browser.ie && renderNatively)`.

johnBartos:
@vseventer Sorry, I should have specified that I was applying a negation to bring ` cue = new window.TextTrackCue(vttCue.startTime, vttCue.endTime, vttCue.text); ` into the conditional

src/js/providers/tracks-mixin.js
```diff
@@ -517,15 +517,35 @@ function addVTTCuesToTrack(track, vttCues) {
...
+    if (Browser.ie) {
+        insertCueInOrder(track, cue);
+    } else {
+        track.addCue(cue);
+    }
+}
+
+function insertCueInOrder(track, vttCue) {
+    const temp = [];
+    const cues = track.cues;
```
egreaves: `track.cues` will be `null` if `track.mode === 'disabled'`. If you need to remove cues, to be safe, you'll need to ensure the mode is either 'hidden' or 'showing' before iterating through the cues. Take a look at [this example](https://github.com/jwplayer/jwplayer/blob/7054d14f331ca7168e5d807c5e6aa7ca14cd5cde/src/js/providers/tracks-mixin.js#L107). Oddly enough, `track.addCue` works even if the track is disabled, which is why there wasn't a need to check the mode before.

src/js/providers/tracks-mixin.js
egreaves:
```suggestion
if (Browser.ie && renderNatively) {
```

PR #3226, opened 2018-12-14
Fix removeEventListener call in ui.js

robwalch:
### This PR will...
Correct how we access `options` passed to `addEventListener(type, listener, options)` for removal and only pass the `useCapture` argument to `removeEventListener()` if it is a boolean.

For UI instance event listeners to be removed:

>[**Matching event listeners for removal**](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#Matching_event_listeners_for_removal)
> Given an event listener previously added by calling `addEventListener()`, you may eventually come to a point at which you need to remove it. Obviously, you need to specify the same `type` and `listener` parameters to `removeEventListener()`, but what about the `options` or `useCapture` parameters?
> 
> While `addEventListener()` will let you add the same listener more than once for the same type if the options are different, the only option `removeEventListener()` checks is the `capture`/`useCapture` flag. Its value must match for `removeEventListener()` to match, but the other values don't.

#### Addresses Issue(s):
JW8-2527

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -378,7 +378,12 @@ function removeHandlers(ui, triggerName) {
     const listenerOptions = options[triggerName];
     if (listeners) {
         Object.keys(listeners).forEach(domEventName => {
-            element.removeEventListener(domEventName, listeners[domEventName], listenerOptions);
+            const useCapture = listenerOptions[domEventName];
+            if (typeof useCapture === 'boolean') {
+                element.removeEventListener(domEventName, listeners[domEventName], useCapture);
+            } else {
+                element.removeEventListener(domEventName, listeners[domEventName]);
+            }
         });
         handlers[triggerName] = null;
         options[triggerName] = null;


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/utils/ui.js
```diff
@@ -378,7 +378,12 @@ function removeHandlers(ui, triggerName) {
     const listenerOptions = options[triggerName];
     if (listeners) {
         Object.keys(listeners).forEach(domEventName => {
-            element.removeEventListener(domEventName, listeners[domEventName], listenerOptions);
+            const useCapture = listenerOptions[domEventName];
```
pajong: I am not sure if we need to check for boolean value.
`the only option removeEventListener() checks is the capture/useCapture flag. Its value must match for removeEventListener() to match, but the other values don't.`
Doesn't this mean that we can just pass in the same options as we did for `addEventListener`, and the options other than capture(boolean)/useCapture(boolean in an object) are ignored?

src/js/utils/ui.js
```diff
@@ -378,7 +378,12 @@ function removeHandlers(ui, triggerName) {
...
+            if (typeof useCapture === 'boolean') {
+                element.removeEventListener(domEventName, listeners[domEventName], useCapture);
+            } else {
+                element.removeEventListener(domEventName, listeners[domEventName]);
```
karimMourra: There's a note on MDN 

> It's worth noting that some browser releases have been inconsistent on this, and unless you have specific reasons otherwise, it's probably wise to use the same values used for the call to addEventListener() when calling removeEventListener().

should we instead always specify the options ?

src/js/utils/ui.js
```diff
@@ -378,7 +378,12 @@ function removeHandlers(ui, triggerName) {
     const listenerOptions = options[triggerName];
     if (listeners) {
         Object.keys(listeners).forEach(domEventName => {
-            element.removeEventListener(domEventName, listeners[domEventName], listenerOptions);
+            const useCapture = listenerOptions[domEventName];
+            if (typeof useCapture === 'boolean') {
```
karimMourra: according to [MDN](https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/removeEventListener#Matching_event_listeners_for_removal) `options = { capture: true }` would also fail to be removed

src/js/utils/ui.js
robwalch: The bug was caused by us passing in an object. IE11 will treat an object as `true` and then not remove the listener. That is the bug that this change fixes.

src/js/utils/ui.js
robwalch: See my comment below https://github.com/jwplayer/jwplayer/pull/3226#discussion_r241848296

Passing in any object will cause the bug in IE 11 that these changes fix.

PR #3227, opened 2018-12-14
Fix issue where fullscreen tooltip won't go away

robwalch:
### This PR will...
Remove redundant controlbar and settings menu tooltip rule that force display of tooltips for all focus state.

There are already rules for showing tooltips on hover and tab focus. Clicking the fullscreen button results in the button having focus. That should not result in the tooltip getting staying open while in fullscreen or after exiting fullscreen.

#### Addresses Issue(s):
JW8-###

### DIFF
--- a/src/css/controls/imports/tooltip.less
+++ b/src/css/controls/imports/tooltip.less
@@ -133,10 +133,6 @@
             }
         }
     }
-
-    .jw-icon:focus > .jw-tooltip {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3230, opened 2018-12-26
show time tooltip on drag 

karimMourra:
### This PR will...
include drag in the list of events required to call `showTimeTooltip`.
On mobile devices we must show the thumbnail when the seek knob is dragged.
n/a
n/a
#### Addresses Issue(s):

JW8-2370

### DIFF
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -115,7 +115,7 @@ class TimeSlider extends Slider {
 
         // Show the tooltip on while dragging (touch) moving(mouse), or moving over(mouse)
         this.ui = (this.ui || new UI(sliderElement))
-            .on('move', this.showTimeTooltip, this)
+            .on('move drag', this.showTimeTooltip, this)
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

karimMourra:
@waxidiotic good catch! It doesn't happen in safari or chrome

karimMourra:
Actually @waxidiotic that error has nothing to do with my changes. It happens when scrubbing any video in Firefox https://bugzilla.mozilla.org/show_bug.cgi?id=1507193

waxidiotic:
@karimMourra Cool, I would still maybe see if `touchmove` is a better option because I think it was meant for this use case but if not, then we should be good.

PR #3233, opened 2018-12-28
Add XHR file protocol error code

karimMourra:
### This PR will...
check if XHR errors are caused by requesting a relative or protocol relative path on a page hosted over the `file:` protocol.
Customers are seeing the unknown XHR error (`006`) in webviews when the page is hosted in the native app. When a playlist feed without a protocol is loaded in the player, XHR assumes the protocol should be `file:`, resulting in an error when the asset is not hosted in the app.
Do you agree with the error code ?
[Docs PR](https://github.com/jwplayer/jwplayer-docs-new/pull/280)
#### Addresses Issue(s):
JW8-2474

### DIFF
--- a/src/js/utils/ajax.js
+++ b/src/js/utils/ajax.js
@@ -169,9 +169,8 @@ function _readyStateChangeHandler(options) {
                 return _ajaxComplete(options)(e);
             }
             // regex checks that the url is relative or protocol relative
-            if (status === 0 && isFileProtocol() && !/^(?:(?:https?|file):\/\/)/.test(options.url)) {
+            if (status === 0 && isFileProtocol() && !/^[a-z][a-z0-9+.-]*:/.test(options.url)) {
                 _error(options, MSG_CANT_PLAY_VIDEO, ERROR_XHR_FILE_PROTOCOL);
-
             }
         }
     };
--- a/src/js/utils/ajax.js
+++ b/src/js/utils/ajax.js
@@ -168,7 +168,7 @@ function _readyStateChangeHandler(options) {
             if (status === 200) {
                 return _ajaxComplete(options)(e);
             }
-            // verify that the error is due to a relative or protocol relative url being requested on page hosted over file: protocol
+            // regex checks that the url is relative or protocol relative
             if (status === 0 && isFileProtocol() && !/^(?:(?:https?|file):\/\/)/.test(options.url)) {
                 _error(options, MSG_CANT_PLAY_VIDEO, ERROR_XHR_FILE_PROTOCOL);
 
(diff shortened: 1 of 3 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/utils/ajax.js
```diff
@@ -158,13 +160,19 @@ function _readyStateChangeHandler(options) {
...
                 return;
             }
-            if (xhr.status === 200) {
+            if (status === 200) {
                 return _ajaxComplete(options)(e);
             }
+            // regex checks that the url is relative or protocol relative
+            if (status === 0 && isFileProtocol() && !/^(?:(?:https?|file):\/\/)/.test(options.url)) {
+                _error(options, MSG_CANT_PLAY_VIDEO, ERROR_XHR_FILE_PROTOCOL);
+
```
waxidiotic: Empty line.

src/js/utils/ajax.js
```diff
@@ -158,13 +160,19 @@ function _readyStateChangeHandler(options) {
...
+            if (status >= 400) {
+                _error(options, MSG_CANT_PLAY_VIDEO, (status < 600) ? status : ERROR_XHR_UNKNOWN);
                 return;
             }
-            if (xhr.status === 200) {
+            if (status === 200) {
                 return _ajaxComplete(options)(e);
             }
+            // regex checks that the url is relative or protocol relative
+            if (status === 0 && isFileProtocol() && !/^(?:(?:https?|file):\/\/)/.test(options.url)) {
```
vseventer: Can you explain what the regex is doing?

src/js/utils/validator.js
```diff
@@ -21,7 +21,21 @@ export function exists(item) {
...
+ */
+export function isFileProtocol() {
+    return getProtocol() === 'file:';
+}
+
+/**
+ * @returns {string} the protocol of the current page
+ */
+function getProtocol() {
+    return window.location.protocol;
```
vseventer: Instead of making this a function, you can assign it to a `const` directly.

src/js/utils/ajax.js
karimMourra: @vseventer there is a comment above the regex; did you miss it or was it unclear when you read it ?

vseventer: Yeah, I didn't realize `isFileProtocol` returns the current page protocol.. Either way, can't the regex be shortened to `/^(https?|file):\/\//`, since the `RegExp.test` method returns a `boolean`?

karimMourra: I agree 100%, `!/^[a-z][a-z0-9+.-]*:/.test(options.url)` is better

PR #3235, opened 2019-01-02
Prevent multiple 'play' and 'providerFirstFrame' events for Dash streams on Firefox

DanFerrer:
### This PR will...

- Exit early from ```playing()``` if ```this.state``` is already set to ```playing```

To prevent inaccurate ```play``` and ```providerFirstFrame``` events for Dash streams when they appear to be stalled but are actually playing in Firefox

#### Addresses Issue(s):

JW8-2547

### DIFF
--- a/src/js/providers/video-listener-mixin.js
+++ b/src/js/providers/video-listener-mixin.js
@@ -48,7 +48,7 @@ const VideoListenerMixin = {
 
         if (!this.seeking && !this.video.paused &&
             (this.state === STATE_STALLED || this.state === STATE_LOADING) &&
-            this.stallTime !== currentTime) {
+            this.stallTime !== this.video.currentTime) {
             this.stallTime = -1;
             this.setState(STATE_PLAYING);
             this.trigger(PROVIDER_FIRST_FRAME);
@@ -103,6 +103,9 @@ const VideoListenerMixin = {
     playing() {
         // When stalling, STATE_PLAYING is only set on timeupdate
         // because Safari and Firefox will fire "playing" before playback recovers from stalling
+        if (this.state === STATE_PLAYING) {
+            return;
+        }
         if (this.stallTime === -1) {
             // Here setting STATE_PLAYING ensures a quick recovery from STATE_LOADING after seeking
             this.setState(STATE_PLAYING);


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/providers/video-listener-mixin.js
```diff
@@ -48,7 +48,7 @@ const VideoListenerMixin = {
 
         if (!this.seeking && !this.video.paused &&
             (this.state === STATE_STALLED || this.state === STATE_LOADING) &&
-            this.stallTime !== currentTime) {
+            this.stallTime !== this.video.currentTime) {
```
robwalch: Why did you make this change when `currentTime` was assigned the value of `this.video.currentTime` above?

src/js/providers/video-listener-mixin.js
```diff
@@ -103,6 +103,9 @@ const VideoListenerMixin = {
     playing() {
         // When stalling, STATE_PLAYING is only set on timeupdate
         // because Safari and Firefox will fire "playing" before playback recovers from stalling
+        if (this.state === STATE_PLAYING) {
```
robwalch: This change will affect all providers and should not be made. Multiple  `PROVIDER_FIRST_FRAME` events are fine because these are internal events handled in the player with 'firstFrame' being the external deduped event.

If we're getting multiple playing events that is a problem with state change deduping or  `this.stallTime` not being treated properly in the shaka provider.

johnBartos: >'firstFrame' being the external deduped event.

Ok, we can make it so that we only guard against firing PLAYING.

>If we're getting multiple playing events that is a problem with state change deduping or this.stallTime not being treated properly in the shaka provider.

The problem is that FF fires `timeupdate` before `playing`, causing the `PLAY` event to be fired twice.

src/js/providers/video-listener-mixin.js
johnBartos: @robwalch `currentTime` can change between allocation and usage. This is the case with the current bug; shaka seeks backwards after. But I can't remember if it was the case that `this.video.currentTime` was equal to `this.stallTime`, and that `currentTime` wasn't; @DanFerrer Can you double check if this is still necessary?

PR #3237, opened 2019-01-02
Update test dependencies

DanFerrer:
### This PR will...

Update test dependencies 

Keep packages up to date

#### Addresses Issue(s):

JW8-2349
JW8-2350

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -81,10 +81,11 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
         },
 
         timeupdate() {
-            if (_positionBeforeSeek !== _videotag.currentTime) {
-                _setPositionBeforeSeek(_videotag.currentTime);
-                VideoEvents.timeupdate.call(_this);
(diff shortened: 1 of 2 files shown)


yarn.lock
```diff
@@ -53,18 +62,38 @@
     "@babel/template" "7.0.0-beta.44"
     "@babel/types" "7.0.0-beta.44"
 
+"@babel/helper-function-name@^7.1.0":
+  version "7.1.0"
```
karimMourra: For all of these it seems like we're pointing to the minor version, but we still have a block pointing to the beta version, i.e. https://github.com/jwplayer/jwplayer/pull/3237/files#diff-8ee2343978836a779dc9f8d6b794c3b2R58
is that safe ?

robwalch: You don't need to worry about changes in the lock file. Especially for child dependencies like this.

robwalch: What you do need to do @DanFerrer is rebase and run `yarn install` on your branch. The branch and the lock file are out of date.

Issue #3240, opened 2019-01-08
Tooltip refactor


karimMourra:
`Tooltip.js` is very confusing because it actually creates a controlbar button containing a tooltip as a child element, instead of just creating a tooltip. Since `VolumeTooltip.js` inherits from `Tooltip.js`, we are confused when accessing `volumeTooltip.element()` since lexically we would expect it to return a tooltip (the volume slider) but it actually returns the mute button.
A few things to consider:
- `simple-tooltip.js` actually creates tooltips. 
- `TimeTip` (defined in `timeslider.js`) inherits from Tooltip.

PR #3241, opened 2019-01-08
Mute button incorrectly announced as slider

karimMourra:
### This PR will...
- rename the `volumeTooltip` controlbar element to `volumeGroup`
- remove the aria information specific to the volume slider from the mute button, and add it to the volumeGroup's overlay (the volume slider)
- set the mute button's role to  'button'

the mute button was being announced by screen readers as a slider, when instead it should be announced as a button. I think this bug was partially caused by the fact that `VolumeTooltip` is a very misleading name since it creates an element containing a tooltip as a child element, instead of creating an actual tooltip. To reduce confusion, `volumeTooltip` was renamed to `volumeGroup` and tech debt item https://github.com/jwplayer/jwplayer/issues/3240 was created to refactor `VolumeTooltip.js` and `Tooltip.js`.
n/a
n/a
#### Addresses Issue(s):
JW8-2567

### DIFF
--- a/src/js/view/controls/components/volumetooltip.js
+++ b/src/js/view/controls/components/volumetooltip.js
@@ -15,8 +15,7 @@ export default class VolumeTooltip extends Tooltip {
 
         const volumeSliderElement = this.volumeSlider.element();
         volumeSliderElement.classList.remove('jw-background-color');
-
-
+        
         const overlay = this.container;
         setAttribute(overlay, 'tabindex', '0');
         setAttribute(overlay, 'aria-label', localization.volumeSlider);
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

waxidiotic:
@karimMourra LGTM. The only thing I noticed is that, when using JAWS, the screen reader always says "press spacebar to activate" (or something similar) but spacebar toggles playback. Has this always been this way?

karimMourra:
@waxidiotic there's a similar behavior on mac, i think we have a ticket to take care of it. i think it says cmd + ctrl + space to use button, which is nonsense but since skillsoft has not complained about it, it has not been prioritized. Ideally there exists sthg like `aria-command` to tell the screen reader what the keys are

PR #3244, opened 2019-01-09
Change hidden vol/time readers from <p> to <span>

pajong:
### This PR will...
Change hidden volume and time update readers from `<p>` to `<span>`

Safari with VoiceOver does not announce the update to the first `<p>` in a hidden `<div>` if more than one `<p>` exist in the same `<div>`, because `<p>` starts a new line.
Changing to `<span>` works because `<span>` is an inline element.

#### Addresses Issue(s):
JW8-2517

### DIFF
--- a/src/templates/player.js
+++ b/src/templates/player.js
@@ -13,8 +13,8 @@ export default (id, ariaLabel) => {
                 `</div>` +
                 `<div class="jw-overlays jw-reset"></div>` +
                 `<div class="jw-hidden-accessibility">` +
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3245, opened 2019-01-10
Remove unnecessary aria-expanded

karimMourra:
### This PR will...
- remove aria-expanded from the tooltips
Per the [documentation](https://www.w3.org/WAI/PF/aria/states_and_properties#aria-expanded), `aria-expanded` should be used on elements which can be interacted with to expand or collapse content (i.e. a drop down list). This label is unnecessary for tooltips since they cannot be interacted with; the user can't tab into a tooltip. The tooltip's purpose is to display information; the same information is contained by the aria-labels on the actual buttons, and is announced by the screen reader.
n/a
https://github.com/jwplayer/jwplayer-plugin-related/pull/370
#### Addresses Issue(s):
JW8-2518

### DIFF
--- a/src/js/view/controls/components/simple-tooltip.js
+++ b/src/js/view/controls/components/simple-tooltip.js
@@ -18,7 +18,6 @@ export function SimpleTooltip(attachToElement, name, text, openCallback, closeCa
                 return;
             }
 
-            tooltipElement.setAttribute('aria-expanded', 'true');
             addClass(tooltipElement, 'jw-open');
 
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3250, opened 2019-01-16
[JW8-2159]Toggle settings menu icon dependent on amount of submenus 

DanFerrer:
### This PR will...

Toggle settings menu on the controlbar dependent on the number submenu items an item has

Improvement for the controlbar. We don't want to see two buttons in the control bar that do the exact same thing, if there is no difference between them (ex: Captions is the only submenu and we show both the captions icon and settings icon)

#### Addresses Issue(s):

JW8-2159

### DIFF
--- a/src/js/view/controls/settings-menu.js
+++ b/src/js/view/controls/settings-menu.js
@@ -54,7 +54,8 @@ export function createSettingsMenu(controlbar, onVisibility, localization) {
 function showSettingsMenuIcon(settingsMenu, controlbar) {
     // Show or hide settings menu icon dependant on amount of submenus
     const submenuNames = settingsMenu.getSubmenuNames();
-    controlbar.elements.settingsButton.toggle(submenuNames.indexOf('quality') > -1 || submenuNames.length > 1);
+    const toggleIcon = submenuNames.indexOf('quality') > -1 || submenuNames.indexOf('playbackRates') > -1;
+    controlbar.elements.settingsButton.toggle(toggleIcon || submenuNames.length > 1);
 }
 
 export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api) {
--- a/src/js/view/controls/settings-menu.js
+++ b/src/js/view/controls/settings-menu.js
@@ -12,8 +12,10 @@ import {
 
 export function createSettingsMenu(controlbar, onVisibility, localization) {
     const settingsButton = controlbar.elements.settingsButton;
-    const onSubmenuAdded = () => {
-        settingsButton.show();
+    const onSubmenuAdded = (submenuNames) => {
+        if (submenuNames.length > 1) {
+            settingsButton.show();
+        }
     };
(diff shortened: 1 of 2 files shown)


src/js/view/controls/settings-menu.js
```diff
@@ -83,6 +86,7 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api)
     const onQualitiesChanged = (changedModel, levels) => {
         if (!levels || levels.length <= 1) {
             removeQualitiesSubmenu(settingsMenu);
+            showSettingsMenuIcon(settingsMenu, controlbar);
```
karimMourra: instead of having the same function call for both scenarios, i would wrap the bottom condition in an else statement and call the function once after the if-else block
```
    const onQualitiesChanged = (changedModel, levels) => {
        if (!levels || levels.length <= 1) {
            removeQualitiesSubmenu(settingsMenu);
        } else {
            const { hd, auto } = model.get('localization');
            addQualitiesSubmenu(
                settingsMenu,
                levels,
                (index) => api.setCurrentQuality(index),
                model.get('currentLevel'),
                hd,
                auto
            );
        }
        showSettingsMenuIcon(settingsMenu, controlbar);
    };
```

src/js/view/controls/settings-menu.js
```diff
@@ -54,6 +51,12 @@ export function createSettingsMenu(controlbar, onVisibility, localization) {
     return settingsMenu;
 }
 
+function showSettingsMenuIcon(settingsMenu, controlbar) {
+    // Show or hide settings menu icon dependant on amount of submenus
+    const submenuNames = settingsMenu.getSubmenuNames();
+    const toggleIcon = submenuNames.indexOf('quality') > -1 || submenuNames.indexOf('playbackRates') > -1;
```
robwalch: How about using `some` instead of `indexOf` twice?
```js
const toggleIcon = submenuNames.some(name => (name === 'quality' || name === 'playbackRates'));
```

src/js/view/controls/settings-menu.js
```diff
@@ -54,6 +51,12 @@ export function createSettingsMenu(controlbar, onVisibility, localization) {
...
+    controlbar.elements.settingsButton.toggle(toggleIcon || submenuNames.length > 1);
```
robwalch: You don't need `some` or `indexOf` to run at all if you check length first. You should always short circuit expensive operations with cheaper ones like this, and encapsulate all `toggleIcon` logic in one place:

```js
const toggleIcon = submenuNames.length > 1 || submenuNames.some(name => (name === 'quality' || name === 'playbackRates'));
```

johnBartos:
test this please

johnBartos:
@mikesalvia @DanFerrer Looks like some automated tests need to be updated for this to pass?

`

Unable to find visible css "div.jw-icon-settings"

`

PR #3251, opened 2019-01-16
[DNM] Remove spinning transition from jw-idle-label class

DanFerrer:
### This PR will...

- Remove ```transition``` and ```transition-property``` from ```.jw-idle.label```
- Move ```filter``` to be applied on idle state

```transition``` was causing the state icons to spin during playback.

#### Addresses Issue(s):

JW8-2614

### DIFF
--- a/src/css/controls/imports/playback-label.less
+++ b/src/css/controls/imports/playback-label.less
@@ -15,17 +15,16 @@
 .jw-idle-label {
     border-radius: 50%;
     color: @white;
-    filter: drop-shadow(1px 1px 5px rgba(12, 26, 71, 0.25));
     font: normal 16px/1 Arial, Helvetica, sans-serif;
     position: relative;
-    transition: background-color @default-transition;
-    transition-property: background-color, filter;
     -webkit-font-smoothing: antialiased;
 }
 
 .jw-state-idle {
     .jw-icon-display {
         &.jw-idle-label {
+            filter: drop-shadow(1px 1px 5px rgba(12, 26, 71, 0.25));
+
             .jw-idle-icon-text {
                 display: block;
             }

### CODE REVIEW

**Comment by johnBartos** [comment]
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

File: src/css/controls/imports/playback-label.less [hunk 1]
```diff
@@ -15,17 +15,16 @@
...
     position: relative;
-    transition: background-color @default-transition;
-    transition-property: background-color, filter;
     -webkit-font-smoothing: antialiased;
 }
 
 .jw-state-idle {
     .jw-icon-display {
         &.jw-idle-label {
+            filter: drop-shadow(1px 1px 5px rgba(12, 26, 71, 0.25));
```
robwalch: `filter` is needed for svgs, but for elements containing only text, we should be using `text-shadow` https://developer.mozilla.org/en-US/docs/Web/CSS/text-shadow

src/css/controls/imports/playback-label.less
```diff
@@ -15,17 +15,16 @@
 .jw-idle-label {
     border-radius: 50%;
     color: @white;
-    filter: drop-shadow(1px 1px 5px rgba(12, 26, 71, 0.25));
     font: normal 16px/1 Arial, Helvetica, sans-serif;
     position: relative;
-    transition: background-color @default-transition;
-    transition-property: background-color, filter;
```
robwalch: So we saw it transitioning the `transform` of the element, not the `background-color`, `filter`. Any idea why removing these properties prevents the element from rotating?

PR #3252, opened 2019-01-18
Set Chromecast item startTime to currentTime to support DVR live edge

robwalch:
### This PR will...

Seek to the live edge when casting Live/DVR Streams.

When casting, the Chromecast module will tell the receiver to seek to the current item's `startTime`. Since receivers seek using `currentTime` we should always set `startTime` to this value.

Changes to the Chromecast provider in commercial enable DVR controls with time events that include currentTime and seekable ranges.
https://github.com/jwplayer/jwplayer-commercial/pull/6208

#### Addresses Issue(s):
JW8-1703

### DIFF
--- a/src/js/program/program-controller.js
+++ b/src/js/program/program-controller.js
@@ -197,7 +197,7 @@ class ProgramController extends Eventable {
         model.attributes.itemReady = false;
 
         const playlistItem = Object.assign({}, item);
-        playlistItem.starttime = model.mediaModel.get('position');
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3254, opened 2019-01-22
Don't expect seekRange to be defined on the player model

robwalch:
### This PR will...
Prevent exceptions while seeking in DVR streams with providers that do not output `seekRange`.

Provider 'time' and 'meta' event overwrite model `seekRange`. If the value is undefined in these provider events, it will be undefined in the model.

https://github.com/jwplayer/jwplayer-commercial/pull/6213

#### Addresses Issue(s):
JW8-1807

### DIFF
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -196,7 +196,7 @@ class TimeSlider extends Slider {
         if (duration === 0) {
             this._api.play(reasonInteraction());
         } else if (this.streamType === 'DVR') {
-            const seekRange = this._model.get('seekRange');
+            const seekRange = this._model.get('seekRange') || { start: 0 };
             const dvrSeekLimit = this._model.get('dvrSeekLimit');
             position = seekRange.start + (-duration - dvrSeekLimit) * percent / 100;
             this._api.seek(position, reasonInteraction());


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/components/timeslider.js
```diff
@@ -196,7 +196,7 @@ class TimeSlider extends Slider {
         if (duration === 0) {
             this._api.play(reasonInteraction());
         } else if (this.streamType === 'DVR') {
-            const seekRange = this._model.get('seekRange');
+            const seekRange = this._model.get('seekRange') || { start: 0 };
```
johnBartos: Do you think it would be better to set seekRange to this default in `program-listeners` instead?

robwalch: In that case I would expect `start` and `end` to be defined, and that would require a lot more work. I went this route because it's the simplest change and `seekRange` reflects the output of provider "time" and "meta" events.

PR #3256, opened 2019-01-22
Unshift cues so that they are added in order

robwalch:
### This PR will...
Unshift cues into a temp array rather than pushing so that they will be re-added in the correct order.

`insertCueInOrder` is supposed remove and re-add cues that come after the `vttCue` argument being inserted into the text track. Pushing the cues into a temp array adds them in reverse order. This results in an invalid exception in Edge and IE when re-adding second to last cue, after the last cue.  

#### Addresses Issue(s):
JW8-2571


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3257, opened 2019-01-23
Fix DVR seeking in html5 provider

robwalch:
### This PR will...
Fix issues with seeking to the live edge and rewinding 10 seconds during DVR playback in Safari with the html5 provider.

`position` and `dvrPosition` should always be `0` or less during DVR playback. `0` would indicate that you are at the live edge. Positive values caused seek time calculated to target the start of a stream instead of the end.

When seek is called with a negative value, we should seek back from the live edge which is `seekRange.end`. Instead we were subtracting from `seekRange.start + seekRange.end` which resulted in drift making the rewind button seek target drift forward the longer a DVR stream was played.

#### Addresses Issue(s):
JW8-716 JW8-2594

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -143,10 +143,10 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
         },
 
         seeking() {
-            const offset = _seekOffset !== null ? _seekOffset : _this.getCurrentTime();
+            const offset = _seekToTime !== null ? _convertTime(_seekToTime) : _this.getCurrentTime();
             const position = _positionBeforeSeek;
-            _setPositionBeforeSeek(offset);
-            _seekOffset = null;
+            _positionBeforeSeek = offset;
+            _seekToTime = null;
             _delayedSeek = 0;
             _this.seeking = true;
             _this.trigger(MEDIA_SEEK, {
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/providers/html5.js
```diff
@@ -143,10 +143,10 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
         },
 
         seeking() {
-            const offset = _seekOffset !== null ? _seekOffset : _this.getCurrentTime();
+            const offset = _seekToTime !== null ? _convertTime(_seekToTime) : _this.getCurrentTime();
```
johnBartos: Is it OK that `_convertTime(_seekToTime)` can also result in a seek?

robwalch: `_convertTime` never results in a seek. We don't know when Safari updates it's `seekable` TimeRange, so we use this method to update the DVR state vars in `updateDvrPosition` if the range has changed.

`_convertTime` is called every time we need a time `event.position` value which is calculated using these DVR state vars so it's perfect for keeping up with changing segment window.

johnBartos: _converTime calls updateDvrPosition, which can call`                     _videotag.currentTime = _seekToTime; `

robwalch: It does not. That is code from `pause` (click the expand button).

PR #3258, opened 2019-01-23
Prevent exception exiting fullscreen via escape key

### LINKED ISSUE #3168: Missing cleanup when closing player in fullscreen
### Expected Behavior
When the player.destroy() method is called, the player should remove any styles it has set on the body element.

### Actual Behavior

If the JwPlayer react component gets unmounted while in fullscreen then it leaves some styling on the body tag disabling scrolling, see https://github.com/micnews/react-jw-player/issues/112. 

According to the issue mentioned above, they call player.destroy() but the player doesn't cleanup as it should.

### Environment
- IE, Chrome, latest versions
- react-JWPlayer  1.18.1
- JWPlayer 8.5.3.

robwalch:
### This PR will...
1. Prevent exception exiting fullscreen via escape key.
2. Cleanup key handler that called a function where `this` is undefined (it's only defined in the `handleKeydown` arrow function).

`document.exitFullscreen` throws an except in Chrome and Firefox when `document.fullscreenElement` is `null`. These changes allow for backwards compatibility with browsers where `document.fullscreenElement` is always `undefined` .
 
We can't try catch because `exitFullscreen` will throw an unhandled promise rejection. We don't need to support the promise implementation of fullscreen methods at this point as that would require a ton of polyfilling and player changes.

#### Addresses Issue(s):
JW8-2587
Fixes #3168

### DIFF
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -233,20 +233,6 @@ export default class Controls {
             api.setVolume(newVol);
         }
 
-        function onEscape() {
-            if (model.get('fullscreen')) {
-                api.setFullscreen(false);
-                this.playerContainer.blur();
-                this.userInactive();
-                return;
-            }
-
-            const related = api.getPlugin('related');
-            if (related) {
-                related.close({ type: 'escape' });
-            }
-        }
-
         const handleKeydown = (evt) => {
             // If Meta keys return
             if (evt.ctrlKey || evt.metaKey) {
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3262, opened 2019-01-24
Change floatOnScroll boolean to float object

pajong:
### This PR will...
Change the name to `float`, and making it an object.
Add a new param `dismissible` in the `float` object to remove the close button for floating player if set to false.

`floatOnScroll` config option used to be boolean.
We want that to be an object so that we are able to have other options to configure different functionalities for floating player.
Some customers do not want the close button to un-float the floating player, so adding an option to disable the close button.

Config schema update: https://github.com/jwplayer/jwplayer-commercial/pull/6238

#### Addresses Issue(s):

JW8-2624 JW8-2631

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -219,7 +219,7 @@ function View(_api, _model) {
 
         if (_floatOnScroll && _floatOnScroll.dismissible !== false) {
             const floatCloseButton = new FloatingCloseButton(_wrapperElement.querySelector('.jw-top'));
-            floatCloseButton.setup(() => _stopFloating(true), _localization.close);
+            floatCloseButton.setup(() => this.stopFloating(true, true), _localization.close);
         }
 
         _model.on('change:hideAdsControls', function (model, val) {
@@ -877,17 +877,19 @@ function View(_api, _model) {
 
             _resizeOnFloat = false;
         } else if (isVisible) {
-            _stopFloating();
+            this.stopFloating();
         }
     }
 
-    function _stopFloating(forever) {
+    this.stopFloating = function(forever, isClicked) {
         if (floatingPlayer === _playerElement) {
             floatingPlayer = null;
 
             if (forever) {
                 _floatOnScroll = null;
-                _api.pause({ reason: 'interaction' });
+                if (isClicked) {
+                    _api.pause({ reason: 'interaction' });
+                }
             }
 
             removeClass(_playerElement, 'jw-flag-floating');
@@ -899,7 +901,7 @@ function View(_api, _model) {
             style(_wrapperElement, { width: null, height: null });
             _this.resize(_model.get('width'), _model.get('aspectratio') ? undefined : _model.get('height'));
         }
-    }
+    };
 
 
     this.destroy = function () {
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -77,7 +77,7 @@ function View(_api, _model) {
     let _resizeOnFloat = false;
     let _stateClassRequestId = -1;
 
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -886,7 +886,8 @@ function View(_api, _model) {
             floatingPlayer = null;
 
             if (forever) {
-                _floatOnScroll = false;
+                _floatOnScroll = null;
```
vseventer: Why `null` rather then `false`?

pajong: It's now an object so I put `null` instead of changing it to a boolean

src/js/view/view.js
```diff
@@ -886,7 +886,8 @@ function View(_api, _model) {
...
+                _api.pause({ reason: 'interaction' });
```
karimMourra: `stopFloating` can be called by the intersection observer, in which case the reason would not be interaction, correct ?
https://github.com/jwplayer/jwplayer/blob/7f093c7fabd2f08e0870543f158998b2172377a7/src/js/view/view.js#L880

src/js/view/view.js
```diff
@@ -77,7 +77,7 @@ function View(_api, _model) {
     let _resizeOnFloat = false;
     let _stateClassRequestId = -1;
 
-    let _floatOnScroll = _model.get('floatOnScroll');
+    let _floatOnScroll = _model.get('float');
```
johnBartos: Nitpick: rename this to `_float`. `_floatOnScroll` implies a boolean

src/js/view/view.js
pajong: addressed

src/js/view/view.js
```diff
@@ -217,9 +217,9 @@ function View(_api, _model) {
         focusHelper = new UI(_playerElement).on('click', function() {});
         fullscreenHelpers = requestFullscreenHelper(_playerElement, document, _fullscreenChangeHandler);
 
-        if (_floatOnScroll) {
+        if (_floatOnScroll && _floatOnScroll.dismissible !== false) {
             const floatCloseButton = new FloatingCloseButton(_wrapperElement.querySelector('.jw-top'));
-            floatCloseButton.setup(() => _stopFloating(true), _localization.close);
+            floatCloseButton.setup(() => this.stopFloating(true, true), _localization.close);
```
johnBartos:
```
floatCloseButton.setup(() => { 
    this.stopFloating(true);
     _api.pause({ reason: 'interaction' }); 
    }, _localization.close); 
```

This would be doing the same thing, right?

src/js/view/view.js
```diff
@@ -77,7 +77,7 @@ function View(_api, _model) {
     let _resizeOnFloat = false;
     let _stateClassRequestId = -1;
 
-    let _floatOnScroll = _model.get('floatOnScroll');
+    let _float = _model.get('float');
```
johnBartos: Hmm I think it would make more sense to call it `_floatingConfig`

src/js/view/view.js
```diff
@@ -877,16 +880,16 @@ function View(_api, _model) {
 
             _resizeOnFloat = false;
         } else if (isVisible) {
-            _stopFloating();
+            this.stopFloating();
         }
     }
 
-    function _stopFloating(forever) {
+    this.stopFloating = function(forever) {
```
johnBartos: Does this method still need to be on the instance?

pajong: Yes, it is being called in commercial

johnBartos: Ok I see, thanks

PR #3264, opened 2019-01-25
[JW8-2594] Support seeking in DVR using currentTime

robwalch:
### This PR will...
Support seeking in DVR using currentTime

In DVR mode we fudge seek time based on how long it's been since we updated `dvrPosition` which is a negative value representing the estimated distance from the live edge. This fudging is only necessary when handling a negative time argument in `seek` because it is an offset of `dvrPosition`. If we get a positive value, we can trust that it is based on `currentTime` and does not need to be fudged.

Also cleaned up handling of time before the media element seek event by making it based on `currentTime` rather than `dvrPosition`, and only updating it on timeupdate when necessary (webkit fullscreen mode).

These changes allow the rewind button seek back 10 seconds more consistently because the logic is based on the absolute `currentTime` instead of the approximate `dvrPosition` which changes with the DVR window (`video.seekable.end()`).

https://github.com/jwplayer/jwplayer-commercial/pull/6230

#### Addresses Issue(s):
JW8-2594

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -81,10 +81,11 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
         },
 
         timeupdate() {
-            if (_positionBeforeSeek !== _videotag.currentTime) {
-                _setPositionBeforeSeek(_videotag.currentTime);
-                VideoEvents.timeupdate.call(_this);
+            // Keep track of position before seek in iOS fullscreen
+            if (_iosFullscreenState && _timeBeforeSeek !== _videotag.currentTime) {
+                setTimeBeforeSeek(_videotag.currentTime);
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

johnBartos:
Looks like a flaky unit test: `Edge_17_17134_0_(Windows_10_0_0).browserstack.api.setup.api.setup triggers "remove" if the api has been previously setup (from browserstack.Edge 17.17134.0 (Windows 10.0.0)) `

PR #3267, opened 2019-01-25
[JW8-2530] [JW8-5612] Update float on scroll resizing on mobile and desktop

DanFerrer:
### This PR will...

- Float when the player is 50% visible and unfloat otherwise.
- Make ```MAX_FLOATING_WIDTH``` and ```MAX_FLOATING_HEIGHT```  ```400px```.
- Height is based off aspect ratio of the floating player:
    -  Adjust height by ratio if the width of the player is equal to the ```MAX_FLOATING_WIDTH```.
    -  Retain height (and width) if  the width of the player is smaller than ```MAX_FLOATING_WIDTH```.
- Use media query for mobile portrait float behavior when screen is less than or equal to ```419px```
- Float player 125px below the top of the page to account for sticky headers.
- Create a new function called ```resizeFloatingPlayer``` to handle resize when switching from```portrait``` to ```landscape``` and vice-versa.
- Make player fill 100% of the viewport if mobile and portrait view.

Improvements to player float on scroll behavior.

#### Addresses Issue(s):

JW8-2530

### DIFF
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -9,6 +9,7 @@
         bottom: 1rem;
         left: auto;
         position: fixed;
+        width: 100%;
         z-index: 1;
 
         @media screen and (max-width: 419px) and (orientation: portrait) {
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -3,14 +3,23 @@
     background-color: #000;
 
     .jw-wrapper {
-        animation: jw-float 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
-        top: auto;
-        right: 1rem;
-        bottom: 1rem;
         left: auto;
+        right: 1rem;
         position: fixed;
         width: 100%;
         z-index: 1;
+
+        &.jw-float-bottom {
+            animation: jw-float-to-bottom 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
+            top: auto;
+            bottom: 1rem;
+        }
+
+        &.jw-float-top {
+            animation: jw-float-to-top 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
+            top: 2rem;
+            bottom: auto;
+        }
 
         @media screen and (max-width: 419px) and (orientation: portrait) {
             left: 0;
@@ -57,9 +66,19 @@
     justify-content: center;
 }
 
-@keyframes jw-float {
+@keyframes jw-float-to-bottom {
     from {
         transform: translateY(100%);
+    }
+
+    to {
+        transform: translateY(0);
+    }
+}
+
+@keyframes jw-float-to-top {
+    from {
+        transform: translateY(-100%);
     }
 
     to {
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -1,3 +1,5 @@
+@import "../../shared-imports/vars.less";
+
 .jw-flag-floating {
     background-size: cover;
     background-color: #000;
@@ -64,8 +66,8 @@
 .jw-float-icon {
     display: none;
     position: absolute;
-    top: 0;
-    right: 0;
+    top: @dismiss-top-padding;
+    right: @dismiss-right-padding;
     align-items: center;
     justify-content: center;
 }
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -13,13 +13,12 @@
 
         @media screen and (max-width: 419px) and (orientation: portrait) {
             animation: jw-float-to-top 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
-            top: 2rem;
+            top: 125px;
             bottom: auto;
             left: 0;
             right: 0;
             margin-left: auto;
             margin-right: auto;
-            width: 100%;
         }
     }
 
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -17,7 +17,8 @@
 
         @media screen and (max-width : 400px) {
             width: 100% !important; /* stylelint-disable-line declaration-no-important */
-            right: auto;
+            left: 0;
+            right: 0;
         }
 
         .jw-flag-touch& {
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -6,26 +6,23 @@
         left: auto;
         right: 1rem;
         position: fixed;
-        width: 100%;
         z-index: 1;
 
-        &.jw-float-bottom {
+        @media screen and (min-width: 420px) {
             animation: jw-float-to-bottom 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
             top: auto;
             bottom: 1rem;
         }
 
-        &.jw-float-top {
+        @media screen and (max-width: 419px) and (orientation: portrait) {
             animation: jw-float-to-top 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
             top: 2rem;
             bottom: auto;
-        }
-
-        @media screen and (max-width: 419px) and (orientation: portrait) {
             left: 0;
             right: 0;
             margin-left: auto;
             margin-right: auto;
+            width: 100%;
         }
     }
 
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -7,12 +7,10 @@
         right: 1rem;
         position: fixed;
         z-index: 1;
+        animation: jw-float-to-bottom 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
+        top: auto;
+        bottom: 1rem;
 
-        @media screen and (min-width: 420px) {
-            animation: jw-float-to-bottom 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
-            top: auto;
-            bottom: 1rem;
-        }
 
         @media screen and (max-width: 419px) and (orientation: portrait) {
             animation: jw-float-to-top 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -16,7 +16,7 @@
         max-height: 400px;
         margin: 0 auto;
 
-        @media screen and (max-width : 400px) {
+        @media screen and (max-width : 480px) {
             width: 100% !important; /* stylelint-disable-line declaration-no-important */
             left: 0;
             right: 0;
@@ -29,7 +29,7 @@
                 bottom: auto;
                 left: 0;
                 right: 0;
-                max-width: none;
+                max-width: none !important; /* stylelint-disable-line declaration-no-important */
                 max-height: none;
             }
         }
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -14,6 +14,7 @@
         right: 1rem;
         max-width: 400px;
         max-height: 400px;
+        margin: 0 auto;
 
         @media screen and (max-width : 400px) {
             width: 100% !important; /* stylelint-disable-line declaration-no-important */
@@ -30,8 +31,6 @@
                 right: 0;
                 max-width: none;
                 max-height: none;
-                margin-left: auto;
-                margin-right: auto;
             }
         }
     }
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -3,22 +3,33 @@
     background-color: #000;
 
     .jw-wrapper {
-        left: auto;
-        right: 1rem;
         position: fixed;
         z-index: 1;
         animation: jw-float-to-bottom 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
         top: auto;
         bottom: 1rem;
+        left: auto;
+        right: 1rem;
+        max-width: 400px;
+        max-height: 400px;
 
-        @media screen and (max-width: 419px) and (orientation: portrait) {
-            animation: jw-float-to-top 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
-            top: 125px;
-            bottom: auto;
-            left: 0;
-            right: 0;
-            margin-left: auto;
-            margin-right: auto;
+        @media screen and (max-width : 400px) {
+            width: 100% !important; /* stylelint-disable-line declaration-no-important */
+            right: auto;
+        }
+
+        .jw-flag-touch& {
+            @media screen and (max-device-width : 480px) and (orientation: portrait) {
+                animation: jw-float-to-top 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
+                top: 125px;
+                bottom: auto;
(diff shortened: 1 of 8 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/view.js
```diff
@@ -868,14 +868,21 @@ function View(_api, _model) {
...
 
-            // Resize within MAX_FLOATING_WIDTH×MAX_FLOATING_HEIGHT bounds, never enlarge.
-            const { width, height } = _this.getSafeRegion(false);
-            const ratio = Math.min(1, MAX_FLOATING_WIDTH / width, MAX_FLOATING_HEIGHT / height);
-            _this.resize(width * ratio, height * ratio, true);
+                // Resize within MAX_FLOATING_WIDTH×MAX_FLOATING_HEIGHT bounds, never enlarge.
+                const { width, height } = _this.getSafeRegion(false);
+                const ratio = Math.min(1, MAX_FLOATING_WIDTH / width, MAX_FLOATING_HEIGHT / height);
+                const floatingWidth = Math.min(MAX_FLOATING_WIDTH, width);
+                const floatingHeight = floatingWidth === MAX_FLOATING_WIDTH ? height * ratio : height;
```
vseventer: @DanFerrer I'm confused what this is trying to do - if I have a video of 398x800, the floating player will be 398x800 with this change? That doesn't seem quite right - it used to be 199x400, which seems more appropriate.

DanFerrer: @vseventer Part of the AC was to keep the dimensions of the player if the players width was less than 400px.

vseventer: @DanFerrer Ah, you're right - think it's an oversight though, as:
- A player sized 400x800 will float 200x400,
- A player sized 398x800 will float 398x800, and take up a lot / almost all of the vertical height.

Just seems strange / unexpected to me.

src/js/view/view.js
```diff
@@ -868,14 +868,21 @@ function View(_api, _model) {
...
+                addClass(_wrapperElement, 'jw-float-bottom');
+                _resizeOnFloat = true;
 
-            // Resize within MAX_FLOATING_WIDTH×MAX_FLOATING_HEIGHT bounds, never enlarge.
-            const { width, height } = _this.getSafeRegion(false);
-            const ratio = Math.min(1, MAX_FLOATING_WIDTH / width, MAX_FLOATING_HEIGHT / height);
-            _this.resize(width * ratio, height * ratio, true);
+                // Resize within MAX_FLOATING_WIDTH×MAX_FLOATING_HEIGHT bounds, never enlarge.
+                const { width, height } = _this.getSafeRegion(false);
+                const ratio = Math.min(1, MAX_FLOATING_WIDTH / width, MAX_FLOATING_HEIGHT / height);
```
karimMourra: @DanFerrer Shouldn't the ratio be width/height ?  We would want the the proportion of the height to the width to be constant, right ?

src/js/view/view.js
```diff
@@ -868,14 +868,21 @@ function View(_api, _model) {
...
+            } else {
+                addClass(_wrapperElement, 'jw-float-bottom');
+                _resizeOnFloat = true;
 
-            // Resize within MAX_FLOATING_WIDTH×MAX_FLOATING_HEIGHT bounds, never enlarge.
-            const { width, height } = _this.getSafeRegion(false);
-            const ratio = Math.min(1, MAX_FLOATING_WIDTH / width, MAX_FLOATING_HEIGHT / height);
-            _this.resize(width * ratio, height * ratio, true);
+                // Resize within MAX_FLOATING_WIDTH×MAX_FLOATING_HEIGHT bounds, never enlarge.
+                const { width, height } = _this.getSafeRegion(false);
```
johnBartos: I think we need unit tests here to clarify the logic. I'm a bit confused by the math - I'm going to work it out on paper but I think tests would be helpful.

johnBartos: Test cases should be when both are under the max, both over, one over/one under, and reciprocal

src/js/view/view.js
```diff
@@ -854,8 +854,8 @@ function View(_api, _model) {
     }
 
     function _updateFloating(intersectionRatio) {
-        // Entirely invisible and no floating player already in the DOM.
-        const isVisible = intersectionRatio === 1;
+        // Player is 50% visible or less and no floating player already in the DOM.
+        const isVisible = intersectionRatio > 0.5;
```
robwalch: "viewable" is `intersectionRatio` rounded, so this should be `>= 0.5` or `Math.round(intersectionRatio) === 1`.

src/js/view/view.js
robwalch: There is no reason to call `getSafeRegion` inside view when you have `_lastWidth` and `_lastHeight` with the values you are looking for.

src/js/view/view.js
robwalch: All you need is this:
```js
const floatingWidth = Math.min(_lastWidth, MAX_FLOATING_WIDTH);
const floatingHeight = Math.min(_lastHeight * floatingWidth / _lastWidth, MAX_FLOATING_HEIGHT);
_this.resize(floatingWidth, floatingHeight, true);
```

src/js/view/view.js
robwalch: Thanks @vseventer

src/js/view/view.js
robwalch: If you actually care about which of x and y ratio, then you have to apply the x ratio to height when x ratio is smaller, or y ratio to width when it is smaller.

cesarrecio-enroute:
test this please

src/css/jwplayer/flags/floatingplayer.less
```diff
@@ -26,17 +36,26 @@
...
+    to {
+        transform: translateY(0);
+    }
+}
+
+@keyframes jw-float-to-top {
+    from {
+        transform: translateY(-100%);
+    }
+
```
robwalch: Why are we using an animation when we can just apply a transition on `transform`?

@radium-v Would you agree that we could just `transition` this `transform` , or is there a reason for using an animation?

radium-v: `transition` would require more class and event management in between the not-floated and floated states since they both sit at the origin (`0, 0`).

Essentially, `transition` doesn't work here because there's no value to transition from. `animation` allows for an immediate starting value when applied, regardless of any existing/nonexisting `transform` state.

src/css/jwplayer/flags/floatingplayer.less
```diff
@@ -6,26 +6,23 @@
         left: auto;
         right: 1rem;
         position: fixed;
-        width: 100%;
         z-index: 1;
 
-        &.jw-float-bottom {
+        @media screen and (min-width: 420px) {
```
robwalch: So there's no default? I don't think you should include this media query because what happens in `landscape` mode at `max-width: 419px`? There should just be small portrait, and everything else.

src/js/view/view.js
```diff
@@ -572,7 +572,7 @@ function View(_api, _model) {
             _model.set('height', playerHeight);
         }
 
-        if (_floatingConfig && _resizeOnFloat) {
+        if (_floatingConfig && _resizeOnFloat && playerStyle.width !== _model.get('containerWidth')) {
```
robwalch: Do you want to style `_playerElement` when `playerStyle.width === _model.get('containerWidth')`? If not, don't add this with `&&`.

src/js/view/view.js
```diff
@@ -870,21 +870,15 @@ function View(_api, _model) {
...
-                // Resize within MAX_FLOATING_WIDTH×MAX_FLOATING_HEIGHT bounds, never enlarge.
-                const { width, height } = _this.getSafeRegion(false);
-                const floatingWidth = Math.min(width, MAX_FLOATING_WIDTH);
-                const floatingHeight = Math.min(height * floatingWidth / width, MAX_FLOATING_HEIGHT);
-                _this.resize(floatingWidth, floatingHeight, true);
+            // Resize within MAX_FLOATING_WIDTH×MAX_FLOATING_HEIGHT bounds, never enlarge.
+            const { width, height } = _this.getSafeRegion(false);
+            const floatingWidth = Math.min(width, MAX_FLOATING_WIDTH);
+            const floatingHeight = Math.min(height * floatingWidth / width, MAX_FLOATING_HEIGHT);
+            _this.resize(floatingWidth, floatingHeight, true);
```
robwalch: The third argument `resize(floatingWidth, floatingHeight, true)` is not being used. Did you mean to pass in `_resizeOnFloat`? That would be better (purer) than using this `_resizeOnFloat` flag.

src/js/view/utils/views-manager.js
```diff
@@ -40,6 +40,9 @@ function matchIntersection(entry, group) {
 function scheduleResponsiveRedraw() {
     cancelAnimationFrame(responsiveRepaintRequestId);
     responsiveRepaintRequestId = requestAnimationFrame(function responsiveRepaint() {
+        views.forEach(view => {
+            view.resizeFloatingPlayer();
```
karimMourra: Why do we have 4 separate forEach blocks ? @robwalch are you ok with @DanFerrer consolidating them ?

src/js/view/view.js
```diff
@@ -875,15 +875,29 @@ function View(_api, _model) {
...
 
             _resizeOnFloat = false;
         } else if (isVisible) {
             _this.stopFloating();
         }
     }
 
+    this.resizeFloatingPlayer = function() {
+        if (floatingPlayer && !_resizeOnFloat) {
+            const rect = bounds(_playerElement);
```
karimMourra: we're already calling `bounds` in `updateBounds`, which calls `element.getBoundingClientRect();`. I believe there are performance issues with calling this since in some browsers it forces a new layout calculation. We should look for a way to minimize the calls to `bounds`

src/js/view/utils/views-manager.js
karimMourra: @DanFerrer if we call `view.resizeFloatingPlayer` after `view.checkResized` we might be able to get away with using `getSafeRegion` instead of `bounds(_playerElement);` in `view.resizeFloatingPlayer`

vseventer: Using `bounds` will not work properly for outstream players regardless, `getSafeRegion` is patched for outstream to return the correct dimensions so using `getSafeRegion` is the preferred approach here.

src/js/view/view.js
vseventer: See comment above regarding `bounds`.

src/js/view/view.js
```diff
@@ -875,15 +875,29 @@ function View(_api, _model) {
...
+            const containerWidth = Math.round(rect.width);
+            const containerHeight = Math.round(rect.height);
+            const floatingWidth = Math.min(containerWidth, MAX_FLOATING_WIDTH);
```
karimMourra: We could make this DRYer by reusing part of `_updateFloating`:
```
const { width, height } = _this.getSafeRegion(false);
const floatingWidth = Math.min(width, MAX_FLOATING_WIDTH);
const floatingHeight = Math.min(height * floatingWidth / width, MAX_FLOATING_HEIGHT);
```
we definitely want our floatingHeight and floatingWidth converters to be in one place

DanFerrer:
@karimMourra I ended up moving my logic inside the ```updateBounds``` function where i check if a floating player is active and if it is, we get the new floating dimensions for it based off the player element.  This way, we can properly resize the floating player on orientation change and fire the proper events when resizing.

@vseventer Let me know if this still interferes with the way outstream player works. My idea behind this is on ```orientationChange```,  while a floating player is floating, we want to resize it based off of the player elements new dimensions , and then fire ```resize``` and ```breakpoint``` at the appropriate time.

Also @robwalch let me know if there are any performance issues with this or if i should change anything else.

src/js/view/view.js
```diff
@@ -118,6 +118,15 @@ function View(_api, _model) {
 
         // Don't update container dimensions to 0, 0 when not in DOM
         if (containerWidth || containerHeight || inDOM) {
+            if (!_resizeOnFloat && floatingPlayer) {
+                // Update floating player based on player elements bounds when updateBounds fires on orientationChange.
+                const { floatingWidth, floatingHeight } = _getFloatingDimensions(containerWidth, containerHeight);
+                containerWidth = floatingWidth;
+                containerHeight = floatingHeight;
+                _resizeOnFloat = true;
+                _resizePlayer(floatingWidth, floatingHeight, true);
```
robwalch: `updateBounds` purpose is just that: to update "containerWidth" and "containerHeight" . It should not resize the player. That should have already happened before `updateBounds`  is called.

robwalch: Can't you just add min/max-width and height  styles to the player in `_updateFloating`:

```js
style(_playerElement, {
    backgroundImage: _preview.el.style.backgroundImage || _model.get('image'),
    maxWidth, // floatingWidth
    maxHeight // floatingHeight
});
```

`updateBounds` should be called by the responsive listener and update the player dimensions after it's floated. We shouldn't need to resize the view at all.

src/js/view/view.js
```diff
@@ -853,12 +852,20 @@ function View(_api, _model) {
     };
 
     function _getCurrentElement() {
-        return _model.get('isFloating') ? _wrapperElement : _playerElement;
+        return _model.get('isFloating') && _resizeOnFloat ? _wrapperElement : _playerElement;
```
vseventer: LGTM - can you just clarify why we need to check for `_resizeOnFloat` here as well? Note that `_getCurrentElement` is used to determine visibility etc. - just want to make sure this doesn't break.

DanFerrer: @vseventer Sure, I'm checking for ```_resizeOnFloat``` because i want to use the ```_playerElement``` in ```updateBounds``` instead of ```_wrapperElement``` when there is a player currently floating and the orientation of the device changes.

This is so floating player can be properly adjusted on either landscape or portrait based on the new dimensions of ```_playerElement```

src/js/view/view.js
```diff
@@ -903,6 +902,20 @@ function View(_api, _model) {
         }
     };
 
+    this.resizeFloatingPlayer = function() {
+        const html = document.documentElement;
+        const body = document.body;
+        const isMobilePortrait = _isMobile && (html.clientWidth || body.clientWidth) < 420;
```
robwalch: `html.clientWidth || body.clientWidth` is the width of the page. It has nothing to do with whether or not the device is in portrait orientation. It's also expensive to test the page dimensions (this check forces layout).

What are you really trying to do here?

Why `< 420` and not `< MAX_FLOATING_WIDTH`?

src/js/view/view.js
```diff
@@ -505,7 +510,12 @@ function View(_api, _model) {
             _api.pause({ reason: 'interaction' });
         });
 
-        controls.on('all', _this.trigger, _this);
+        controls.on('all', (type, event) => {
+            if (event && event.doNotForward) {
```
pajong: Is this to prevent `controller` from firing `dismissFloating` event?
Outstream listens to the `dismissFloating` view event: https://github.com/jwplayer/jwplayer-commercial/blob/master/src/js/view/view-outstream.js#L46

robwalch: Commercial view and this are the same.  This prevents events from being forwarded through the public API. dismissFloating is not a documented event or a new event that we want to make public.

robwalch:
@pajong @DanFerrer Added a commit to support a couple more cases:
- Support `jwplayer().resize(x, y)` while floating (the player container is still resized, and the floating player is updated)
- Maintain player sizes smaller than 400px wide on narrow pages (player width is 100% on narrow pages, but we maintain a max-width of the player's width)

src/js/view/view.js
pajong: But the commercial `view-outstream` is listening to `view.on('dismissFloating`)`, which is triggered in the line below with `_this.trigger`. Shouldn't the check be moved to `controller`?

robwalch: I was thinking commercial is also listening to controls (not the instance it's extending). I will look into this.

PR #3269, opened 2019-01-28
Enable captions on muted start with config

pajong:
### This PR will...
- Enable captions by default if mute is true on initial start
-- Choose the first available caption
- Caption should still be enabled if the first playlist item does not have captions, and the second playlist item does
-- Even if unmuted before the start of the second playlist item
- As soon as the user sets the captions to `off`, this feature does not have an effect for following playlist items

- A/B testing

#### Addresses Issue(s):

JW8-2641

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -9,6 +9,7 @@ const Captions = function(_model) {
     let _tracks = [];
     let _tracksById = {};
     let _unknownCount = 0;
+    let _this = this;
 
     // Reset and load external captions on playlist item
     _model.on('change:playlistItem', (model) => {
@@ -51,6 +52,8 @@ const Captions = function(_model) {
         let track = null;
         if (captionsMenuIndex !== 0) {
             track = _tracks[captionsMenuIndex - 1];
+        } else {
+            this.enableOnLoad = false;
         }
         model.set('captionsTrack', track);
     }, this);
@@ -115,7 +118,11 @@ const Captions = function(_model) {
         // Because there is no explicit track for "Off"
         //  it is the implied zeroth track
         if (label === 'Off') {
-            _model.set('captionsIndex', 0);
+            if (_this.enableOnLoad && _tracks.length) {
+                _model.set('captionsIndex', 1);
+            } else {
+                _model.set('captionsIndex', 0);
+            }
             return;
         }
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -212,6 +212,10 @@ Object.assign(Controller.prototype, {
                 }
             });
 
+            if (_model.get('mute') && _model.get('enableCaptionsOnMute')) {
+                _captions.enableOnLoad = true;
+            }
+
             _view.init();
         };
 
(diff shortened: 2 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -212,6 +212,10 @@ Object.assign(Controller.prototype, {
                 }
             });
 
+            if (_model.get('mute') && _model.get('enableCaptionsOnMute')) {
+                _captions.enableOnLoad = true;
```
johnBartos: Can you just `model.set('captionsIndex', 0);` here, instead of having this flag?

src/js/controller/controller.js
```diff
@@ -511,6 +511,10 @@ Object.assign(Controller.prototype, {
                     });
                 }
 
+                if (_this.getMute()) {
```
pajong: We still need the flag. If we do `_model.set('captionsIndex', 1)` here, it does not get applied to the 2nd playlist item in case the first one does not have captions. The `captionsIndex` on model is set to 1, but it is being replaced when the `itemReady` event fires for the next playlist item, that results the `captionsIndex` to be set to 0 again: https://github.com/jwplayer/jwplayer/pull/3269/files#diff-7c257d0959b87bf538adcbf0e4fa5504R119

johnBartos: Can we shift the flag to the model?

robwalch: What is the config option set by the ab test? Use that.

src/js/controller/captions.js
```diff
@@ -108,14 +110,18 @@ const Captions = function(_model) {
...
     }
 
-    function _selectDefaultIndex() {
+    function _selectDefaultIndex(selectAvailableTrack) {
         let captionsMenuIndex = 0;
         const label = _model.get('captionLabel');
 
         // Because there is no explicit track for "Off"
         //  it is the implied zeroth track
         if (label === 'Off') {
```
robwalch: `label === 'Off'` when the user elected to turn off captions. This should always take precedence. We should only turn on caption after this return statement if the ab test set an argument in the model that captions should be enabled AND we autostarted muted.

src/js/controller/captions.js
```diff
@@ -43,14 +43,16 @@ const Captions = function(_model) {
...
+        _setCaptionsList(this.captionsByDefault);
     }, this);
 
     // Listen for captions menu index changes from the view
     _model.on('change:captionsIndex', (model, captionsMenuIndex) => {
         let track = null;
         if (captionsMenuIndex !== 0) {
             track = _tracks[captionsMenuIndex - 1];
+        } else {
+            this.captionsByDefault = false;
```
robwalch: Use the model.

src/js/controller/controller.js
```diff
@@ -511,6 +511,10 @@ Object.assign(Controller.prototype, {
                     });
                 }
 
+                if (_this.getMute() && _model.get('enableCaptionsOnMutedAutostart')) {
```
robwalch: Why not just clear "enableCaptionsOnMutedAutostart"? We don't need another property.

src/js/controller/captions.js
robwalch: If you call `localStorage.clear()` and refresh your test page, `_model.get('captionLabel')` will return `undefined`.

PR #3272, opened 2019-01-29
Enable captions on autostart muted with config

pajong:
### This PR will...
- Add new boolean config `enableDefaultCaptions` which when set to true:
-- Enable captions by default if mute is true on initial start
-- Choose the first available caption
- Caption should still be enabled if the first playlist item does not have captions, and the second playlist item does
-- Even if unmuted before the start of the second playlist item
- As soon as the user sets the captions to `off`, this feature does not have an effect for following playlist items

- A/B testing

Closed the previous PR: https://github.com/jwplayer/jwplayer/pull/3269/files

#### Addresses Issue(s):
JW8-2641

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -131,9 +131,8 @@ const Captions = function(_model) {
             }
         }
         // set the index without the side effect of storing the Off label in _selectCaptions
-        if (_model.get('enableDefaultCaptions') && _model.get('autostart')) {
-            _setCurrentIndex(1);
-            return;
+        if (_model.get('enableDefaultCaptions') && _model.get('autostart') && !captionsMenuIndex) {
+            captionsMenuIndex = 1;
         }
         _setCurrentIndex(captionsMenuIndex);
     }
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -131,6 +131,10 @@ const Captions = function(_model) {
             }
         }
         // set the index without the side effect of storing the Off label in _selectCaptions
+        if (_model.get('enableDefaultCaptions') && _model.get('autostart')) {
+            _setCurrentIndex(1);
+            return;
+        }
         _setCurrentIndex(captionsMenuIndex);
     }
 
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/captions.js
```diff
@@ -131,6 +131,10 @@ const Captions = function(_model) {
             }
         }
         // set the index without the side effect of storing the Off label in _selectCaptions
+        if (_model.get('enableDefaultCaptions') && _model.get('autostart')) {
+            _setCurrentIndex(1);
```
johnBartos: Set `captionsMenuIndex = 1` and delete the return

src/js/controller/captions.js
```diff
@@ -131,6 +131,10 @@ const Captions = function(_model) {
             }
         }
         // set the index without the side effect of storing the Off label in _selectCaptions
+        if (_model.get('enableDefaultCaptions') && _model.get('autostart')) {
```
johnBartos: I think this is fine as long as we're fine with override a viewer's language choice when the current video does not have captions of that choice available

src/js/controller/controller.js
```diff
@@ -511,6 +511,10 @@ Object.assign(Controller.prototype, {
                     });
                 }
 
+                if (!_this.getMute()) {
```
johnBartos: You can shift this conditional into the above block to avoid calling `getMute` twice

johnBartos:
Tested with the following config on Chrome:

```
{
    playlist: [
        {
              "file": "//playertest.longtailvideo.com/adaptive/paradise/paradise-novtt.m3u8"
        },
        {
                "file": "http://content.bitsontherun.com/videos/bkaovAYt-52qL9xLP.mp4"
        }
    ],
  "mute": true,
  "autostart": true,
  enableDefaultCaptions: true
}
```

PR #3273, opened 2019-01-30
[JW8-2648] Exempt jw-plugin-googima from jw-plugin styling

waxidiotic:
### This PR will...

Exempt elements with both `jw-plugin` and `jw-plugin-googima` classes from the `jw-plugin`-specific CSS rules.

Since we now set `opacity: 1` on IMA ads earlier (https://github.com/jwplayer/jwplayer-ads-googima/pull/453), the ad creative was appearing higher in the player very quickly before repositioning to the expected position.

This was caused by:
`.jw-plugin {
  position: absolute;
  bottom: (@mobile-touch-target * 1.5);
}`

n/a

n/a

#### Addresses Issue(s):

JW8-2648

### DIFF
--- a/src/css/jwplayer/imports/jwplayerlayout.less
+++ b/src/css/jwplayer/imports/jwplayerlayout.less
@@ -61,8 +61,10 @@
 }
 
 .jw-plugin {
-    position: absolute;
-    bottom: (@mobile-touch-target * 1.5);
+    &:not(.jw-plugin-googima) {
+        position: absolute;
+        bottom: (@mobile-touch-target * 1.5);
+    }
 
     .jw-banner {
         max-width: 100%;


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/css/jwplayer/imports/jwplayerlayout.less
```diff
@@ -61,8 +61,10 @@
 }
 
 .jw-plugin {
-    position: absolute;
-    bottom: (@mobile-touch-target * 1.5);
+    &:not(.jw-plugin-googima) {
+        position: absolute;
+        bottom: (@mobile-touch-target * 1.5);
+    }
```
robwalch: Rather than this `not` rule here for an element not in this project, add a new position or bottom rule in https://github.com/jwplayer/jwplayer-ads-googima/blob/master/src/css/styles.less

PR #3280, opened 2019-02-06
[JW8-2595] Return in getElementWindow() if element is null

waxidiotic:
JW8-2595

### This PR will...

Return `window` if `getElementWindow(element)` (in `ui.js`) if `element` is falsy.

* In IE 11 and Edge, `getElementWindow(element)` was being called when `element === null`.
* This appears to be caused by when IE11 and Edge fire pointer events compared to other browsers. In IE 11/Edge, `this.destroy()` in `dismissButton.js` is setting the element to `null` before `interactEndHandler` is bring run.
* This caused a JavaScript exception in those environments.
* Functionality was not affected but we should avoid JavaScript exceptions when possible.
* The stack trace went back to minified jQuery code so I couldn't find the exact cause.

n/a

n/a

#### Addresses Issue(s):

JW8-2595

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -159,6 +159,9 @@ function initInteractionListeners(ui) {
 
     const interactEndHandler = (e) => {
         clearTimeout(longPressTimeout);
+        if (!ui.el) {
+            return;
+        }
         releasePointerCapture(ui);
         removeHandlers(ui, WINDOW_GROUP);
         if (ui.dragged) {
@@ -348,10 +351,6 @@ const eventRegisters = {
 };
 
 export function getElementWindow(element) {
-    if (!element) {
-        return window;
-    }
-
     const document = element.ownerDocument || element;
     return (document.defaultView || document.parentWindow || window);
 }
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -348,6 +348,10 @@ const eventRegisters = {
 };
 
 export function getElementWindow(element) {
+    if (!element) {
+        return;
+    }
+
     const document = element.ownerDocument || element;
     return (document.defaultView || document.parentWindow || window);
 }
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/utils/ui.js
```diff
@@ -348,6 +348,10 @@ const eventRegisters = {
 };
 
 export function getElementWindow(element) {
+    if (!element) {
+        return;
```
robwalch: If this does not return anything then there will be another exception with the code that uses the `undefined` result. Please see what where we are in the debug code after this point, and don't use the harness - use the events page (no jQuery).

Once you've finished investigating, `return window` here.

waxidiotic: @robwalch Thanks for the suggestion on using the event page. I was able to pinpoint the issue to being when IE11 and Edge fire pointer events. In those browsers, the `destroy()` method in `dismissButton.js` (in commercial) is setting the element to `null` before the `interactEndHandler` can run. So by the time it does run, the element is `null`.

I updated to make sure `ui.js` returns `window`.

robwalch: This sounds like a bug in `ui.off`. If we call `removeHandlers(this, WINDOW_GROUP)` in `destroy() => this.off()` then `interactEndHandler` should not get triggered (or is it already running?)

```js
    off(name, callback, context) {
        if (!name) {
            removeHandlers(this, WINDOW_GROUP);
            const { handlers } = this;
            Object.keys(handlers).forEach(triggerName => {
                removeHandlers(this, triggerName);
            });
        } else if (eventsApi(name)) {
            removeHandlers(this, name);
        }
        return super.off(name, callback, context);
    }
```

PR #3281, opened 2019-02-07
JW8-2183 Audio mode  horizontal volume slider

pajong:
### This PR will...
- Add horizontal volume slider to control bar only player
- Rename `Tooltip` `VolumeTooltip` `Timetip` to `TooltipIcon` `VolumeTooltipIcon` `TimetipIcon`, since they are creating an icon that has a tooltip on hover. Instead of `container`, those classes now contain `tooltip` which makes more sense.

#### Addresses Issue(s):

JW8-2183

### DIFF
--- a/src/css/controls/flags/audioplayer.less
+++ b/src/css/controls/flags/audioplayer.less
@@ -88,18 +88,18 @@
             }
 
             &.jw-open {
-                transition: width 300ms @default-timing-function;
                 width: @volume-rail-length;
 
                 .jw-slider-volume {
                     padding: 0 12px;
+                    transition: opacity 300ms;
                     opacity: 1;
                 }
 
                 ~ .jw-slider-time {
                     flex: 1 1 auto;
                     width: auto;
-                    transition: opacity 1000ms;
+                    transition: opacity 300ms, width 300ms;
                 }
             }
         }
--- a/src/css/controls/flags/audioplayer.less
+++ b/src/css/controls/flags/audioplayer.less
@@ -80,6 +80,7 @@
             height: 17px;
             align-items: center;
             width: 0;
+            outline: none;
 
             &.jw-tab-focus:focus {
                 outline: @ui-focus;
--- a/src/css/controls/flags/audioplayer.less
+++ b/src/css/controls/flags/audioplayer.less
@@ -77,9 +77,10 @@
         }
 
         .jw-horizontal-volume-container {
+            transition: width 300ms @default-timing-function;
+            width: 0;
             height: 17px;
             align-items: center;
-            width: 0;
             outline: none;
 
             &.jw-tab-focus:focus {
@@ -87,7 +88,7 @@
             }
 
             &.jw-open {
-                transition: width 1000ms @default-timing-function;
+                transition: width 300ms @default-timing-function;
                 width: @volume-rail-length;
 
                 .jw-slider-volume {
--- a/src/css/controls/flags/audioplayer.less
+++ b/src/css/controls/flags/audioplayer.less
@@ -16,6 +16,10 @@
         min-height: @mobile-touch-target;
     }
 
+    .jw-spacer {
+        display: none;
+    }
+
     .jw-preview, // overrides 'block' from .jw-state-idle
     .jw-display, // overrides 'block' from .jw-flag-touch.jw-state-paused
     .jw-title, // overrides 'block' from .jw-state-idle
@@ -63,14 +67,27 @@
             order: 1;
         }
 
+        .jw-icon-volume {
+            margin-right: 0;
+            transition: margin-right @default-transition;
+
+            .jw-overlay {
+                display: none;
+            }
+        }
+
         .jw-horizontal-volume-container {
             height: 17px;
             align-items: center;
             width: 0;
 
+            &.jw-tab-focus:focus {
+                outline: @ui-focus;
+            }
+
             &.jw-open {
                 transition: width 1000ms @default-timing-function;
-                width: 100%;
+                width: @volume-rail-length;
 
                 .jw-slider-volume {
                     padding: 0 12px;
@@ -80,13 +97,12 @@
                 ~ .jw-slider-time {
                     flex: 1 1 auto;
                     width: auto;
-                    transition: opacity 1000ms @default-timing-function;
+                    transition: opacity 1000ms;
                 }
             }
         }
 
         .jw-slider-volume {
-            width: 100%;
             opacity: 0;
 
             .jw-knob {
@@ -96,6 +112,13 @@
             ~ .jw-icon-volume {
                 margin-right: @volume-rail-length;
             }
+        }
+    }
+
+    &.jw-breakpoint-1,
+    &.jw-breakpoint-2 {
+        .jw-horizontal-volume-container.jw-open ~ .jw-slider-time {
+            opacity: 0;
         }
     }
 
--- a/src/css/controls/flags/audioplayer.less
+++ b/src/css/controls/flags/audioplayer.less
@@ -37,7 +37,7 @@
             display: none;
         }
 
-        .jw-icon-volume:not(.jw-icon-tooltip),
+        .jw-icon-volume,
         .jw-icon-playback,
         .jw-icon-next,
         .jw-icon-rewind,
@@ -57,9 +57,45 @@
         }
 
         .jw-slider-time {
-            display: flex;
             flex: 0 1 auto;
             align-items: center;
+            display: flex;
+            order: 1;
+        }
+
+        .jw-horizontal-volume-container {
+            height: 17px;
+            align-items: center;
+            width: 0;
+
+            &.jw-open {
+                transition: width 1000ms @default-timing-function;
+                width: 100%;
+
+                .jw-slider-volume {
+                    padding: 0 12px;
+                    opacity: 1;
+                }
+
+                ~ .jw-slider-time {
+                    flex: 1 1 auto;
+                    width: auto;
+                    transition: opacity 1000ms @default-timing-function;
+                }
+            }
+        }
+
+        .jw-slider-volume {
+            width: 100%;
+            opacity: 0;
+
+            .jw-knob {
+                transform: translate(-50%, -50%);
+            }
+
+            ~ .jw-icon-volume {
+                margin-right: @volume-rail-length;
+            }
         }
     }
 
--- a/src/js/view/controls/components/tooltipicon.js
+++ b/src/js/view/controls/components/tooltipicon.js
@@ -3,8 +3,8 @@ import ariaLabel from 'utils/aria';
 import { toggleClass } from 'utils/dom';
 import svgParse from 'utils/svgParser';
 
-export default class Tooltip {
-    constructor(name, ariaText, elementShown, svgIcons) {
+export default class TooltipIcon {
+    constructor(name, ariaText, elementShown, svgIcons, container) {
         Object.assign(this, Events);
         this.el = document.createElement('div');
         let className = 'jw-icon jw-icon-tooltip ' + name + ' jw-button-color jw-reset';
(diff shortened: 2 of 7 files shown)


src/js/view/controls/controlbar.js
```diff
@@ -137,8 +138,12 @@ export default class Controlbar {
 
         // Do not initialize volume slider or tooltip on mobile
         if (!this._isMobile) {
-            volumeGroup = new VolumeTooltip(_model, 'jw-icon-volume', vol,
-                cloneIcons('volume-0,volume-50,volume-100'));
+            if (_model.get('audioMode')) {
+                volumeContainer = document.createElement('div');
```
pajong: This creates a `div` for the horizontal volume slider to go in. Passing this `div` into `VolumeTooltipIcon` puts the tooltip for the icon into the `div`.

johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/components/volumetooltipicon.js
```diff
@@ -1,22 +1,25 @@
...
+        super(name, ariaText, true, svgIcons, container);
 
         const localization = _model.get('localization');
+        const audioMode = _model.get('audioMode');
+        const orientation = audioMode ? 'horizontal' : 'vertical';
 
         this._model = _model;
-        this.volumeSlider = new Slider('jw-slider-volume jw-volume-tip', 'vertical');
+        this.volumeSlider = new Slider('jw-slider-volume jw-volume-tip', orientation);
+        //toggleClass(this.volumeSlider.element(), 'jw-volume-tip', !audioMode);
```
karimMourra: can we remove this ?

src/js/view/controls/components/tooltipicon.js
```diff
@@ -61,6 +61,7 @@ export default class Tooltip {
             this.trigger('open-' + this.componentType, evt, { isOpen: true });
             this.isOpen = true;
             toggleClass(this.el, this.openClass, this.isOpen);
+            toggleClass(this.tooltip, this.openClass, this.isOpen);
```
karimMourra: why does the tooltip need the open class now ? in the css we only check for `jw-horizontal-volume-container.jw-open`

src/js/view/controls/components/volumetooltipicon.js
```diff
@@ -0,0 +1,85 @@
...
+        setAttribute(styleElement, 'aria-orientation', orientation);
+        setAttribute(styleElement, 'aria-valuemin', 0);
+        setAttribute(styleElement, 'aria-valuemax', 100);
+        setAttribute(styleElement, 'role', 'slider');
+
+        // Apply a click interaction listener to help with focus styling
+        const { openTooltip, closeTooltip } = callbacks;
+        this.uiOver = new UI(styleElement)
+            .on('click', function() {}, callbacks)
+            .on('over', openTooltip, callbacks)
```
karimMourra: seems odd to pass in tooltip functions (`openTooltip` and closeTooltip). Since slider is eventable, could we trigger events ?

src/js/view/controls/components/volumetooltipicon.js
```diff
@@ -0,0 +1,85 @@
...
+export default class VolumeTooltipIcon extends TooltipIcon {
+    constructor(_model, name, ariaText, svgIcons, horizontalContainer) {
+        super(name, ariaText, true, svgIcons);
+
+        this._model = _model;
+        this.horizontalContainer = horizontalContainer;
+
+        const volumeLabel = _model.get('localization').volumeSlider;
+        this.horizontalSlider = new VolumeSlider('horizontal', volumeLabel, horizontalContainer, this);
+        this.volumeSlider = new VolumeSlider('vertical', volumeLabel, this.tooltip, this);
```
karimMourra: let's rename to verticalSlider for clearer naming

src/js/view/controls/components/volumetooltipicon.js
karimMourra: it would also decouple the volumeSlider from `VolumeTooltipIcon`

src/js/view/controls/components/volumetooltipicon.js
```diff
@@ -0,0 +1,85 @@
...
+            .on('out', closeTooltip, this)
+            .on('focus', openTooltip, this)
+            .on('blur', closeTooltip, this);
+
+        this._model.on('change:volume', this.onVolume, this);
+    }
+
+    openTooltip(evt) {
+        super.openTooltip(evt);
+        toggleClass(this.horizontalContainer, this.openClass, this.isOpen);
```
karimMourra: Is this coupled because volumeSlider calls openTooltip? If so we should listen for events coming from the slider

karimMourra:
@radium-v could we get your blesssing on the css ?

radium-v:
@karimMourra The horizontal slider still gets focus even when it's not visible; try tabbing through the kitchen sink and you'll see after highlighting the volume slider, the focus ring disappears but it's on the invisible horizontal slider.

pajong:
@radium-v Would you expect it to be unfocused?

radium-v:
> @radium-v Would you expect it to be unfocused?

If it's not intended to be visible in certain views then it shouldn't be focusable either. Not sure why both a vertical and horizontal slider are present when the same object can be used while swapping DOM order and a few classes (which would be rare since a player doesn't typically go from controlbar-only to regular view).

pajong:
> > @radium-v Would you expect it to be unfocused?
> 
> If it's not intended to be visible in certain views then it shouldn't be focusable either. Not sure why both a vertical and horizontal slider are present when the same object can be used while swapping DOM order and a few classes (which would be rare since a player doesn't typically go from controlbar-only to regular view).

It doesn't typically go from controlbar-only to regular, but it is easy to go the other way around. Resizing to height of 30 changes the mode to controlbar-only.
We thought that changing the DOM structure will be a lot more expensive, especially when the mode can change with just a resize. Please let us know if you want to discuss about this.

We are going to remove the focus on the horizontal slider when it closes.

src/js/view/controls/components/volumetooltipicon.js
pajong: This opens up the horizontal slider when the icon is focused/hovered.

karimMourra: i think we should use the specific value instead of `this.isOpen` since that flag is handled by the tooltip. If tooltip logic ever changes, we'll have issues here. 
`toggleClass(this.horizontalContainer, this.openClass, true);`

src/js/view/controls/components/volumetooltipicon.js
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    openSlider(evt) {
+        super.openTooltip(evt);
+        toggleClass(this.horizontalContainer, this.openClass, this.isOpen);
+    }
+
+    closeSlider(evt) {
+        super.closeTooltip(evt);
+        toggleClass(this.horizontalContainer, this.openClass, this.isOpen);
```
karimMourra: `toggleClass(this.horizontalContainer, this.openClass, false);`

pajong:
> i think we can use dot notation now: `model.audioMode`

I don't think this is true

pajong:
test this please

karimMourra:
@pajong when tabbing from the volume button to the horizontal bar, the bar closes and opens again. I think it should stay open. @aliciahurst do you agree?

aliciahurst:
@karimMourra @pajong, feedback:
1. The volume slider should use the same smooth transition to close as it did to open
2. I agree. When you tab from 
3. I think the transition open and closed should be faster. Currently it's set to 1000, let's try it with 300. Karim and I tested this in the browser and it worked well.

Thanks!!

src/css/controls/flags/audioplayer.less
```diff
@@ -77,17 +77,18 @@
         }
 
         .jw-horizontal-volume-container {
+            transition: width 300ms @default-timing-function;
```
karimMourra: @pajong time slider should also get this; it doesn't close with a transition

pajong: I applied transition to the horizontal volume container when closing, so it has the same effect. Let me know if I am wrong.

karimMourra: what i meant is that the time slider closes instantaneously, instead of closing with a transition. The transitions for time and volume sliders should be the same

src/js/view/controls/controlbar.js
```diff
@@ -121,13 +121,14 @@ export default class Controlbar {
...
         let volumeGroup;
+        let horizontalVolumeContainer;
         let muteButton;
         let feedShownId = '';
 
         const vol = localization.volume;
 
         // Do not show the volume toggle in the mobile SDKs or <iOS10
-        if (!_model.get('sdkplatform') && !(OS.iOS && OS.version.major < 10)) {
+        if (this._isMobile && !_model.get('sdkplatform') && !(OS.iOS && OS.version.major < 10)) {
```
robwalch: This block should go in an `else if` mobile:
```js
if (!this._isMobile) {
} else if (!_model.get('sdkplatform') && !(OS.iOS && OS.version.major < 10)) {
```

src/js/view/controls/controls.js
```diff
@@ -269,12 +272,16 @@ export default class Controls {
                     api.playToggle(reasonInteraction());
                     break;
                 case 37: // left-arrow, if not adMode and settings menu is hidden
-                    if (!adMode && menuHidden) {
```
robwalch: Is this a pattern we use in `controls`? Why not intercept in the slider element to handle custom keyboard behavior? adding it to `ui` in the slider will deal with focus automatically and you don't have to add new code and queries here.

src/js/view/controls/components/volumetooltipicon.js
```diff
@@ -0,0 +1,113 @@
...
+
+        this.horizontalSlider.on('update', function (evt) {
+            this.trigger('update', evt);
+        }, this);
+
+        this.horizontalSlider.uiOver.on('keydown', (evt) => {
+            const event = evt.sourceEvent;
+            switch (event.keyCode) {
+                case 37:
+                    event.stopPropagation();
```
pajong: Discuss

src/css/controls/flags/audioplayer.less
```diff
@@ -57,9 +61,66 @@
...
+            margin-right: 0;
+            transition: margin-right @default-transition;
+
+            .jw-overlay {
+                display: none;
+            }
+        }
+
+        .jw-horizontal-volume-container {
+            transition: width 300ms @default-timing-function;
```
robwalch: Can we use `transform` with `scaleX(0.0-1.0)` instead of `width`?

src/js/view/controls/components/volumetooltipicon.js
robwalch: Why? This way is good. It's this or add more logic to the controls which would then need deep access to this module.

src/css/controls/flags/audioplayer.less
```diff
@@ -16,6 +16,10 @@
         min-height: @mobile-touch-target;
     }
 
+    .jw-spacer {
+        display: none;
+    }
```
robwalch: Why would you do this? Do you want all the controls to be centered in the player?

screen shot 2019-02-27 at 3 34 24 pm

src/css/controls/flags/audioplayer.less
pajong: This pushes the other elements when the horizontal volume is not open.

PR #3282, opened 2019-02-11
Autopause Refinements

waxidiotic:
### This PR will...

* Decouple `autoPause` from `autostart`. Now, autostarting will not be required to autoPause.
* If `autoPause.viewability` is not set in the config, it will default to `true`. This means autoPause can be set up with an empty object.
* Add logic to controller and have it check each time the player's viewability change. If `autoPause` is enabled and the player goes out of view, the autoPause functionality will take affect.
* Ensure ads will continue playing when scrolling out of view. If autoPause is enabled and player is still out of view when the ad break ends, the player will then pause.
* Added `'viewability'` as a `pauseReason` and `playReason` and set it if paused/resumed due to autoPause.

* Publishers want to be able to automatically pause when not in view to save on streaming costs and ensure higher engagement.
* Publishers do not want ads to pause when scrolling out of view due to its affect on ad complete rates.

There is no impact on anything (i.e. analytics) for changing the `pauseReason` and `playReason` reasons, right?

https://github.com/jwplayer/jwplayer-commercial/pull/6287

#### Addresses Issue(s):

JW8-2531

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -84,6 +84,13 @@ const Config = function(options, persisted) {
     config.language = language;
     config.intl = intl;
 
+    // If autoPause is configured with an empty block,
+    // default autoPause.viewability to true.
+    let autoPause = config.autoPause;
+    if (autoPause && !autoPause.viewability) {
+        autoPause.viewability = true;
+    }
+
     let rateControls = config.playbackRateControls;
 
     if (rateControls) {
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -87,8 +87,8 @@ const Config = function(options, persisted) {
 
     // If autoPause is configured with an empty block,
     // default autoPause.viewability to true.
-    let autoPause = allOptions.autoPause || {};
-    config.autoPause.viewability = ('viewability' in autoPause) ? !!autoPause.viewability : true;
+    let autoPause = allOptions.autoPause;
+    config.autoPause.viewability = (autoPause && ('viewability' in autoPause)) ? !!autoPause.viewability : true;
 
     let rateControls = config.playbackRateControls;
 
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -88,7 +88,9 @@ const Config = function(options, persisted) {
     // If autoPause is configured with an empty block,
     // default autoPause.viewability to true.
     let autoPause = allOptions.autoPause;
-    config.autoPause.viewability = (autoPause && ('viewability' in autoPause)) ? !!autoPause.viewability : true;
+    if (autoPause) {
+        config.autoPause.viewability = ('viewability' in autoPause) ? !!autoPause.viewability : true;
+    }
 
     let rateControls = config.playbackRateControls;
 
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -87,7 +87,7 @@ const Config = function(options, persisted) {
     // If autoPause is configured with an empty block,
     // default autoPause.viewability to true.
     let autoPause = config.autoPause;
-    if (autoPause && !autoPause.viewability) {
+    if (autoPause && !autoPause.hasOwnProperty('viewability')) {
         autoPause.viewability = true;
     }
 
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -10,6 +10,7 @@ import { getLanguage, getCustomLocalization, applyTranslation, normalizeIntl } f
 // Defaults
 // Alphabetical order
 const Defaults = {
+    autoPause: { viewability: false },
     autostart: false,
     bandwidthEstimate: null,
     bitrateSelection: null,
@@ -87,9 +88,7 @@ const Config = function(options, persisted) {
     // If autoPause is configured with an empty block,
     // default autoPause.viewability to true.
     let autoPause = config.autoPause;
-    if (autoPause && !autoPause.hasOwnProperty('viewability')) {
-        autoPause.viewability = true;
-    }
+    autoPause.viewability = ('viewability' in autoPause) ? !!autoPause.viewability : true;
 
     let rateControls = config.playbackRateControls;
 
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -87,8 +87,8 @@ const Config = function(options, persisted) {
 
     // If autoPause is configured with an empty block,
     // default autoPause.viewability to true.
-    let autoPause = config.autoPause;
-    autoPause.viewability = ('viewability' in autoPause) ? !!autoPause.viewability : true;
+    let autoPause = allOptions.autoPause || {};
+    config.autoPause.viewability = ('viewability' in autoPause) ? !!autoPause.viewability : true;
 
     let rateControls = config.playbackRateControls;
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -298,7 +298,7 @@ Object.assign(Controller.prototype, {
 
             // Autostart immediately if we're not waiting for the player to become viewable first.
             if (_model.get('autostart') === true && !_model.get('playOnViewable')) {
-                _autoStart({ reason: 'autostart' });
+                _autoStart('autostart');
             }
             apiQueue.flush();
         }
@@ -338,7 +338,7 @@ Object.assign(Controller.prototype, {
         function _checkPlayOnViewable(model, viewable) {
             if (model.get('playOnViewable')) {
                 if (viewable) {
-                    _autoStart({ reason: 'viewability' });
+                    _autoStart('viewability');
                 } else if (OS.mobile) {
                     _this.pause({ reason: 'autostart' });
                 }
@@ -352,7 +352,7 @@ Object.assign(Controller.prototype, {
 
             const adState = _getAdState();
             if (!adState) {
-                _this.pause({ reason: 'viewability' });
+                _this.pause({ reason: 'viewable' });
             } else if (model.get('activeTab') && adState !== 'paused') {
                 const instream = _this._instreamAdapter;
                 if (instream) {
@@ -534,7 +534,7 @@ Object.assign(Controller.prototype, {
                     _model.set('enableDefaultCaptions', false);
                 }
 
-                return _play(reason).catch(() => {
+                return _play({ reason }).catch(() => {
                     if (!_this._instreamAdapter) {
                         _model.set('autostartFailed', true);
                     }
@@ -593,7 +593,7 @@ Object.assign(Controller.prototype, {
             if (state !== STATE_PAUSED && state !== STATE_IDLE) {
                 const pauseReason = _getReason(meta);
                 _model.set('pauseReason', pauseReason);
-                _model.set('playOnViewable', (pauseReason === 'viewability'));
+                _model.set('playOnViewable', (pauseReason === 'viewable'));
             }
 
             const adState = _getAdState();
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -353,12 +353,12 @@ Object.assign(Controller.prototype, {
             const adState = _getAdState();
             if (!adState) {
                 _this.pause({ reason: 'viewability' });
-            } else if (adState !== 'paused') {
+            } else if (model.get('activeTab') && adState !== 'paused') {
                 const instream = _this._instreamAdapter;
                 if (instream) {
                     instream.noResume = true;
-                    model.set('playOnViewable', true);
                 }
+                model.set('playOnViewable', true);
             }
         }
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -269,7 +269,6 @@ Object.assign(Controller.prototype, {
 
             _model.change('viewable', viewableChange);
             _model.change('viewable', _checkPlayOnViewable);
-            _model.change('viewable', _checkPauseOnViewable);
             _model.once('change:autostartFailed change:mute', function(model) {
                 model.off('change:viewable', _checkPlayOnViewable);
             });
@@ -310,6 +309,8 @@ Object.assign(Controller.prototype, {
             });
 
             preload();
+
+            _checkPauseOnViewable(model, viewable);
         }
 
         function preload() {
@@ -348,12 +349,11 @@ Object.assign(Controller.prototype, {
             _model.set('pauseOnViewable', true);
         }
 
-        function _checkPauseOnViewable() {
+        function _checkPauseOnViewable(model, viewable) {
             const adState = _getAdState();
-            const viewable = _model.get('viewable');
-            const isPaused = _model.get('state') === 'paused';
+            const isPaused = model.get('state') === 'paused';
 
-            if (_model.get('pauseOnViewable') && !viewable && !adState && !isPaused) {
+            if (!viewable && model.get('pauseOnViewable') && !adState && !isPaused) {
                 _this.pause({ reason: 'viewability' });
             }
         }
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -351,9 +351,7 @@ Object.assign(Controller.prototype, {
 
         function _checkPauseOnViewable(model, viewable) {
             const adState = _getAdState();
-            const isPaused = model.get('state') === STATE_PAUSED;
-
-            if (!viewable && model.get('pauseOnViewable') && !adState && !isPaused) {
+            if (!viewable && model.get('pauseOnViewable') && !adState) {
                 _this.pause({ reason: 'viewability' });
             }
         }
@@ -530,12 +528,6 @@ Object.assign(Controller.prototype, {
                     _model.set('enableDefaultCaptions', false);
                 }
 
-                // Enable autoPause behavior.
-                const autoPause = _model.get('autoPause');
-                if (autoPause && autoPause.viewability) {
-                    _model.set('playOnViewable', true);
-                }
-
                 return _play(reason).catch(() => {
                     if (!_this._instreamAdapter) {
                         _model.set('autostartFailed', true);
@@ -591,9 +583,12 @@ Object.assign(Controller.prototype, {
             _actionOnAttach = null;
             checkAutoStartCancelable.cancel();
 
-            const pauseReason = _getReason(meta);
-            _model.set('pauseReason', pauseReason);
-            _model.set('playOnViewable', (pauseReason === 'viewability'));
+            const state = _getState();
+            if (state !== STATE_PAUSED && state !== STATE_IDLE) {
+                const pauseReason = _getReason(meta);
+                _model.set('pauseReason', pauseReason);
+                _model.set('playOnViewable', (pauseReason === 'viewability'));
+            }
 
             const adState = _getAdState();
             if (adState) {
@@ -601,7 +596,7 @@ Object.assign(Controller.prototype, {
                 return;
             }
 
-            switch (_model.get('state')) {
+            switch (state) {
                 case STATE_ERROR:
                     return;
                 case STATE_PLAYING:
@@ -617,7 +612,7 @@ Object.assign(Controller.prototype, {
         }
 
         function _isIdle() {
-            const state = _model.get('state');
+            const state = _getState();
             return (state === STATE_IDLE || state === STATE_COMPLETE || state === STATE_ERROR);
         }
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -269,6 +269,7 @@ Object.assign(Controller.prototype, {
 
             _model.change('viewable', viewableChange);
             _model.change('viewable', _checkPlayOnViewable);
+            _model.change('viewable', _checkPauseOnViewable);
             _model.once('change:autostartFailed change:mute', function(model) {
                 model.off('change:viewable', _checkPlayOnViewable);
             });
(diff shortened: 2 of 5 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

waxidiotic:
test this please

waxidiotic:
test this please

src/js/controller/controller.js
```diff
@@ -337,12 +338,26 @@ Object.assign(Controller.prototype, {
             if (model.get('playOnViewable')) {
                 if (viewable) {
                     _autoStart();
-                } else if (OS.mobile || (model.get('autoPause') && model.get('autoPause').viewability)) {
+                } else if (OS.mobile) {
                     _this.pause({ reason: 'autostart' });
```
robwalch: Do we really want to keep this mobile behavior?

src/js/controller/controller.js
```diff
@@ -337,12 +338,26 @@ Object.assign(Controller.prototype, {
...
                 }
             }
         }
 
+        if (_model.get('autoPause') && _model.get('autoPause').viewability) {
+            _model.set('pauseOnViewable', true);
```
robwalch: Why don't you do this in config.js so that it gets copied to the model? Do you ever set `pauseOnViewable` to false? Why don't you always add `autoPause` to the model, but set `viewability` to false when it's not set in the config. Then you can use `_model.autoPause.viewability` everywhere without additional `&&` checks.

src/js/controller/controller.js
```diff
@@ -337,12 +338,26 @@ Object.assign(Controller.prototype, {
...
+        }
+
+        function _checkPauseOnViewable() {
```
robwalch: `viewable` is passed in by the change callback. There is no reason to `get` it again:
`_checkPauseOnViewable(model, viewable) {`

src/js/controller/controller.js
```diff
@@ -337,12 +338,26 @@ Object.assign(Controller.prototype, {
...
+            const adState = _getAdState();
+            const viewable = _model.get('viewable');
+            const isPaused = _model.get('state') === 'paused';
+
+            if (_model.get('pauseOnViewable') && !viewable && !adState && !isPaused) {
```
robwalch: Written this way your code will always do 4 model `get` calls. Write your logic so that the browser doesn't have to do more work when the logic is short-circuited:

```js
if (!viewable && _model.get('pauseOnViewable') && !_getAdState()) {
```

Also why bother checking for the paused state? You can call pause when already paused. What's going to happen? The only change would be to "pauseReason" on the model.

src/js/controller/controller.js
```diff
@@ -577,7 +592,9 @@ Object.assign(Controller.prototype, {
             _actionOnAttach = null;
             checkAutoStartCancelable.cancel();
 
-            _model.set('pauseReason', _getReason(meta));
+            const pauseReason = _getReason(meta);
+            _model.set('pauseReason', pauseReason);
+            _model.set('playOnViewable', (pauseReason === 'viewability'));
```
robwalch: So the only reason you added "playOnViewable" and "pauseOnViewable" model properties was to access them in instream? Instream has access to the controller. You should not use the model for things that (1)  you don't need to share throught the application (2) you can access as properties from other objects (3) you don't need "change" listeners for.

src/js/program/program-controller.js
```diff
@@ -266,7 +266,10 @@ class ProgramController extends Events {
             return;
         }
 
-        backgroundMediaController.mediaModel.attributes.mediaState = 'buffering';
+        const autoPause = model.get('autoPause');
+
+        // If background loading and using autoPause, we want to show the paused state after ads.
+        backgroundMediaController.mediaModel.attributes.mediaState = autoPause ? 'paused' : 'buffering';
```
robwalch: We do? I thought we nixed that idea.

If we so then just set `instream.noResume` to `true` and get rid of your changes in instream.

src/js/api/config.js
```diff
@@ -84,6 +84,13 @@ const Config = function(options, persisted) {
     config.language = language;
     config.intl = intl;
 
+    // If autoPause is configured with an empty block,
+    // default autoPause.viewability to true.
+    let autoPause = config.autoPause;
+    if (autoPause && !autoPause.viewability) {
```
vseventer: This doesn't work if I set-up with `autoPause: { viewability: false }`, as this block changes it to `true`.

src/js/controller/controller.js
```diff
@@ -337,12 +338,26 @@ Object.assign(Controller.prototype, {
...
         }
 
+        if (_model.get('autoPause') && _model.get('autoPause').viewability) {
+            _model.set('pauseOnViewable', true);
+        }
+
+        function _checkPauseOnViewable() {
+            const adState = _getAdState();
+            const viewable = _model.get('viewable');
+            const isPaused = _model.get('state') === 'paused';
```
vseventer: In general, all states can be imported as constants (e.g. `STATE_PAUSED`).

src/js/controller/controller.js
```diff
@@ -577,7 +592,9 @@ Object.assign(Controller.prototype, {
             _actionOnAttach = null;
             checkAutoStartCancelable.cancel();
 
-            _model.set('pauseReason', _getReason(meta));
+            const pauseReason = _getReason(meta);
```
vseventer: What about `playReason` when playback is resumed b/c autoPause?

src/js/controller/controller.js
waxidiotic: This is how we pause and resume on mobile web when not viewable.

src/js/controller/controller.js
waxidiotic: We have to check if we're paused because if the user pauses by interaction and then scrolls out of view, the `pauseReason` gets overridden to `viewability` so when the user scrolls the player back into view, it resumes which we dont want.

src/js/program/program-controller.js
waxidiotic: The whole reason for the change in here is because the content would be paused after the ad but the player would be in a buffering state.

src/js/controller/controller.js
waxidiotic: `playOnViewable` already existed on the model before these changes. I originally tried just using `playOnViewable` to implement the autoPause functionality but it would not work properly and involved a lot more logic than would be needed by just adding another property on the model.

I'm also setting it on the model because I believe we do need to share this throughout other components. Please see my replies to your feedback in the program-controller and instream-adapter files.

src/js/controller/controller.js
waxidiotic: 👍

src/js/controller/controller.js
```diff
@@ -269,6 +269,7 @@ Object.assign(Controller.prototype, {
 
             _model.change('viewable', viewableChange);
             _model.change('viewable', _checkPlayOnViewable);
+            _model.change('viewable', _checkPauseOnViewable);
```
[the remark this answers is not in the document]
waxidiotic: 👍

src/js/controller/controller.js
waxidiotic: We don't currently have access to the model inside of config.js so I didn't want to add it unnecessarily. 

As for setting `viewability` to false - I'm confused. It should be defaulting to `true` if it's not set in the config. That's part of the AC.

src/js/api/config.js
waxidiotic: 👍

src/js/controller/controller.js
waxidiotic: 👍

src/js/controller/controller.js
waxidiotic: 👍

src/js/controller/controller.js
```diff
@@ -482,7 +497,7 @@ Object.assign(Controller.prototype, {
             return 'external';
         }
 
-        function _autoStart() {
+        function _autoStart(autostartReason) {
```
vseventer: Would be nicer to name the argument `reason`.

waxidiotic: Wanted to, but it already exists in an inner scope.

vseventer: That refers to the `reason` in the `catch` block within `_autoStart`, which can be refactored to:
```
if (!checkAutoStartCancelable.cancelled()) {
    const code = getPlayAttemptFailedErrorCode(error);
    _this.trigger(AUTOSTART_NOT_ALLOWED, {
        reason: error.reason,
        code,
        error
    });
}
```

waxidiotic: Cool. Done.

src/js/controller/controller.js
robwalch: @waxidiotic Everything in the model is based on the config. It doesn't need access to the model.

1. When `autoPause` is undefined, add `{ autoPause: { viewability: false }` to the config (and thus the model.
2. When `autoPause` is defined add it and make sure viewability is the default correctly:
```js
{
  autoPause: {
    viewability: ('viewability' in autoPause) ? !!autoPause.viewability: true
}
```

src/js/controller/controller.js
robwalch: Hmm... maybe we shouldn't set a pauseReason if already paused - Because any call to pause could overwrite the initial pauseReason.

src/js/controller/controller.js
robwalch: instream has access to the controller, so why not `controller.on/off('viewable', controller.play)`?

src/js/program/program-controller.js
robwalch: But that's what `noResume` does. It prevents the attempt to play after an ad, so there is no buffering.

src/js/api/config.js
```diff
@@ -84,6 +84,13 @@ const Config = function(options, persisted) {
     config.language = language;
     config.intl = intl;
 
+    // If autoPause is configured with an empty block,
+    // default autoPause.viewability to true.
+    let autoPause = config.autoPause;
+    if (autoPause && !autoPause.hasOwnProperty('viewability')) {
```
robwalch: `'viewability' in autoPause`. See https://github.com/jwplayer/jwplayer/pull/3282#discussion_r257038288

src/js/controller/controller.js
robwalch: Sure, but why not just enable autopause on mobile by default? Why do we need more code to do the same thing?

src/js/program/program-controller.js
waxidiotic: If I revert the changes in program-controller to be back to `backgroundMediaController.mediaModel.attributes.mediaState = autoPause ? 'paused' : 'buffering';` and set `instream.noResume = true` in the controller, then the media state will still be set to buffering. Yeah, the video won't play but the player will still be showing the buffering state.

src/js/api/config.js
```diff
@@ -84,6 +85,11 @@ const Config = function(options, persisted) {
     config.language = language;
     config.intl = intl;
 
+    // If autoPause is configured with an empty block,
+    // default autoPause.viewability to true.
+    let autoPause = config.autoPause;
+    autoPause.viewability = ('viewability' in autoPause) ? !!autoPause.viewability : true;
```
robwalch: Now `config.autoPause` will always be defined because you added it to `Defaults`. So for this to work, you want to check what was passed into setup which is `allOptions`:

```js
let autoPause = allOptions.autoPause; // set by user
config.autoPause.viewability = ('viewability' in autoPause) ? !!autoPause.viewability : true;
```

src/js/controller/controller.js
```diff
@@ -601,7 +618,7 @@ Object.assign(Controller.prototype, {
         }
 
         function _isIdle() {
-            const state = _model.get('state');
+            const state = _getState();
```
robwalch: `_getState` includes ad state. Are you sure you want to change `_model.get('state')` to `_getState()`?

src/js/controller/controller.js
```diff
@@ -577,15 +589,20 @@ Object.assign(Controller.prototype, {
             _actionOnAttach = null;
             checkAutoStartCancelable.cancel();
 
-            _model.set('pauseReason', _getReason(meta));
+            const state = _getState();
```
robwalch: Use `_model.get('state')`. Calling `_getState()` also calls `_getAdState()` which we don't want.

waxidiotic:
test this please

src/js/controller/controller.js
```diff
@@ -336,13 +338,30 @@ Object.assign(Controller.prototype, {
         function _checkPlayOnViewable(model, viewable) {
             if (model.get('playOnViewable')) {
                 if (viewable) {
-                    _autoStart();
-                } else if (OS.mobile || (model.get('autoPause') && model.get('autoPause').viewability)) {
+                    _autoStart({ reason: 'viewability' });
```
robwalch: Can we make this `'viewable'` (no object wrapper - it's repetitive).

src/js/controller/controller.js
```diff
@@ -298,7 +298,7 @@ Object.assign(Controller.prototype, {
 
             // Autostart immediately if we're not waiting for the player to become viewable first.
             if (_model.get('autostart') === true && !_model.get('playOnViewable')) {
-                _autoStart();
+                _autoStart({ reason: 'autostart' });
```
robwalch: External `play` and `pause` methods need `meta` with `reason`. This call only needs the `reason` string:

`_autoStart('autostart');`

src/js/controller/controller.js
```diff
@@ -515,13 +534,7 @@ Object.assign(Controller.prototype, {
...
                 }
 
-                // Enable autoPause behavior.
-                const autoPause = _model.get('autoPause');
-                if (autoPause && autoPause.viewability) {
-                    _model.set('playOnViewable', true);
-                }
-
-                return _play({ reason: 'autostart' }).catch(() => {
+                return _play(reason).catch(() => {
```
robwalch: Just pass in a string so we can do:

`return _play({ reason }).catch(() => {`

src/js/controller/controller.js
```diff
@@ -577,15 +596,20 @@ Object.assign(Controller.prototype, {
             _actionOnAttach = null;
             checkAutoStartCancelable.cancel();
 
-            _model.set('pauseReason', _getReason(meta));
+            const state = _model.get('state');
+            if (state !== STATE_PAUSED && state !== STATE_IDLE) {
+                const pauseReason = _getReason(meta);
+                _model.set('pauseReason', pauseReason);
+                _model.set('playOnViewable', (pauseReason === 'viewable'));
+            }
```
robwalch: This new code should be a function that only changes the model when we pause ads or the program controller. This way we only update the model before calling one of the pause methods below, and we don't need the new "state" checks you added above (because we're already dealing with state in `switch (state)` and `adState`:

```js
if (adState && adState !== STATE_PAUSED) {
    _updatePauseReason(meta);
    _api.pauseAd(true, meta);
    return;
}
...
case STATE_BUFFERING: {
    _updatePauseReason(meta);
    _programController.pause();
    break;
...

...outside of the _pause function:
function _updatePauseReason(meta) {
    const pauseReason = _getReason(meta);
    _model.set('pauseReason', pauseReason);
    _model.set('playOnViewable', (pauseReason === 'viewable'));
}
```

src/js/program/program-controller.js
```diff
@@ -266,7 +266,11 @@ class ProgramController extends Events {
             return;
         }
 
-        backgroundMediaController.mediaModel.attributes.mediaState = 'buffering';
+        let mediaState = backgroundMediaController.mediaModel.attributes.mediaState;
+        if (mediaState !== 'paused') {
+            mediaState = 'buffering';
```
robwalch: Setting `mediaState` to buffering here does nothing. You have to update `backgroundMediaController.mediaModel.attributes.mediaState`.

src/js/controller/controller.js
```diff
@@ -309,6 +309,8 @@ Object.assign(Controller.prototype, {
             });
 
             preload();
+
+            _checkPauseOnViewable(model, viewable);
```
robwalch: Is this now the only place this is called? Why are you passing in the model? At this point this should be it's own callback for "change: viewable" only added when `autoPause.viewability` is `true`.

src/js/controller/controller.js
```diff
@@ -336,13 +338,37 @@ Object.assign(Controller.prototype, {
...
                 }
             }
         }
 
+        function _checkPauseOnViewable(model, viewable) {
+            const instream = _this._instreamAdapter;
+
+            // If player left view and came back before ad completes,
+            // the player should begin/resume playing content.
+            if (viewable && instream && instream.noResume) {
```
robwalch: You could use `_getAdState() === PAUSED` instead of checking `instream.noResume`.

src/js/controller/controller.js
```diff
@@ -336,13 +338,37 @@ Object.assign(Controller.prototype, {
...
+                instream.noResume = false;
+                model.set('playOnViewable', true);
+                return;
+            } else if (viewable || !model.get('autoPause').viewability) {
```
robwalch: `_checkPauseOnViewable` should never be called when `model.get('autoPause').viewability` is `false`.

src/js/controller/controller.js
```diff
@@ -336,13 +338,37 @@ Object.assign(Controller.prototype, {
...
+                return;
+            }
+
+            const adState = _getAdState();
+            if (!adState) {
+                _this.pause({ reason: 'viewable' });
+            } else if (model.get('activeTab') && adState !== 'paused') {
```
robwalch: Why the check for activeTab?

src/js/controller/controller.js
```diff
@@ -336,13 +338,37 @@ Object.assign(Controller.prototype, {
...
+                if (instream) {
+                    instream.noResume = true;
+                }
+                model.set('playOnViewable', true);
+            }
+        }
```
robwalch: There's too much repetition and disconnected logic in this function. You should rewrite it.

Here's what a simplified version looks like: 
```js
if (model.get('autoPause').viewability) {
    _model.change('viewable', _checkPauseOnViewable);
}
...

function _checkPauseOnViewable(model, viewable) {
    if (adState) {
         _this._instreamAdapter. noResume = !viewable;
    }
    model.set('playOnViewable', !viewable);
    if (!viewable) {
        _this.pause({ reason: 'viewable' });
    }
}
```

src/js/program/program-controller.js
waxidiotic: So setting `mediaState = backgroundMediaController.mediaModel.attributes.mediaState` and then setting `mediaState` to something else doesn't work?

src/js/controller/controller.js
waxidiotic: Because opening a new tab causes the viewability of the player to change. I was finding a bug in which the following behavior:

1. Click on an ad during playback. Ad pauses and a new tab opens.
2. Go back to the page with the player on it.
3. Scroll down so the player is not viewable any more.
4. Scroll back up so the player is viewable.
5. Ad will resume playback without user interaction.

src/js/controller/controller.js
waxidiotic: Is `instream.noResume` always `true` when the `adState` is paused?

src/js/controller/controller.js
waxidiotic: Can't do this because it will pause if an ad is currently playing.

src/js/controller/controller.js
waxidiotic: 👍

src/js/controller/controller.js
waxidiotic: 👍

src/js/api/config.js
```diff
@@ -84,6 +85,13 @@ const Config = function(options, persisted) {
     config.language = language;
     config.intl = intl;
 
+    // If autoPause is configured with an empty block,
+    // default autoPause.viewability to true.
+    let autoPause = allOptions.autoPause;
+    if (autoPause) {
+        config.autoPause.viewability = ('viewability' in autoPause) ? !!autoPause.viewability : true;
```
johnBartos: Why do you need to use `in`? That operator can be a bit tricky because it also checks the prototype chain.

waxidiotic: @johnBartos I was originally using `hasOwnProperty` but Rob suggested this way.

johnBartos: ok cool

src/js/controller/controller.js
robwalch: No, but it could be if you want the player to not call `play` to resume after an ad break. I would expect that to change either here or in  @vseventer changes https://github.com/jwplayer/jwplayer/pull/3295 if you need it to be (`instream.noResume` shouldn't stop playback or go to idle though - that should only happen when calling `jwplayer().stop()`

Currently `instream.noResume` defaults to `false` because ads play as a results of content playback and should resume content when complete.

src/js/controller/controller.js
robwalch: That doesn't sound like a bug to me. If you click on an ad you are not pausing the ad, the player is. Viewing the player again should resume playback (it should have resumed in step 2 so that's a different issue).

src/js/controller/controller.js
robwalch: The code above may not be complete. But it is a much smaller template for you to base the logic on. Complete the logic as necessary, without splitting the logic for instream an ad state in multiple places.

src/js/program/program-controller.js
robwalch: it's a variable.

src/js/controller/controller.js
waxidiotic: @robwalch - I just clarified with @aliciahurst because it wasn't super-clear in the ticket and she confirmed that the behavior should remain what it is now which is that the ad remains paused when the viewer comes back to the page after clicking an ad.

src/js/controller/controller.js
```diff
@@ -336,15 +340,30 @@ Object.assign(Controller.prototype, {
         function _checkPlayOnViewable(model, viewable) {
             if (model.get('playOnViewable')) {
                 if (viewable) {
-                    _autoStart();
-                } else if (OS.mobile || (model.get('autoPause') && model.get('autoPause').viewability)) {
+                    _play('viewable');
```
robwalch: The problem with this is that autostart players still depend on "playOnViewable". They really shouldn't. If they are supposed to call `_autoStart('viewable');` then a separate callback, or some additional logic here would allow us to call either `_autoStart('viewable');` or `_play('viewable');` as appropriate.

In the model this is happening (not ideal since `_autoStart` should only be called once):
```js
    // Mobile players always wait to become viewable.
    // Desktop players must have autostart set to viewable
    setAutoStart(autoStart) {
        }

        const autoStartOnMobile = OS.mobile && this.get('autostart');
        this.set('playOnViewable', autoStartOnMobile || this.get('autostart') === 'viewable');
```

PR #3283, opened 2019-02-11
[JW8-2506]Retain pause state when/after seeking 

DanFerrer:
### This PR will...

- Prevent video from playing while scrubbing through the player when paused
- Prevent video from playing on seek when paused.

Feature to improve User Experience of the player.

https://github.com/jwplayer/jwplayer-commercial/pull/6335

#### Addresses Issue(s):

JW8-2506

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -611,10 +611,6 @@ Object.assign(Controller.prototype, {
             if (!_model.get('scrubbing') && state !== STATE_PLAYING) {
                 meta = meta || {};
                 meta.startTime = pos;
-
-                if (_model.get('dvrLive')) {
-                    this.play(meta);
-                }
             }
         }
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -52,7 +52,7 @@ Object.assign(Controller.prototype, {
         let _actionOnAttach;
         let _stopPlaylist = false;
         let _interruptPlay;
-        let _stateBeforeScrubbing = null;
+        let _resumeAfterScrubbing = null;
         let checkAutoStartCancelable = cancelable(_checkAutoStart);
         let updatePlaylistCancelable = cancelable(noop);
 
@@ -144,9 +144,9 @@ Object.assign(Controller.prototype, {
 
         _model.on('change:scrubbing', function(model, state) {
             if (state) {
-                _stateBeforeScrubbing = _model.get('state');
+                _resumeAfterScrubbing = _model.get('state') !== STATE_PAUSED;
                 _pause();
-            } else if (_stateBeforeScrubbing === STATE_PLAYING) {
+            } else if (_resumeAfterScrubbing) {
                 _play({ reason: 'interaction' });
             }
         });
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -144,8 +144,6 @@ Object.assign(Controller.prototype, {
         _model.on('change:scrubbing', function(model, state) {
             if (state) {
                 _pause();
-            } else {
-                _play({ reason: 'interaction' });
             }
         });
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -611,6 +611,10 @@ Object.assign(Controller.prototype, {
             if (!_model.get('scrubbing') && state !== STATE_PLAYING) {
                 meta = meta || {};
                 meta.startTime = pos;
+
+                if (_model.get('dvrLive')) {
+                    this.play(meta);
+                }
             }
         }
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -142,8 +142,10 @@ Object.assign(Controller.prototype, {
         _this.on(AD_PAUSE, event => changeReason(null, event.pauseReason));
 
         _model.on('change:scrubbing', function(model, state) {
-            if (state && model.get('state') !== STATE_PLAYING) {
-                _pause();
+            if (model.get('state') !== STATE_PLAYING) {
+                if (state) {
+                    _pause();
+                }
             }
         });
 
@@ -603,13 +605,17 @@ Object.assign(Controller.prototype, {
         }
 
         function _seek(pos, meta) {
-            if (_model.get('state') === STATE_ERROR) {
+            const state = _model.get('state');
+            if (state === STATE_ERROR) {
                 return;
             }
             _programController.position = pos;
-            if (!_model.get('scrubbing') && _model.get('state') !== STATE_PLAYING) {
+            if (!_model.get('scrubbing') && state !== STATE_PLAYING) {
                 meta = meta || {};
                 meta.startTime = pos;
+                if (state === STATE_COMPLETE || _model.get('mediaModel').get('duration') - pos < 1) {
+                    this.play(meta);
+                }
             }
         }
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -142,7 +142,7 @@ Object.assign(Controller.prototype, {
         _this.on(AD_PAUSE, event => changeReason(null, event.pauseReason));
 
         _model.on('change:scrubbing', function(model, state) {
-            if (state) {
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -142,10 +142,8 @@ Object.assign(Controller.prototype, {
         _this.on(AD_PAUSE, event => changeReason(null, event.pauseReason));
 
         _model.on('change:scrubbing', function(model, state) {
-            if (state) {
+            if (state && model.get('state') !== STATE_PLAYING) {
```
karimMourra: i think we can now access state with `model.state` thanks to rob's simple-model attribute changes.

src/js/controller/controller.js
```diff
@@ -606,14 +604,17 @@ Object.assign(Controller.prototype, {
...
+            if (state === STATE_ERROR) {
                 return;
             }
             _programController.position = pos;
-            if (!_model.get('scrubbing') && _model.get('state') !== STATE_PLAYING) {
+            if (!_model.get('scrubbing') && state !== STATE_PLAYING) {
                 meta = meta || {};
                 meta.startTime = pos;
-                this.play(meta);
+                if (state === STATE_COMPLETE || _model.get('mediaModel').get('duration') - pos < 1) {
```
karimMourra: is this the desired behavior for complete ? Why would it be different than idle ?

src/js/controller/controller.js
robwalch: maybe... almost.

src/js/controller/controller.js
robwalch: You can definitely do `_model.mediaModel`.

src/js/controller/controller.js
robwalch: Why not always call pause? Checking the state shouldn't be necessary. Don't add extra code that we don't need. There shouldn't be any harm in calling pause when paused.

src/js/controller/controller.js
```diff
@@ -606,14 +604,18 @@ Object.assign(Controller.prototype, {
...
                 return;
             }
             _programController.position = pos;
-            if (!_model.get('scrubbing') && _model.get('state') !== STATE_PLAYING) {
+            if (!_model.get('scrubbing') && state !== STATE_PLAYING) {
                 meta = meta || {};
                 meta.startTime = pos;
-                this.play(meta);
+                if (_model.get('streamType') === 'VOD' && (state === STATE_COMPLETE || _model.get('mediaModel').get('duration') - pos < 1)) {
+                    // Only resume VOD playback when  state is complete or position is within 1 second of duration.
```
robwalch: Why not just check the state? Isn't `state === STATE_COMPLETE` enough?

The player never reaches `STATE_COMPLETE` for Live and DVR streams, and never would be complete before the end of a stream,  so this extra logic is not needed.

src/js/controller/controller.js
DanFerrer: I was checking ```model.get('state')``` because we don't want to call pause when seeking while the video is still playing.

robwalch: The tag needs to be paused while scrubbing or you will hear audio and see playback jitteres as you scrub (drag the slider).

src/js/controller/controller.js
```diff
@@ -606,14 +604,18 @@ Object.assign(Controller.prototype, {
...
             }
             _programController.position = pos;
-            if (!_model.get('scrubbing') && _model.get('state') !== STATE_PLAYING) {
+            if (!_model.get('scrubbing') && state !== STATE_PLAYING) {
                 meta = meta || {};
                 meta.startTime = pos;
-                this.play(meta);
+
+                if (_model.get('dvrLive')) {
+                    this.play(meta);
```
karimMourra: why do we play on dvr ?

src/js/controller/controller.js
robwalch: Still waiting on this feedback to be addressed

src/js/controller/controller.js
DanFerrer: @karimMourra ```play``` is called when ```dvrLive``` is true for cases when we press the live button to get the live edge. The stream should be seeked to the live edge and then resume playback

robwalch:
@DanFerrer please rebase against master taking note of these changes https://github.com/jwplayer/jwplayer/pull/3282
You shouldn't need to check the player state before calling `_pause()`

src/js/controller/controller.js
```diff
@@ -605,14 +605,14 @@ Object.assign(Controller.prototype, {
...
-            if (_model.get('state') === STATE_ERROR) {
+            const state = _model.get('state');
+            if (state === STATE_ERROR) {
                 return;
             }
             _programController.position = pos;
-            if (!_model.get('scrubbing') && _model.get('state') !== STATE_PLAYING) {
+            if (!_model.get('scrubbing') && state !== STATE_PLAYING) {
                 meta = meta || {};
                 meta.startTime = pos;
```
robwalch: If the state is idle you should call play (seek to start playback).

src/js/view/controls/controlbar.js
```diff
@@ -527,6 +527,10 @@ export default class Controlbar {
             const dvrPosition = Math.min(this._model.get('position'), -1);
             const dvrSeekLimit = this._model.get('dvrSeekLimit');
 
+            this._model.once('seeked', () => {
```
robwalch: If you add this callback you then need to remove it in case the playlist item changes before seeking completes or in the case of an error.

Instead of doing that just call play. There is no need to wait for the callback. Calling `seek` and `play` (like the line you removed in the controller) results in playback after seeking.

src/js/controller/controller.js
```diff
@@ -52,6 +52,7 @@ Object.assign(Controller.prototype, {
         let _actionOnAttach;
         let _stopPlaylist = false;
         let _interruptPlay;
+        let _stateOnScrubbing = null;
```
karimMourra: @DanFerrer i don't think we need to store the state, since calling `_pause()` or `_play()` shouldn't alter the state; otherwise we'd have on pause events firing while scrubbing

src/js/controller/controller.js
```diff
@@ -144,7 +145,8 @@ Object.assign(Controller.prototype, {
         _model.on('change:scrubbing', function(model, state) {
             if (state) {
                 _pause();
-            } else {
+                _stateOnScrubbing = _model.get('state');
+            } else if (_stateOnScrubbing === STATE_PLAYING) {
```
robwalch: This looks like a mistake even though it's not (updating `_stateOnScrubbing` right above the else).  See @karimMourra's comment https://github.com/jwplayer/jwplayer/pull/3283/files#r261038692

pajong: I started looking into it, and it looks like we are changing the state (and firing `pause` event - checked in the latest develop). Without the pause, we can see the video frame changing while we are scrubbing.
Also checked `_model.get('state')` when we are done with scrubbing, and it returns `paused` independent of the state before scrubbing.

src/js/controller/controller.js
```diff
@@ -143,8 +144,9 @@ Object.assign(Controller.prototype, {
 
         _model.on('change:scrubbing', function(model, state) {
             if (state) {
+                _stateBeforeScrubbing = _model.get('state');
                 _pause();
-            } else {
+            } else if (_stateBeforeScrubbing === STATE_PLAYING) {
                 _play({ reason: 'interaction' });
```
robwalch: If you scrub while buffering, the player was trying to play. So only checking` _stateBeforeScrubbing === STATE_PLAYING` is incomplete.

```js
_model.on('change:scrubbing', function(model, state) {
  if (state) {
    _resumeAfterScrubbing = _model.get('state') !== STATE_PAUSED;
     _pause();
  } else if (_resumeAfterScrubbing) { 
   _play...
```

PR #3284, opened 2019-02-12
JW8 - add title secondary in player and info overlay

quebello:
### This PR will ...
modify the code to add a secondary title (subtitle) of a media in the player and in the content of the info overlay

For our project, we need to display a subtitle (secondary title) in the player and in the info overlay.
For example, show the episode title under the serie title.

no

no

#### Addresses Issue(s):

JW8-3285 https://github.com/jwplayer/jwplayer/issues/3285

### DIFF
--- a/src/css/jwplayer/imports/title.less
+++ b/src/css/jwplayer/imports/title.less
@@ -8,7 +8,8 @@
 
 // if we set color on the parent, it is overwritten by the reset
 .jw-title-primary,
-.jw-title-secondary {
+.jw-title-secondary,
+.jw-title-description {
     color: @white;
     // ensure there's padding even when the logo is positioned in the top-left
     padding-left: @ui-margin;
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -18,6 +18,7 @@ const Defaults = {
     defaultPlaybackRate: 1,
     displaydescription: true,
     displaytitle: true,
+    displaytitleSecondary: true,
     displayPlaybackLabel: false,
     height: 360,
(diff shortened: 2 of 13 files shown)


karimMourra:
@quebello why not use `jw-title-secondary`?

johnBartos:
Hi @quebello, thanks for the PR. However, we won't be merging this because you can do it with our existing API. The `description` and `title` fields accept HTML input. For example:

```
jwplayer().setup({
    decription: '<h2>Hello!</h2>
});
```

Unfortunately I just found a bug with this feature - the HTML is added, but the text is stripped out. The last working version is 8.7.3. We'll put in a bug ticket and fix it in the near future. Thanks for helping us find this!

quebello:
Hello, in our case the description is a short description about de media.

johnBartos:
@quebello Are the existing title & description config options not sufficient for your usecase?

quebello:
@johnBartos unfornunately not, our business is showed/shared a lot of media contents and we have the possibility to share our embed with jwplayer and this feature meets our needs and maybe for other people ^^. I dont know if you understand ?

johnBartos:
@quebello Before you pass it to JW just put the secondary title into HTML, and pass that in. It sounds like you know what the title is (because you would have passed it into the new config option), so I'm not sure I understand why this is a blocker

quebello:
@johnBartos the first problem if i put some html in the title or description field, the html is hardcoded in the overlay info. Its so ugly ^^, i think. This is why i suggest a pretty solution :)

PR #3299, opened 2019-02-22
Provide information about keyboard shortcut

zetagame:
### This PR will...

- Standardize modal dimensions
- Force modals to close on "escape" press.
- Create a "keyboard shortcuts" modal that opens on "shift + ?"
- Screen reader will announce "Press shift ? to access a list of keyboard shortcuts" the first time player is focused (non-persistent)
- Add keyboard shortcuts link to right click menu

Did localization in english file. Is there a global one, or will it override other languages where it isn't present?
None that I know of
#### Addresses Issue(s):
https://jwplayer.atlassian.net/browse/JW8-2503
JW8-2503


zetagame:
I still need to write tests, but I wanted to get this up here as soon as possible

johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/controls.js
```diff
@@ -304,6 +321,14 @@ export default class Controls {
                 case 70: // f-key
                     api.setFullscreen();
                     break;
+                case 191: // ? key
+                    if (!this.shortcutsTooltip) {
```
karimMourra: shouldn't this check if it is closed ?

src/js/view/controls/controls.js
```diff
@@ -336,6 +361,10 @@ export default class Controls {
 
         // Hide controls when focus leaves the player
         const blurCallback = (evt) => {
+            //  Remove new shortcut tooltip after first read.
+            if (!OS.mobile) {
+                this.playerContainer.removeAttribute('aria-describedby');
```
karimMourra: i don't think we should remove this. If you're worried about screen readers announcing it multiple times, we could look into using aria-live or alternatives

src/js/view/view.js
```diff
@@ -57,7 +57,11 @@ function View(_api, _model) {
     });
 
     const _localization = _model.get('localization');
-    const _playerElement = createElement(playerTemplate(_model.get('id'), _localization.player));
+    const _playerElement = createElement(playerTemplate(
+        _model.get('id'), 
+        _localization.player.label,
```
karimMourra: player is not an object, this might be why your commercial tests fail

src/js/view/controls/controls.js
```diff
@@ -188,6 +194,7 @@ export default class Controls {
         if (OS.mobile) {
             this.div.appendChild(settingsMenu.element());
         } else {
+            this.playerContainer.setAttribute('aria-describedby', 'jw-new-features-explanation');
```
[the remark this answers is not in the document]
zetagame: It's my understanding that they want screen readers to read that description when focusing on the player the first time (in the non-persistence sense of "first time")

src/js/view/controls/controls.js
zetagame: Nope. It's checking if it exists on the object. If it does, it's already been opened and we can simply toggle the visibility.

src/js/view/controls/controls.js
zetagame: It's my understanding that they want screen readers to read that description when focusing on the player the first time (in the non-persistence sense of "first time"). It has to be removed here to prevent it from occurring every time the player is focused.

src/js/view/view.js
zetagame: Good catch, I missed this when I moved the copy out of localization

src/assets/translations/en.js
```diff
@@ -29,6 +29,9 @@ export default {
     "exitFullscreen": "Exit Fullscreen",
     "fullscreen": "Fullscreen",
     "hd": "Quality",
+    "infos" : {
```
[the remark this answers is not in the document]
zetagame: done!

src/css/controls/imports/rightclick.less
```diff
@@ -8,6 +8,25 @@
     display: none;
     position: absolute;
     white-space: nowrap;
+    width: 284px;
```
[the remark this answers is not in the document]
zetagame: There's a separate ticket for consolidating to a single modal

src/css/controls/imports/info-overlay.less
```diff
@@ -8,8 +8,25 @@
     user-select: text;
     overflow: hidden;
     flex-direction: column;
-    max-width: calc(50% + @mobile-touch-target);
-    max-height: calc(50% + @mobile-touch-target);
+    width: 284px;
```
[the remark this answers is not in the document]
zetagame: There were styles in place for that that I meant to move to the modal.less. Just added them in the latest commit.

src/js/view/controls/controls.js
jnatalzia: Also a little confused. on this class `this.shortcutsTooltip` instantiated as null and only updated if we're not on mobile. if the tooltip is falsy, there's no falsy value in JS that will allow you to call a method on it. Are we just not hitting this execution path at all? in just testing?

could also be misreading this. seems like your comment is related to the state of the modal but i don't see anything in this file where the modal state is tied directly to this instance var

src/js/view/controls/controls.js
jnatalzia: is there danger in a broad brush removal of this attribute on the player container? are there other elements that we may want to repeat the screen-read on focus for? given it's a shared component, that is.

src/js/view/controls/controls.js
zetagame: line 125
` if (!OS.mobile) {
            this.shortcutsTooltip = new ShortcutsTooltip(this.playerContainer, api, model);
        }`

jnatalzia: yeah so that's when it's instantiated, only on desktop. if we're on mobile though, we're going to hit this part of the code where it'll try to call `this.shortcutsTooltip.open();` on a nonexistent object. is there another place where this is instantiated or is this an execution path we can't hit on mobile? either way i don't know why this block would be needed

jnatalzia: happy to talk through in person if i'm not making sense!

test/unit/controls/shortcuts-tooltip.js
```diff
@@ -0,0 +1,63 @@
...
+        document.body.appendChild(player);
+    });
+    afterEach(() => {
+        document.body.removeChild(player);
+        shortcutsTooltip = null;
+        player = null;
+    });
+    it('should be hidden initially', () => {
+        const isInitiallyHidden = isHidden(shortcutsTooltip.el);
+        expect(isInitiallyHidden).to.deep.equal(true)
```
jnatalzia: shouldn't need deep equality on booleans

DanFerrer:
@zetagame Rebase your PR against master to fix the localization tests.

src/css/controls/imports/modal.less
```diff
@@ -0,0 +1,32 @@
+.jw-modal {
+    width: 284px;
+
+    .jw-breakpoint-7 &,
+    .jw-breakpoint-6 &,
+    .jw-breakpoint-5 & {
+        .rect(284px, 232px);
```
karimMourra: No need to use rect since the width is already set on line 2; we can just specify the height here.

src/css/controls/imports/modal.less
```diff
@@ -0,0 +1,32 @@
...
+
+    .jw-breakpoint-2 &,
+    .jw-flag-small-player & {
+        bottom: 0;
+        right: 0;
+        height: 100%;
+        width: 100%;
+        max-height: none;
+        max-width: none;
+        z-index: 1;
```
karimMourra: we avoid setting z-index to avoid conflicts with customer's layers
 do you really need this ?

src/js/view/controls/controls.js
```diff
@@ -316,6 +337,11 @@ export default class Controls {
                 case 70: // f-key
                     api.setFullscreen();
                     break;
+                case 191: // ? key
+                    if (!OS.mobile) {
```
karimMourra: there are no keys in mobile. I think we should check for existence `if (this.shortcutsTooltip) {`
this also covers the mobile case, since shortcutsTooltip is only instantiated when `!OS.mobile`

src/js/view/controls/controls.js
```diff
@@ -348,6 +374,14 @@ export default class Controls {
 
         // Hide controls when focus leaves the player
         const blurCallback = (evt) => {
+            //  Remove new shortcut tooltip after first read.
+            if (!OS.mobile) {
```
karimMourra: This callback runs every time focus leaves the player, yet the logic you are adding should be called only once.
 I think we should have a separate callback which would be removed. Also, let's look into an alternative for `aria-describedby`, there might be a better tool. I think with `aria-live` we can alter the frequency with which something gets announced.

src/js/view/controls/rightclick.js
```diff
@@ -50,6 +51,12 @@ export default class RightClick {
                 link: 'http://www.adobe.com/software/flash/about/'
             });
         }
+        if (!OS.mobile) {
```
karimMourra: we should check `if (this.shortcutsTooltip)` instead. That way, we keep all the conditions required for `this.shortcutsTooltip` to be instantiated in one place. Say we decide to exclude it from a specific browser, you would have to add that condition in multiple places.

src/css/controls/imports/modal.less
zetagame: Unfortunately, we use z-index: 1 all over the place, so I do have to.

image

src/js/view/controls/controls.js
zetagame: As per our earlier discussion today, I'll leave this in for now until there's a better way to do it.

src/js/view/controls/rightclick.js
zetagame: Agreed, I think it's a more logical way to do it. Fixed!

src/js/view/controls/controls.js
zetagame: Agreed, that would be more sustainable.

src/css/controls/imports/modal.less
karimMourra: My bad, you are right! We should avoid using z-index >1, as you have probably already noticed! Great catch!

src/js/view/controls/rightclick.js
```diff
@@ -50,6 +51,12 @@ export default class RightClick {
                 link: 'http://www.adobe.com/software/flash/about/'
             });
         }
+        if (this.shortcutsTooltip) {
+            const secondToLastIndex = menu.items.length - 1;
```
karimMourra: we can improve minification if we make an items const: `const items = menu.items`. We should declare it on line 45, since it can be used as early as line 48. 
We can then get rid of the `secondToLastIndex` var since `items.length - 1` is still pretty readable

src/js/view/controls/shortcuts-tooltip.js
```diff
@@ -0,0 +1,118 @@
...
+        removeClass(template, 'jw-open');     
+        document.removeEventListener('click', documentClickHandler);
+        container.focus();
+        isOpen = false;
+        if (lastState === STATE_PLAYING) {
+            api.play(settingsInteraction);
+        }
+    };
+    const documentClickHandler = (e) => {
+        if (!/jw-shortcuts/.test(e.target.className)) {
```
karimMourra: dope.js

src/js/view/controls/shortcuts-tooltip.js
```diff
@@ -0,0 +1,118 @@
...
+                description: 'seek to %'
+            }
+        ];
+        const closeButton = button('jw-shortcuts-close', () => {
+            close();
+        }, model.get('localization').close, [cloneIcon('close')]);
+        
+        //  Iterate all shortcuts to create list of them. 
+        shortcuts.map(shortcut => {
+            const keyWrapper = document.createElement('li');
```
karimMourra: Wouldn't a template be better ?

zetagame: Done!

src/css/controls/imports/modal.less
```diff
@@ -0,0 +1,29 @@
...
+        height: 232px;
+    }
+
+    .jw-breakpoint-4 &,
+    .jw-breakpoint-3 & {
+        height: 192px;
+    }
+
+    .jw-breakpoint-2 & {
+        height: 179px;
```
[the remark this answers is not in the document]
zetagame: Done!

src/js/view/controls/rightclick.js
zetagame: Done!

src/css/controls/imports/info-overlay.less
```diff
@@ -8,8 +8,6 @@
     user-select: text;
     overflow: hidden;
     flex-direction: column;
-    max-width: calc(50% + @mobile-touch-target);
```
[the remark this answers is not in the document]
zetagame: Yes. Part of the ticket is to remove the dynamic sizing. If it's too small, it'll go to full width.

PR #3300, opened 2019-02-26
[JW8-5170] Ad and Chapters Cues Should Both Appear on Timeline

waxidiotic:
JW8-5170

### This PR will...

* Create new `getCues()` API method that returns the list of cues on the model.
* Create new `addCues()` API method that gets the current cues and appends new cues.
* Make sure chapter cues are set to the model.
* Also call `resetChapters()` inside of `reset()` so it occurs on each playlist item

* When ad plugins were adding cues at different times than chapter cues, whatever cues were added first were cleared.

n/a

https://github.com/jwplayer/jwplayer-ads-vast/pull/485
https://github.com/jwplayer/jwplayer-ads-googima/pull/466
https://github.com/jwplayer/jwplayer-ads-freewheel/pull/152

#### Addresses Issue(s):

JW8-5170

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -981,6 +981,8 @@ Object.assign(Controller.prototype, {
             _model.set('cues', cues);
         };
 
+        this.getCues = _model.get('cues');
+
         this.updatePlaylist = function(playlist, feedData) {
             try {
                 const filteredPlaylist = filterPlaylist(playlist, _model, feedData);
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -978,8 +978,14 @@ Object.assign(Controller.prototype, {
         };
 
         this.setCues = function (cues) {
-            const existingCues = _model.get('cues');
-            const newCues = existingCues ? existingCues.concat(cues) : cues;
+            let newCues = [];
+            if (cues && cues.length) {
+                // If setCues() is called with empty array, we want to clear the cues.
+                // Otherwise, add the new cues to the existing cues.
+                // This allows setting of cues by ad plugins or setCues() API.
+                const existingCues = _model.get('cues');
+                newCues = existingCues ? existingCues.concat(cues) : cues;
+            }
             _model.set('cues', newCues);
         };
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -977,16 +977,14 @@ Object.assign(Controller.prototype, {
             _programController.controls = mode;
         };
 
+        this.addCues = function (cues) {
+            const existingCues = this.getCues();
+            const newCues = existingCues ? existingCues.concat(cues) : cues;
+            this.setCues(newCues);
+        };
+
         this.setCues = function (cues) {
-            let newCues = [];
-            if (cues && cues.length) {
-                // If setCues() is called with empty array, we want to clear the cues.
-                // Otherwise, add the new cues to the existing cues.
-                // This allows setting of cues by ad plugins or setCues() API.
-                const existingCues = _model.get('cues');
-                newCues = existingCues ? existingCues.concat(cues) : cues;
-            }
-            _model.set('cues', newCues);
+            _model.set('cues', cues);
         };
 
         this.getCues = _model.get('cues');
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -978,8 +978,8 @@ Object.assign(Controller.prototype, {
         };
 
         this.setCues = function (cues) {
-            const existingCues = this._model.get('cues');
-            const newCues = existingCues.concat(cues);
+            const existingCues = _model.get('cues');
+            const newCues = existingCues ? existingCues.concat(cues) : cues;
             _model.set('cues', newCues);
         };
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -978,7 +978,9 @@ Object.assign(Controller.prototype, {
         };
 
         this.setCues = function (cues) {
-            _model.set('cues', cues);
+            const existingCues = this._model.get('cues');
+            const newCues = existingCues.concat(cues);
+            _model.set('cues', newCues);
         };
 
         this.getCues = _model.get('cues');
--- a/src/js/view/controls/components/chapters.mixin.js
+++ b/src/js/view/controls/components/chapters.mixin.js
@@ -52,6 +52,11 @@ const ChaptersMixin = {
         }
 
         this.cues.forEach((cue) => {
+            this._model.get('cues').forEach((existingCue) => {
+                if (cue.time === existingCue.begin) {
+                    cue.text = existingCue.text;
+                }
+            });
             cue.align(duration);
             cue.el.addEventListener('mouseover', () => {
                 this.activeCue = cue;
--- a/src/js/view/controls/components/chapters.mixin.js
+++ b/src/js/view/controls/components/chapters.mixin.js
@@ -35,6 +35,9 @@ const ChaptersMixin = {
         if (Array.isArray(data)) {
             data.forEach((obj) => this.addCue(obj));
             this.drawCues();
+            const existingCues = this._model.get('cues');
+            const playerCues = existingCues.concat(data);
+            this._model.set('cues', playerCues);
         }
     },
 
--- a/src/js/view/controls/components/chapters.mixin.js
+++ b/src/js/view/controls/components/chapters.mixin.js
@@ -39,7 +39,9 @@ const ChaptersMixin = {
         if (Array.isArray(data)) {
             data.forEach((obj) => this.addCue(obj));
             this.drawCues();
-            this._model.set('cues', data);
+            const existingCues = this._model.get('cues');
+            const playerCues = existingCues.concat(data);
+            this._model.set('cues', playerCues);
         }
     },
 
--- a/src/js/view/controls/components/chapters.mixin.js
+++ b/src/js/view/controls/components/chapters.mixin.js
@@ -41,7 +41,20 @@ const ChaptersMixin = {
     chaptersFailed: function () {},
 
     addCue: function (obj) {
+        this.cues.forEach((cue) => {
+            if (cue.time === obj.begin) {
+                this.removeCue(cue);
+            }
+        });
         this.cues.push(new Cue(obj.begin, obj.text));
+    },
+
+    removeCue: function (cueToRemove) {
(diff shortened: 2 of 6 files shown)


waxidiotic:
The one failure can't be reproduced in PR build's harness.

johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

aizaztoppa:
test this please

src/js/view/controls/components/timeslider.js
```diff
@@ -280,7 +280,6 @@ class TimeSlider extends Slider {
     }
 
     addCues(model, cues) {
-        this.resetChapters();
```
pajong: I don't think this is safe. 
`api.setCues` changes the `cues` in `model`, which then calls this function. This function used to clear the already added cues. Removing this code will change the api behavior.

waxidiotic: @robwalch - Your thoughts on this? @pajong and I were talking. He's worried that is a customer calls `setCues()` which isn't documented, it would cause duplicate cues. But, we have to make sure that multiple parts of the codebase are able to call `setCues()` (i.e. ads and chapters).

waxidiotic: New thought: could change `setCues()` in controller.js to clear the cues if there are any before setting them. I believe the only thing that would be able to call `setCues()` would be the JavaScript API. The mixins and ad plugins themselves don't call it.

pajong: The ads plugins call `setCues()`
e.g.
https://github.com/jwplayer/jwplayer-ads-vast/blob/master/src/js/vastPlugin.js#L468
https://github.com/jwplayer/jwplayer-ads-googima/blob/master/src/PlaylistItemManager.js#L181

waxidiotic: Yeah, I already realized that it wouldn't work. :(

waxidiotic: @jnatalzia - Curious your thoughts on this too.

jnatalzia: I don't have too much context on exactly what `resetChapters` here means but what's the risk around leaving it here to preserve existing functionality and adding it to `reset`?

waxidiotic: @jnatalzia - It would cause the issue that is the cause of this work which is that the chapter cues will be cleared when the ad plugins call `addCues()`.

robwalch: @waxidiotic I am concerned that this only works if setCues is only called twice when you have chapter cues and adds. Now you can't use `setCues` to reset the cues (API change which Jong pointed out.)

What we should do is add a `getCues` method if we don't have one, and the ads or chapters should get cues, add their cues to a copy of that array, and then set cues to the new updated list.

waxidiotic:
test this please

waxidiotic:
test this please

src/js/view/controls/components/chapters.mixin.js
```diff
@@ -10,6 +10,10 @@ class Cue {
     }
 
     align(duration) {
+        if (!this.time) {
+            return;
+        }
```
robwalch: Why is this necessary?

`this.time = time` is set in the constructor. You should never be instantiating cues without a time. Isn't `0` still valid?

src/js/view/controls/components/chapters.mixin.js
```diff
@@ -35,15 +39,29 @@ const ChaptersMixin = {
         if (Array.isArray(data)) {
             data.forEach((obj) => this.addCue(obj));
             this.drawCues();
+            const existingCues = this._model.get('cues');
+            const playerCues = existingCues.concat(data);
+            this._model.set('cues', playerCues);
```
robwalch: Why do you draw the cues and then update the model?

src/js/view/controls/components/chapters.mixin.js
```diff
@@ -35,15 +39,29 @@ const ChaptersMixin = {
...
     },
 
     chaptersFailed: function () {},
 
     addCue: function (obj) {
+        this.cues.forEach((cue) => {
+            if (cue.time === obj.begin) {
+                this.removeCue(cue);
+            }
+        });
```
robwalch: We didn't need to bother with duplicates before. Why do we now, and why is the chapter mixin responsible for this? Why are chapter cues more important than other cues?

src/js/view/controls/components/chapters.mixin.js
```diff
@@ -35,15 +39,29 @@ const ChaptersMixin = {
...
         this.cues.push(new Cue(obj.begin, obj.text));
     },
 
+    removeCue: function (cueToRemove) {
```
robwalch: This doesn't need to be its own function exposed on the object. Update the model, then reset and redraw all cues.

robwalch: > Fix a JS exception that was occurring when clearing out the queues using `jwplayer().setCues([{}]);`

Can't you just call `jwplayer().setCues([]);`? `{}` is not a valid cue. What calls it? Don't add a check to each cue, when you can handle this at a higher level and just reset.

> Prevent duplicate cues at the same time on the timeslider
Cues with the same time but different content are not duplicates. Establish rules about which types of cues take priority. Handle this when drawing the full list of cues not by removing and replacing elements one at a time.

src/js/view/controls/components/chapters.mixin.js
waxidiotic: I'm not sure why it would matter. Can you explain?

src/js/view/controls/components/chapters.mixin.js
waxidiotic: You told me last week that we don't want duplicate cues to appear if there are two cues set at the same time. Also, it's not necessarily chapters over ads - it's just what gets added first.

src/js/view/controls/components/chapters.mixin.js
waxidiotic: @robwalch 
The question I have would be how to tell what is adding cues. It just happens to be that chapters are added when `playlistItem` fires while the ad plugins add them when the playback is initiated. I assume that we would always want ad cues to take priority over chapter cues, right? If so, I can check in `drawCues()`.

This is a little stream of consciousness but I want to make sure I'm thinking along the same lines.

src/js/view/controls/components/chapters.mixin.js
waxidiotic: I removed this. Will handle dupes in `drawCues()`.

src/js/view/controls/components/chapters.mixin.js
waxidiotic: Refactored this to just set the cue to the model when the chapters are loaded. Changing the cues on the model will then call addCue() and drawCues().

src/js/view/controls/components/chapters.mixin.js
waxidiotic: This was because I was calling `setCues()` wrong to test.

src/js/controller/controller.js
```diff
@@ -978,9 +978,19 @@ Object.assign(Controller.prototype, {
...
         this.setCues = function (cues) {
-            _model.set('cues', cues);
+            let newCues = [];
+            if (cues && cues.length) {
+                // If setCues() is called with empty array, we want to clear the cues.
+                // Otherwise, add the new cues to the existing cues.
+                // This allows setting of cues by ad plugins or setCues() API.
+                const existingCues = _model.get('cues');
+                newCues = existingCues ? existingCues.concat(cues) : cues;
+            }
```
robwalch: This goes back to what the original feedback @pajong and I left - `setCues` should not append to cues.

You could add a new method called `addCues` that does this, and then avoid the repetition of calling `setCues(getCues().append(newCues))` in every place that currently calls `setCues()` to add chapter or ads markers.

src/js/view/controls/components/chapters.mixin.js
```diff
@@ -33,8 +33,8 @@ const ChaptersMixin = {
     chaptersLoaded: function (evt) {
         const data = srt(evt.responseText);
         if (Array.isArray(data)) {
-            data.forEach((obj) => this.addCue(obj));
-            this.drawCues();
+            // Add chapter cues directly to model which will trigger addCue()
+            this._model.set('cues', data);
```
robwalch: This should append the new cues, or call a new `addCues` method which does the appending. Since you have direct access to the model, you could slice model cues, append and set, and then depend on the `change:cues` event to handle the update in the time slider chapters mixin.

waxidiotic:
test this please

src/js/controller/controller.js
```diff
@@ -977,10 +977,18 @@ Object.assign(Controller.prototype, {
             _programController.controls = mode;
         };
 
+        this.addCues = function (cues) {
+            const existingCues = this.getCues();
+            const newCues = existingCues ? existingCues.concat(cues) : cues;
```
robwalch: 1. Let's default the cues in model to an empty array so that we don't need this ternary operator.
2. In `setCues` let's validate the input - it must be an array, otherwise we should not update the model.

src/js/view/controls/components/chapters.mixin.js
```diff
@@ -33,8 +33,10 @@ const ChaptersMixin = {
     chaptersLoaded: function (evt) {
         const data = srt(evt.responseText);
         if (Array.isArray(data)) {
-            data.forEach((obj) => this.addCue(obj));
-            this.drawCues();
+            // Add chapter cues directly to model which will trigger addCue()
+            const existingCues = this._model.get('cues');
+            const newCues = existingCues ? existingCues.concat(data) : data;
```
robwalch: Guarantee that model 'cues' is always an array so you don't have to check if existingCues is defined (see above)

src/js/view/controls/components/timeslider.js
```diff
@@ -280,12 +280,14 @@ class TimeSlider extends Slider {
...
+        // Only reset chapters if setCues() is called with empty array (to clear cues)
```
robwalch: 1. Can we rename `addCues` to set or update cues?
2. Shouldn't you always reset (`resetChapters` is really `resetCues` no?) before adding each new cue?

waxidiotic:
Replaced with https://github.com/jwplayer/jwplayer/pull/3313

PR #3307, opened 2019-03-04
Allow a 'disabled' property which stops any floating behavior

jnatalzia:
### This PR will...
Allow a `disabled` property in the `floating` config block to stop any floating behavior.

When running A/B tests, we don't have any way to stop the test from running on a specific player. Since we don't want to force this behavior for all publishers players where we are running the a/b test, we need to give them an option to disable it.

Is it non-standard to mutate config variables like this?

Also I checked the `controls` and `view` test files and didn't see any functionality related to the floating block. Given this is a fairly benign change, I didn't see a need to write unit tests for this config update. In the event either a.) there is a file or section in the tests related to this functionality I can update or b.) we view this as important enough to check with a unit test I'd be more than happy to add them.

https://github.com/jwplayer/jwplayer-commercial/pull/6340

#### Addresses Issue(s):

JW8-5615

### DIFF
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -120,7 +120,7 @@ export default class Controls extends Events {
             if (visible) {
                 //  Focus modal close button on open
                 this.div.querySelector('.jw-info-close').focus();
-            }  
+            }
         });
         //  Add keyboard shortcuts if not on mobi;e
         if (!OS.mobile) {
@@ -141,7 +141,9 @@ export default class Controls extends Events {
         }
 
         // Floating Close Button
-        const floatingConfig = model.get('floating');
+        let floatingConfig = model.get('floating');
+        floatingConfig = floatingConfig && floatingConfig.disabled ? null : floatingConfig;
+
         if (floatingConfig) {
             const floatCloseButton = new FloatingCloseButton(element, model.get('localization').close);
             const doNotForward = true;
@@ -296,7 +298,7 @@ export default class Controls extends Events {
                     }
                     if (this.shortcutsTooltip) {
                         this.shortcutsTooltip.close();
-                        
+
                     }
                     break;
                 case 13: // enter
@@ -398,7 +400,7 @@ export default class Controls extends Events {
             }
             this.playerContainer.removeEventListener('blur', onRemoveShortcutsDescription, true);
         };
-        
+
         if (this.shortcutsTooltip) {
             this.playerContainer.addEventListener('blur', onRemoveShortcutsDescription, true);
             this.onRemoveShortcutsDescription = onRemoveShortcutsDescription;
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

jnatalzia:
@pajong will this impact any of your open floating work?

src/js/view/controls/controls.js
```diff
@@ -141,7 +141,9 @@ export default class Controls extends Events {
         }
 
         // Floating Close Button
-        const floatingConfig = model.get('floating');
+        let floatingConfig = model.get('floating');
+        floatingConfig = floatingConfig.disabled ? null : floatingConfig;
```
pajong: `model.get('floating`)` is `undefined` when it is not set in config.
`floatingConfig.disabled` will throw exception in that case.

pajong:
@jnatalzia There also needs to be a config schema update in commercial repo when we have a new config option: https://github.com/jwplayer/jwplayer-commercial/blob/master/schemas/api/input/setup/config_schema.json#L2186

jnatalzia:
@pajong thanks for the feedback, updated and linked new commercial pr

https://github.com/jwplayer/jwplayer-commercial/pull/6340

src/js/view/controls/controls.js
```diff
@@ -141,7 +141,9 @@ export default class Controls extends Events {
         }
 
         // Floating Close Button
-        const floatingConfig = model.get('floating');
+        let floatingConfig = model.get('floating');
+        floatingConfig = floatingConfig && floatingConfig.disabled ? null : floatingConfig;
+
         if (floatingConfig) {
```
robwalch: What if the logic in this block was a function that returned early when `floatingConfig.disabled` is true? I'm thinking that would avoid repeating the `floatingConfig` check (covered once in `if`) and the ternary operators (just exit when `floatingConfig.disabled`).

src/js/view/view.js
```diff
@@ -76,6 +76,7 @@ function View(_api, _model) {
     let _stateClassRequestId = -1;
 
     let _floatingConfig = _model.get('floating');
+    _floatingConfig = _floatingConfig && _floatingConfig.disabled ? null : _floatingConfig;
```
robwalch: This should be normalized in the config so you don't need to repeat this kind of logic. Either set "floating" to null or always define `floating` with  `disabled = true` so you can depend on it being defined and never have to repeat this `_floatingConfig &&` check.

PR #3308, opened 2019-03-04
[JW8-5615] Normalize the disabling of floating in the config

jnatalzia:
### This PR will...
Update the floating disable to be normalized through the config and centralize the logic

The previous method worked but was not ideal

I like the explicit null in the config when this happens given we are replacing an existing object that _was_ defined but could be persuaded otherwise

Nope

#### Addresses Issue(s):

JW8-5615

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -92,6 +92,13 @@ const Config = function(options, persisted) {
         config.autoPause.viewability = ('viewability' in autoPause) ? !!autoPause.viewability : true;
     }
 
+    // If floating is configured to be disabled,
+    // remove the floating block from the config
+    const floating = config.floating;
+    if (floating && !!floating.disabled) {
+        config.floating = null;
+    }
+
     let rateControls = config.playbackRateControls;
 
     if (rateControls) {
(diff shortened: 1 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

jnatalzia:
test this please

src/js/api/config.js
```diff
@@ -92,6 +92,13 @@ const Config = function(options, persisted) {
         config.autoPause.viewability = ('viewability' in autoPause) ? !!autoPause.viewability : true;
     }
 
+    // If floating is configured to be disabled,
+    // remove the floating block from the config
+    let floating = allOptions.floating;
```
DanFerrer: Do we/Are we going to reassign ```floating``` at all. We can use ```const``` instead of ```let``` if thats not the case?

Also we can use ```config``` insrtead of ```allOptions``` since we've already merged ```allOptions``` and ```Defaults``` with ```config``` at this point.

jnatalzia: great calls! will update

PR #3309, opened 2019-03-04
[JW8-2627] Do not allow dragging ads on floating player

pajong:
### This PR will...
- Disable dragging floating player when playing ads
- Stop floating when the state changes to `IDLE` `COMPLETE` or `ERROR`
- Do not float when the state is `COMPLETE`
- Do not let the page scroll on mobile when dragging the floating player

- Ads have issues dragging due to the ad click functionality.
- We do not want to float player that is not worthy of floating

#### Addresses Issue(s):

JW8-2627


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/view.js
```diff
@@ -885,7 +896,9 @@ function View(_api, _model) {
 
                 updateFloatingSize();
 
-                _this.floatingUI = new FloatingDragUI(_wrapperElement);
+                if (!_model.get('instreamMode')) {
```
jnatalzia: could we potentially still drag in outstream mode?

src/js/view/view.js
```diff
@@ -940,7 +953,10 @@ function View(_api, _model) {
                 top: null,
                 bottom: null
             });
-            _this.floatingUI.destroy();
+            if (_this.floatingUI) {
```
jnatalzia: you're duplicating this functionality in two places, since it uses the same locally scoped var we can abstract it out in the event we need to update it in the future. also ensures that any future changes to this conditional only require 1 update

src/js/view/view.js
pajong: `instreamMode` is the mode set to signify that we are playing ads

src/js/view/view.js
```diff
@@ -895,6 +905,13 @@ function View(_api, _model) {
         }
     }
 
+    function _destroyFloatingUI() {
```
karimMourra: can we rename this to disableDragging ?

src/js/view/view.js
```diff
@@ -871,7 +878,8 @@ function View(_api, _model) {
         // Player is 50% visible or less and no floating player already in the DOM.
         const shouldFloat = intersectionRatio < 0.5;
         if (shouldFloat) {
-            if (_model.get('state') !== STATE_IDLE && floatingPlayer === null) {
+            const state = _model.get('state');
+            if (state !== STATE_IDLE && state !== STATE_COMPLETE && floatingPlayer === null) {
```
karimMourra: what about error state ?

src/js/view/view.js
```diff
@@ -790,6 +793,10 @@ function View(_api, _model) {
             _controls.destroyInstream(_model);
         }
 
+        if (floatingPlayer === _playerElement) {
+            _this.floatingUI = new FloatingDragUI(_wrapperElement);
```
robwalch: Why don't you just manage dragging via  `_this.floatingUI.disable()` and `_this.floatingUI.enable()`? That way this logic is encapsulated in `FloatingDragUI ` instead of here in the view, and you don't have to (duplicate code) create new instances.

PR #3310, opened 2019-03-04
Do not apply new max-width to floating player

pajong:
### This PR will...
Not apply new max-width to the floating player

Applying new max-width made the floating player larger on desktop, which we did not want

It looks like the code was added for mobile - I did a quick run on mobile and it looked fine removing this, but if there were some specific cases, please let me know

#### Addresses Issue(s):

JW8-###

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -901,9 +901,7 @@ function View(_api, _model) {
         const width = _model.get('width');
         const height = _model.get('height');
         const styles = getPlayerSizeStyles(width);
-        if (isNumber(width)) {
-            styles.maxWidth = width;
-        }
+
         if (!_model.get('aspectratio')) {
             const containerWidth = _model.get('containerWidth');
             const containerHeight = _model.get('containerHeight');


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/view.js
```diff
@@ -901,9 +901,7 @@ function View(_api, _model) {
         const width = _model.get('width');
         const height = _model.get('height');
         const styles = getPlayerSizeStyles(width);
-        if (isNumber(width)) {
```
jnatalzia: @DanFerrer does this revert some of your aspect ratio work from earlier? are there any considerations for the mobile case?

karimMourra: @pajong it seems to me like the problem is a missing size constraint. If it appears too big on desktop, i would say we should enforce a specific width or a smaller max-width, instead of removing max-width

DanFerrer: @jnatalzia This shouldn't break anything I'm aware of. We already have size constraints for ```max-width``` in ```floatingplayer.less``` at ```400px```with the exception of mobile portrait where we set the ```max-width``` to ```none```

PR #3312, opened 2019-03-05
[JW8-5648] Strip object and iframe elements from HTML input

robwalch:
### This PR will...
Address vulnerability to XSS via object and iframe tags in UI elements that accept HTML input.

Prevent XSS injection via object tag data and iframes.

#### Addresses Issue(s):
JW8-5648

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -40,7 +40,7 @@ export function htmlToParentElement(html) {
     // Delete event handler attributes that could execute XSS JavaScript
     const insecureElements = parsedElement.querySelectorAll('img,svg');
 
-    for (let i = 0; i < insecureElements.length; i++) {
+    for (let i = insecureElements.length; i--;) {
         const element = insecureElements[i];
         sanitizeElementAttributes(element);
     }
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/utils/dom.js
```diff
@@ -49,10 +49,10 @@ export function htmlToParentElement(html) {
 }
 
 export function sanitizeScriptNodes(element) {
-    const scripts = element.querySelectorAll('script');
-    for (let i = 0; i < scripts.length; i++) {
-        const script = scripts[i];
-        script.parentNode.removeChild(script);
+    const nodes = element.querySelectorAll('script,object,iframe');
```
jnatalzia: any tests to update?

src/js/utils/dom.js
```diff
@@ -49,10 +49,10 @@ export function htmlToParentElement(html) {
...
+    for (let i = 0; i < nodes.length; i++) {
+        const node = nodes[i];
+        node.parentNode.removeChild(node);
```
zetagame: Dope implementation! Approving, but be careful of where this is utilized. If we fire it inside a 3rd party iframe (like IMA), it'll throw a cross origin iframe script blocking error.

src/js/utils/dom.js
robwalch: I will add tests to [dom-tests.js](https://github.com/jwplayer/jwplayer/blob/master/test/unit/dom-test.js#L84)

robwalch: done

src/js/utils/dom.js
```diff
@@ -49,17 +49,17 @@ export function htmlToParentElement(html) {
...
     }
     return element;
 }
 
 export function sanitizeElementAttributes(element) {
     const attributes = element.attributes;
-    for (let i = 0; i < attributes.length; i++) {
+    for (let i = attributes.length; i--;) {
```
robwalch: This is an important fix. `attributes.length` is decremented as attributes are removed, so we have to start at the end and cycle through in reverse to remove all matches.

PR #3313, opened 2019-03-07
[JW8-5170] Add addCues() api to append cues instead of clearing them

waxidiotic:
JW8-5170

### This PR will...

* Create new `getCues()` API method that returns the list of cues on the model.
* Create new `addCues()` API method that gets the current cues and appends new cues.
* Make sure chapter cues are set to the model.
* Also call `resetChapters()` inside of `reset()` so it occurs on each playlist item

* When ad plugins were adding cues at different times than chapter cues, whatever cues were added first were cleared.

n/a

https://github.com/jwplayer/jwplayer-ads-vast/pull/485
https://github.com/jwplayer/jwplayer-ads-googima/pull/466
https://github.com/jwplayer/jwplayer-ads-freewheel/pull/152

#### Addresses Issue(s):

JW8-5170

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -977,9 +977,17 @@ Object.assign(Controller.prototype, {
             _programController.controls = mode;
         };
 
+        this.addCues = function (cues) {
+            const existingCues = this.getCues();
+            const newCues = existingCues ? existingCues.concat(cues) : cues;
+            this.setCues(newCues);
+        };
+
         this.setCues = function (cues) {
             _model.set('cues', cues);
         };
+
+        this.getCues = _model.get('cues');
 
         this.updatePlaylist = function(playlist, feedData) {
             try {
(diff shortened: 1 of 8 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -977,8 +977,20 @@ Object.assign(Controller.prototype, {
...
+        };
+
         this.setCues = function (cues) {
-            _model.set('cues', cues);
+            if (Array.isArray(cues)) {
+                _model.set('cues', cues);
+            }
+        };
+
+        this.getCues = function() {
```
robwalch: You don't need this method because the api doesn't use it, and you can just use the model `_model.get('cues');` in `addCues` above.

PR #3316, opened 2019-03-08
Issue on translation FR

ludoAlcala:
Fixes translation in french
Paratager -> Partager

JW8-###

### DIFF
--- a/src/assets/translations/fr.json
+++ b/src/assets/translations/fr.json
@@ -53,7 +53,7 @@
         "copied": "Copié",
         "email": "Courriel",
         "embed": "Intégrer",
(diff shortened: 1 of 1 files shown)


karimMourra:
@ludoAlcala thank you for reporting this! The fix will go out in 8.8.0 . In the meantime to override this, you can use our `intl` block. In your config:
```
{
    ...
    intl: {
        fr: {
            sharing: {
                heading: 'Partager'
            }
        }
    }
}
```

karimMourra:
Note: Per @aliciahurst, internal source of truth has been modified to reflect these changes

PR #3320, opened 2019-03-08
[JW8-5643] Limit prevent scrolling to mobile

pajong:
### This PR will...
Limit prevent scrolling for drag floating player to mobile devices.

"preventScrolling" option on desktop calls `el.setPointerCapture`, resulting the upcoming events to be only captured in that element.
Since `_wrapperElement` is the element (which contains controls), any events after `pointerdown` on controls are ignored, making the controls not clickable.
Mobile doesn't have the issue, because `el.setPointerCapture` is only called on `pointerdown`

#### Addresses Issue(s):
JW8-5643

### DIFF
--- a/src/js/view/floating-drag-ui.js
+++ b/src/js/view/floating-drag-ui.js
@@ -1,5 +1,6 @@
 import UI from 'utils/ui';
 import { style } from 'utils/css';
+import { OS } from 'environment/environment';
 
 export default class FloatingDragUI {
     constructor(element) {
@@ -20,7 +21,7 @@ export default class FloatingDragUI {
         let innerWidth;
         const auto = 'auto';
         const { element } = this;
-        const ui = this.ui = new UI(element, { preventScrolling: true })
+        const ui = this.ui = new UI(element, { preventScrolling:  OS.mobile })
             .on('dragStart', () => {
                 playerLeft = element.offsetLeft;
                 playerTop = element.offsetTop;
--- a/src/js/view/floating-drag-ui.js
+++ b/src/js/view/floating-drag-ui.js
@@ -2,6 +2,8 @@ import UI from 'utils/ui';
 import { style } from 'utils/css';
 import { OS } from 'environment/environment';
 
+const _isMobile = OS.mobile;
+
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/floating-drag-ui.js
```diff
@@ -20,7 +21,7 @@ export default class FloatingDragUI {
         let innerWidth;
         const auto = 'auto';
         const { element } = this;
-        const ui = this.ui = new UI(element, { preventScrolling: true })
+        const ui = this.ui = new UI(element, { preventScrolling:  OS.mobile })
```
DanFerrer: Remove extra space between ```:``` and ```OS.mobile```.

src/js/view/floating-drag-ui.js
```diff
@@ -1,5 +1,8 @@
 import UI from 'utils/ui';
 import { style } from 'utils/css';
+import { OS } from 'environment/environment';
+
+const _isMobile = OS.mobile;
```
robwalch: Don't make this a const. Use `OS.mobile` inline. It's a getter so it's better to call it later than up here where you slow down the eval of controls script.

robwalch: @pajong @DanFerrer ^^

pajong: Please see what I wrote in the PR message - it prevents click on desktop

robwalch: @pajong that is an issue with UI.js, not with setting `preventScrolling`. As I pointed out, you want to prevent scrolling on all touch devices, not just mobile. We should either update the pointer event handling or use a different element for dragging.

PR #3325, opened 2019-03-11
Floating Player Improvements [JW8-2530] [JW8-5643]

robwalch:
### This PR will...
- [JW8-2530] Maintain responsive player width with margins when floating
- [JW8-2530] Make the header margin for floating players on mobile pages 62px
- [JW8-5643] Fix UI controls while dragging floating player

- The player's size when floating should match the player bounds before floating (100% + margins)
- `height: 62px;` is a common main-header size 
- The player container and wrapper should not be draggable. These elements contain controls which are also draggable.

### DIFF
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -17,7 +17,7 @@
         margin: 0 auto;
 
         @media screen and (max-width : 480px) {
-            width: 100% !important; /* stylelint-disable-line declaration-no-important */
+            width: 100%;
             left: 0;
             right: 0;
         }
@@ -25,11 +25,11 @@
         .jw-flag-touch& {
             @media screen and (max-device-width : 480px) and (orientation: portrait) {
                 animation: jw-float-to-top 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
-                top: 125px;
+                top: 62px;
                 bottom: auto;
                 left: 0;
                 right: 0;
-                max-width: none !important; /* stylelint-disable-line declaration-no-important */
+                max-width: none;
                 max-height: none;
             }
         }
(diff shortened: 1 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/css/jwplayer/flags/floatingplayer.less
```diff
@@ -17,19 +17,19 @@
         margin: 0 auto;
 
         @media screen and (max-width : 480px) {
-            width: 100% !important; /* stylelint-disable-line declaration-no-important */
```
jnatalzia: we should really do an `important` audit _especially_ being an embedded solution this could be real bad in some cases

src/js/utils/ui.js
```diff
@@ -112,6 +112,7 @@ function initInteractionListeners(ui) {
             if (!passive) {
                 const { pointerId } = e;
                 ui.pointerId = pointerId;
+                el.setPointerCapture(pointerId);
```
jnatalzia: no safari support, i assume this is a known?

also really cool api! haven't seen this before

src/css/jwplayer/flags/floatingplayer.less
robwalch: We most commonly use `important` to override inline styles added by JS. Good news here is we're not using it anymore.

src/js/utils/ui.js
robwalch: ui.js normalizes input across browsers for Mouse, Touch and Pointer events. It listens for Mouse or Pointer events and Touch events when available.

PR #3327, opened 2019-03-13
Improve horizontal volume slider styles

pajong:
### This PR will...
- Add more padding to the right side of the horizontal volume slider when it opens, so that it is not so close with the currentTime/duration ui
- Remove tab focus outline on the container of the slider rail
- Apply css to correct classes so that `focus`/`hover` styles to increase the slider size works

- Styles were not applied to correct classes, resulting `hover` and `focus` styles to be not applied
- Horizontal volume is closer to the currentTime/duration mark than to the mute icon

#### Addresses Issue(s):
JW8-5651

### DIFF
--- a/src/css/controls/imports/slider.less
+++ b/src/css/controls/imports/slider.less
@@ -120,6 +120,10 @@
     align-items: center;
     background: transparent none;
     padding: 0 12px;
+}
+
+.jw-slider-time,
+.jw-horizontal-volume-container {
     z-index: 1;
     outline: none;
 
--- a/src/css/controls/imports/slider.less
+++ b/src/css/controls/imports/slider.less
@@ -1,5 +1,12 @@
 @import "../../shared-imports/vars.less";
 
+.jw-slider-components() {
+    backface-visibility: hidden;
+    height: 100%;
+    transform: translate(0, -50%) scale(1, 0.6);
+    transition: transform 150ms ease-in-out;
+}
+
 .jw-slider-container {
     display: flex;
     align-items: center;
@@ -122,6 +129,14 @@
     padding: 0 12px;
 }
 
+.jw-slider-time .jw-cue {
+    .jw-slider-components();
+    background-color: @background-color;
+    cursor: pointer;
+    position: absolute;
+    width: 6px;
+}
+
 .jw-slider-time,
 .jw-horizontal-volume-container {
     z-index: 1;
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/css/controls/imports/slider.less
```diff
@@ -120,6 +120,10 @@
     align-items: center;
     background: transparent none;
     padding: 0 12px;
+}
+
+.jw-slider-time,
+.jw-horizontal-volume-container {
```
robwalch: @pajong `jw-horizontal-volume-container` will never contain cues or a buffer bar right? I'm wondering how much additional css does this produce and couldn't we keep buffer and cue rules outside of vertical and volume sliders?

pajong:
comparison

@robwalch String comparison (left is before the cue changes, right is after the change) is attached above. It actually increases the line number of the compiled css by 1, but removes the cue styles not needed (and cleaner). We can undo the changes if the line number increase is a concern.

robwalch:
@pajong which is longer? Length (number of characters) is what we care about, not number of lines.

robwalch:
@pajong, the problem with your current changes is you are duplicating half of the styles now (backface..., height...) those should not be split out into a new block.

src/css/controls/imports/slider.less
```diff
@@ -120,24 +127,25 @@
...
     outline: none;
 
     .jw-rail,
     .jw-buffer,
-    .jw-progress,
-    .jw-cue {
-        backface-visibility: hidden;
-        height: 100%;
-        transform: translate(0, -50%) scale(1, 0.6);
-        transition: transform 150ms ease-in-out;
```
robwalch: Leave this part. It more efficient to share these styles (is there a way in less to omit the .jw-horizontal-volume-container > cues line?)

src/css/controls/imports/slider.less
```diff
@@ -120,24 +127,25 @@
...
-    }
-
-    .jw-cue {
-        background-color: @background-color;
-        cursor: pointer;
-        position: absolute;
-        width: 6px;
```
robwalch: Only move this part out to `.jw-slider-time .jw-cue ` to reduce the css output.

src/css/controls/imports/slider.less
pajong: I tried to find a way to omit `.jw-horizontal-volume-container > cues`, but couldn't find any solution yet

pajong:
comparison

@robwalch Above is the string comparison of the first solution with the current - I couldn't find the way to omit `.jw-horizontal-volume-container > cues`. I will update the PR if I find the way

PR #3328, opened 2019-03-13
[JW8-5672] Clear cues on model when playlist item changes

waxidiotic:
### This PR will...

* Clear the cues on the model when the playlist item changes.
* Do not call `updateCues()` separately as it will be called when the cues on the model change

* Cues were not being cleared off the model when new playlist items are loaded. This results in the cues from the last playlist item to remain and be added to the new set of cues.

n/a

n/a

#### Addresses Issue(s):

JW8-5672

### DIFF
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -177,7 +177,7 @@ class TimeSlider extends Slider {
             return;
         }
         this.reset();
-        this.updateCues(model, []);
+        model.set('cues', []);
 
         const tracks = playlistItem.tracks;
         each(tracks, function (track) {
@@ -280,12 +280,6 @@ class TimeSlider extends Slider {
     }
 
     updateCues(model, cues) {
-        // If updateCues is called with cues === [], clear the cues on the model.
-        // This will clear out the cues on the model on new playlist items.
-        if (cues && cues.length === 0) {
-            model.set('cues', cues);
-        }
-
         this.resetCues();
         if (cues && cues.length) {
             cues.forEach((ele) => {
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -177,7 +177,7 @@ class TimeSlider extends Slider {
             return;
         }
         this.reset();
-        this.updateCues(model, model.get('cues'));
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/components/timeslider.js
```diff
@@ -177,7 +177,7 @@ class TimeSlider extends Slider {
             return;
         }
         this.reset();
-        this.updateCues(model, model.get('cues'));
+        this.updateCues(model, []);
```
jnatalzia: would it be easier to get rid of the logic in `updateCues` and call `resetCues` here?

waxidiotic: `resetCues()` doesn't clear the cues on the model. I tried doing it but the build kept on failing with maximum stack call errors.

jnatalzia: oh duh, it gets called right after your change - my bad

robwalch: Just update the model and let the `change` callback handle `updateCues`"
 
`model.set('cues', []);`

src/js/view/controls/components/timeslider.js
```diff
@@ -280,6 +280,12 @@ class TimeSlider extends Slider {
     }
 
     updateCues(model, cues) {
+        // If updateCues is called with cues === [], clear the cues on the model.
+        // This will clear out the cues on the model on new playlist items.
+        if (cues && cues.length === 0) {
+            model.set('cues', cues);
```
robwalch: This line executes this method when the model is changed, so you should never change the model in this method `.on('change:cues', this.updateCues, this)`

src/js/view/controls/components/timeslider.js
robwalch: No what you want to do is get rid of the logic `updateCues` and update the model here. `updateCues` is fired when the model is changed.

waxidiotic:
test this please

PR #3330, opened 2019-03-14
[JW8-5649] Lazy DOM parsing for UI elements that are not shown on setup

robwalch:
### This PR will...
- Lazy DOM parsing in tooltips
- Use textContent instead of innerText to set UI text content

Most users will never see tooltips - at least not all of them. So we should not populate these elements until they are first shown. This will speed up player setup.

#### Addresses Issue(s):
JW8-5649

### DIFF
--- a/src/js/view/controls/components/simple-tooltip.js
+++ b/src/js/view/controls/components/simple-tooltip.js
@@ -1,4 +1,4 @@
-import { addClass, removeClass, replaceInnerHtml } from 'utils/dom';
+import { toggleClass, replaceInnerHtml } from 'utils/dom';
 
 export function SimpleTooltip(attachToElement, name, text, openCallback, closeCallback) {
     const tooltipElement = document.createElement('div');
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/components/simple-tooltip.js
```diff
@@ -28,18 +31,32 @@ export function SimpleTooltip(attachToElement, name, text, openCallback, closeCa
...
 
-    instance.setText(text);
+    const redraw = (opened) => {
+        if (opened && instance.dirty) {
+            replaceInnerHtml(textElement, instance.text);
+            instance.dirty = false;
+        }
+
+        instance.opened = opened;
+        toggleClass(tooltipElement, 'jw-open', open);
```
jnatalzia: your third arg should be `opened` right? otherwise you're passing a function each time which will fail the check in `toggleClass` for a boolean meaning it will always toggle

robwalch: updated

zetagame: approved

PR #3331, opened 2019-03-14
Clear model cues on playlist item in controller

pajong:
### This PR will...
Clear model cues on `playlistItem`change in `controller`, rather than in `timeslider`

Timeslider's `model` is the `viewModel`, which updates the `playlistItem` every time we enter/exit ads mode.
We only want to clear cues on the `playlistItem` when the actual content item changes.
Moving the `model.set('cues', []);` to be called with `_model` in `controller` (which is the actual player model, not viewModel).

PLEASE NOTE: model from `model.set('cues, [])` is different from `_model` in controller (https://github.com/jwplayer/jwplayer/compare/bugfix/reset-cues-on-playlistItem?expand=1#diff-a181c6e20596563a55e5710b8f4dbab7R250)

#### Addresses Issue(s):

JW8-5672

### DIFF
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -177,7 +177,7 @@ class TimeSlider extends Slider {
             return;
         }
         this.reset();
-        model.set('cues', []);
+        this.updateCues(model, model.get('cues'));
 
         const tracks = playlistItem.tracks;
         each(tracks, function (track) {
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -176,7 +176,7 @@ class TimeSlider extends Slider {
 
     onPlaylistItem(model, playlistItem) {
         this.reset();
-        this.updateCues(model, model.get('cues'));
+        model.set('cues', []);
 
         const tracks = playlistItem.tracks;
         each(tracks, function (track) {
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -101,16 +101,12 @@ class TimeSlider extends Slider {
                     this.updateAriaText();
                 }
             })
-            .change('playlistItem', this.onPlaylistItem, this)
             .change('position', this.onPosition, this)
             .change('buffer', this.onBuffer, this)
             .change('streamType', this.onStreamType, this);
 
         // Clear cues on player model's playlistItem change event
-        const { _model } = this._model;
-        _model.change('playlistItem', (model) => {
-            model.set('cues', []);
-        });
+        this._model.player.change('playlistItem', this.onPlaylistItem, this);
 
         const sliderElement = this.el;
         setAttribute(sliderElement, 'tabindex', '0');
@@ -179,9 +175,6 @@ class TimeSlider extends Slider {
     }
 
     onPlaylistItem(model, playlistItem) {
-        if (!playlistItem) {
-            return;
-        }
         this.reset();
         this.updateCues(model, model.get('cues'));
 
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -106,6 +106,12 @@ class TimeSlider extends Slider {
             .change('buffer', this.onBuffer, this)
             .change('streamType', this.onStreamType, this);
 
+        // Clear cues on player model's playlistItem change event
+        const { _model } = this._model;
+        _model.change('playlistItem', (model) => {
+            model.set('cues', []);
+        });
+
         const sliderElement = this.el;
         setAttribute(sliderElement, 'tabindex', '0');
         setAttribute(sliderElement, 'role', 'slider');
--- a/test/unit/controlbar-test.js
+++ b/test/unit/controlbar-test.js
@@ -14,6 +14,10 @@ describe('Control Bar', function() {
     model.mediaController = {};
     model.mediaController.on = sinon.stub();
     model.mediaController.on.returnsThis();
+    const playerModel = new SimpleModel();
+    playerModel.change = sinon.stub();
+    playerModel.change.returnsThis();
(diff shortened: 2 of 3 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/components/timeslider.js
```diff
@@ -177,7 +177,7 @@ class TimeSlider extends Slider {
             return;
         }
         this.reset();
-        model.set('cues', []);
+        this.updateCues(model, model.get('cues'));
```
waxidiotic: May not need this because `updateCues()` is called whenever the playlistItem on the model is changed.

waxidiotic: Nevermind, the cues are removed from the UI when entering ads mode.

src/js/view/controls/components/timeslider.js
```diff
@@ -106,6 +106,12 @@ class TimeSlider extends Slider {
             .change('buffer', this.onBuffer, this)
             .change('streamType', this.onStreamType, this);
 
+        // Clear cues on player model's playlistItem change event
```
pajong: I was just going to do `this._model._model`, but it looked ugly and lint complained about dangling `_`. I can disable linter for this line if that is preferred

src/js/view/controls/components/timeslider.js
```diff
@@ -106,6 +106,12 @@ class TimeSlider extends Slider {
...
+        const { _model } = this._model;
+        _model.change('playlistItem', (model) => {
```
robwalch: `this._model.player` get's the player view model. You should not access the internal model directly.

src/js/view/controls/components/timeslider.js
robwalch: Why are you not removing `.change('playlistItem', this.onPlaylistItem, this)`?

test/unit/controlbar-test.js
```diff
@@ -14,6 +14,10 @@ describe('Control Bar', function() {
     model.mediaController = {};
     model.mediaController.on = sinon.stub();
     model.mediaController.on.returnsThis();
+    const playerModel = new SimpleModel();
+    playerModel.change = sinon.stub();
+    playerModel.change.returnsThis();
+    model._model = playerModel;
```
robwalch: Remove. Use `.player`

src/js/view/controls/components/timeslider.js
```diff
@@ -177,7 +183,7 @@ class TimeSlider extends Slider {
             return;
```
robwalch: This check can be removed once you use the player view model.

src/js/view/controls/components/timeslider.js
robwalch: `this._model.player.on('change:playlistItem', this.onPlaylistItem, this);`

You'll need to update the 'Control Bar' tests as well since they use a mock SimpleModel that will need a player property added.

src/js/view/controls/components/timeslider.js
```diff
@@ -179,9 +175,6 @@ class TimeSlider extends Slider {
     }
 
     onPlaylistItem(model, playlistItem) {
-        if (!playlistItem) {
```
waxidiotic: @pajong Are we sure we don't need this check for `!playlistItem`? I looked back in the history and don't see the exact reason why this was put here though.

test/unit/controlbar-test.js
```diff
@@ -14,6 +14,10 @@ describe('Control Bar', function() {
     model.mediaController = {};
     model.mediaController.on = sinon.stub();
     model.mediaController.on.returnsThis();
+    const playerModel = new SimpleModel();
+    playerModel.change = sinon.stub();
+    playerModel.change.returnsThis();
+    model.player = playerModel;
```
pajong: this is still needed, as `SimpleModel` does not have an instance of player model.

src/js/view/controls/components/timeslider.js
pajong: It was there because we used to use `viewModel`, instead of `playerViewModel`. Now that we use the player view model, we always get `playlistItem`

src/js/view/controls/components/timeslider.js
```diff
@@ -101,11 +101,13 @@ class TimeSlider extends Slider {
                     this.updateAriaText();
                 }
             })
-            .change('playlistItem', this.onPlaylistItem, this)
             .change('position', this.onPosition, this)
             .change('buffer', this.onBuffer, this)
             .change('streamType', this.onStreamType, this);
 
+        // Clear cues on player model's playlistItem change event
+        this._model.player.change('playlistItem', this.onPlaylistItem, this);
```
pajong: Looked into using `on`, but it is not triggered on the initial setup, resulting `this.onPlaylistItem` to not trigger on the 1st playlist item

test/unit/controlbar-test.js
robwalch: 👍

src/js/view/controls/components/timeslider.js
```diff
@@ -173,11 +175,8 @@ class TimeSlider extends Slider {
     }
 
     onPlaylistItem(model, playlistItem) {
-        if (!playlistItem) {
-            return;
-        }
         this.reset();
-        model.set('cues', []);
+        this.updateCues(model, model.get('cues'));
```
robwalch: Where are cues being reset now?

pajong: Thank you! It was working fine when I was because `cues` are being updated when they load.

PR #3334, opened 2019-03-15
Clear cues before firing playlistItem event

pajong:
### This PR will...
- Reset cues before firing `playlistItem` event
- Remove `resetCues` from `reset` function so that it is not called on `change:playlistItem`

`playerViewModel`'s `change:playlistItem` fires AFTER the player `playlistItem` event, resulting ads plugins to add the cues before the timeslider's `onPlaylistItem` function call.

The cues are being reset every time there is an update to `cues` in `model`, so we do not need to call `resetCues` on playerViewModel's `change:playlistItem` - it removes already updated cues.

@robwalch We either have to do this, or listen to the player model in `timeslider` - the player view model playlist item change comes too late when non-VMAP ads are set up.

The config to reproduce the issue (2nd playlistItem): 
```
{
    "playlist": [
        {
            "sources": [
                {
                    "file": "https://content.jwplatform.com/manifests/YXkJrbkQ.m3u8"
                }
            ],
            "adschedule": {
                "adbreak1": {
                    "offset": 20,
                    "ad": {
                        "tag": "//playertest.longtailvideo.com/mid.xml"
                    }
                }
            }
        },
        {
            "sources": [
                {
                    "file": "https://content.jwplatform.com/manifests/FR1xB6OK.m3u8"
                }
            ],
            "adschedule": {
                "adbreak1": {
                    "offset": 15,
                    "ad": {
                        "tag": "//playertest.longtailvideo.com/mid.xml"
                    }
                }
            }
        }
    ],
    "advertising": {
        "client": "../../bin-debug/vast.js",
        "skipoffset": 2
    }
}
```

#### Addresses Issue(s):

JW8-5672

### DIFF
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -176,7 +176,6 @@ class TimeSlider extends Slider {
 
     onPlaylistItem(model, playlistItem) {
         this.reset();
-        this.updateCues(model, model.get('cues'));
 
         const tracks = playlistItem.tracks;
         each(tracks, function (track) {
@@ -311,7 +310,6 @@ class TimeSlider extends Slider {
 
     reset() {
         this.resetThumbnails();
-        this.resetCues();
         this.timeTip.resetWidth();
         this.textLength = 0;
     }
--- a/src/js/view/controls/components/timeslider.js
+++ b/src/js/view/controls/components/timeslider.js
@@ -176,7 +176,7 @@ class TimeSlider extends Slider {
 
     onPlaylistItem(model, playlistItem) {
         this.reset();
-        model.set('cues', []);
+        this.updateCues(model, model.get('cues'));
 
         const tracks = playlistItem.tracks;
         each(tracks, function (track) {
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/components/timeslider.js
```diff
@@ -176,7 +176,7 @@ class TimeSlider extends Slider {
 
     onPlaylistItem(model, playlistItem) {
         this.reset();
-        model.set('cues', []);
+        this.updateCues(model, model.get('cues'));
```
robwalch: Since the cues are already empty, do we need this line anymore? `updateCues` will get called when cues are added.

pajong: Yes, the `reset` clears all the cues, and the cues are being added from the ad plugins before this call

waxidiotic: @pajong - I think what Rob is saying is that whenever the cues on the model changes, `updateCues()` is called so we may not need this.

PR #3338, opened 2019-03-20
[JW8-5260] Fix tab-nabbing vulnerability by using noreferrer rel

zetagame:
### This PR will...
Implement a new "open" util function that allows custom configs.

Allows a cross-platform implementation equivelant to window.open, but with support for noreferrer, as well as any additional options we may want to add in the future.

#### Addresses Issue(s):

JW8-5260

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -209,3 +209,11 @@ export function nextSibling(element) {
 export function previousSibling(element) {
     return element.previousElementSibling;
 }
+
+export function open(link, target, additionalOptions = {}) {
+    let a = document.createElement('a');
+    a.href = link;
+    a.target = target;
+    a = Object.assign(a, { additionalOptions });
+    a.click();
+}
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -214,6 +214,6 @@ export function open(link, target, additionalOptions = {}) {
     let a = document.createElement('a');
     a.href = link;
     a.target = target;
-    a = Object.assign(a, { additionalOptions });
+    a = Object.assign(a, additionalOptions);
     a.click();
 }
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -210,7 +210,7 @@ export function previousSibling(element) {
     return element.previousElementSibling;
 }
 
-export function open(link, target, additionalOptions = {}) {
+export function openLink(link, target, additionalOptions = {}) {
     let a = document.createElement('a');
     a.href = link;
     a.target = target;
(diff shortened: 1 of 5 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/utils/dom.js
```diff
@@ -209,3 +209,11 @@ export function nextSibling(element) {
 export function previousSibling(element) {
     return element.previousElementSibling;
 }
+
+export function open(link, target, additionalOptions = {}) {
+    let a = document.createElement('a');
```
jnatalzia: let's write some tests!

src/js/utils/dom.js
```diff
@@ -209,3 +209,11 @@ export function nextSibling(element) {
...
+    a.href = link;
+    a.target = target;
+    a = Object.assign(a, additionalOptions);
+    a.click();
```
jnatalzia: are there any potential browser security features that would block this click vs. a `window.open`?

src/js/utils/dom.js
jnatalzia: you should be able to mock `document.createElement` temporarily for the purpose of the test so you can assert the result of these changes.

src/js/utils/dom.js
```diff
@@ -209,3 +209,11 @@ export function nextSibling(element) {
 export function previousSibling(element) {
     return element.previousElementSibling;
 }
+
+export function open(link, target, additionalOptions = {}) {
```
jnatalzia: this is super pedantic but `openLink` might be more descriptive here - less potential for variable collisions and whatnot too

src/js/utils/dom.js
zetagame: Nope! The only issues that prevent click are on really old versions of safari we don't officially support:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/click

jnatalzia: 👍

test/unit/dom-test.js
```diff
@@ -393,4 +394,28 @@ describe('dom', function() {
         expect('bounds should not be empty when element has layout').to.not.equal(bounds(element), emptyBound);
     });
 
+    it.only('opens a link', function() {
```
jnatalzia: remove the `only` here

PR #3340, opened 2019-03-21
Continue floating on complete

jnatalzia:
### This PR will...
Update the player so it no longer stops floating when entering the complete state.

Players should remain in the floating position on complete unless scrolling away from an originally positioned player and in final complete state (no playlist). In that case the player should not float.

Nope

None

#### Addresses Issue(s):

JW8-5717

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -726,7 +726,6 @@ function View(_api, _model) {
             case STATE_IDLE:
             case STATE_ERROR:
             case STATE_COMPLETE:
-                _this.stopFloating();
                 if (_captionsRenderer) {
                     _captionsRenderer.hide();
                 }
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -722,12 +722,10 @@ function View(_api, _model) {
         }
         replaceClass(_playerElement, /jw-state-\S+/, 'jw-state-' + state);
 
-        if (state === STATE_ERROR) {
-            _this.stopFloating();
-        }
-
         switch (state) {
             case STATE_ERROR:
+                _this.stopFloating();
+                 /* falls through to update captions renderer */
             case STATE_IDLE:
             case STATE_COMPLETE:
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/view.js
```diff
@@ -723,14 +723,18 @@ function View(_api, _model) {
...
             case STATE_ERROR:
-            case STATE_COMPLETE:
                 _this.stopFloating();
                 if (_captionsRenderer) {
                     _captionsRenderer.hide();
                 }
                 break;
+            case STATE_IDLE:
+            case STATE_COMPLETE:
+                if (_captionsRenderer) {
```
jnatalzia: really not crazy about the duplicate logic but the linter yells if there's not a break statement in the first `if STATE_ERROR` case

robwalch: I would prefer an `if (state === STATE_ERROR) {  _this.stopFloating(); }` block to these duplicate cases.

jnatalzia: updating!

robwalch: Oh now I see. The other way to do it is put STATE_ERROR first and add a `/* falls through */` comment where the break would be. Ex:

https://github.com/jwplayer/jwplayer/blob/ab663ef80aacba02d295b784ecbdb93461a0ff4c/src/js/program/program-listeners.js#L98

jnatalzia: Yeah, our linter doesn't allow two of the same case in a single switch though so we couldn't do that since the fallthrough would _also_ need to catch `STATE_ERROR`. that was my first fix but got a big red ❌

PR #3344, opened 2019-03-26
Normalize floating config in the model

jnatalzia:
### This PR will...
Normalize the config in the model to ensure it is fully up to date before we setup.

The normalization of the floating block that we were doing before was not working with the a/b test module as that module modifies the config after the initialization but before setup. We need a step to ensure that by the time that data is integrated into the model that is it fully up to date with the expected config.

Nah

Nah

#### Addresses Issue(s):

JW8-5775

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -216,8 +216,8 @@ class Model extends SimpleModel {
     _normalizeConfig(cfg) {
         const floating = cfg.floating;
 
-        if (!floating || !!floating.disabled) {
-            cfg.floating = null;
+        if (floating && !!floating.disabled) {
+            delete cfg.floating;
         }
     }
 }
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -19,6 +19,8 @@ class Model extends SimpleModel {
     }
 
     setup(config) {
+        config = config || {};
+        this._normalizeConfig(config);
         Object.assign(this.attributes, config, INITIAL_PLAYER_STATE);
         this.providerController = new Providers(this.getConfiguration());
         this.setAutoStart();
@@ -209,6 +211,14 @@ class Model extends SimpleModel {
             return;
         }
         this.set('bandwidthEstimate', bwEstimate);
+    }
+
+    _normalizeConfig(cfg) {
+        const floating = cfg.floating;
+
+        if (!floating || !!floating.disabled) {
+            cfg.floating = null;
+        }
     }
 }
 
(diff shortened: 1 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/model.js
```diff
@@ -210,6 +212,14 @@ class Model extends SimpleModel {
         }
         this.set('bandwidthEstimate', bwEstimate);
     }
+
+    _normalizeConfig(cfg) {
+        const floating = cfg.floating;
+
+        if (!floating || !!floating.disabled) {
+            cfg.floating = null;
```
jnatalzia: @robwalch i know we talked about using an explicit object here (`floating: { disabled: false }` when enabled `floating: { disabled: true }` when not) however given the implementation, i would have to update 3-4 checks against the model's floating config as it expects a null value currently when disabling.

i could make those updates, however i do worry about how that leads to future changes where we'll have to add more conditional logic around the `disabled` attribute in places where we're checking against it, vs just doing it once in the model and forgetting about it. thoughts?

src/js/controller/model.js
```diff
@@ -19,6 +19,8 @@ class Model extends SimpleModel {
     }
 
     setup(config) {
+        config = config || {};
```
jnatalzia: currently this is in place because we have failing unit tests that try to setup a model without a config (previously it wasn't breaking behavior). the alternative is to early exit in `_normalizeConfig`, this seemed like a better catch-all

src/js/controller/model.js
jnatalzia: i suppose if we're certain we have a floating block we could remove the first half of the `floatingConfig && !floatingConfig.disabled` check. if `_normalizeConfig` were to always set a floating block could we be certain we would always have one?

zetagame: Do we have additional customizations for floating? If not maybe it should be a boolean with either floating: true, floating: false. Totally understand if existing use cases prevent this.

jnatalzia: `dismissable` is a config option that allows the player to be dismissed

robwalch: What do you think about doing `delete cfg.floating` rather than setting it to `null`? That would remove the property and prevent it from showing up when cloning the model attributes with `getConfig()`.

jnatalzia: I did the explicit null so we have a track record that it was updated by the code rather than undefined which doesn't tell us if a block was initially passed in - though we'd be able to
 get that from setup config so i think i like this approach better, will update!

jnatalzia: updated if you get a chance for another look @robwalch @zetagame

PR #3347, opened 2019-03-27
Add config option for disabling the next up overlay for an A/B test

jnatalzia:
### This PR will...
Adds normalization functionality to disabled the next up overlay upon model instantiation

we need the ability to disable this feature through an a/b test config option

None that I can think of

Nope

#### Addresses Issue(s):

JW8-5657

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -219,6 +219,10 @@ class Model extends SimpleModel {
         if (floating && !!floating.disabled) {
             delete cfg.floating;
         }
+
+        if (typeof cfg.abHideNextUpDisplay !== 'undefined') {
+            cfg.nextUpDisplay = !cfg.abHideNextUpDisplay;
+        }
     }
 }
 
--- a/test/unit/model/model-test.js
+++ b/test/unit/model/model-test.js
@@ -35,30 +35,57 @@ describe('Model', function() {
     });
 
     describe('#_normalizeConfig', () => {
-        it('should leave the floating block untouched when disabled is not true', () => {
-            let startingCfg = {
-                floating: { disabled: false }
-            };
-            model._normalizeConfig(startingCfg);
-            expect(startingCfg.floating).to.deep.eq({ disabled: false });
+        describe('floating block', () => {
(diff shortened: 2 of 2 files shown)


src/js/controller/model.js
```diff
@@ -219,6 +219,10 @@ class Model extends SimpleModel {
         if (floating && !!floating.disabled) {
             delete cfg.floating;
         }
+
+        if (typeof cfg.abHideNextUpDisplay !== 'undefined') {
```
jnatalzia: I opted to add to the `_normalizeConfig` method in the model for the same reason as `floating`. It centralizes this logic and ensures that any downstream consumers of the model config do not have to know or care about ab tests. It also (potentially) centralizes our ab config code to make it easier in future to clean up

test/unit/model/model-test.js
```diff
@@ -35,30 +35,57 @@ describe('Model', function() {
     });
 
     describe('#_normalizeConfig', () => {
-        it('should leave the floating block untouched when disabled is not true', () => {
-            let startingCfg = {
-                floating: { disabled: false }
-            };
-            model._normalizeConfig(startingCfg);
-            expect(startingCfg.floating).to.deep.eq({ disabled: false });
+        describe('floating block', () => {
```
jnatalzia: only change to the floating block tests was scoping them to their own describe block

johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/model.js
robwalch: But config.js doesn't remove or chan `nextUpDisplay, it only ensures a default of true.

I don't see how this helps if we want to avoid allowing a test to override an option specified by the user.

jnatalzia: The ab-test PR would be a follow-on however your comment makes me think that these changes aren't necessary at all. The ab code can simply set `nextUpDisplay: true/false` depending on the variant selected. Will close

PR #3356, opened 2019-04-08
Consider all falsy status codes a CORS error

johnBartos:
Fetch returns with an undefined status code if encountering a CORS error, whereas XHR returns with 0.

Should I put this code in the Hls.js provider (`code = code || 0`), or handle it here? I figure it'll be helpful for other providers too.

No

#### Addresses Issue(s):
Exposed by JW8-2565

### DIFF
--- a/src/js/providers/utils/network-error-parser.js
+++ b/src/js/providers/utils/network-error-parser.js
@@ -12,7 +12,8 @@ export default function parseNetworkError(baseCode, statusCode, url) {
         code += clampStatus(statusCode);
     } else if (('' + url).substring(0, 5) === 'http:' && document.location.protocol === 'https:') {
         code += 12;
-    } else if (statusCode === 0) {
+    } else if (!statusCode) {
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/providers/utils/network-error-parser.js
```diff
@@ -12,7 +12,8 @@ export default function parseNetworkError(baseCode, statusCode, url) {
         code += clampStatus(statusCode);
     } else if (('' + url).substring(0, 5) === 'http:' && document.location.protocol === 'https:') {
         code += 12;
-    } else if (statusCode === 0) {
+    } else if (!statusCode) {
```
robwalch: We always expected a number here. Maybe we should handle this higher up the chain - like in HLS.js for example.

PR #3359, opened 2019-04-09
[DNM] [JW8-8782] autostart / autoPause Improvements

waxidiotic:
### This PR will...

* Call `_checkPauseOnViewable()` inside of `attachMedia()`, if `autoPause.viewability` is enabled.
* Set `noResume = true` on the instream adapter on any autostart player on mobile, regardless of `autoPause` setting.

* If `autoPause` and `autostart` was enabled on desktop, if the player begins pre-roll playback out of view, the content will begin playing after the ad (it shouldn't - the player should be paused after the pre-roll).
* Regardless if `autoPause` is set, if `autostart` is enabled on mobile, the content would play after ad breaks if not visible.

I noticed that, inside of `_play()`, we are calling `checkPauseOnViewable()` on the next playlist item. This would cause players to pause on the next item if the player is not visible, regardless of `autoPause` setting. Pretty sure we should add a check to see if `autoPause.viewability` is set here, but wanted to double-check.

https://github.com/jwplayer/jwplayer-commercial/pull/6460 (manual test page)

#### Addresses Issue(s):

JW8-8782

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -2,9 +2,9 @@
 Thanks for considering contributing to our repo! We appreciate all contributions to the player, large or small. Please take the time to read through this document before making a contribution to ensure a fun and effective process for everyone involved.
 
 ## General Guidelines
-Are you having trouble getting started with JW Player, configuration, or customization? If so, please check the [support docs](https://support.jwplayer.com/) before submitting an issue here. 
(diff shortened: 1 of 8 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3360, opened 2019-04-09
Unit tests - Ensure translations are always complete

karimMourra:
### This PR will...
add a unit test to check that all JSON translations have the same keys as the english defaults. The comparison checks that each property is of the same type, and uses recursion when a property is an object.
This unit test will ensure that fields are added to the default localization, without forgetting the translation. It will also prevent contributors from submitting incomplete translations.   
n/a
n/a
#### Addresses Issue(s):

n/a

### DIFF
--- a/test/unit/language-test.js
+++ b/test/unit/language-test.js
@@ -257,25 +257,28 @@ describe('languageUtils', function() {
             expect(languageCodes).to.deep.equal(translatedLanguageCodes);
         });
 
-        it('should have same structure as localization default', function() {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3365, opened 2019-04-16
[JW8-8812] Implement percentage based Next Up offset

zetagame:
### This PR will...
Allow a user to enter a percentage value for next up offset. This will be positive (distance from start of video) not negative (distance from end of video).

Currently it is only possible to offset next up overlay by a positive or negative number.

Where can I change the regex match for the harness incorrect format warning?

https://github.com/jwplayer/jwplayer-commercial/pull/6538
#### Addresses Issue(s):

JW8-8812


src/js/view/controls/nextuptooltip.js
```diff
@@ -164,10 +164,22 @@ export default class NextUpTooltip {
             return;
         }
 
-        // Use nextupoffset if set or default to 10 seconds from the end of playback
-        let offset = seconds(model.get('nextupoffset') || -10);
-        if (offset < 0) {
-            // Determine offset from the end. Duration may change.
+        let offset;
+        let isPercentage;
+        let configOffset = model.get('nextupoffset');
```
zetagame: make configOffset a const

johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/nextuptooltip.js
```diff
@@ -164,10 +164,22 @@ export default class NextUpTooltip {
...
+
+        if (configOffset) {
+            isPercentage = configOffset[configOffset.length - 1] === '%';
+            if (isPercentage) {
+                //  if it's a percentage, normalize the value and create offset.
+                offset = seconds((parseInt(configOffset) / 100) * duration);
```
zetagame: Consider using parseFloat instead of parseInt, in case they don't use whole number percentages.

src/js/view/controls/nextuptooltip.js
```diff
@@ -164,10 +164,22 @@ export default class NextUpTooltip {
...
+            } else {
+                offset = seconds(model.get('nextupoffset'));
```
zetagame: already stored as configOffset, don't re-call from the model.

src/js/view/controls/nextuptooltip.js
```diff
@@ -164,10 +164,22 @@ export default class NextUpTooltip {
...
+            }
+        } else {
+            offset = seconds(-10);
+        }
+        if (!isPercentage) {
             offset += duration;
```
zetagame: applying seconds format should be last step

zetagame: Only add duration if it's a negative number

src/js/view/controls/nextuptooltip.js
```diff
@@ -164,13 +164,29 @@ export default class NextUpTooltip {
...
+                    offset = (percentOffset / 100) * duration;
+                }
+            } else {
+                offset = configOffset;
+            }
+            //  If offset is hh:mm:ss at this point, make sure to convert to seconds.
+            offset = seconds(offset); 
         }
 
+        if (!isPercentage && offset < 0) {
```
zetagame: Redundant percentage check. Remove this.

robwalch:
I think we allow any positive number (not limited to integers).

src/js/view/controls/nextuptooltip.js
```diff
@@ -164,10 +164,27 @@ export default class NextUpTooltip {
...
         }
 
-        // Use nextupoffset if set or default to 10 seconds from the end of playback
-        let offset = seconds(model.get('nextupoffset') || -10);
+        const configOffset = model.get('nextupoffset');
+        let offset = -10;
+        let isPercentage;
+
+        if (configOffset) {
+            isPercentage = configOffset[configOffset.length - 1] === '%';
```
robwalch: `configOffset` is not guaranteed to be a string. `(10).length` and `(10)[1]` are `undefined`.

robwalch:
> Where can I change the regex match for the harness incorrect format warning?

See config_schema.json in jwplayer-commercial. When you make changes to setup options you should update the schema and copy the changes in the worker used by ace editor in the harness.

This PR adds gapro options:
https://github.com/jwplayer/jwplayer-commercial/pull/6498/files#diff-3d14f8140f197cf28024ef9081d262bc
And copies those file contents into this worker used by ace editor:
https://github.com/jwplayer/jwplayer-commercial/pull/6498/files#diff-53ca262ee25a58f899150f6b4c60c218

src/js/view/controls/nextuptooltip.js
zetagame: this would set isPercentage to undefined which would evaluate 'false' in the if.

zetagame: I believe this would evaluate to false in the isPercentage check and prevent any of the string methods from being called.

zetagame: While that is the case, given a non-string value configOffset[configOffset.length - 1] would be undefined, and isPercentage will be false, preventing any of the new logic from executing.

robwalch: You missed the most important part of the review. Use `offsetToSeconds`.

zetagame:
test this please

yarn-error.log
```diff
@@ -0,0 +1,8413 @@
+Arguments:
```
robwalch: The yarn error log should not be added. (you can add it to .gitignore to avoid this in the future)

src/js/view/controls/nextuptooltip.js
```diff
@@ -164,10 +164,15 @@ export default class NextUpTooltip {
...
+        const configOffset = model.get('nextupoffset');
+        let offset = -10;
+
+        if (configOffset) {
+            offset = offsetToSeconds(configOffset, duration);
+        }
+
         if (offset < 0) {
-            // Determine offset from the end. Duration may change.
+            //  Determine offset from the end. Duration may change.
```
robwalch: extra space

zetagame: fixed

PR #3368, opened 2019-04-18
[JW8-8782] autoPause/autostart refinements

waxidiotic:
### This PR will...

* Call `_checkPauseOnViewable()` inside of `attachMedia()`, if `autoPause.viewability` is enabled.
* Set `noResume = true` on the instream adapter on any autostart player on mobile, regardless of `autoPause` setting.

* If `autoPause` and `autostart` was enabled on desktop, if the player begins pre-roll playback out of view, the content will begin playing after the ad (it shouldn't - the player should be paused after the pre-roll).
* Regardless if `autoPause` is set, if `autostart` is enabled on mobile, the content would play after ad breaks if not visible.

* There is a manual test page in the commercial repository.
* There is also a link in the JIRA ticket for a Google Sheet I created with various scenarios.

n/a

#### Addresses Issue(s):

JW8-8782

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -345,19 +345,10 @@ Object.assign(Controller.prototype, {
             }
         }
 
-        function _checkPlayOnViewable(model, viewable) {
-            if (model.get('playOnViewable')) {
-                if (viewable) {
-                    const reason = 'viewable';
-                    if (model.get('state') === STATE_IDLE) {
-                        _autoStart(reason);
-                    } else {
-                        _play({ reason });
-                    }
-                } else if (OS.mobile && !_getAdState()) {
-                    _this.pause({ reason: 'autostart' });
-                    _model.set('playOnViewable', true);
-                }
+        function _pauseAfterAd(viewable) {
+            _this._instreamAdapter.noResume = !viewable;
+            if (!viewable) {
+                _updatePauseReason({ reason: 'viewable' });
             }
         }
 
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -503,7 +514,7 @@ Object.assign(Controller.prototype, {
                     mediaPool.prime();
                 }
 
-                if (playReason === 'playlist') {
+                if (playReason === 'playlist' && _model.get('autoPause').viewability) {
```
zetagame: is _model.get('autoPause') always defined and an object?

waxidiotic: Yup, we default it to:

```json
autoPause: {
  viewability: false
}
```
https://github.com/jwplayer/jwplayer/blob/master/src/js/api/config.js#L13

waxidiotic:
test this please

src/js/controller/controller.js
```diff
@@ -354,17 +369,16 @@ Object.assign(Controller.prototype, {
...
                 }
-            }
-        }
 
-        function _pauseWhenNotViewable(viewable) {
-            if (!viewable) {
-                _this.pause({ reason: 'viewable' });
-                _model.set('playOnViewable', !viewable);
+                if (OS.mobile && adState) {
+                    // If during an ad on mobile, we should always be paused after the ad,
```
jnatalzia: just want to be clear here, regardless of any autopause setting we _always_ want to pause after an ad? is this for a specific case? it feels to me like in most cases we would want to go straight to the actual content after an ad

waxidiotic: @jnatalzia If it's not viewable and `model.get('playOnViewable') === true`, then yes, the content should pause after the ad.

jnatalzia: ahhhh missed the `playOnViewable` with the weird diff 👍

egreaves:
> The changes here look good. I do worry about the complexity of autopause/start functionality and where that's documented. Do you have a sense as to if there's a good way to describe the behavior that we can keep up to date? Would help refactoring the functionality in future if we get to a point after incremental updates where there's a more logical overhaul

 @jnatalzia Agreed. @aliciahurst created a matrix for this in the ticket.

waxidiotic:
> > The changes here look good. I do worry about the complexity of autopause/start functionality and where that's documented. Do you have a sense as to if there's a good way to describe the behavior that we can keep up to date? Would help refactoring the functionality in future if we get to a point after incremental updates where there's a more logical overhaul
> 
> @jnatalzia Agreed. @aliciahurst created a matrix for this in the ticket.

@egreaves - I also shared the test results Google Sheet I made with @jnatalzia .

PR #3371, opened 2019-04-23
Fix native fullscreen handler for ad program controller

pajong:
### This PR will...
- Listen to `Provider` for `fullscreenchange` event instead of listening to `model` within `AdProgramController`

- Native fullscreen is handled and triggered by `html5.js`. Listening to `model` to trigger `fullscreenchange` event does not do anything. 

- Are there any cases where we would need to listen to `model` instead of `Provider`?

#### Addresses Issue(s):
JW8-2326

### DIFF
--- a/src/js/program/ad-program-controller.js
+++ b/src/js/program/ad-program-controller.js
@@ -113,6 +113,7 @@ export default class AdProgramController extends ProgramController {
             event.oldstate = model.get(PLAYER_STATE);
             adMediaModelContext.set('mediaState', event.newstate);
         });
+        this.off();
         this.on('fullscreenchange', this._nativeFullscreenHandler, this);
         adMediaModelContext.on('change:mediaState', (changeAdModel, state) => {
             this._stateHandler(state);
@@ -143,6 +144,7 @@ export default class AdProgramController extends ProgramController {
     destroy() {
         const { model, mediaPool, playerModel } = this;
         model.off();
+        this.off();
 
         // We only use one media element from ads; getPrimedElement will return it
         const mediaElement = mediaPool.getPrimedElement();
--- a/src/js/program/ad-program-controller.js
+++ b/src/js/program/ad-program-controller.js
@@ -3,6 +3,7 @@ import ProgramController from 'program/program-controller';
 import Model from 'controller/model';
 import changeStateEvent from 'events/change-state-event';
 import SharedMediaPool from 'program/shared-media-pool';
+import { NATIVE_FULLSCREEN } from "../events/events";
 
 export default class AdProgramController extends ProgramController {
     constructor(model, mediaPool) {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/program/ad-program-controller.js
```diff
@@ -113,6 +113,7 @@ export default class AdProgramController extends ProgramController {
             event.oldstate = model.get(PLAYER_STATE);
             adMediaModelContext.set('mediaState', event.newstate);
         });
+        this.off();
```
robwalch: `this.off` removes all event listeners, not just 'fullscreenchange' for context `this`.

robwalch: `this.off` removes all event listeners, not just 'fullscreenchange' for context `this`.

src/js/program/ad-program-controller.js
```diff
@@ -113,6 +113,7 @@ export default class AdProgramController extends ProgramController {
...
         this.on('fullscreenchange', this._nativeFullscreenHandler, this);
```
robwalch: You really should listening to events on `this`. It means you are ignoring the the code path that actually triggers the event.

Instead either add a check for 'fullscreenchange' in the `provider.on('all'` handler, or add a new event listener on the provider or source(s) of the event itself.

In events.js
`NATIVE_FULLSCREEN = 'fullscreenchange'`

robwalch: Note that this event should be routed through the router.So if it's not, that is what you should be fixing.

src/js/program/ad-program-controller.js
```diff
@@ -114,6 +114,9 @@ export default class AdProgramController extends ProgramController {
             event.oldstate = model.get(PLAYER_STATE);
             adMediaModelContext.set('mediaState', event.newstate);
         });
+        provider.on(NATIVE_FULLSCREEN, (event) => {
+            this._nativeFullscreenHandler(Object.assign({}, event, { type: NATIVE_FULLSCREEN }));
```
robwalch: That looks like it's doing too much.

How about this:
```js
provider.on(NATIVE_FULLSCREEN, this._nativeFullscreenHandler, this);

// Then, since you already know the type of event you are triggering:
_nativeFullscreenHandler(evt) {
    this.model.trigger(NATIVE_FULLSCREEN, evt);
    this.trigger(FULLSCREEN, {
        fullscreen: evt.jwstate
    });
}
```

PR #3373, opened 2019-04-23
Allow setting of playbackRates and playbackRateControls

jnatalzia:
### This PR will...
Allow users to reset playbackRates and playbackRateControls mid-playback

The SDKs need to be able to remove unsupported playback rates as well as disable controls until rates are known.

None that I can think of beyond checking to see if I overlooked a simpler solution

Nope 

#### Addresses Issue(s):

JW8-2685

### DIFF
--- a/src/js/view/controls/settings-menu.js
+++ b/src/js/view/controls/settings-menu.js
@@ -12,11 +12,10 @@ import {
 
 export function createSettingsMenu(controlbar, onVisibility, localization) {
     const settingsButton = controlbar.elements.settingsButton;
-    const onMenuEmpty = () => {
-        settingsButton.hide();
-    };
-
-    const settingsMenu = SettingsMenu(onVisibility, onMenuEmpty, localization);
+    const settingsMenu = SettingsMenu(onVisibility, {
+        hide: () => settingsButton.hide(),
+        show: () => settingsButton.show()
+    }, localization);
 
     controlbar.on('settingsInteraction', (submenuName, isDefault, event) => {
         const submenu = settingsMenu.getSubmenu(submenuName);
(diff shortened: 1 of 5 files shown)


src/js/view/controls/settings-menu.js
```diff
@@ -204,6 +203,10 @@ export function setupSubmenuListeners(settingsMenu, controlbar, viewModel, api)
         }
     }, settingsMenu);
 
+    model.change('playbackRateControls', () => {
```
robwalch: Does this needs to be invoked immediately? `model.change` is vs `model.on('change:...` which is only invoked after a change to the current value.

src/js/view/controls/settings-menu.js
```diff
@@ -12,11 +12,10 @@ import {
...
 export function createSettingsMenu(controlbar, onVisibility, localization) {
     const settingsButton = controlbar.elements.settingsButton;
-    const onMenuEmpty = () => {
-        settingsButton.hide();
-    };
-
-    const settingsMenu = SettingsMenu(onVisibility, onMenuEmpty, localization);
+    const settingsMenu = SettingsMenu(onVisibility, {
+        hide: () => settingsButton.hide(),
+        show: () => settingsButton.show()
```
robwalch: Is this hiding all settings or just the rate tab in the settings?

Ideally we only hide/show or add/remove the tab, and the menu button is shown or hidden based on the number of settings tabs.

robwalch: I get it now.

src/js/view/controls/settings-menu.js
jnatalzia: Ooo nice, will update

Issue #2005, opened 2017-04-24
Default captions styles should not include max-height calc()


robwalch:
In order for captions to display correctly in chromeless mode we need to use the ".jw-flag-controls-hidden" class. (See #2004)

By default captions should take up the entire player. Only when the controlbar is shown should captions need to be moved by setting max-height. In general the styling of any element whose position is adjusted when controls are visible, should default to layout without controls.

Currently this is only possible using ":not()" selectors that rule out controls being hidden.

Another similar problem is that the controls-hidden rules are duplicated within the user-inactive less file. See ".jw-flag-user-inactive".

Unifying rules for showing and hiding controls in one less file would help reduce duplication.

robwalch:
Should be addressed with https://jwplayer.atlassian.net/browse/JW8-2660

PR #3379, opened 2019-04-25
Fix iOS native fullscreen captions position

pajong:
### This PR will...
- Remove max-height to the native captions on iOS native fullscreen.
- Remove `max-height` on default captions, and apply it when the controls are showing
- Remove css functions that are not being used

- The styles added to the native captions container positions the captions too high on iPhone fullscreen.
- [Tech debt:](https://github.com/jwplayer/jwplayer/issues/2005) we should not put `max-height` by default and then removing that when the controlbar is active. Instead, we should only apply when the controlbar is active.

- This PR is adding a new class `jw-flag-ios-fullscreen`: please comment if there is a better way or better suggestions

#### Addresses Issue(s):

JW8-2660


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/view.js
```diff
@@ -558,6 +558,7 @@ function View(_api, _model) {
             const provider = model.getVideo() || instreamProvider;
             if (provider && provider.setFullscreen) {
                 provider.setFullscreen(state);
+                toggleClass(_playerElement, 'jw-flag-ios-fullscreen', state);
```
robwalch: This class should be added in response to the native fullscreen event, not following the request to go fullscreen.

src/js/view/view.js
```diff
@@ -636,6 +636,10 @@ function View(_api, _model) {
         return provider.getFullScreen();
     }
 
+    function _nativeFullscreenChangeHandler(event) {
+        toggleClass(_playerElement, 'jw-flag-ios-fullscreen', event.jwstate);
```
pajong: Do we need to do something like this?
`const newState = (event.jwstate !== undefined) ? event.jwstate : _isNativeFullscreen();`
I see that the event is coming from the provider and it is always adding `jwstate`, but the `_fullscreenChangeHandler` below gets the newState like that. Am I missing something?

robwalch: You should only add the class when `jwstate` is `true`, and the os is `ios`.

pajong: `NATIVE_FULLSCREEN` event only fires with `event.jwstate = true` if it is iOS fullscreen.
This is the only place where `NATIVE_FULLSCREEN` is fired: (https://github.com/jwplayer/jwplayer/blob/master/src/js/providers/html5.js#L670)

robwalch: You think so, but it’s not guaranteed. For that reason, you should verify the iOS.

Issue #3370, opened 2019-04-23
Unable to get property 'contains' of undefined or null reference


esetnik:
### Expected Behavior

No error on Edge browser

### Actual Behavior

Sentry is reporting errors on Edge browser:

Unable to get property 'contains' of undefined or null reference
```
{snip}  t.hideMenu()},3e3),!1}},{key:"hideMenu",value:function(e){e&&this.el.contains(e.target)||(Object(l.n)(this.playerContainer,"jw-flag-rightcl {snip}
```

### Steps to reproduce

Unknown

### Environment

- *Windows 10*
- *Edge 17.17134*
- */player/v/8.8.2/jwplayer.core.controls.html5.js*

jnatalzia:
Hey @esetnik do you have any additional information? I understand that Sentry is reporting errors but does this happen on all plays on this browser? Specific embeds? Certain configurations? 

It's hard for us to narrow down the issue unless we know the scope. Any additional information (especially reproduction steps) would be extremely helpful here.

esetnik:
@jnatalzia totally understand that this is tricky to debug without steps to reproduce. I don't have any more detail about this issue right now as it's only occurred once and I can't reproduce it.

I did a bit of digging and see a potential issue which could be causing this. The offending line is most likely: https://github.com/jwplayer/jwplayer/blob/c1b9ce108832be0a7d98f19c1ea737c90aadf73e/src/js/view/controls/rightclick.js#L107
```
    hideMenu(evt) {
        if (evt && this.el.contains(evt.target)) {
            // Do not hide menu when clicking inside menu
            return;
        }

        removeClass(this.playerContainer, 'jw-flag-rightclick-open');
        removeClass(this.el, 'jw-open');
    }
```
`this.el.contains()` would only be called when `evt` is truthy. And the only time `hideMenu` is called with a value is when it is bound to `hideMenuHandler`https://github.com/jwplayer/jwplayer/blob/c1b9ce108832be0a7d98f19c1ea737c90aadf73e/src/js/view/controls/rightclick.js#L134

There are 4 instances where `hideMenuHandler` is added as an event listener:
https://github.com/jwplayer/jwplayer/blob/c1b9ce108832be0a7d98f19c1ea737c90aadf73e/src/js/view/controls/rightclick.js#L169
https://github.com/jwplayer/jwplayer/blob/c1b9ce108832be0a7d98f19c1ea737c90aadf73e/src/js/view/controls/rightclick.js#L170
[...]

PR #3382, opened 2019-04-30
prevent crash in async click handler when el is null

esetnik:
### This PR will...
Prevents an asynchronous crash in `hideMenu` when document click handler results in player remove().

On Edge browser if the user calls remove() inside the document click handler there will be crash because the hideMenu function is called with PointerEvent `evt` after destroy() sets `this.el` to null.

#### Addresses Issue(s):
#3370

### DIFF
--- a/src/js/view/controls/rightclick.js
+++ b/src/js/view/controls/rightclick.js
@@ -119,7 +119,7 @@ export default class RightClick {
     }
 
     hideMenu(evt) {
-        if (evt && this.el.contains(evt.target)) {
+        if (evt && this.el && this.el.contains(evt.target)) {
             // Do not hide menu when clicking inside menu
             return;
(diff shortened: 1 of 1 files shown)


src/js/view/controls/rightclick.js
```diff
@@ -104,7 +104,7 @@ export default class RightClick {
     }
 
     hideMenu(evt) {
-        if (evt && this.el.contains(evt.target)) {
+        if (evt && this.el && this.el && this.el.contains(evt.target)) {
```
robwalch: It looks like `this.el && this.el &&` was repeated by accident.

esetnik: Sorry about that, don't know how it snuck in there twice 😄

PR #3386, opened 2019-05-01
remove touchstart event listener

esetnik:
### This PR will...
Remove the registered `touchstart` event listener from wrapper element

When `addHideMenuHandlers()` is called it adds an event listener for `touchstart` events on the `wrapperElement` but this event listener is never removed.

See https://github.com/jwplayer/jwplayer/blob/master/src/js/view/controls/rightclick.js#L184


robwalch:
Thanks @esetnik. Good catch!

PR #3393, opened 2019-05-03
Update Cues on setControls

zetagame:
### This PR will...
Update cues when they are not up-to-date when then onPlaylistItem event fires (the event currently fires regardless of whether or not a playlist is present, and is used to update chapters).

When controls are disabled and re-enabled via setControls, the cues are removed from timeSlider and never re-added.

Is there a property I could use to prevent having to get cues from the model?

#### Addresses Issue(s):

JW8-8833


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/components/timeslider.js
```diff
@@ -177,6 +177,12 @@ class TimeSlider extends Slider {
     onPlaylistItem(model, playlistItem) {
         this.reset();
 
+        //  If cues have been cleared from slider but exist on model, update cues.
+        const cues = model.get('cues');
+        if (!this.cues.length && cues.length) {
```
jnatalzia: i would be more curious as to where the cues visuals are getting out of sync with the model cues, i would hope that anytime the cues change on the model, that change is reflected in the ui. feels like finding that will be a more holistic fix rather than forcing a sync at this point in the execution flow. let me know if i'm misunderstanding the issue!

zetagame: Here's the event order to better understand my decision to put this back here.

Setup:
1. ) Time slider sets up. In setup, event listener for onPlaylistItem is added.
2.) Vast plugin sends cues, which triggers add cues -> setCues -> updateCues -> drawCues
3.) Duration hasn't been updated yet, so drawCues returns without doing anything.
4.) durationchange triggers series of sets/additional triggers leading to onDuration cb. Much of this is obfuscated by backbone.events.
5.) onDuration triggers drawCues, now they're painted in!
6.) jwplayer().setControls() to disable controls.
7.) jwplayer().setControls() to enable controls.
8.) Process repeats, but without the vast sendCues step.

src/js/utils/scriptloader.js
```diff
@@ -18,21 +18,20 @@ function makeStyleLink(styleUrl) {
     return link;
 }
 
-function makeScriptTag(scriptUrl) {
+function makeScriptTag(scriptUrl, timeout) {
```
jnatalzia: expected changes?

src/js/view/controls/components/timeslider.js
jnatalzia: Shouldn't `onDuration` trigger draw cues or is the problem that `setControls` removes the cues? I'm trying to figure out where the cues are being removed. If they are removed in `setControls` that's a little strange to me. It seems like we could hide the controls without being destructive to the underlying cues on the model

zetagame: > Shouldn't onDuration trigger draw cues or is the problem that setControls removes the cues?

That's correct, onDuration triggers the cues. setControls indirectly removes the cues. See below for additional info.

>  I'm trying to figure out where the cues are being removed

Do you mean the elements, or timeSlider.cues? The elements are destroyed. The entire timeSlider object is recreated. Because vastPlugin doesn't redeliver the adSchedule (and I don't think it should), the model cues on the model need to be referenced. 

> It seems like we could hide the controls without being destructive to the underlying cues on the model

The model is unaffected, which is why I'm getting the cues from it. I can change the behavior of setControls to be more of a "hideControls" function, and simply toggle a display:none style or jw-hidden class, but I don't think that's the intended purpose of setControls. Controlbar already has a toggle function for visually hiding it.

jnatalzia: what's the benefit adding the cues in on playlist item vs when the controls are re-enabled?

zetagame: There's a few reasons I chose that block:
It's keeping it in the same place that we're re-adding chapters, so it's easier to see where the work is happening. 

Prevents an additional drawing of cues on initial setup.

One less listener on the model.

loadChapter's cues concatenate with cues pre-existing on the timeSlider, this ensures they're always present for that step.

Prevents any edge-cases that may have been addressed when loadThumbnails and loadChapters were added to the onPlaylistItem cb.

PR #3397, opened 2019-05-07
Replace window resize listener with more holistic solution

jnatalzia:
### This PR will...
Replace the current responsive resize listener with one that adapts to player container changes

We do not track resizes of the player if it is `100%` width and a parent container changes size

Yes. I double checked the listener was only firing once and checked window resize and parent conditions but could be some perf impact I'm overlooking

Nope

#### Addresses Issue(s):

JW8-2383

### DIFF
--- a/src/js/view/utils/resize-listener.js
+++ b/src/js/view/utils/resize-listener.js
@@ -51,7 +51,6 @@ export default class ResizeListener {
         this.resizeRaf = -1;
         this.lastWidth = 0;
         this.currentWidth = element.offsetWidth;
-
         this.scrollListener = (e) => {
             let resizeRaf = this.resizeRaf;
             if (resizeRaf) {
--- a/src/js/view/utils/resize-listener.js
+++ b/src/js/view/utils/resize-listener.js
@@ -1,6 +1,6 @@
-import { requestAnimationFrame, cancelAnimationFrame } from 'os/utils/request-animation-frame';
-import { createElement } from 'os/utils/dom';
-import { css, style } from 'os/utils/css';
+import { requestAnimationFrame, cancelAnimationFrame } from 'utils/request-animation-frame';
+import { createElement } from 'utils/dom';
+import { css, style } from 'utils/css';
 
 export default class ResizeListener {
 
--- a/src/js/view/utils/views-manager.js
+++ b/src/js/view/utils/views-manager.js
@@ -93,7 +93,6 @@ function removeFromGroup(view, group) {
 
 document.addEventListener('visibilitychange', onVisibilityChange);
 document.addEventListener('webkitvisibilitychange', onVisibilityChange);
-window.addEventListener('orientationchange', scheduleResponsiveRedraw);
 
 if (isAndroidChrome && hasOrientation) {
     window.screen.orientation.addEventListener('change', onOrientationChange);
@@ -102,7 +101,6 @@ if (isAndroidChrome && hasOrientation) {
 window.addEventListener('beforeunload', () => {
     document.removeEventListener('visibilitychange', onVisibilityChange);
     document.removeEventListener('webkitvisibilitychange', onVisibilityChange);
-    window.removeEventListener('orientationchange', scheduleResponsiveRedraw);
 
     if (isAndroidChrome && hasOrientation) {
         window.screen.orientation.removeEventListener('change', onOrientationChange);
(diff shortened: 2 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

jnatalzia:
test this please

src/js/view/utils/views-manager.js
```diff
@@ -103,7 +102,6 @@ if (isAndroidChrome && hasOrientation) {
 window.addEventListener('beforeunload', () => {
     document.removeEventListener('visibilitychange', onVisibilityChange);
     document.removeEventListener('webkitvisibilitychange', onVisibilityChange);
-    window.removeEventListener('resize', scheduleResponsiveRedraw);
     window.removeEventListener('orientationchange', scheduleResponsiveRedraw);
```
jnatalzia: left this in - but debating it. not sure if it's redundant like `resize` with the changes here. 

also are there concerns with moving from the `scheduleResponsiveRedraw` function here to `_responsiveListener` (which also schedules in a way)

robwalch: `scheduleResponsiveRedraw` notifies all players that they need to check if they've been resized. If each player has it's own `ResizeListener` then `scheduleResponsiveRedraw` and the "orientationchange" listener are no longer needed.

jnatalzia: Will remove!

src/js/view/utils/resize-listener.js
```diff
@@ -1,6 +1,6 @@
-import { requestAnimationFrame, cancelAnimationFrame } from 'os/utils/request-animation-frame';
-import { createElement } from 'os/utils/dom';
-import { css, style } from 'os/utils/css';
+import { requestAnimationFrame, cancelAnimationFrame } from 'utils/request-animation-frame';
+import { createElement } from 'utils/dom';
+import { css, style } from 'utils/css';
```
robwalch: good catch

src/js/view/view.js
```diff
@@ -297,6 +299,10 @@ function View(_api, _model) {
             viewsManager.observe(_playerElement);
         }
         _model.set('inDom', inDOM);
+
+        if (!this.resizeListener) {
```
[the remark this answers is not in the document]
jnatalzia: could the latter suggestion backfire if the user calls `resize` with a `%` based width? is that a possible use of the api?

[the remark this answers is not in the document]
robwalch: Well at that point you would need to instantiate a ResizeListener.

jnatalzia:
@robwalch did some light memory profiling on the weight of an unused resize listener.

If I'm reading this right (and I def could not be, these tools can be tough to read) it's a little over a kb in memory to house the listener and all references associated with it. 

Screen Shot 2019-05-09 at 9 24 22 AM

Given we're removing two window listeners and a function in memory I'm of the mind we're basically replacing the memory we took out. Not as great as removing it entirely for cases where it's not needed however I am still concerned about the implications of tearing out a window resize listener that always existed and not having something to replace it in the static width case. I also think with the current flex behavior of the player (as well as potentially behavior we don't know about) it's far more risky and not work the 1kb to try and find all the cases where a resize listener wouldn't be required.

Happy to circle back around, I want to make sure this is in a good place even if it means sorting through those edge cases but figured I'd show my findings first and get your opinion

robwalch:
@jnatalzia It a good trade off. The thing to keep in mind is that the window and orientation listeners were attached to the singleton views-manager and serviced all players. If someone creates 10 players, with the new resize listener each player will get one. I think that's OK since each will be fully responsive with the new element listener.

robwalch:
@jnatalzia Here's a good page to use to analyze responsive performance with multiple players 

http://player-develop-test-jenkins.longtailvideo.com/builds/lastSuccessfulBuild/archive/test/public/moviepilot/

PR build
http://player-pr-test-jenkins.longtailvideo.com/builds/21485/archive/test/public/moviepilot/

src/js/view/utils/views-manager.js
```diff
@@ -37,27 +35,6 @@ function matchIntersection(entry, group) {
     }
 }
 
-function scheduleResponsiveRedraw() {
```
robwalch: Could we do something where the ResizeListener's callbacks were debounced calls to this method?

jnatalzia:
test this please

PR #3398, opened 2019-05-07
Move interaction check into a util module

robwalch:
### This PR will...
- Move interaction check into a util module
- Make the default "reason" for play and pause reasons "unknown"

Play and pause reasons are always specified internally via our UI and plugins. When no reason is provided in play, pause, seek or playlist item change calls, "unknown" will be assigned.

In the commercial player, external calls will always pass in "external" or "interaction" when the external call is made in a user input event (see https://github.com/jwplayer/jwplayer-commercial/pull/6582/files#diff-b1c33615795eeebb83bdbba042caedceR82).

#### Addresses Issue(s):
JW8-8871


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3399, opened 2019-05-07
Trigger buffer events during ads

karimMourra:
### This PR will...
add a listener for `buffer` on instream and trigger a 'buffer' event when an ad is loaded.
Currently when a preRoll is loaded, the player state goes from `idle` to `buffering` but a buffer event is never triggered. I think a buffer event should fire because a developer should be able to rely on our events to detect state changes.

This bug was noticed while working on performance improvements for the iOS SDK. In the past, we would use the `getState` API to determine the state. This resulted in a large amount of javascript injections which are costly. To reduce the amount of javascript injections, and since WKWebView handles javascript asynchronously, we have decided to rely on the player's events to track the state. Given that the buffer event doesn't fire on ads, our state tracking gets out of sync. 

If firing buffer events during ad playback damages our event orders, we could look into changing the state in the SDK to buffer on other events such as playAttempt or adStart. Our preference though would be to rely on the player's buffer event. 
Should we have a adBuffer event ?
n/a
#### Addresses Issue(s):
SDK-4152

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -104,7 +104,7 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
 
         // This enters the player into instream mode
         _model.set('instream', _adProgram);
-        _adProgram.model.set('state', STATE_BUFFERING);
+        _adProgram.model.set(PLAYER_STATE, STATE_BUFFERING);
 
         // don't trigger api play/pause on display click
         const clickHandler = _view.clickHandler();
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -1058,6 +1058,9 @@ Object.assign(Controller.prototype, {
         this.createInstream = function() {
             this.instreamDestroy();
             this._instreamAdapter = new InstreamAdapter(this, _model, _view, mediaPool);
+            this._instreamAdapter.on('buffer', function (evt) {
+                _this.trigger(evt.type, evt);
+            });
             return this._instreamAdapter;
         };
 
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -826,8 +826,9 @@ Object.assign(Controller.prototype, {
             }
 
             _model.set('fullscreen', state);
-            if (_this._instreamAdapter && _this._instreamAdapter._adModel) {
-                _this._instreamAdapter._adModel.set('fullscreen', state);
+            const instream = _this._instreamAdapter;
+            if (instream && instream._adModel) {
+                instream._adModel.set('fullscreen', state);
             }
         }
 
@@ -1057,11 +1058,11 @@ Object.assign(Controller.prototype, {
 
         this.createInstream = function() {
             this.instreamDestroy();
-            this._instreamAdapter = new InstreamAdapter(this, _model, _view, mediaPool);
-            this._instreamAdapter.on('buffer', function (evt) {
+            const instream = this._instreamAdapter = new InstreamAdapter(this, _model, _view, mediaPool);
+            instream.on('buffer', function (evt) {
                 _this.trigger(evt.type, evt);
             });
-            return this._instreamAdapter;
+            return instream;
         };
 
         this.skipAd = function() {


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

karimMourra:
@reviewer: an integration test failed due to flakiness

karimMourra:
@pajong `off` is called in the instream's `destroy`. @robwalch thanks, I will address shortly

src/js/controller/instream-adapter.js
```diff
@@ -89,6 +89,7 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         _adProgram.setup();
 
         _adProgram.on('all', _instreamForward, this);
+        _adProgram.on('buffer', triggerAdBuffer, this);
```
robwalch: We already handle state changes in `_handleStateChange`.

src/js/controller/instream-adapter.js
```diff
@@ -417,6 +418,10 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         }
     }
 
+    function triggerAdBuffer(evt) {
+        _controller.trigger(AD_BUFFER, evt);
```
robwalch: Can we deal with this in `_handleStateChange`?

src/js/controller/instream-adapter.js
pajong: @robwalch Isn't this different from state event? (`buffering` vs `buffer`).
This `buffer` event comes from `program-listener` every time the buffer changes: https://github.com/jwplayer/jwplayer/blob/master/src/js/program/program-listeners.js#L73

robwalch: The point is to match the player state changes.

Player events | Ad mode events
------------ | -------------
buffer | adBuffer
play | adPlay
pause | adPause

src/js/controller/instream-adapter.js
robwalch: Also see `_stateHandler` in ad-program-controller

```
    _stateHandler(state) {
        const { model } = this;
        switch (state) {
            case STATE_PLAYING:
            case STATE_PAUSED:
                model.set(PLAYER_STATE, state);
                break;
            default:
                break;
        }
    }
```

robwalch: and 

```
function normalizeState(newstate) {
    if (newstate === STATE_LOADING || newstate === STATE_STALLED) {
        return STATE_BUFFERING;
    }
    return newstate;
}
```

robwalch: What I expect to see is `_triggerAdPlayPause` renamed to `_triggerAdState` and add handling for buffering states
    function _triggerAdPlayPause(event) {
        const { newstate } = event;
        switch (newstate) {
            STATE_PLAYING:
            _controller.trigger(AD_PLAY, event);
           break;
          STATE_PAUSED:
            _controller.trigger(AD_PAUSE, event);
            break;
          STATE_LOADING:
          STATE_STALLED:
          STATE_BUFFERING:
            _controller.trigger(AD_BUFFER, event); // make any changes necessary to event properties
            break;
        }
    }

src/js/controller/instream-adapter.js
karimMourra: i'm confused by `handleStateChange`; the `newState` is coming from the provider, but the `oldState` is coming from the model. But the model's state is set to `buffering` before the provider state change , so `handleStateChange` detects a change from `buffering` to `loading` which is incorrect. Meanwhile the player's state is still `idle`. What should the flow be ?
https://github.com/jwplayer/jwplayer/pull/3169/files#diff-4b5d9c794240aa20cb620353eb027d56R114

karimMourra:
@robwalch I addressed your feedback. 
adBuffer does not fire on IMA and Freewheel, I need to set the buffering state on the ad plugins' `instreamProvider` before the adImpression is fired, possibly on the plugin's `loaded` event

PR #3405, opened 2019-05-09
Cleanup dependencies in package.json

### LINKED ISSUE #2017: Cleanup package.json
Half the time of our build is spent on `npm install`. Let's clean up all unused packages to lower our time

robwalch:
### This PR will...
Remove jsQuery, require.js (plus karma-require) and underscore dependencies.

These dependencies are not being used and are not needed.

#### Addresses Issue(s):
Fixes #2017


karma.conf.js
```diff
@@ -13,19 +13,12 @@ const webpackTestConfig = merge(webpackConfig, {
...
-            root: '$'
-        },
         sinon: 'sinon'
     },
     resolve: {
         alias: {
             'test/underscore': path.resolve(__dirname + '/node_modules/underscore/underscore.js'),
             'utils/video': path.resolve(__dirname + '/test/mock/video.js'),
-            // Tests using jQuery: api-test, jwplayer-selectplayer-test, setup-test
-            jquery: path.resolve(__dirname + '/node_modules/jquery/dist/jquery.js'),
```
jnatalzia: do we need to update any test pages?

robwalch: There are only two pages in test and neither use `$`.

package.json
```diff
@@ -43,7 +42,6 @@
     "karma-junit-reporter": "1.2.0",
```
zetagame: Do we use karma-junit-reporter anywhere?

package.json
```diff
@@ -59,13 +57,11 @@
     "postcss-safe-parser": "3.0.1",
```
zetagame: Same question for postcss-safe-parser

package.json
robwalch: Jenkins

robwalch: Run ‘yarn why’. Is it a peer dependency?

package.json
robwalch: Run ‘yarn why’. Is it a peer dependency?

robwalch:
```
$ yarn why postcss-safe-parser                                                                                                                                                                                                            
 🤔  Why do we have the module "postcss-safe-parser"...?
=> Found "postcss-safe-parser@3.0.1"
info Has been hoisted to "postcss-safe-parser"
info Reasons this module exists
   - Specified in "devDependencies"
   - Hoisted from "stylelint#postcss-safe-parser"
```

PR #3406, opened 2019-05-10
Responsive size listener performance

jnatalzia:
### This PR will...
Improve performance of the responsive resize listener

It performs poorly, especially with multiple players on page

None I can think of

No

#### Addresses Issue(s):

JW8-2383

### DIFF
--- a/src/js/view/utils/resize-listener.js
+++ b/src/js/view/utils/resize-listener.js
@@ -68,11 +68,6 @@ export default class ResizeListener {
         };
 
         requestAnimationFrame(() => {
-            // TODO: this should be moved to the widget. The player element is always relative.
-            if (getComputedStyle(element).position === 'static') {
-                style(element, { position: 'relative' });
-            }
-
             this.currentWidth = element.offsetWidth;
             this.resetTriggers();
             element.addEventListener('scroll', this.scrollListener, true);
--- a/src/js/view/utils/resize-listener.js
+++ b/src/js/view/utils/resize-listener.js
@@ -2,27 +2,36 @@ import { requestAnimationFrame, cancelAnimationFrame } from 'utils/request-anima
 import { createElement } from 'utils/dom';
 import { css, style } from 'utils/css';
 
-export default class ResizeListener {
+const topLeft = {
+    display: 'block',
+    position: 'absolute',
+    top: 0,
+    left: 0
+};
 
-    constructor(element, callback) {
-        const topLeft = {
-            display: 'block',
-            position: 'absolute',
-            top: 0,
-            left: 0
-        };
+const stretch = {
+    width: '100%',
+    height: '100%'
+};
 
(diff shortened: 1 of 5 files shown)


src/js/view/utils/resize-listener.js
```diff
@@ -49,26 +54,29 @@ export default class ResizeListener {
...
                 this.callback(e, currentWidth);
             });
-            this.resizeRaf = resizeRaf;
             this.resetTriggers();
         };
 
-        element.addEventListener('scroll', this.scrollListener, true);
-        this.resetTriggers();
+        requestAnimationFrame(() => {
+            // TODO: this should be moved to the widget. The player element is always relative.
```
jnatalzia: assume you mean shelf widget here @robwalch ?

robwalch: Yep. The player must be styled `position: relative` or the absolute positioned elements contained within will fill the page: https://github.com/jwplayer/jwplayer/blob/master/src/css/jwplayer/imports/jwplayerlayout.less#L6

src/js/view/view.js
```diff
@@ -341,6 +338,9 @@ function View(_api, _model) {
         // Triggering 'resize' resulting in player 'ready'
         _lastWidth = _lastHeight = null;
         this.checkResized();
+        if (!this.resizeListener) {
+            this.resizeListener = new ResizeListener(_playerElement, viewsManager.scheduleResponsiveRedraw);
```
robwalch: The most optimal way to fit this into view setup would be for this classes constructor to only do the DOM creation and styling - that can happen in `view.setup`. And then defer the reads until here (or just after `updateBounds` above).

test/unit/setup-test.js
```diff
@@ -31,18 +33,25 @@ describe('api.setup', function() {
...
+            api.remove();
+        }
         if (fixture.parentNode) {
             fixture.parentNode.removeChild(fixture);
         }
         for (let i = instances.length; i--;) {
             instances[i].remove();
         }
         sandbox.restore();
+        api = null;
```
robwalch: Before these changes, were were not tearing down all api instances (and their views) properly, which left `view` instances in the `viewManager`s list of views.

johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/view.js
```diff
@@ -1013,6 +1009,7 @@ function View(_api, _model) {
         clearCss(_model.get('id'));
         if (this.resizeListener) {
             this.resizeListener.destroy();
+            delete this.resizeListener;
```
robwalch: we usually null these on destroy

PR #3408, opened 2019-05-13
[JW8-2383/JW8-2551] Respond to external resizing on ready

robwalch:
### This PR will...
- Respond to external resizing on ready (JW8-2551)
- Batch resize-listener DOM style reads and updates using view methods
- Move static css into jwplayer css bundle added by webpack
- Insert resize listener element behind other player elements inside the wrapper element
- ~Let the resize listener handle floating and fullscreen player resizes~ (can't let resizing fail when we know to run the update loop for specific view)

This ensures that the player responds even to style changes made on the ready event while the resize-listener is setting up.

The batching improves performance when there is more than one player on the page by avoiding layout thrashing caused by recalculating styles and forcing layout.

https://github.com/jwplayer/jwplayer-commercial/pull/6639

#### Addresses Issue(s):
JW8-2551

### DIFF
--- a/src/js/view/utils/resize-listener.js
+++ b/src/js/view/utils/resize-listener.js
@@ -89,12 +89,10 @@ export default class ResizeListener {
             if (index !== -1) {
                 instances.splice(index, 1);
             }
-            this.element.removeEventListener('scroll', this.scrollListener, true);
+            this.element.removeEventListener('scroll', scrollListener, true);
             this.element.removeChild(this.hiddenElement);
-            this.scrollListener =
-                this.view =
+            this.view =
                 this.model = null;
-
         }
     }
 }
--- a/src/js/view/utils/resize-listener.js
+++ b/src/js/view/utils/resize-listener.js
@@ -1,44 +1,56 @@
 import { requestAnimationFrame, cancelAnimationFrame } from 'utils/request-animation-frame';
 import { createElement } from 'utils/dom';
-import { css, style } from 'utils/css';
+import { style } from 'utils/css';
 
-const topLeft = {
-    display: 'block',
-    position: 'absolute',
-    top: 0,
-    left: 0
-};
+const instances = [];
+let resizeRaf = -1;
 
-const stretch = {
-    width: '100%',
-    height: '100%'
-};
-
-let contractTriggerCssAdded = false;
-
-const lazyAddCss = function() {
-    if (!contractTriggerCssAdded) {
-        contractTriggerCssAdded = true;
-        css('.jw-contract-trigger::before', Object.assign({
-            content: '',
-            overflow: 'hidden',
-            width: '200%',
-            height: '200%'
-        }, topLeft));
-    }
-};
+function scrollListener() {
+    cancelAnimationFrame(resizeRaf);
+    resizeRaf = requestAnimationFrame(() => {
+        instances.forEach(resizeListener => {
+            resizeListener.updateBounds();
+        });
+        instances.forEach(resizeListener => {
+            resizeListener.contractElement.scrollLeft = resizeListener.width * 2;
+        });
+        instances.forEach(resizeListener => {
+            style(resizeListener.expandChild, { width: resizeListener.width + 1 });
+        });
+        instances.forEach(resizeListener => {
+            resizeListener.expandElement.scrollLeft = resizeListener.width + 1;
+        });
+        instances.forEach(resizeListener => {
+            resizeListener.checkResize();
+        });
+    });
+}
 
 export default class ResizeListener {
 
-    constructor(element, callback) {
-        lazyAddCss();
-
+    constructor(element, callback, initialWidth) {
+        if (!initialWidth) {
+            initialWidth = element.offsetWidth;
+        }
         const hiddenHtml = '<div style="opacity:0;visibility:hidden;overflow:hidden;">' + // resizeElement
             '<div>' + // expandElement
             '<div style="height:1px;">' + // expandChild
             '</div></div>' +
             '<div class="jw-contract-trigger">' + // contractElement
             '</div></div>';
+
+        const topLeft = {
+            display: 'block',
+            position: 'absolute',
+            top: 0,
+            left: 0
+        };
+
+        const stretch = {
+            width: '100%',
+            height: '100%'
+        };
+
         const resizeElement = createElement(hiddenHtml);
         const expandElement = resizeElement.firstChild;
         const expandChild = expandElement.firstChild;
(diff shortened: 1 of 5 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/utils/resize-listener.js
```diff
@@ -1,44 +1,56 @@
...
-    }
-};
+function scrollListener() {
+    cancelAnimationFrame(resizeRaf);
+    resizeRaf = requestAnimationFrame(() => {
+        instances.forEach(resizeListener => {
+            resizeListener.updateBounds();
+        });
+        instances.forEach(resizeListener => {
+            resizeListener.contractElement.scrollLeft = resizeListener.width * 2;
```
jnatalzia: assuming this is faster/less thrashy than having one `instances.forEach` block which does all the scroll/style updates for each listener in a single go. and also assuming that's because only one of these would actually force layout (style) and as a result we want to separate it from the compute of the scroll left? though that would at least paint right?

src/js/view/utils/resize-listener.js
```diff
@@ -50,40 +62,39 @@ export default class ResizeListener {
...
-                const currentWidth = this.currentWidth = element.offsetWidth;
-                if (this.lastWidth === currentWidth) {
-                    return;
-                }
-                this.callback(e, currentWidth);
-            });
-            this.resetTriggers();
-        };
+        this.width = initialWidth;
+        this.lastWidth = initialWidth;
```
jnatalzia: `this.lastWidth = this.width = ...`?

src/js/view/view.js
```diff
@@ -657,7 +658,6 @@ function View(_api, _model) {
             _model.set('fullscreen', newState);
         }
 
-        _responsiveListener();
```
jnatalzia: nice

src/js/view/utils/resize-listener.js
```diff
@@ -50,40 +62,39 @@ export default class ResizeListener {
...
+        if (element.firstChild) {
```
jnatalzia: why doesn't ie have `prepend` support 😑

src/js/view/utils/resize-listener.js
robwalch: Seems like we pay mostly on the last `scrollLeft` update after a `style` update. Batching prevent thrashing between the back and forth of setting both for each player.

robwalch: @jnatalzia I just pushed some updates that remove the `scheduleResponsiveRedraw` callback and instead implement it in the scroll listener.

This is now a view resize listener, and the player and widget views both implement `updateBounds`, `updateStyles` (handled when updating the scroll listener styles) and `checkResized`.

This is a nice boost for both single and multiple players since it removes the thrashing caused by the read writes of the resize-listener followed by the read writes of `scheduleResponsiveRedraw`.

src/js/view/utils/resize-listener.js
```diff
@@ -50,44 +67,34 @@ export default class ResizeListener {
...
+        if (this.view) {
+            const index = instances.indexOf(this);
+            if (index !== -1) {
+                instances.splice(index, 1);
+            }
             this.element.removeEventListener('scroll', this.scrollListener, true);
             this.element.removeChild(this.hiddenElement);
             this.scrollListener =
-                this.callback = null;
+                this.view =
```
jnatalzia: `this.scrollListener = null`?
`this.view = null`?

src/js/view/utils/resize-listener.js
```diff
@@ -1,44 +1,61 @@
...
+            const width = resizeListener.view.model.get('containerWidth');
+            resizeListener.resized = resizeListener.width !== width;
+            resizeListener.width = width;
+        });
+        instances.forEach(resizeListener => {
+            resizeListener.contractElement.scrollLeft = resizeListener.width * 2;
+        });
+        instances.forEach(resizeListener => {
+            style(resizeListener.expandChild, { width: resizeListener.width + 1 });
+            if (resizeListener.resized && resizeListener.view.model.get('visibility')) {
```
jnatalzia: why the viewability check if we're not updating on viewable? the scroll listener is only listening for the scroll on the internal container, couldn't we hit a case where we resize with player offscreen and then without resizing again scroll up to make the player viewable and the sizing is off?

robwalch: In view.js `_model.on('change:visibility', redraw);` calls `view.updateStyles()`. There's no need to invalidate the layout for players not in view.

src/js/view/utils/resize-listener.js
robwalch: updated

src/js/view/utils/resize-listener.js
```diff
@@ -50,44 +67,32 @@ export default class ResizeListener {
...
+        }
+        element.addEventListener('scroll', scrollListener, true);
+        instances.push(this);
+        scrollListener();
     }
 
     destroy() {
-        if (this.callback) {
-            this.element.removeEventListener('scroll', this.scrollListener, true);
+        if (this.view) {
```
jnatalzia: wouldn't we want to splice this out regardless or whether or not there's a view? not too clear on why the destroy is dependent on this view check when we don't have a similar check in the constructor where this stuff is set up.the event listener could dangle here or anything else. i might be overlooking something as well!

src/js/view/view.js
```diff
@@ -157,6 +157,9 @@ function View(_api, _model) {
         const containerHeight = _model.get('containerHeight');
         const floating = _model.get('isFloating');
         if (containerWidth !== _lastWidth || containerHeight !== _lastHeight) {
+            if (!this.resizeListener) {
```
jnatalzia: i see this function being called only a few places in view, most of them pertaining to orientation/fullscreen changes along with floats. would this set up with a non-floating player in a normal scenario? i know there's a call in `init`, are the `_lastWidth`s (etc) not correct at the time of that init call and this causes it to get set up? want to be sure we're not missing any execution paths

src/js/view/utils/resize-listener.js
robwalch: `view` is assigned a value in the constructor, and set to `null` when destroyed. The check is so we don't run any of the destroy logic twice. It's the `removeChild` call that would fail if called twice, but there's not reason to run any of the other code if `view` or `model` was nulled.

src/js/view/view.js
robwalch: This always runs. The player fires "ready" only after `view.init` results in this code succeeding and firing a "resize" event. So what I'm doing here is deferring the creation of the resize listener until right before the ready event, when we know the player is in the DOM and we have valid `containerWidth` and `containerHeight` which the listener needs to setup properly.

PR #3410, opened 2019-05-15
Break out autoplay behavior into overrideable functions

jnatalzia:
### This PR will...
Allow controllers to specifically short-circuit autoplay behavior that is the default in the open source player

We need to give the commercial controller the ability to disable the default behavior of the open source player given that it automatically advances any playlist which would override any options in `related` for playlists.

Yes all of it please

https://github.com/jwplayer/jwplayer-commercial/pull/6652

#### Addresses Issue(s):

JW8-9009

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -698,22 +698,32 @@ Object.assign(Controller.prototype, {
             _item(_model.get('item') + 1, meta);
         }
 
-        function _completeHandler() {
+        function _completeCancelled() {
             if (!_isIdle()) {
                 // Something has made an API call before the complete handler has fired.
-                return;
+                return true;
             } else if (_stopPlaylist) {
                 // Stop called in onComplete event listener
                 _stopPlaylist = false;
+                return true;
+            }
+
+            return false;
+        }
+
+        function _shouldAutoAdvance() {
+            // If it's the last item in the playlist
+            const idx = _model.get('item');
+            return idx !== _model.get('playlist').length - 1;
+        }
+        function _completeHandler() {
+            if (_this.completeCancelled()) {
                 return;
             }
 
-            _actionOnAttach = _completeHandler;
+            _actionOnAttach = _this.completeHandler;
 
-            const idx = _model.get('item');
-            let overrideAutoplay = _model.get('__autoplayOverride');
-            if (overrideAutoplay || idx === _model.get('playlist').length - 1) {
-                // If it's the last item in the playlist
+            if (!_this.shouldAutoAdvance()) {
                 if (_model.get('repeat')) {
                     _next({ reason: 'repeat' });
                 } else {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/controller.js
```diff
@@ -710,7 +710,8 @@ Object.assign(Controller.prototype, {
             _actionOnAttach = _completeHandler;
 
             const idx = _model.get('item');
-            if (idx === _model.get('playlist').length - 1) {
+            let overrideAutoplay = _model.get('__autoplayOverride');
+            if (overrideAutoplay || idx === _model.get('playlist').length - 1) {
```
jnatalzia: i wanted to use a different conditional here to be cleaner which would base whether or not to stop on the presence of a `nextUp` item however this does not cover the open source case where `nextUp` never gets set on the model (no recs). in that case you would still need the playlist check but could check `nextUp` in the event recs is present. super messy. this keeps default OS behavior while giving the application a chance to override this behavior if we want to do something smarter.

I tried this out with a extension method in commercial but since the entirety of this function would have been repeated (or a flag used as a parameter) it didn't seem like the right use case for it. along with that, the extended `onComplete` method in commercial would have had to know recs state at any point in time to make the right determination which seems like way more of the plugin's responsibility. 

This allows the _plugin_ to have control over the autoplay override and stops unnecessary code duplication or parameterization in the commercial controller

robwalch: I think it would be better to break this method up into the components you need to override in commercial.

```
this.completeHandler = function() {
   if (this.completeCancelled()) { // encapsulate _isIdle & _stopPlaylist check
      return;
  }
  _actionOnAttach = _completeHandler;
  const idx = _model.get('item');
  const lastItemInPlaylist = idx === _model.get('playlist').length - 1;
  if (lastItemInPlaylist) { // this could be a method too if you want to add a nextUp check
    this.lastItemComplete();
    return;
  }
  // Aren't we just trying to interrupt this method?
  _this.nextItem();
}
```

robwalch: I'm not a fan of communicating between modules using the model (side-effects). The above would allow for a more functional approach.

jnatalzia:
You would still need commercial to understand all the logic that the
related plugin needs internally to determine when to call these which feels
to me like trading one duplication for another. Though I def agree the
model side effect is not ideal.

jnatalzia:
Especially if we make perf improvements and conditionally create the
autoplay/ui components then you def need that logic in both related &
commercial which is worrying

src/js/controller/controller.js
```diff
@@ -697,21 +697,32 @@ Object.assign(Controller.prototype, {
...
-                return;
+                return true;
             }
 
-            _actionOnAttach = _completeHandler;
+            return false;
+        }
 
+        function _shouldAutoAdvance() {
+            // If it's the last item in the playlist
```
robwalch: then it should not auto advance?

jnatalzia: oh duh, good catch

PR #3412, opened 2019-05-16
[JW8-8766] Only turn off viewable listener when autoPause is not enabled

waxidiotic:
JW8-8766

### This PR will...

* Move the logic to turn off the `change:viewable` listener that calls `_checkPlayOnViewable()` to a conditional that checks whether or not `autoPause` is enabled

* If a player is setup to be muted and to have autoPause enabled, if it was unmuted before scrolling the player out of view and then back, the content would not resume.

n/a

n/a

#### Addresses Issue(s):

JW8-8766

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -278,12 +278,13 @@ Object.assign(Controller.prototype, {
 
             _model.change('viewable', viewableChange);
             _model.change('viewable', _checkPlayOnViewable);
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

PR #3416, opened 2019-05-20
Avoid disabling a track if it is already disabled JW8-2590

johnBartos:
For an unknown reason, Edge will throw an exception if trying to set track modes if it is already disabled. This prevents Hls.js from switching live streams; cues are cleared on `stop` and `play`, and on the second call to remove cues, an exception is thrown.

Try/catch also works. Maybe that's better than the guard?

No

#### Addresses Issue(s):

JW8-2590

### DIFF
--- a/src/js/providers/tracks-mixin.js
+++ b/src/js/providers/tracks-mixin.js
@@ -573,7 +573,7 @@ function _removeCues(renderNatively, tracks) {
             // Cues are inaccessible if the track is disabled. While hidden,
             // we can remove cues while the track is in a non-visible state
             // Set to disabled before hidden to ensure active cues disappear
-            if (track.mode !== 'disabled') {
+            if (!Browser.ie || track.mode !== 'disabled') {
                 // Avoid setting the track to disabled if it is already so. This prevents an exception when trying
                 // to set the mode on Edge
                 track.mode = 'disabled';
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/providers/tracks-mixin.js
```diff
@@ -573,8 +573,12 @@ function _removeCues(renderNatively, tracks) {
             // Cues are inaccessible if the track is disabled. While hidden,
             // we can remove cues while the track is in a non-visible state
             // Set to disabled before hidden to ensure active cues disappear
-            track.mode = 'disabled';
-            track.mode = 'hidden';
+            if (track.mode !== 'disabled') {
```
robwalch: "Cues are inaccessible if the track is disabled"

That's why we need to change the mode to "hidden", so that we can access the cues and remove them.

Would this check work if we tested for the track being hidden instead?
`if (track.mode !== 'hidden')`

egreaves: What Rob said.

If it's just `'disabled'` that causes Edge to barf, then I'd suggest:

```
if (track.mode !== 'disabled') {
    track.mode = 'disabled';
}
track.mode = 'hidden';
```

johnBartos: That didn't work. I don't know what the bug is but setting it to disabled is not the problem. It's setting to hidden while disabled

johnBartos: @robwalch I tried that and it didn't work. The state was disabled when entering that block, not hidden

robwalch: yeah we already talked through this.

PR #3418, opened 2019-05-23
[JW8-9168] openLink FF fix with IE support

zetagame:
### This PR will...
Extend the new openLink method to work on IE.

IE doesn't support custom events. The new method uses custom events in order to work on FF without causing reflows.

#### Addresses Issue(s):

JW8-9168

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -217,7 +217,7 @@ export function openLink(link, target, additionalOptions = {}) {
     a.target = target;
     a = Object.assign(a, additionalOptions);
 
-    // Must support all browsers.
+    // Firefox is the only modern browser that doesn't support clicking orphaned anchors.
     if (Browser.firefox) {
         a.dispatchEvent(new MouseEvent('click', { bubbles: true, cancelable: true, view: window }));
     } else {
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -215,5 +215,11 @@ export function openLink(link, target, additionalOptions = {}) {
     a.href = link;
     a.target = target;
     a = Object.assign(a, additionalOptions);
-    a.dispatchEvent(new MouseEvent(`click`, { bubbles: true, cancelable: true, view: window }));
+
+    // Must support all browsers.
+    try {
+        a.dispatchEvent(new MouseEvent('click', { bubbles: true, cancelable: true, view: window }));
+    } catch (e) {
+        a.click();
+    }
 }
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -1,5 +1,6 @@
 import { trim } from 'utils/strings';
 import { isString, contains, difference, isBoolean } from 'utils/underscore';
+import { Browser } from '../environment/environment';
 
 let parser;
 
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/utils/dom.js
```diff
@@ -215,5 +215,11 @@ export function openLink(link, target, additionalOptions = {}) {
     a.href = link;
     a.target = target;
     a = Object.assign(a, additionalOptions);
-    a.dispatchEvent(new MouseEvent(`click`, { bubbles: true, cancelable: true, view: window }));
+
+    // Must support all browsers.
+    try {
```
jnatalzia: is there any way of detecting custom event support rather than a try catch?

zetagame: yes, but not in a way that works for both IE and firefox. I could use document.customEvent as a check, but the property exists on both IE and firefox. a.click() won't work on firefox because the anchor is never appended to the DOM

zetagame: Implemented an OS check instead of try/catch

src/js/utils/dom.js
```diff
@@ -215,5 +216,11 @@ export function openLink(link, target, additionalOptions = {}) {
     a.href = link;
     a.target = target;
     a = Object.assign(a, additionalOptions);
-    a.dispatchEvent(new MouseEvent(`click`, { bubbles: true, cancelable: true, view: window }));
+
+    // Must support all browsers.
+    if (Browser.firefox) {
+        a.dispatchEvent(new MouseEvent('click', { bubbles: true, cancelable: true, view: window }));
+    } else {
+        a.click();
```
robwalch: Shouldn't `ie` and `.click()` be the exception and not the norm?

src/js/utils/dom.js
```diff
@@ -215,5 +216,11 @@ export function openLink(link, target, additionalOptions = {}) {
     a.href = link;
     a.target = target;
     a = Object.assign(a, additionalOptions);
-    a.dispatchEvent(new MouseEvent(`click`, { bubbles: true, cancelable: true, view: window }));
+
+    // Must support all browsers.
```
robwalch: > // Must support all browsers.
That kind of goes without saying.

Why only create an event in `firefox`? I thought we wanted to do that in all modern browsers, and just call click in `ie`.

src/js/utils/dom.js
zetagame: Because the bug was that it was not working in Firefox. A fix was implemented (dispatchEvent), but that fix didn't work in IE. So the method was updated to use the original implementation (that worked everywhere besides Firefox), but with a fallback for Firefox.

src/js/utils/dom.js
zetagame: Firefox is the only modern browser that doesn't support clicking orphaned anchors.

robwalch: Yes, but we don't need a comment that says "Must support all browsers".

PR #3423, opened 2019-05-31
[JW8-9245] Catch playPromise.

vseventer:
### This PR will...
Catch the `playPromise` for subsequent ad pod items.

Subsequent ad pod items might error (e.g. mediaError), which right now logs an uncaught exception in the console. This PR fixes that.

Note the actual error is handled by the plugins `playbackErrorHandler`, so we're not ignoring the error altogether.

jwplayer/jwplayer-ads-vast#541

#### Addresses Issue(s):
JW8-9245

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -174,7 +174,7 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
 
     function _loadNextItem() {
         _arrayIndex++;
-        _this.loadItem(_array);
+        _this.loadItem(_array).catch(function() { });
     }
 
     function _instreamForward(type, data) {


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/instream-adapter.js
```diff
@@ -174,7 +174,7 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
 
     function _loadNextItem() {
         _arrayIndex++;
-        _this.loadItem(_array);
+        _this.loadItem(_array).catch(function() { });
```
waxidiotic: Does it make sense to use `noop` here or no since we're only using it this one time in this file?

pajong: Would be nice to use `noop` since there is another place where we can use it: https://github.com/jwplayer/jwplayer/pull/3423/files#diff-79d1ee28e72759418dcb9c1e8855dd7cL56

robwalch: this is a no-op. There is a lot of overhead in importing modules, so no real gain in importing `utils.noop`.

pajong: I was meaning having 
`
const noop = function() { };
`
to use it within the file

PR #3401, opened 2019-05-08
[JW8-9022] Workaround iOS background bug 

robwalch:
### This PR will...
Add a workaround iOS background bug  https://bugs.webkit.org/show_bug.cgi?id=195452

### Why is this Pull Request needed?
If iOS Safari throws a decode error when returning to a tab after it's been backgrounded, we can attempt to reload the video from where it left off until there is a fix for this issue https://bugs.webkit.org/show_bug.cgi?id=195452

The retry requires that it's a decode error, the OS is iOS, and that some video has loaded and played successfully. We will not retry if the currentTime equals 0(this could be an issue if you seek back to the start and then background Safari).
 
#### Addresses Issue(s):
JW8-9022

PR #3427, opened 2019-06-05
Implement retry logic in html5 provider

robwalch:
### This PR will...
Implement retry logic in html5 provider

Safari makes a lot of manifest requests after loading the page and player, re-requesting manifests almost every time the tab focus or system audio context changes. Since delivery of content can be compromised for one of these requests (bad status, cached result for expired signed url) we need to build in some resiliency for handling these errors.

These changes build on top of retry logic for decode errors (code 3) https://github.com/jwplayer/jwplayer/pull/3401 by adding a retry for not supported error (code 2). This means we can recover before or after playback has started. It also adds support for 3 tries, and resets the retry count after 2 consecutive timeupdate events.

Updates to this HLS proxy script allow for testing against a stream that alternates between returning a healthy stream and one which contains child manifests that 403:
https://github.com/jwplayer/jwplayer-commercial/pull/6733

#### Addresses Issue(s):
JW8-9327

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -2,7 +2,7 @@ import { qualityLevel } from 'providers/data-normalizer';
 import { Browser, OS } from 'environment/environment';
 import { isAndroidHls } from 'providers/html5-android-hls';
 import {
-    STATE_IDLE, STATE_PLAYING, STATE_STALLED, MEDIA_META_CUE_PARSED, MEDIA_META, MEDIA_ERROR,
+    STATE_IDLE, STATE_PLAYING, STATE_STALLED, MEDIA_META_CUE_PARSED, MEDIA_META, MEDIA_ERROR, WARNING,
     MEDIA_VISUAL_QUALITY, MEDIA_TYPE, MEDIA_LEVELS, MEDIA_LEVEL_CHANGED, MEDIA_SEEK, NATIVE_FULLSCREEN, STATE_LOADING
 } from 'events/events';
 import VideoEvents from 'providers/video-listener-mixin';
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/providers/html5.js
```diff
@@ -86,6 +93,10 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
         },
 
         timeupdate() {
+            if (_this.currentTime) {
+                // Reset error retries after concurrent timeupdate events
+                _this.retries = 0;
```
johnBartos: Reset on init

src/js/providers/html5.js
```diff
@@ -194,14 +205,19 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
             const error = video.error;
             const errorCode = (error && error.code) || -1;
 
-            if (errorCode === 3 && _this.currentTime !== -1 && OS.iOS) {
-                // Workaround iOS bug https://bugs.webkit.org/show_bug.cgi?id=195452
+            if (errorCode === 3 || errorCode === 4 && _this.retries < 3) {
```
waxidiotic: So we would trigger this warning if `errorCode === 3` OR `errorCode === 4 && _this.retries < 3`?

src/js/providers/html5.js
robwalch: done

src/js/providers/html5.js
robwalch: Nice catch. Fixed!

PR #3431, opened 2019-06-10
[DNM] [JW8-9133] Add skipAd to api prototype, remove from controller

waxidiotic:
JW8-9133

### This PR will...

* Add `skipAd()` to API prototype so that the logic can be handled by the ads plugins themselves instead of the controller
* Removes `skipAd()` logic from the controller

* For VPAIDs, we want to be able to call the VPAID creative's own `skipAd()` method and also check to see if the VPAID deems itself as skippable or not.
* This also creates parity with our `skipAd()` works with the IMA plugin

n/a

https://github.com/jwplayer/jwplayer-ads-vast/pull/543
https://github.com/jwplayer/jwplayer-ads-freewheel/pull/155

#### Addresses Issue(s):

JW8-9133

### DIFF
--- a/src/js/api/api.js
+++ b/src/js/api/api.js
@@ -843,15 +843,6 @@ export default function Api(element) {
         },
 
         /**
-         * Calls `skipAd` on the active instream adapter instance if present.
-         * @returns {Api} The Player API instance.
-         */
-        skipAd() {
-            core.skipAd();
-            return this;
-        },
-
-        /**
          * Stops any active playback.
          * @returns {Api} The Player API instance.
          */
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/api/api.js
```diff
@@ -842,15 +842,6 @@ export default function Api(element) {
             return core.createInstream();
         },
 
-        /**
-         * Calls `skipAd` on the active instream adapter instance if present.
-         * @returns {Api} The Player API instance.
-         */
-        skipAd() {
-            core.skipAd();
-            return this;
```
robwalch: Should there still be a way for customers to force an ad to be skipped (or end an ad break)? This removes that capability for the vast plugin in some cases.

waxidiotic: We've been discussing this very thing and the consensus is that we should only allow skipping the ad if:

- there is a `skipoffset` configured
- we have reached the `skipoffset`
- if a VPAID, the VPAID's `getAdSkippableState()` method returns `true`

cc @pajong @vseventer @egreaves - thoughts?

waxidiotic:
Closing as this has been pushed back to 8.10.0.

PR #3436, opened 2019-06-13
[JW8-9161] Debounce submenu open to prevent duplicate tap events

zetagame:
### This PR will...
Debounce the submenu open call, preventing the duplicate tap event from focusing an un-clicked menu option.

For a currently unknown reason, clicking a button that opens a submenu triggers 2-3 tap events.

Do you know where the additional taps are dispatched? If so, let me know as I'd prefer to take a more holistic approach.

#### Addresses Issue(s):

JW8-9161

### DIFF
--- a/src/js/view/controls/settings-menu.js
+++ b/src/js/view/controls/settings-menu.js
@@ -1,3 +1,4 @@
+import { debounce } from '../../utils/underscore';
 import { SettingsMenu } from 'view/controls/components/settings/menu';
 import {
     addCaptionsSubmenu,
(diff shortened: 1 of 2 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/controls/settings-menu.js
```diff
@@ -43,7 +46,8 @@ export function createSettingsMenu(controlbar, onVisibility, localization) {
                 // Activate the first submenu if clicking the default button
                 settingsMenu.activateFirstSubmenu(nonKeyboardInteraction);
             }
-            settingsMenu.open(isDefault, event);
+        
+            delayedOpen(isDefault, event);
```
jnatalzia: i understand the fix as is but i wonder if this is a case where event propagation is causing the multiple firings. this line in particular 
> clicking a button that opens a submenu triggers 2-3 tap events.
makes me wonder if it's just bubbling and an easier fix would be to `stopPropagation`? maybe you've already tried it

zetagame: As per our convo, stopPropogation won't help here because it's a single listener on a specific button causing the issue, it's not bubbling. PreventDefault isn't applicable without performance issues since we would have to set ui.passive to false, which would decrease performance of all event handling on settings menu.

PR #3449, opened 2019-07-18
Always pass the captions tracks to the captions renderer live restart rendering

jnatalzia:
### This PR will...
Ensure that captions tracks are always passed to the renderer

We destroy and recreate the hlsjs instance when live is stopped and resumed. Since the caption tracks were not always passed, when we were to clear them on destroy the renderer would not have access to them when spinning back up, requiring the user to turn them off and then on again

Nah

Nah
#### Addresses Issue(s):

JW8-9105

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -779,10 +779,10 @@ Object.assign(Controller.prototype, {
             return _programController.audioTracks;
         }
 
-        function _setCurrentCaptions(index) {
+        function _setCurrentCaptions(index, tracks) {
             index = parseInt(index, 10) || 0;
 
-            _model.persistVideoSubtitleTrack(index);
+            _model.persistVideoSubtitleTrack(index, tracks);
             _programController.subtitles = index;
 
             _this.trigger(CAPTIONS_CHANGED, {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

jnatalzia:
@robwalch I found three other places where it's used but I don't think there's any need to change those

https://github.com/jwplayer/jwplayer/blob/63b959105fed48f921e137f6b5e76313d0c43893/src/js/api/api.js#L634 - Sets the currently selected caption through API, no change to the tracks so need to set the ones that are persisted

https://github.com/jwplayer/jwplayer/blob/63b959105fed48f921e137f6b5e76313d0c43893/src/js/view/controls/controls.js#L333 - Same thing as above, sets the currently selected caption track, doesn't set any new tracks

https://github.com/jwplayer/jwplayer/blob/63b959105fed48f921e137f6b5e76313d0c43893/src/js/program/program-listeners.js#L154 - this is the other place where we were calling persist but it's already being passed the tracks so no need for change

PR #3455, opened 2019-07-23
[JW8-9533] Add autoPause option to pause ads

waxidiotic:
JW8-9533

### This PR will...

- Cause the player to automatically pause during ad playback when player becomes not viewable, if `autoPause.viewability` and `autoPause.pauseAds` configuration options are both `true`
- If ads are paused due to the player becoming not viewable, they should resume when the player becomes viewable.

- Previously, ad playback would continue and then pause afterwards if player becomes not viewable.
- Customers have requested the ability to automatically pause ad playback in outstream players.

n/a

https://github.com/jwplayer/jwplayer-commercial/pull/6890

#### Addresses Issue(s):

JW8-9533

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -10,8 +10,11 @@ import { getLanguage, getCustomLocalization, applyTranslation, normalizeIntl } f
 /* eslint camelcase: 0 */
 // Defaults
 // Alphabetical order
-export const Defaults = {
-    autoPause: { viewability: false },
+const Defaults = {
+    autoPause: {
+        viewability: false,
+        pauseAds: false
+    },
     autostart: false,
     bandwidthEstimate: null,
     bitrateSelection: null,
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -10,7 +10,7 @@ import { getLanguage, getCustomLocalization, applyTranslation, normalizeIntl } f
 /* eslint camelcase: 0 */
 // Defaults
 // Alphabetical order
-const Defaults = {
+export const Defaults = {
     autoPause: {
         viewability: false,
         pauseAds: false
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -364,7 +364,7 @@ Object.assign(Controller.prototype, {
             if (model.get('playOnViewable')) {
                 if (viewable) {
                     const reason = 'viewable';
-                    if (model.get('state') === STATE_IDLE) {
+                    if (model.get('state') === STATE_IDLE && !(model.get('autoPause').pauseAds && adState)) {
                         _autoStart(reason);
                     } else {
                         _play({ reason });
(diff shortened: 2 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

waxidiotic:
test this please

src/js/controller/controller.js
```diff
@@ -489,7 +493,8 @@ Object.assign(Controller.prototype, {
 
             const adState = _getAdState();
             const pauseReason = _model.get('pauseReason');
-            if (adState && pauseReason === 'viewable' && playReason !== 'interaction') {
+            const autoPauseAds = _model.get('autoPause').pauseAds;
+            if (adState && !autoPauseAds && pauseReason === 'viewable' && playReason !== 'interaction') {
```
robwalch: Why are you adding `pauseAds` logic in the `play` command? I would expect this logic to be outside of the `play` method where necessary. Doing it here will block API calls from toggling ad playback when pauseAds is false which we don't want.

Even the logic before this change seems really sketchy and should be scoped to (inside) the next `if(adState)` block.

waxidiotic: I didn't consider this `pauseAds` 'logic' because this isn't pausing anything - it's more the logic for resuming playback if `pauseAds` is set. However, if you feel like it would be better if I abstract this out, then I will do so.

src/js/controller/controller.js
```diff
@@ -366,7 +366,7 @@ Object.assign(Controller.prototype, {
             if (model.get('playOnViewable')) {
                 if (viewable) {
                     const reason = 'viewable';
-                    if (model.get('state') === STATE_IDLE) {
+                    if (model.get('state') === STATE_IDLE && !(model.get('autoPause').pauseAds && adState)) {
```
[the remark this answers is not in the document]
waxidiotic: @robwalch - I have no problem doing this but I'm just curious as to the reason for it. Is it just to make the code cleaner?

src/js/controller/controller.js
robwalch: > Doing it here will block API calls from toggling ad playback when pauseAds is false which we don't want

waxidiotic:
test this please

src/js/controller/controller.js
waxidiotic: If `pauseAds` is true and I pause an ad by scrolling out of view, I'm able to call `.play()` in the console and have it resume.

waxidiotic: @robwalch - I just tested a bunch with `pauseAds: false` and everything worked as expected.

robwalch: It's seems like the `playReason` should be `"viewable"` too here. No?

What if you undid the changes to this if statement, and instead did not call play() on viewable when `!autoPauseAds`? (`_checkPlayOnViewable` -> `if (viewable) {`)

waxidiotic: Are you talking about in this conditional?

`else if (!adState || model.get('autoPause').pauseAds)`

If so, then I'm not quite understanding because we would still need to call `play()` if `pauseAds: false` if we're not in ad playback.

waxidiotic:
test this please

src/js/api/config.js
```diff
@@ -10,8 +10,11 @@ import { getLanguage, getCustomLocalization, applyTranslation, normalizeIntl } f
 /* eslint camelcase: 0 */
 // Defaults
 // Alphabetical order
-export const Defaults = {
-    autoPause: { viewability: false },
+const Defaults = {
```
robwalch: woops!

waxidiotic:
test this please

waxidiotic:
test this please

PR #3456, opened 2019-07-26
[JW8-9936] Sanitize Right Click Menu links before appending to DOM.

DanFerrer:
### This PR will...

- Sanitize attributes of all elements instead of just ```img``` and ```svg``` tags.
- Pass ```true```  when calling```cloneNode```. 

- To prevent injecting JavaScript in the right click menu through config options.
- To make sure the text of an html element is copied over when calling ```appendHtml```. We currently don't do that if you pass in html that has text in it. (ex: the description of a video is ```<div>foobar</div>```. Foobar would get stripped out but the empty div would still be appended to the DOM as the description.

- Could looping through every node, slow down setup time?

#### Addresses Issue(s):

JW8-9936

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -25,7 +25,7 @@ function appendHtml(element, html) {
     const fragment = document.createDocumentFragment();
     const nodes = htmlToParentElement(html).childNodes;
     for (let i = 0; i < nodes.length; i++) {
-        fragment.appendChild(nodes[i].cloneNode());
+        fragment.appendChild(nodes[i].cloneNode(true));
     }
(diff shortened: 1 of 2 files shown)


src/js/utils/dom.js
```diff
@@ -39,7 +39,7 @@ export function htmlToParentElement(html) {
     // Delete script nodes
     sanitizeScriptNodes(parsedElement);
     // Delete event handler attributes that could execute XSS JavaScript
-    const insecureElements = parsedElement.querySelectorAll('img,svg');
```
zetagame: could be parsedElement.childNodes

PR #3458, opened 2019-07-30
[JW8-8984] Change player aspect ratio in between videos

robwalch:
### This PR will...
Resize the player when `setConfig` is called with `width` `height` or `aspectratio`. The "resize" event if fired immediately after a change to `width` and `height` or `aspectratio`.

So that developers can update any combination of `width`, `height`, `aspectratio` and `stretching` after the player is setup. So, if you just want to update `aspectratio` between playlist items:

```
{
        width: '100%',
        aspectratio: '16:9',
        events: {
// If your feed has width and height info you could do this on "playlistItem"
            visualQuality: function(e) {
                if (e.level && e.level.width) {
                    this.setConfig({
                        aspectratio: e.level.width + ':' + e.level.height
                    });
                }
            }
        },
        playlist: '//video-feed.json'
}
``` 

Here's an example that will resize the player to 640x360 while ads are playing, with ads filling the player viewport, and then switch to a responsive width with an aspect ratio that matches video when content is playing:
```
{
        // `aspectratio` is only supported with responsive widths
        //  (width must be a percentage)
        width: '100%',
        aspectratio: '16:9',
        events: {
            visualQuality: function(e) {
                if (e.level && e.level.width) {
                    this.setConfig({
                        width: '100%',
                        aspectratio: e.level.width + ':' + e.level.height
                    });
                }
            },
            adBreakStart: function() {
                this.setConfig({
                    width: 640,
                    height: 360,
                    stretching: 'fill'
                });
            },
            adBreakEnd: function() {
                this.setConfig({
                    width: 640,
                    height: 360,
                    stretching: 'uniform'
                });
            }
        },
        playlist: '//video-feed.json',
        advertising: {
            skipoffset: 1,
            tag: 'ad-tag.xml'
        },
    }
```

https://github.com/jwplayer/jwplayer-commercial/pull/6910

#### Addresses Issue(s):
JW8-8984

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -1,3 +1,4 @@
+import { normalizeAspectRatio, normalizeSize } from 'api/config-normalization';
 import { loadFrom, getScriptPath } from 'utils/playerutils';
 import { serialize } from 'utils/parser';
 import { isValidNumber, isNumber, pick, isBoolean } from 'utils/underscore';
@@ -44,13 +45,6 @@ function _deserialize(options) {
         }
         options[key] = serialize(options[key]);
     });
-}
-
-export function normalizeSize(val) {
-    if (val.slice && val.slice(-2) === 'px') {
-        val = val.slice(0, -2);
-    }
-    return val;
 }
 
 
@@ -168,27 +162,5 @@ const Config = function(options, persisted) {
     config.backgroundLoading = isBoolean(config.backgroundLoading) ? config.backgroundLoading : Features.backgroundLoading;
     return config;
 };
-
-export function normalizeAspectRatio(ar, width) {
-    if (width.toString().indexOf('%') === -1) {
-        return 0;
-    }
-    if (typeof ar !== 'string' || !ar) {
-        return 0;
-    }
-    if (/^\d*\.?\d+%$/.test(ar)) {
-        return ar;
-    }
-    const index = ar.indexOf(':');
-    if (index === -1) {
-        return 0;
-    }
-    const w = parseFloat(ar.substr(0, index));
-    const h = parseFloat(ar.substr(index + 1));
-    if (w <= 0 || h <= 0) {
-        return 0;
-    }
-    return (h / w * 100) + '%';
-}
 
 export default Config;
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -46,7 +46,7 @@ function _deserialize(options) {
     });
 }
 
-function _normalizeSize(val) {
+export function normalizeSize(val) {
     if (val.slice && val.slice(-2) === 'px') {
         val = val.slice(0, -2);
     }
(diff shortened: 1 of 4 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/api/set-config.js
```diff
@@ -19,30 +10,55 @@ function setAutoStart(model, controller, autoStart) {
 
 export default (controller, newConfig) => {
     const model = controller._model;
+    const attributes = model.attributes;
 
-    if (!size(newConfig)) {
-        return;
+    if (newConfig.height) {
```
jnatalzia: thoughts on maybe breaking this out into a diff func? feels a little strange to have 2 separate config mutators in the same func which seem responsible for different ways of attacking the problem (explicit logic around sizing vs a switch case depending on attribute)

src/js/view/view.js
```diff
@@ -481,6 +481,10 @@ function View(_api, _model) {
         style(aspectRatioContainer, {
             paddingTop: aspectratio || null
         });
+        if (_this.isSetup && aspectratio) {
```
jnatalzia: does this fix an existing bug or is this a new requirement with these `set after load` changes?

src/js/api/config.js
```diff
@@ -169,7 +169,7 @@ const Config = function(options, persisted) {
     return config;
 };
 
-function _evaluateAspectRatio(ar, width) {
+export function normalizeAspectRatio(ar, width) {
```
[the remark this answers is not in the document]
robwalch: I think so. They are still used here first. I see no reason to move them. They are not utils. Config normalization should be done in config.

src/js/api/set-config.js
robwalch: `width` `height` and `aspectratio` need to be handled at once. The only place where it makes sense to do this is in `view.resize` but I don't want to change it at this point. These cannot be handled one at a time in the switch statement - since we could end up resizing 3 times.

src/js/view/view.js
robwalch: new requirement. `onAspectRatioChange` is called on view setup where the player element is already being styled and we don't need or want to call `_responsiveUpdate`. After setup, we need to do this. We do it when `resize` is called, but `resize` only handles `width` and `height` changes with no `aspectratio`. I kept these two ways of resizing separate to avoid regressions.

src/js/api/set-config.js
jnatalzia: totally agree, was more curious if you thought it helpful to break that out into another function for code clarity purposes

src/js/view/view.js
jnatalzia: ahh got it. as an aside, i read `this.isSetup` as in `this is in the setup phase` not `this has been setup` - non-critical just peanut gallery-ing

robwalch: `this.isSpartaaaa`

PR #3460, opened 2019-08-06
[JW8-10020] Only show Client Id In info-overlay if opted into tracking

DanFerrer:
### This PR will...

- Append ```clientId``` to the ```info-overlay``` template if the publisher has opted into tracking. We wont show it otherwise.

We don't want to create the client id portion of the template nor show the client id if tracking isnt enabled.

#### Addresses Issue(s):

JW8-10020


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

bobbyjw:
@nebnitt please take a look when you have sec;  nebnitt aka Dan Saronson will likely be working on the ping plugin side of things and I want to be sure we have a meeting of the minds here  🤝

src/js/view/controls/info-overlay.js
```diff
@@ -6,7 +6,9 @@ import { cloneIcon } from 'view/controls/icons';
 import { timeFormat } from 'utils/parser';
 
 export default function (container, model, api, onVisibility) {
-    const template = createElement(InfoOverlayTemplate());
+    const analytics = api.getPlugin('jwpsrv');
+    const shouldTrackUser = analytics && !analytics.doNotTrackUser();
```
nebnitt: Looks like there's an assumption here that `doNotTrackUser()` will _always_ exist in the ping plugin. I can make sure that's the case -- I imagine this adds negligible overhead so it should be OK. There's an alternate path, I think, where we only call `doNotTrackUser()` in cases where the player config has the `doNotTrack` option, and then only the "do not track" version of the plugin has the function exposed. But that may be needlessly complex, I think this is fine. 

LGTM!

jnatalzia: we should absolutely not assume it will always be present - in case we ever push plugins out over older versions or mistag (we have done this in the past)

nebnitt: good point, thanks Joe. In that case let's make sure to check for the existence of the function before calling it!

src/js/view/controls/info-overlay.js
```diff
@@ -57,7 +68,10 @@ export default function (container, model, api, onVisibility) {
             }
             durationContainer.textContent = durationText;
         }, instance);
-        clientIdContainer.textContent = `Client ID: ${getClientId()}`;
+
+        if (clientIdContainer) {
```
jnatalzia: what is the benefit to showing/hiding the div vs. making `textContent = ''` here? are theree styling implications to leaving the div there but not having any info inside of it?

DanFerrer: I figured we wouldn't want to create the div to begin with if client id isn't available

jnatalzia: ah fair!

jnatalzia: ah totally!

jnatalzia: ahh fair!

robwalch: Joe's suggestion would require less code as the template would remain unchanged (smaller) and you would call `shouldTrackUser()` here (lazier) rather than up top.
`clientIdContainer.textContent = shouldTrackUser() ? '' : `Client ID: ${getClientId()}`;`

test/unit/client-id-test.js
```diff
@@ -0,0 +1,34 @@
+// import InfoOverlay from 'view/controls/info-overlay';
+// import MockModel from 'mock/mock-model';
+// import MockApi from 'mock/mock-api';
+// import { registerPlugin } from 'plugins/plugins';
+
```
robwalch: Did you mean to include this file? It's completely commented out.

src/js/view/controls/templates/info-overlay.js
```diff
@@ -5,6 +5,6 @@ export default () => (
             `<div class="jw-reset-text jw-info-duration" dir="auto"></div>` +
             `<div class="jw-reset-text jw-info-description" dir="auto"></div>` +
         `</div>` +
-        `<div class="jw-reset jw-info-clientid"></div>` +
+        '<div class="jw-reset jw-info-clientid"></div>' +
```
robwalch: this is odd

test/unit/client-id-test.js
```diff
@@ -0,0 +1,51 @@
+import InfoOverlay from 'view/controls/info-overlay';
+import MockModel from 'mock/mock-model';
+import MockApi from 'mock/mock-api';
+
+describe('Info Overlay Client ID', function () {
+    let model = new MockModel();
```
robwalch: const

test/unit/client-id-test.js
```diff
@@ -0,0 +1,51 @@
...
+    model.setup({});
+
+    function noop () {}
+
+    function createAPI(doNotTrackUser) {
+        let api = new MockApi();
+        let oldGetPlugin = api.getPlugin;
```
robwalch: consts

test/unit/client-id-test.js
```diff
@@ -0,0 +1,51 @@
...
+    }
+
+    afterEach(() => {
+        localStorage.removeItem('jwplayerLocalId');
+    });
+
+    it('should not show client id in the info overlay if doNotTrackUser returns true', function() {
+        let api = createAPI(true);
+        let element = document.createElement('div');
+        let infoOverlay = new InfoOverlay(element, model, api, noop);
```
robwalch: more consts

test/unit/client-id-test.js
```diff
@@ -0,0 +1,51 @@
...
+        infoOverlay.open();
+        expect(element.querySelector('.jw-info-clientid').innerText).to.eql('');
+
+    });
+    it('should show client id in the info overlay if doNotTrackUser returns false', function() {
+        localStorage.setItem('jwplayerLocalId', 'test123');
+
+        let api = createAPI(false);
+        let element = document.createElement('div');
+        let infoOverlay = new InfoOverlay(element, model, api, noop);
```
robwalch: even more consts

src/js/view/controls/templates/info-overlay.js
jnatalzia: will revert

PR #3462, opened 2019-08-07
Adds the name parameter

canvasplay:
This just seems to be missing as the parameter is already passed in by the submenu component.

### This PR will...

Add an additional class to jw-settings-submenu panels.

jw-settings-submenu panels are indistinguishable via CSS.

Maybe... className naming conventions. Is the suffix approach ok?

#### Addresses Issue(s):

JW8-3461

### DIFF
--- a/src/js/view/controls/templates/settings/submenu.js
+++ b/src/js/view/controls/templates/settings/submenu.js
@@ -1,6 +1,6 @@
-export default () => {
+export default (name) => {
     return (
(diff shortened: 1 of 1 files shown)


robwalch:
Thanks for the contribution @canvasplay. This makes sense as long as the name in `jw-settings-submenu-${name}` matches the names in classes on each tab button )ex: `.jw-settings-quality .jw-submenu-quality`, `jw-settings-playbackRates ` etc...).

robwalch:
@DanFerrer @jnatalzia merging so that users can customize individual menus.

PR #3467, opened 2019-08-19
[JW8-9807] Reset custom font-size after exiting fullscreen on iOS

robwalch:
### This PR will...
Force iOS to resize captions shadow-dom elements when setting a custom font-size.

When exiting fullscreen on iOS, the browser ignores the important font-size rules that we attempt to restore. Setting a rule to inherit font-size first, forces the browser to update layout.

#### Addresses Issue(s):
JW8-9807

### DIFF
--- a/src/js/view/captionsrenderer.js
+++ b/src/js/view/captionsrenderer.js
@@ -273,12 +273,15 @@ const CaptionsRenderer = function (viewModel) {
 
     function _setShadowDOMFontSize(playerId, fontSize) {
         // Set Shadow DOM font size (needs to be important to override browser's in line style)
-        _windowStyle.fontSize = fontSize;
         const selector = `#${playerId} .jw-video::-webkit-media-text-track-display`;
-        if (fontSize && OS.iOS) {
-            // Force layout after exiting fullscreen
-            css(selector, { fontSize: 'inherit' }, playerId, true);
+        if (fontSize) {
+            fontSize += 'px';
+            if (OS.iOS) {
+                // Force layout after exiting fullscreen
+                css(selector, { fontSize: 'inherit' }, playerId, true);
+            }
         }
+        _windowStyle.fontSize = fontSize;
         css(selector, _windowStyle, playerId, true);
     }
 
--- a/src/js/view/captionsrenderer.js
+++ b/src/js/view/captionsrenderer.js
@@ -200,7 +200,7 @@ const CaptionsRenderer = function (viewModel) {
 
         let fontSize;
         if (_model.get('fullscreen') && OS.iOS) {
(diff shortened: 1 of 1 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/view/captionsrenderer.js
```diff
@@ -200,7 +200,7 @@ const CaptionsRenderer = function (viewModel) {
 
         let fontSize;
         if (_model.get('fullscreen') && OS.iOS) {
-            fontSize = 'inherit';
+            fontSize = null;
```
robwalch: When going fullscreen we want to clear the custom rules, not set them to inherit. This only appeared to work as expected because the value 'inherit' was made invalid by appending 'px' below.

src/js/view/captionsrenderer.js
```diff
@@ -273,8 +273,16 @@ const CaptionsRenderer = function (viewModel) {
 
     function _setShadowDOMFontSize(playerId, fontSize) {
         // Set Shadow DOM font size (needs to be important to override browser's in line style)
-        _windowStyle.fontSize = fontSize + 'px';
-        css('#' + playerId + ' .jw-video::-webkit-media-text-track-display', _windowStyle, playerId, true);
+        const selector = `#${playerId} .jw-video::-webkit-media-text-track-display`;
+        if (fontSize) {
+            fontSize += 'px';
```
robwalch: The `css` module will handle adding 'px' numeric style values, but it rounds the values up. Add 'px' here so we can use fractional font-sizes. If `fontSize` is null, then we're clearing styles.

src/js/view/captionsrenderer.js
jnatalzia: 🍀

src/js/view/captionsrenderer.js
```diff
@@ -273,8 +273,16 @@ const CaptionsRenderer = function (viewModel) {
...
+            if (OS.iOS) {
+                // Force layout after exiting fullscreen
```
jnatalzia: is there a way to scope this function to only run on fs exit?

robwalch: We could check if `fontSize` was set to `null` because of what happens before this function is run:
```
if (_model.get('fullscreen') && OS.iOS) {
     fontSize = null;
```
But I'd rather not add side-effects or pass in the model. Ensuring these styles update on both enter and exit of fullscreen is best.

Issue #3474, opened 2019-08-29
Expose the language attribute on tracks returned by `getCaptionsList()`


drewwyatt:
### Expected Behavior

Expose track `language` attribute along with `label` and `id` when calling `getCaptionsList()`.

### Actual Behavior

JWPlayer exposes only `label` and `id`.

- [documentation](https://developer.jwplayer.com/jw-player/docs/javascript-api-reference/#jwplayergetcaptionslist)
- [relevant code](https://github.com/jwplayer/jwplayer/blob/master/src/js/controller/captions.js#L98-L110)

### Motivation

`LANGUAGE` is [part of the HLS specification for alternate media](https://developer.apple.com/documentation/http_live_streaming/example_playlists_for_http_live_streaming/adding_alternate_media_to_a_playlist) as an optional attribute used to "[identify] the primary language used in the media selection." This differs from `NAME` (returned as `label` by JWPlayer), which the HLS specification indicates is a "[description of] the primary language used in the media selection."

It’s generally good practice to be distinguish a display name (for user presentation) from a programatic identifier, so as to be able to vary these independently. The ability to do so via separate `name` (`label`) and `language` attributes corresponding to the HLS spec is common in video players (examples of some are provided below).

User presentation makes the `NAME` attribute an especially poor candidate for unique language identifiers (like locales) because even in instances where JWPlayer's default UI is being overridden, the `name` attribute will still be exposed to users when the media is played on devices not controlled by JWPlayer (e.g. iOS Safari's web player, Chromecast, AppleTV) where `NAME` is assumed to be user-facing (as the HLS specification suggests). 

### Examples

This behavior can be seen in the following players

- [Brightcove](https://docs.brightcove.com/brightcove-player/current-release/TextTrack.html) via `TextTrack`'s `language` property
- [AVFoundation (iOS/macOS/tvOS/watchOS)](https://developer.apple.com/documentation/avfoundation/avmediaselectionoption) via `AVMediaSelectionOption`'s `locale` property
- [ExoPlayer](https://github.com/google/ExoPlayer/blob/44293e8f45fc62ff918001b4425c44d0ded5ffc9/library/core/src/main/java/com/google/android/exoplayer2/Format.java#L160) via `TrackGroup` `Format`'s `language` property

jnatalzia:
Hi Drew!

[...]

PR #3475, opened 2019-08-29
add language to captions returned by _captionsMenu

drewwyatt:
### This PR will...

Address #3474 and add the `language` property to tracks returned by `_captionsMenu()` (and, by extension, `getCaptionsList()`).

`LANGUAGE` is [part of the HLS specification for alternate media](https://developer.apple.com/documentation/http_live_streaming/example_playlists_for_http_live_streaming/adding_alternate_media_to_a_playlist) as an optional attribute used to "[identify] the primary language used in the media selection." This differs from `NAME` (returned as `label` by JWPlayer), which the HLS specification indicates is a "[description of] the primary language used in the media selection."

It’s generally good practice to be distinguish a display name (for user presentation) from a programatic identifier, so as to be able to vary these independently. The ability to do so via separate `name` (`label`) and `language` attributes corresponding to the HLS spec is common in video players (examples of some are provided below).

User presentation makes the `NAME` attribute an especially poor candidate for unique language identifiers (like locales) because even in instances where JWPlayer's default UI is being overridden, the `name` attribute will still be exposed to users when the media is played on devices not controlled by JWPlayer (e.g. iOS Safari's web player, Chromecast, AppleTV) where `NAME` is assumed to be user-facing (as the HLS specification suggests). 

None that I know of.

No.

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -103,7 +103,8 @@ const Captions = function(_model) {
         for (let i = 0; i < _tracks.length; i++) {
             list.push({
                 id: _tracks[i]._id,
-                label: _tracks[i].name || 'Unknown CC'
+                label: _tracks[i].name || 'Unknown CC',
+                language: _tracks[i].language,
             });
         }
         return list;


jnatalzia:
Related: https://github.com/jwplayer/jwplayer/issues/3474

drewwyatt:
hey @robwalch, I'm able to see the `language` property populate whenever I load a playlist that has subtitle media tracks where `LANGUAGE` is specified (it's an optional attribute). I was able to see language show up when loading [this playlist](https://gist.github.com/drewwyatt/886eeeb1366e15b32fd3fac1e09bd725#file-playlist-m3u8-L9-L12).

robwalch:
Cool! I think we can include this in the next minor release.

PR #3476, opened 2019-09-04
[JW8-10237] Await Shaka destroy

vseventer:
### This PR will...
Wait for Shaka's destroy handler before starting ad playback.

Shaka was manipulating the shared videotag at the same time as the HTML5 provider, causing ads not to play.

N/A

jwplayer/jwplayer-commercial#7028

#### Addresses Issue(s):
JW8-10237

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -32,6 +32,7 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
     let _arrayOptions;
     let _arrayIndex = 0;
     let _data = {};
+    let _destroyPromise = Promise.resolve();
     let _options = {};
     let _skipAd = _instreamItemNext;
     let _backgroundLoadTriggered = false;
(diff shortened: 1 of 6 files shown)


johnBartos:
<table>
  <thead>
    <tr>
      <th width="50"></th>
    </tr>
  </thead>
  <tbody><tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  
<tr>
      <td>:warning:</td>
      <td>

  </td>
    </tr>
  </tbody>
</table>



<p align="right">
</p>

src/js/controller/instream-adapter.js
```diff
@@ -98,6 +99,12 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         // Make sure the original player's provider stops broadcasting events (pseudo-lock...)
         _controller.detachMedia();
 
+        // Shaka cleans the videotag async, so wait until its destroy promise fulfills.
+        const provider = _model.get('provider');
+        if (!_model.get('backgroundLoading') && provider && provider.name === 'shaka') {
+            _destroyPromise = _model._provider.destroy();
```
robwalch: This would result in a second call to provider.destroy(). Can we get the promise from the first one (if `detachMedia` returns a promise, then we have a destroy promise)?

vseventer: `detachMedia` assigns `programController.attached = false`, which is a setter. This setter calls `provider.destroy`, so I'm not sure how we could get the destroy promise this way.

robwalch: Change it to a method (+ read-only getter). We should not use setters for async actions.

PR #3477, opened 2019-09-05
Float on scroll on mobile when player reaches floating point

jnatalzia:
### This PR will...
Update mobile float on scroll so that the player starts to scroll when it reaches expected floating position. Currently it starts when viewability is <50%

It's a better experience by far on mobile

Nah

None

#### Addresses Issue(s):

JW8-10219

### DIFF
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -84,10 +84,6 @@
 }
 
 @keyframes jw-float-to-top {
-    from {
-        transform: translateY(-100%);
-    }
-
     to {
         transform: translateY(0);
     }
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -25,7 +25,7 @@
 
         .jw-flag-touch& {
             @media screen and (max-device-width : 480px) and (orientation: portrait) {
-                animation: jw-float-to-top 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
+                animation: none;
                 top: 62px;
                 bottom: auto;
                 left: 0;
@@ -78,12 +78,6 @@
         transform: translateY(100%);
     }
 
-    to {
-        transform: translateY(0);
-    }
-}
-
-@keyframes jw-float-to-top {
     to {
         transform: translateY(0);
     }
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -224,3 +224,7 @@ export function openLink(link, target, additionalOptions = {}) {
         a.click();
     }
 }
+
+export function browserIsLandscape() {
+    return window.innerWidth > window.innerHeight;
+}
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -225,6 +225,11 @@ export function openLink(link, target, additionalOptions = {}) {
     }
 }
 
-export function browserIsLandscape() {
-    return window.innerWidth > window.innerHeight;
+export function deviceIsLandscape() {
+    const ort = window.screen.orientation;
+    const isLandscape = ort ?
+        ort.type === 'landscape-primary' || ort.type === 'landscape-secondary'
+        : false;
+
+    return isLandscape || (window.orientation === 90 || window.orientation === -90);
 }
--- a/src/js/view/utils/views-manager.js
+++ b/src/js/view/utils/views-manager.js
@@ -105,7 +105,9 @@ export default {
             scrollHandlerInitialized = true;
             window.addEventListener('scroll', onScroll);
         }
-        scrollHandlers.push(handler);
+        if (scrollHandlers.indexOf(handler) === -1) {
+            scrollHandlers.push(handler);
+        }
     },
     removeScrollHandler: function(handler) {
         scrollHandlers.splice(scrollHandlers.indexOf(handler), 1);
--- a/src/js/view/utils/views-manager.js
+++ b/src/js/view/utils/views-manager.js
@@ -105,12 +105,14 @@ export default {
             scrollHandlerInitialized = true;
             window.addEventListener('scroll', onScroll);
         }
-        if (scrollHandlers.indexOf(handler) === -1) {
-            scrollHandlers.push(handler);
-        }
+
+        scrollHandlers.push(handler);
     },
     removeScrollHandler: function(handler) {
-        scrollHandlers.splice(scrollHandlers.indexOf(handler), 1);
+        let idx = scrollHandlers.indexOf(handler);
+        if (idx !== -1) {
+            scrollHandlers.splice(idx, 1);
+        }
     },
     addWidget: function(widget) {
         widgets.push(widget);
--- a/src/js/view/utils/views-manager.js
+++ b/src/js/view/utils/views-manager.js
@@ -9,6 +9,7 @@ const hasOrientation = 'screen' in window && 'orientation' in window.screen;
 const isAndroidChrome = OS.android && Browser.chrome;
 
 let intersectionObserver;
+let scrollHandlerInitialized = false;
 
 function lazyInitIntersectionObserver() {
     const IntersectionObserver = window.IntersectionObserver;
@@ -85,15 +86,12 @@ if (isAndroidChrome && hasOrientation) {
 window.addEventListener('beforeunload', () => {
     document.removeEventListener('visibilitychange', onVisibilityChange);
     document.removeEventListener('webkitvisibilitychange', onVisibilityChange);
-    document.removeEventListener('webkitvisibilitychange', onVisibilityChange);
     window.removeEventListener('scroll', onScroll);
 
     if (isAndroidChrome && hasOrientation) {
         window.screen.orientation.removeEventListener('change', onOrientationChange);
     }
 });
-
-window.addEventListener('scroll', onScroll);
 
 export default {
     add: function(view) {
@@ -103,6 +101,10 @@ export default {
         removeFromGroup(view, views);
     },
     addScrollHandler: function(handler) {
+        if (!scrollHandlerInitialized) {
+            scrollHandlerInitialized = true;
+            window.addEventListener('scroll', onScroll);
+        }
         scrollHandlers.push(handler);
     },
     removeScrollHandler: function(handler) {
--- a/src/js/view/utils/views-manager.js
+++ b/src/js/view/utils/views-manager.js
@@ -1,5 +1,6 @@
 import activeTab from 'utils/active-tab';
 import { Browser, OS } from 'environment/environment';
+import { deviceIsLandscape } from 'utils/dom';
 
 const views = [];
 const widgets = [];
@@ -46,8 +47,7 @@ function onOrientationChange() {
         }
 
         const state = model.get('state');
-        const orientation = window.screen.orientation.type;
-        const isLandscape = orientation === 'landscape-primary' || orientation === 'landscape-secondary';
+        const isLandscape = deviceIsLandscape();
 
         if (!isLandscape && state === 'paused' && view.api.getFullscreen()) {
             view.api.setFullscreen(false);
--- a/src/js/view/utils/views-manager.js
+++ b/src/js/view/utils/views-manager.js
@@ -3,6 +3,7 @@ import { Browser, OS } from 'environment/environment';
 
 const views = [];
 const widgets = [];
+const scrollHandlers = [];
 const observed = {};
 const hasOrientation = 'screen' in window && 'orientation' in window.screen;
 const isAndroidChrome = OS.android && Browser.chrome;
@@ -68,6 +69,12 @@ function removeFromGroup(view, group) {
     }
 }
 
+function onScroll(e) {
+    scrollHandlers.forEach(handler => {
+        handler(e);
+    });
+}
+
 document.addEventListener('visibilitychange', onVisibilityChange);
 document.addEventListener('webkitvisibilitychange', onVisibilityChange);
 
@@ -78,11 +85,15 @@ if (isAndroidChrome && hasOrientation) {
 window.addEventListener('beforeunload', () => {
     document.removeEventListener('visibilitychange', onVisibilityChange);
     document.removeEventListener('webkitvisibilitychange', onVisibilityChange);
+    document.removeEventListener('webkitvisibilitychange', onVisibilityChange);
+    window.removeEventListener('scroll', onScroll);
 
     if (isAndroidChrome && hasOrientation) {
         window.screen.orientation.removeEventListener('change', onOrientationChange);
     }
 });
+
+window.addEventListener('scroll', onScroll);
 
 export default {
     add: function(view) {
@@ -90,6 +101,12 @@ export default {
     },
     remove: function(view) {
         removeFromGroup(view, views);
+    },
+    addScrollHandler: function(handler) {
+        scrollHandlers.push(handler);
+    },
+    removeScrollHandler: function(handler) {
+        scrollHandlers.splice(scrollHandlers.indexOf(handler), 1);
     },
     addWidget: function(widget) {
         widgets.push(widget);
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -340,8 +340,48 @@ function View(_api, _model) {
         playerViewModel.change('playlistItem', onPlaylistItem);
         // Triggering 'resize' resulting in player 'ready'
         _lastWidth = _lastHeight = null;
+
+        // Setup floating scroll handler
+        if (_floatingConfig && OS.mobile) {
+            viewsManager.addScrollHandler(throttledFloatScrollHandler);
+        }
+
         this.checkResized();
     };
+
+    // Functions for handler float on scroll (mobile)
+    const topOffset = 62;
+    let canFire = true;
+    let debounceTO;
+    function checkFloatOnScroll() {
+        const floating = _model.get('isFloating');
+        const hasCrossedThreshold = playerBounds.top < topOffset ?
+            playerBounds.top <= window.scrollY :
+            playerBounds.top <= window.scrollY + topOffset;
+
+        if (!floating && hasCrossedThreshold) {
+            _updateFloating(0);
+        } else if (floating && !hasCrossedThreshold) {
+            _updateFloating(1);
+        }
+    }
+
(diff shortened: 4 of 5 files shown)


src/js/view/view.js
```diff
@@ -340,6 +340,45 @@ function View(_api, _model) {
         playerViewModel.change('playlistItem', onPlaylistItem);
         // Triggering 'resize' resulting in player 'ready'
         _lastWidth = _lastHeight = null;
+
+        // Setup floating scroll handler
+        const topOffset = 62;
```
jnatalzia: i hate this but it's a magic number in the css too

src/js/view/view.js
```diff
@@ -340,6 +340,45 @@ function View(_api, _model) {
...
+        if (_floatingConfig && _api.getEnvironment().OS.mobile) {
+            let canFire = true;
+            let debounceTO;
+
+            const checkFloat = () => {
+                const floating = _model.get('isFloating');
+                const hasCrossedThreshold = playerBounds.top < topOffset ?
```
jnatalzia: if the player isn't 62px down the page, we need to float when the top of player is at top of viewport

src/js/view/view.js
```diff
@@ -340,6 +340,45 @@ function View(_api, _model) {
...
+                    playerBounds.top <= window.scrollY :
+                    playerBounds.top <= window.scrollY + topOffset;
+
+                if (!floating && hasCrossedThreshold) {
+                    _updateFloating(1, true);
```
jnatalzia: not in love with this but felt like over-refactoring to pull all but one line into a sep func

src/js/view/view.js
```diff
@@ -340,6 +340,45 @@ function View(_api, _model) {
...
+                } else if (floating && !hasCrossedThreshold) {
+                    _updateFloating(1, false);
+                }
+            };
+
+            const throttledScroll = () => {
```
jnatalzia: throttle to stop the player from killing the page - debounce the last call to ensure if a user stops scrolling when throttled we get one more check that can re-fix the player (important for players at top of page)

src/js/view/view.js
```diff
@@ -893,9 +932,9 @@ function View(_api, _model) {
         return _model.get('isFloating') ? _wrapperElement : _playerElement;
     }
 
-    function _updateFloating(intersectionRatio) {
+    function _updateFloating(intersectionRatio, forceFloat) {
```
jnatalzia: again not in love with this, will give it some thought but happy to get suggestions

src/js/view/view.js
```diff
@@ -340,6 +340,45 @@ function View(_api, _model) {
...
+
+                canFire = false;
+                checkFloat();
+
+                setTimeout(() => {
+                    canFire = true;
+                }, 50);
+            };
+
+            window.addEventListener('scroll', throttledScroll);
```
robwalch: Is this listener only ever added once? Where is it removed?

Global listeners like this should be handled once in views-manager.js and update all player views.

src/js/view/view.js
```diff
@@ -880,7 +919,7 @@ function View(_api, _model) {
         const intersectionRatio = Math.round(entry.intersectionRatio * 100) / 100;
         _model.set('intersectionRatio', intersectionRatio);
 
-        if (_floatingConfig) {
+        if (_floatingConfig && !_api.getEnvironment().OS.mobile) {
```
robwalch: `import { OS } from 'environment/environment';` don't use the api.

src/js/view/view.js
```diff
@@ -340,6 +340,45 @@ function View(_api, _model) {
         playerViewModel.change('playlistItem', onPlaylistItem);
         // Triggering 'resize' resulting in player 'ready'
         _lastWidth = _lastHeight = null;
+
+        // Setup floating scroll handler
+        const topOffset = 62;
+        if (_floatingConfig && _api.getEnvironment().OS.mobile) {
```
robwalch: `import { OS } from 'environment/environment';` don't use the api.

src/js/view/view.js
robwalch: That magic number is supposed to account for site headers. But the new ticket suggests we should get rid of that (even though it could break mobile sites currently using this feature).

src/js/view/utils/views-manager.js
```diff
@@ -78,19 +87,29 @@ if (isAndroidChrome && hasOrientation) {
...
+window.addEventListener('scroll', onScroll);
+
 export default {
     add: function(view) {
         views.push(view);
     },
     remove: function(view) {
         removeFromGroup(view, views);
     },
+    addScrollHandler: function(handler) {
```
jnatalzia: want to unit test this but unsure of the best way to make `scrollHandlers` accessible to the outside world. is a synthetic scroll event possible with our setup? could test that way with mocked funcs

src/css/jwplayer/flags/floatingplayer.less
```diff
@@ -84,10 +84,6 @@
 }
 
 @keyframes jw-float-to-top {
```
robwalch: get rid of this because it doesn't do anything now.

robwalch: Can we do a transition on top or translateY to go from current pos (or 0) to the magic number.

src/js/view/utils/views-manager.js
```diff
@@ -78,19 +85,29 @@ if (isAndroidChrome && hasOrientation) {
 window.addEventListener('beforeunload', () => {
     document.removeEventListener('visibilitychange', onVisibilityChange);
     document.removeEventListener('webkitvisibilitychange', onVisibilityChange);
+    document.removeEventListener('webkitvisibilitychange', onVisibilityChange);
```
robwalch: Duplicate line?

src/js/view/utils/views-manager.js
```diff
@@ -78,19 +85,29 @@ if (isAndroidChrome && hasOrientation) {
...
+    window.removeEventListener('scroll', onScroll);
 
     if (isAndroidChrome && hasOrientation) {
         window.screen.orientation.removeEventListener('change', onOrientationChange);
     }
 });
 
+window.addEventListener('scroll', onScroll);
```
robwalch: npick but can this line go up to ln 80 to be grouped with other `addEventListener` calls?

src/js/view/utils/views-manager.js
robwalch: There are some examples of synthetic browser events in ui-test.js:

https://github.com/jwplayer/jwplayer/blob/ac5c60ac26757d01378552ed99eba9934735a56d/test/unit/ui-test.js#L34

src/js/view/utils/views-manager.js
robwalch: Maybe we don't need to do this until a scroll handler is added?

src/js/view/view.js
```diff
@@ -340,9 +340,49 @@ function View(_api, _model) {
...
         this.checkResized();
     };
 
+    // Functions for handler float on scroll (mobile)
+    const topOffset = 62;
+    let canFire = true;
+    let debounceTO;
+    function checkFloatOnScroll() {
+        const floating = _model.get('isFloating');
+        const hasCrossedThreshold = playerBounds.top < topOffset ?
```
robwalch: Is there a condition where we should exit early or always stop floating?
Examples:
- player is not in DOM
- bounds are empty (width or height === 0)

And should this function be invoked when inDOM or bounds change?

src/js/view/utils/views-manager.js
jnatalzia: nice catch

src/js/view/utils/views-manager.js
jnatalzia: great calls

src/js/view/view.js
jnatalzia: we remove the scroll handler when the player is destroyed so that takes care of #1 (other cleanup handled by other code) - for point #2 what function would you suggest to put that call into? it's a good catch, we should recheck when bounds change

src/js/view/view.js
```diff
@@ -905,6 +946,22 @@ function View(_api, _model) {
...
 
+                if (mobileFloatIntoPlace) {
+                    // Creates a dynamic animation where the top of the current player
+                    // Smoothly transitions into the expected floating space in the event
+                    // we can't start floating at 62px
+                    style(_wrapperElement, {
+                        transform: `translateY(-${FLOATING_TOP_OFFSET - playerBounds.top}px)`
+                    });
+
+                    setTimeout(() => {
```
jnatalzia: i know this many style calls is usually a no-no but given we are only `transform`ing the element, this will only impact `composite` step of the browser rendering. it's by far the cheapest as there is no heavy lifting, simply a movement of the painted layer

src/js/view/view.js
```diff
@@ -945,30 +1002,50 @@ function View(_api, _model) {
...
+            if (mobileFloatIntoPlace) {
+                // Reverses a dynamic animation where the top of the current player
+                // Smoothly transitions into the expected static space in the event
+                // we didn't start floating at 62px
+                style(_wrapperElement, {
+                    transform: `translateY(-${FLOATING_TOP_OFFSET - playerBounds.top}px)`,
+                    'transition-timing-function': 'ease-out'
+                });
+
+                setTimeout(resetFloatingStyles, 150);
```
jnatalzia: also know that there is hesitance around using `setTimeout` in general let alone for tracking the end of a floating animation however this approach allows us to dynamically ensure the player floats to the proper location.

given the fact that it's only `150ms` as well, there is little room for traditional concern around how the web threads lock up when exiting a tab or how it can get held up by other browser execution.

if anyone has an additional idea, more than happy to talk about it! I thought about using `requestAnimationFrame` but i would have to track deltas until we hit `150ms` which felt like overkill but i could be convinced otherwise

src/js/view/view.js
robwalch: 1. A player instance can be removed (or not added) to the DOM without being destroyed (See `_model.set('inDom', inDOM);` in view.js).
2. One that doesn't cause recursion 😱 `updateBounds` is where we actually set `playerBounds`, but `checkResized` or `updateStyles` is where we usually want to respond to those changes. Using `throttledFloatScrollHandler` would be good so we're not over doing it 😌

src/js/view/view.js
robwalch: Maybe use `transitionend` ? https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/transitionend_event

jnatalzia: :+1: great idea. works like a charm, but need a way of resetting styles for browsers that don't support that event, or could use 
```
webkitTransitionEnd otransitionend oTransitionEnd msTransitionEnd transitionend
```
but i'm not in love with that. let's chat monday!

src/js/view/view.js
jnatalzia: Let's chat monday, not totally clear on the second - are you suggesting adding a call to the throttled handler to `updateBounds`? just wanna be sure i'm clear before i implement!

robwalch: Consider:
```js
    function _responsiveUpdate() {
        if (!_this.isSetup) {
            return;
        }
        _this.updateBounds();
        _this.updateStyles();
        _this.checkResized();
    }
```
`playerBounds` is changed in `updateBounds`, classes and styles are then modified in `updateStyles` and events are dispatched in `checkResized`. When would be the best time to run `throttledFloatScrollHandler` in a responsive update or `resize` call?

src/js/view/view.js
robwalch: Don't forget `moz` 🦊 Maybe we can make a prefix util and also use it in css.js `transform()` and request-fullscreen-helper.js?

src/js/view/view.js
jnatalzia: Given bounds updates are really what we want to respond to and we are only setting/removing styles in the case where we _are_ floating it seems like the right call to do it in `updateStyles` so we can update all the styles in one function call. going to put it there, let me know what you think

src/js/utils/dom.js
```diff
@@ -224,3 +224,7 @@ export function openLink(link, target, additionalOptions = {}) {
         a.click();
     }
 }
+
+export function browserIsLandscape() {
```
jnatalzia: does this exist elsewhere?

src/js/view/utils/views-manager.js
```diff
@@ -91,6 +100,18 @@ export default {
     remove: function(view) {
         removeFromGroup(view, views);
     },
+    addScrollHandler: function(handler) {
+        if (!scrollHandlerInitialized) {
+            scrollHandlerInitialized = true;
+            window.addEventListener('scroll', onScroll);
+        }
+        if (scrollHandlers.indexOf(handler) === -1) {
```
jnatalzia: this is a safeguard for any faulty logic, maybe not needed?

src/js/view/view.js
```diff
@@ -340,9 +350,59 @@ function View(_api, _model) {
...
+
+        if (!floating && hasCrossedThreshold) {
+            _updateFloating(0, enoughRoomForFloat);
+        } else if (floating && !hasCrossedThreshold) {
+            _updateFloating(1, enoughRoomForFloat);
+        }
+    }
+
+    function throttledMobileFloatScrollHandler() {
+        if (!fosMobileBehavior()) {
```
jnatalzia: just like how the below `updateFloating` is responsible for checking whether or not we should follow this floating behavior, i used that same pattern here to deal with the mobile portrait vs. landscape issue (landscape acts like desktop but up to this point it's all been handled in css)

src/js/view/view.js
```diff
@@ -340,9 +350,59 @@ function View(_api, _model) {
         playerViewModel.change('playlistItem', onPlaylistItem);
         // Triggering 'resize' resulting in player 'ready'
         _lastWidth = _lastHeight = null;
+
+        // Setup floating scroll handler
+        if (_floatingConfig && OS.mobile) {
+            _model.change('inDom', (model, val) => {
```
jnatalzia: instead of trying to catch every place where we're dealing with `inDOM` here, I used a catch-all handler in setup, that way the floating view doesn't care where in dom is changed, it can respond regardless

src/js/view/view.js
```diff
@@ -149,6 +154,11 @@ function View(_api, _model) {
 
         _resizeMedia(containerWidth, containerHeight);
         _captionsRenderer.resize();
+
+
+        if (_floatingConfig) {
```
jnatalzia: now that the throttled handler checks for mobile and orientation, no need to do it here

src/js/utils/dom.js
robwalch: I wouldn't use `innerWidth` for this.

robwalch: Just in views-manager `onOrientationChange`. It's actually checked differently on older iOS devices using `window.orientation` we have a check in webvr polyfill http://www.williammalone.com/articles/html5-javascript-ios-orientation/

Android
```js
const orientation = window.screen.orientation.type;
const isLandscape = orientation === 'landscape-primary' || orientation === 'landscape-secondary';
```
iOS
```js
const isLandscape = window.orientation == 90 || window.orientation == -90;
``

File: src/js/utils/dom.js [hunk 21]
```diff
@@ -1,5 +1,6 @@
 import { trim } from 'utils/strings';
 import { isString, contains, difference, isBoolean } from 'utils/underscore';
+import { getPropertyAtPath } from 'utils/parser';
```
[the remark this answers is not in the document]
jnatalzia: great find!

src/css/jwplayer/flags/floatingplayer.less
```diff
@@ -25,7 +25,7 @@
 
         .jw-flag-touch& {
             @media screen and (max-device-width : 480px) and (orientation: portrait) {
-                animation: jw-float-to-top 150ms cubic-bezier(0, 0.25, 0.25, 1) forwards 1;
+                animation: none;
```
robwalch: Is `animation: none` necessary? What animation are you overriding or cancelling here?

src/js/view/utils/views-manager.js
robwalch: I prefer to always remove listeners before adding to avoid duplicates. You could add the `if (scrollHandlers.indexOf(handler) === -1) {` check below in `removeScrollHandler`.

src/js/view/view.js
robwalch: Isn't the only thing you need to do is exit early in `throttledMobileFloatScrollHandler` if not inDOM?

src/js/view/view.js
```diff
@@ -340,9 +350,59 @@ function View(_api, _model) {
         playerViewModel.change('playlistItem', onPlaylistItem);
         // Triggering 'resize' resulting in player 'ready'
         _lastWidth = _lastHeight = null;
+
+        // Setup floating scroll handler
+        if (_floatingConfig && OS.mobile) {
+            _model.change('inDom', (model, val) => {
+                if (val) {
+                    viewsManager.addScrollHandler(throttledMobileFloatScrollHandler);
```
robwalch: So the player just got added to the DOM, but nothing happens until a scroll event...

src/js/view/view.js
```diff
@@ -905,6 +965,22 @@ function View(_api, _model) {
 
                 addClass(_playerElement, 'jw-flag-floating');
 
+                if (mobileFloatIntoPlace) {
+                    // Creates a dynamic animation where the top of the current player
+                    // Smoothly transitions into the expected floating space in the event
+                    // we can't start floating at 62px
+                    style(_wrapperElement, {
+                        transform: `translateY(-${FLOATING_TOP_OFFSET - playerBounds.top}px)`
```
robwalch: Should this include `transition: null` or `'none'` to clear transitions previously added?

src/css/jwplayer/flags/floatingplayer.less
jnatalzia: https://github.com/jwplayer/jwplayer/blob/d45ac57c42f68649de32de20892913b68007676f/src/css/jwplayer/flags/floatingplayer.less#L11

src/js/view/view.js
jnatalzia: those get cleared by the `stopFloating` which should in all cases be called before we hit tthis execution path again

src/js/view/view.js
jnatalzia: ahhh good call, will update this and fix the early exist in the scroll handler

PR #3478, opened 2019-09-09
Utilize internal audio/video determination in html5 provider

jnatalzia:
### This PR will...
Fix an issue where audio only mp4s were displaying a big black box instead of the poster image

We should certainly display the poster image for any audio only case - also important for tracking to ensure we have the media type correct

None I can think of, events test page looks good (single `mediaType` evt)

Nah

#### Addresses Issue(s):

JW8-10172

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -524,7 +524,7 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
             });
         }
         if (_levels.length && _levels[0].type !== 'hls') {
-            _this.sendMediaType(_levels);
+            _setMediaType();
         }
     }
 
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -524,7 +524,7 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
             });
         }
         if (_levels.length && _levels[0].type !== 'hls') {
-            _this.sendMediaType(_levels);
```
jnatalzia: this outsourced `mediaType` determination to the default provider, this change makes it the responsibility of the html5 provider andd removes a gating that required `hls` for the html5 provider to make the call

src/js/providers/html5.js
```diff
@@ -629,7 +629,7 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
         _this.eventsOn_();
         // the loadeddata event determines the mediaType for HLS sources
         if (_levels.length && _levels[0].type !== 'hls') {
-            this.sendMediaType(_levels);
+            _setMediaType();
```
robwalch: There's no point it calling `_setMediaType()` here. We called `this.sendMediaType(_levels);` if we had levels. `init` is called before the video is loaded so we should not call methods that require the video tag to have loaded anything.

robwalch: The drawback here is for a setup with an audio file. Now we can't start loading the poster until the audio file is loaded.

Issue #3440, opened 2019-06-20
e.getTime is not a function


esetnik:
### Expected Behavior
iOS v9.3.5 should not have any playback errors

### Actual Behavior
On Mobile Safari v9.3.5 there is an error when the player is initialized
```
TypeError:
e.getTime is not a function. (In 'e.getTime()', 'e.getTime' is undefined)
```
```
{snip} ion(){if(R.getStartDate){var e=R.getStartDate(),t=e.getTime();if(t!==w.startDateTime&&!isNaN(t)){w.startDateTime=t;var i=e.toISOString(),n=w {snip}
```

I traced the error to https://github.com/jwplayer/jwplayer/blob/a18716723f59f5c98c16263e4c2ac65efb95d4ff/src/js/providers/html5.js#L344 but I'm not sure why `getStartDate()` is not returning a `Date` instance. As per the Safari documentation https://developer.apple.com/documentation/webkitjs/htmlmediaelement/1634352-getstartdate is implemented as of Mobile Safari 9.0+.

I did a bit of searching and found some issues which might be related:
https://github.com/videojs/video.js/issues/4311

### Steps to reproduce

1. Use any sample code on iOS Safari v9.3.5 and begin playback

### Environment

- *iPad*
- *iOS v9.3.5*
- *Mobile Safari*
- *JWPlayer v8.9.0 | /player/v/8.9.0/jwplayer.core.controls.polyfills.html5.js*

waxidiotic:
Hi @esetnik - 

JW Player 8 is officially supported on iOS 11, or higher. Please see our [browser support article|https://support.jwplayer.com/articles/browser-support] for more information.

With that being said, I will take a brief look at this to see if I can find something out. I'll let you know.

Thank you!

waxidiotic:
I was able to reproduce this behavior, which seems to stem from calling `.getTime()` on a value of `NaN`. However, it does not reproduce in iOS 11 or 12, nor does it seem to affect performance or usability when reproducible in iOS 9.

esetnik:
Yes I didn't notice any performance or usability issue, but this generates a lot of spurious errors in our error reporting system. I think it should be sufficient to just check for isNan before attempting to call `.getTime()`. This will get rid of the error. I'm happy to submit a PR if you like.

waxidiotic:
Sure, you may submit a PR and we’ll review it. Thanks!

PR #3489, opened 2019-09-26
Prevent .getTime() from being called on NaN value

talkshoplivedev:
### This PR will...
Prevent `e.getTime is not a function` error from being generated for < iOS 11 when calling `.getTime()` on a NaN value

While this issue seemingly doesn't affect performance or usability, the purpose of this PR is to prevent generating spurious errors in an effort to reduce logging in error reporting systems.

No

Not that we know of

#### Addresses Issue(s):
Fixes: https://github.com/jwplayer/jwplayer/issues/3440

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -340,7 +340,7 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
     function checkStartDateTime() {
         if (_videotag.getStartDate) {
             const startDate = _videotag.getStartDate();
-            const startDateTime = startDate.getTime();
+            const startDateTime = !isNaN(startDate) ? startDate.getTime() : NaN;
             if (startDateTime !== _this.startDateTime && !isNaN(startDateTime)) {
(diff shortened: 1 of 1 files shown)


src/js/providers/html5.js
```diff
@@ -340,7 +340,7 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
     function checkStartDateTime() {
         if (_videotag.getStartDate) {
             const startDate = _videotag.getStartDate();
-            const startDateTime = startDate.getTime();
+            const startDateTime = !isNaN(startDate) ? startDate.getTime() : NaN;
```
karimMourra: instead of checking for `isNaN` how about checking if `getTime` is defined ? 
`const startDateTime = startDate.getTime  ? startDate.getTime() : NaN;`

talkshoplivedev: Much better. Will update.

talkshoplivedev:
@robwalch is there an estimation on when this might be merged?

PR #3491, opened 2019-09-30
[JW8-10550] Always show float close button when floating and dismissible is true

pajong:
### This PR will...
- Make float close button always visible even when the controls are hidden from inactivity
- If the controls are set to false by the api `jwplayer().setControls(false)`, the close button is also hidden

- AdExchange requirements

- For TE: please test on different scenarios
-- Outstream, when `dismissible` for outstream is set to `true`/ `false`
-- Floating when `dismissible` is set to false, the icon should not appear

#### Addresses Issue(s):
JW8-10550

### Checklist

### DIFF
--- a/src/css/jwplayer/flags/floatingplayer.less
+++ b/src/css/jwplayer/flags/floatingplayer.less
@@ -61,6 +61,7 @@
         .jw-logo {
             display: none;
         }
+
         .jw-float-icon {
             display: flex;
         }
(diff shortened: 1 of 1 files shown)


pajong:
@waxidiotic Checked with Alicia, tt should be hidden if controls is explicitly set to false, as written in the PR description.

waxidiotic:
> @waxidiotic Checked with Alicia, tt should be hidden if controls is explicitly set to false, as written in the PR description.

Oh - I read it wrong.

PR #3494, opened 2019-10-01
Focus the player container on play btn click

jnatalzia:
### This PR will...
Add a focus call which ensures the player container remains focused as the idle state ui disappears after a play button click

The player did not remain focused after clicking on the play button in idle state

Looking at how [the display background](https://github.com/jwplayer/jwplayer/blob/master/src/js/view/view.js#L449-L461) accomplishes this, there is no difference other than the `userActive` call present when clicking the play button. I tried to remove this to no avail so landed on a new `focus()` call. Any other ideas welcome!

Nah

#### Addresses Issue(s):

JW8-10349

### Checklist

### DIFF
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -108,6 +108,7 @@ export default class Controls extends Events {
                 this.trigger(DISPLAY_CLICK);
                 this.userActive(1000);
                 api.playToggle(reasonInteraction());
+                this.playerContainer.focus();
             });
 
             this.div.appendChild(displayContainer.element());


src/js/view/controls/controls.js
```diff
@@ -108,6 +108,7 @@ export default class Controls extends Events {
                 this.trigger(DISPLAY_CLICK);
                 this.userActive(1000);
                 api.playToggle(reasonInteraction());
+                this.playerContainer.focus();
```
robwalch: What if we removed this block completely?

jnatalzia:
@robwalch updated to remove the click handlers and defer to the background. despite looking into it i can't seem to determine why this button in particular is causing the focus to change, the rewind/next buttons that occupy the same div do not have this issue (tested this change with them, all is well)

i can continue to dig if you think it's worthwhile but given this removes some event handlers _and_ solves the problem i'm comfortable with this fix

jnatalzia:
test this please

jnatalzia:
@robwalch no luck figuring out root cause. combed through the performance panel/stack trace and didn't see any smoking gun. noticed that on mobile it stayed focused until the `userInactive` timeout however those same calls are not made on desktop for the same flow.

i don't think it's worth spending more time on. let me know what you think!

PR #3497, opened 2019-10-02
[JW8-10218] Fix iOS float close button interacting with element below it

pajong:
### This PR will...
- Use `addEventListener` function rather than using `UI` class to handle the click

- On iOS Safari, there is an intermittent timing issue where the `click` event applies to the element below the player as the player is removed from the position when the float close icon is being clicked
- `UI` tries to see if there is any dragging activity when `click` is detected, and we suspect that is causing the `click` to be applied to the element below after the floating player is removed

- I could not really pin down at where the timing issue is happening. I came up with the solution after creating a test page with two divs one on top of the other, with `addEventListener` - it never had any issues with the event being applied to the bottom div.
- If there is a better solution, please let me know, but I still think this is good because we do not have to register `UI` for simple click handlers that just triggers an event

#### Addresses Issue(s):
JW8-10218

### Checklist

### DIFF
--- a/src/js/view/floating-close-button.js
+++ b/src/js/view/floating-close-button.js
@@ -1,4 +1,5 @@
 import { createElement } from 'utils/dom';
+import UI from 'utils/ui';
 import floatingCloseButton from 'templates/floating-close-button';
 import { cloneIcon } from 'view/controls/icons';
 import Events from 'utils/backbone.events';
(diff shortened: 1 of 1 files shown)


src/js/view/floating-close-button.js
```diff
@@ -11,16 +10,24 @@ export default class FloatingCloseButton extends Events {
         this.element = createElement(floatingCloseButton(ariaLabel));
 
         this.element.appendChild(cloneIcon('close'));
-        this.ui = new UI(this.element).on('click tap enter', () => {
```
zetagame: Removing this will cause an error on line 149 of controls.js, as `floatCloseButton.on` will be undefined.

`        if (floatingConfig) {
            const floatCloseButton = new FloatingCloseButton(element, model.get('localization').close);
            const doNotForward = true;
            floatCloseButton.on(USER_ACTION, () => this.trigger('dismissFloating', { doNotForward }));

            if (floatingConfig.dismissible !== false) {
                addClass(this.playerContainer, 'jw-floating-dismissible');
            }
        }`

src/js/view/floating-close-button.js
```diff
@@ -11,16 +10,24 @@ export default class FloatingCloseButton extends Events {
...
-            this.trigger(USER_ACTION);
-        });
+
+        this.interactionHandler = this.userActionHandler.bind(this);
+        this.element.addEventListener('click', this.interactionHandler);
```
zetagame: take a look at 'directSelect' option in ui.js

pajong:
test this please

PR #3498, opened 2019-10-03
fix(): updates Dutch translations to more accurate representation of …

joranquinten:
…the language

### This PR will 
Apply a more accurate translation for Dutch

The current translation uses English capitalisation rules which are not native to the Dutch language.

### Checklist

### DIFF
--- a/src/assets/translations/nl.json
+++ b/src/assets/translations/nl.json
@@ -25,7 +25,7 @@
         "protectedContent": "Deze beschermde video kan niet worden afgespeeld.",
         "technicalError": "Deze video kan niet worden afgespeeld door een technisch probleem."
     },
-    "exitFullscreen": "Volledig Scherm verlaten",
+    "exitFullscreen": "Volledig scherm verlaten",
     "fullscreen": "Fullscreen",
     "hd": "Kwaliteit",
     "liveBroadcast": "Live",
@@ -39,7 +39,7 @@
     "play": "Start",
     "playback": "Start",
     "playbackRates": "Snelheid",
-    "player": "Video speler",
+    "player": "Videospeler",
     "poweredBy": "Ondersteund door",
     "prev": "Vorige",
     "related": {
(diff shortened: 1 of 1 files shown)


src/assets/translations/nl.json
```diff
@@ -25,29 +25,29 @@
...
-    "nextUp": "Volgende Video",
+    "nextUp": "Volgende video",
     "notLive": "Niet live",
     "off": "Uit",
     "pause": "Pause",
     "play": "Start",
     "playback": "Start",
     "playbackRates": "Snelheid",
-    "player": "Video Speler",
+    "player": "Video speler",
```
stragulus: This should be a single compound word, 'videospeler'

src/assets/translations/nl.json
```diff
@@ -25,29 +25,29 @@
         "protectedContent": "Deze beschermde video kan niet worden afgespeeld.",
         "technicalError": "Deze video kan niet worden afgespeeld door een technisch probleem."
     },
-    "exitFullscreen": "Volledig Scherm Verlaten",
+    "exitFullscreen": "Volledig Scherm verlaten",
```
stragulus: I would not capitalize Scherm here, just like in another example where "Volledig scherm" is used further below and is not capitalized. I don't think this is a strict rule, but it's definitely more consistent.

joranquinten: I kinda chose it because of the meaning of 'Volledig Scherm' being a mode on the JW player, but will change it.

src/assets/translations/nl.json
joranquinten: Agreed

robwalch:
@stragulus Please rebase / fix conflicts

aliciahurst:
Hi @joranquinten, thank you for your submission. This isn't documented, but we require translation submissions to originate in CSV format rather than JSON. We'll work on adding a template for that going forward. In the meantime, I've already converted your submission via copy and paste. We will review your translations internally and include any changes in an upcoming release. Thank you!

PR #3504, opened 2019-10-11
Focus the right element based on floating state

jnatalzia:
### This PR will...
Fix an issue where floating player focus caused the page to jump

Focus updates to the player broke things

Is this the best way to do this? I'm reticent to dive further into this like I did last week.

Nope

#### Addresses Issue(s):

JW8-10624

### Checklist

### DIFF
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -110,7 +110,7 @@ export default class Controls extends Events {
                 this.trigger(DISPLAY_CLICK);
                 this.userActive(1000);
                 api.playToggle(reasonInteraction());
-                this.playerContainer.focus();
+                model.get('isFloating') ? this.wrapperElement.focus() : this.playerContainer.focus();
             });
 
             this.div.appendChild(displayContainer.element());


src/js/view/controls/controls.js
```diff
@@ -110,7 +110,7 @@ export default class Controls extends Events {
                 this.trigger(DISPLAY_CLICK);
                 this.userActive(1000);
                 api.playToggle(reasonInteraction());
-                this.playerContainer.focus();
+                model.get('isFloating') ? this.wrapperElement.focus() : this.playerContainer.focus();
```
robwalch: Since the wrapper is always in the player, couldn't we just always focus the wrapper?

jnatalzia:
@robwalch yeah would def prefer that - the focus doesn't work as expected when on the wrapper element though.

looking into it, i tried a bunch of code variations to attempt to get the wrapper focus to work without floating. what _did_ appear to work was changing the position attribute of the wrapper from `absolute` to `fixed`. i'm assuming this is also why focusing the wrapper when it is active works in that it is no longer absolutely positioned. unfortunately that means in the current setup we'll need the ternary

i have never heard of absolute elements being unfocusable but it wouldn't surprise me. any other ideas from your end?

PR #3509, opened 2019-10-23
[JW8-10646] Implement platform detection for iPadOS 13

zetagame:
### This PR will...
Add detection for iPadOS 13

iPadOS 13 uses a desktop user agent by default.

#### Addresses Issue(s):

JW8-10646

### Checklist

### DIFF
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -19,9 +19,7 @@ export const isFF = lazyUserAgentMatch(/gecko\//i);
 export const isIETrident = lazyUserAgentMatch(/trident\/.+rv:\s*11/i);
 export const isIPod = lazyUserAgentMatch(/iP(hone|od)/i);
 // Second check is for iPadOS 13, which uses desktop UA
-export const isIPad = () => function () {
-    return (userAgentMatch(/iPad/i) || (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1));
-};
+export const isIPad = () => userAgentMatch(/iPad/i) || (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1);
 export const isOSX = lazyUserAgentMatch(/Macintosh/i);
 // Check for Facebook App Version to see if it's Facebook
 export const isFacebook = lazyUserAgentMatch(/FBAV/i);
--- a/src/js/utils/browser.js
+++ b/src/js/utils/browser.js
@@ -19,7 +19,9 @@ export const isFF = lazyUserAgentMatch(/gecko\//i);
 export const isIETrident = lazyUserAgentMatch(/trident\/.+rv:\s*11/i);
 export const isIPod = lazyUserAgentMatch(/iP(hone|od)/i);
 // Second check is for iPadOS 13, which uses desktop UA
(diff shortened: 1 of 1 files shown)


src/js/utils/browser.js
```diff
@@ -18,7 +18,8 @@ export function isFlashSupported() {
 export const isFF = lazyUserAgentMatch(/gecko\//i);
 export const isIETrident = lazyUserAgentMatch(/trident\/.+rv:\s*11/i);
 export const isIPod = lazyUserAgentMatch(/iP(hone|od)/i);
-export const isIPad = lazyUserAgentMatch(/iPad/i);
+// Second check is for iPadOS 13, which uses desktop UA
+export const isIPad = lazyUserAgentMatch(/iPad/i) || (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1);
```
waxidiotic: I'm assuming the `lazyUserAgentMatch(/iPad/i)` will match for iPads running iOS 12 and earlier, right? 

And are we sure iPhones running iOS 13 don't return `navigator.platform === 'MacIntel'`? I'm pretty sure it won't but I can see an exhaustive list of possible values anywhere.

robwalch: We'll get more `isIPad = true`.

Will `isOSX` be true too?

robwalch: `isIPad` is a function

zetagame: > I'm assuming the `lazyUserAgentMatch(/iPad/i)` will match for iPads running iOS 12 and earlier, right?
> 
> And are we sure iPhones running iOS 13 don't return `navigator.platform === 'MacIntel'`? I'm pretty sure it won't but I can see an exhaustive list of possible values anywhere.

navigator.platform is iPhone on iOS 13.

zetagame: > `isIPad` is a function

I don't understand. It was a constant with a boolean value before, and it still is. Do you mean it should now be a function?

zetagame: > `isIPad` is a function. What you want to do is:
> 
> `export const isIPad = () => lazyUserAgentMatch(/iPad/i) || (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1);`
> 
> and as mentioned above, if possible, confirm this is not OSX/macos.

Good catch on the function return. I believe navigator.maxTouchPoints > 1 ensures this won't return true on OSX, but I can add an additional check.

zetagame: > `isIPad` is a function. What you want to do is:
> 
> `export const isIPad = () => lazyUserAgentMatch(/iPad/i) || (navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1);`
> 
> and as mentioned above, if possible, confirm this is not OSX/macos.

If we're making isIpad an arrow function, I believe we should use userAgentMatch instead of lazyUserAgentMatch, since it should be a function that returns a boolean, not a function that returns a function. I've made an additional update to reflect this.

robwalch: Yes. My example wouldn't work anyway cause it needs to actually execute the match function `lazyUserAgentMatch(/iPad/i)()`

robwalch: `navigator.maxTouchPoints > 1`, but will our osx check evaluate to true on iPad?

zetagame:
test this please

PR #3512, opened 2019-10-28
[JW8-10667] SanitizeElementAttributes protects against href XSS injection

zetagame:
### This PR will...
Add detection for inline javascript in hrefs.

To prevent execution of arbitrary JS in config properties like "abouttext".

#### Addresses Issue(s):

JW8-10667

### Checklist

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -65,6 +65,12 @@ export function sanitizeElementAttributes(element) {
         if (/^on/.test(name)) {
             element.removeAttribute(name);
         }
+        if (/href/.test(name)) {
+            const link = attributes[i].value;
+            if (/javascript:|javascript&colon;/.test(link)) {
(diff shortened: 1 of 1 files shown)


src/js/utils/dom.js
```diff
@@ -65,6 +65,12 @@ export function sanitizeElementAttributes(element) {
         if (/^on/.test(name)) {
             element.removeAttribute(name);
         }
+        if (/href/.test(name)) {
+            const link = attributes[i].value;
+            if (/javascript:|javascript&colon;/.test(link)) {
```
jnatalzia: this def fixes it and makes sense. are there any other ways of gettting js into a tag we could patch while we're in here? also why `javascript:|javascript&colon;` trying to save people with `javascript` in their urls i'd guess?

zetagame: Yep!

PR #3520, opened 2019-11-05
Cleanup cast button creation

robwalch:
### This PR will...
Remove duplicate code in controlbar.js by using button.js for cast button

Code from button.js was copy-pasted into controlbar.js. Reusing `button` here is easy. It removes 23 lines and a module import in controlbar.

I noticed this while pulling up the two modules side-by-side while discussing button patterns with @zetagame.

Zack, since you're looking at making changes in this area, I thought you could take this on.

### Checklist

### DIFF
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -6,7 +6,6 @@ import TimeSlider from 'view/controls/components/timeslider';
 import VolumeTooltipIcon from 'view/controls/components/volumetooltipicon';
 import button from 'view/controls/components/button';
 import { SimpleTooltip } from 'view/controls/components/simple-tooltip';
-import ariaLabel from 'utils/aria';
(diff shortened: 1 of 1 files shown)


src/js/view/controls/controlbar.js
```diff
@@ -49,45 +47,24 @@ function createCastButton(castToggle, localization) {
         SimpleTooltip(airplayButton.element(), 'airplay', localization.airplay);
 
         return airplayButton;
+    } else if (Browser.chrome || OS.iOS) {
```
jnatalzia:
```
    if (!Browser.chrome || OS.iOS) {
        return;
    }
```
was the previous code, should this be `&& !OS.iOS`

robwalch: Thanks! Changed it to use the same check we use to webpack load the Chromecast controller in jwplayer-commercial.

PR #3523, opened 2019-11-07
VTT Caption Positioning Fixes

jnatalzia:
### This PR will...
Fix positioning in the following ways:
* Ensure caption width only extends as far as the words, not the full caption container
* Fix the position finding algorithm such that it correct calculates the best place for the caption (previously was selecting the worst)
* Fix the position finding algorithm by ensuring it combines both horizontal and vertical axis to position in the true best position

Using `line` attribute screws with all the above issues and surfaced quite a few bugs in our caption rendering

All of it - this took a lot of research and trial and error to get right. want to be sure i'm not missing something

Nope

#### Addresses Issue(s):

JW8-10682

### Checklist

### DIFF
--- a/src/js/polyfills/webvtt.js
+++ b/src/js/polyfills/webvtt.js
@@ -674,23 +674,6 @@ BoxPosition.getSimpleBoxPosition = function (obj) {
     return ret;
 };
 
-// Helper function for visualizing box position
-// Helpful when debugging
-function drawBox(b, container, color) { // eslint-disable-line no-unused-vars
-    let pos = b.toCSSCompatValues(container);
-    let el = document.createElement('div');
-    el.style.top = pos.top + 'px';
-    el.style.bottom = pos.bottom + 'px';
-    el.style.left = pos.left + 'px';
-    el.style.right = pos.right + 'px';
-    el.style.width = pos.width + 'px';
-    el.style.height = pos.height + 'px';
-    el.style.border = '2px solid ' + color;
-    el.style.position = 'absolute';
-    document.querySelector('.jw-text-track-container').appendChild(el);
-}
-
-
 // Move a StyleBox to its specified, or next best, position. The containerBox
 // is the box that contains the StyleBox, such as a div. boxPositions are
 // a list of other boxes that the styleBox can't overlap with.
@@ -740,7 +723,6 @@ function moveBoxToLinePosition(window, styleBox, containerBox, boxPositions, num
             let otherAxis = bestAxis.indexOf('y') === -1 ? ['-y', '+y'] : ['-x', '+x'];
             return findBestPosition(finalPos, otherAxis, posFindRecursiveCount+1);
         }
-
         return finalPos;
     }
     let boxPosition = new BoxPosition(styleBox);
--- a/src/js/polyfills/webvtt.js
+++ b/src/js/polyfills/webvtt.js
@@ -458,15 +458,13 @@ function CueStyleBox(window, cue) {
     switch (cue.align) {
         case 'start':
         case 'left':
+        case 'end':
+        case 'right':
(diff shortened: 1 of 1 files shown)


src/js/polyfills/webvtt.js
```diff
@@ -696,16 +715,26 @@ function moveBoxToLinePosition(window, styleBox, containerBox, boxPositions, num
...
             // Reset the box position to the specified position.
             b = new BoxPosition(specifiedPosition);
         }
-        return bestPosition || specifiedPosition;
-    }
 
+        let finalPos = bestPosition || specifiedPosition;
+        // If we detect a best position based on one axis, oftentimes combining with the other axis
+        // results in a much better position.
+        // TODO: Make more performant by checking axis permutations above
```
jnatalzia: there is certainly a way to refactor this to be O(n) however the complexity mixed with the amount of time already sunk into this make me believe this option is the best way forward until we want to spend more time on this rendering

src/js/polyfills/webvtt.js
```diff
@@ -655,17 +652,37 @@ BoxPosition.getSimpleBoxPosition = function (obj) {
...
+        height: trueHeight,
+        bottom: obj.bottom || (top + trueHeight),
         width: obj.width || width
     };
     return ret;
 };
 
+// Helper function for visualizing box position
+// Helpful when debugging
+function drawBox(b, container, color) { // eslint-disable-line no-unused-vars
```
jnatalzia: this is not used by the code so would get shipped with the player however the utility of this function in debugging these captions box positions is worth the extra couple kb of delivery imo

src/js/polyfills/webvtt.js
```diff
@@ -457,15 +458,13 @@ function CueStyleBox(window, cue) {
...
             textPos = cue.position;
             break;
         case 'middle':
         case 'center':
-            textPos = (cue.position === 'auto' ? 50 : cue.position) - (cue.size / 2);
-            break;
-        case 'end':
-        case 'right':
-            textPos = cue.position - cue.size;
+            textPos = (cue.position === 'auto' ? 0 : cue.position);
```
jnatalzia: this assumes that `align:end` with `position:0%` expects the text to be right aligned but present at the beginning of the captions container, if this is not correct - can update

jnatalzia: just checked the vtt spec - looks like this needs to be reverted

src/js/polyfills/webvtt.js
```diff
@@ -454,19 +455,22 @@ function CueStyleBox(window, cue) {
     // https://w3c.github.io/webvtt/#ref-for-enumdef-positionalignsetting-1
     // The cue.align property is settable and other browsers use it as the offset from which the cue.position
     // value is applied.
+    let transform = '';
```
jnatalzia: transform ensures that we respect the cue alignment when determining position. now that these are inline-block elements the size is not known at time of instantiation. this new code adds a transform to properly place the cue based on alignment and then removes it when the stylebox is finally moved to prevent it from being applied twice (once during exact px calculations and then after when we go to display)

src/js/polyfills/webvtt.js
robwalch: How do you execute it? Why does it need to be part of the player/this code, vs a snippet in your browser used for testing?

jnatalzia: fair questions - you would have to call it yourself below. i suppose it's more confusing than anything leaving it in - i just found it extremely helpful. might add as a window global

jnatalzia:
test this please

jnatalzia:
test this please

robwalch:
> test this please

Click ❗️to go to the job and then "Rebuild"

PR #3528, opened 2019-11-13
Fix keyboard check to ensure working toggle on key press

jnatalzia:
### This PR will...
Update our keyboard press check to expect a class not a full class list

I broke the keyboard toggle for keyboard shortcuts

Nah

Nah
#### Addresses Issue(s):

JW8-10720

### Checklist

### DIFF
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -207,9 +207,9 @@ export default class Controls extends Events {
                 if (isKeyEvent) {
                     settingsButton.element().focus();
                 } else if (evt) {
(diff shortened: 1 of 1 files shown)


src/js/view/controls/controls.js
```diff
@@ -317,7 +317,7 @@ export default class Controls extends Events {
                     break;
                 case 13: // enter
                 case 32: // space
-                    if (document.activeElement.className === 'jw-switch' && evt.keyCode === 32) {
+                    if (document.activeElement.classList.contains('jw-switch') && evt.keyCode === 32) {
```
zetagame: evt.keycode is always 32 here, we can remove the extra check. Looks good after that!

PR #3530, opened 2019-11-13
[JW8-10890] Tizen Browser Support (2016, 2018, 2019)

robwalch:
### This PR will...
Address issues that prevent the player from setting up and playing through playlists on 2016 Webkit based Tizen Samsung TV browsers. Old Mac owners who use Safari 5-8 can rejoice as these fixes apply to there as well.

- Add partial webkit flexbox fallbacks (partial, some elements need more work)
- Disables background loading for playlist progression to work (Ideally the user agent browser flag would not be necessary, but this is a quick fix for the purpose of exposing the need for better feature detection and broader support for disabling background loading)
- Disable autoplay test when `new Blob` thows
- Uses polyfill for `parser.parseFromString('', 'text/html')`

### Other issues not addressed
- CSS opacity animations/transitions should be disabled completely
- Motion thumbnail video elements show up black before and after playback (and trick play should be based on background loading)
- Playback toggles to paused when pressing the fullscreen button
- Testing on 2017 models
- Testing in an application bundle
- Testing ads (vast/ima/fw)
- Testing shelf (overlay works fine)
- hls.js support not included in this PR

#### Findings on 2018-2019 Samsung TVs (6-7 Series)
- Background loading must also be disabled to prevent DOM video from going black
- HLS streams are supported with `video.src = '//url.m3u8'`
- The hls.js provider functions correctly when our player recognizes the user-agent as Chrome

No.

#### Addresses Issue(s):
[JW8-10890](https://jwplayer.atlassian.net/browse/JW8-10890) / JW8-10701


src/js/providers/html5.js
```diff
@@ -476,9 +476,16 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
...
+                    currentQuality = i;
+                    break;
                 }
             }
+            if (!currentQuality) {
+                // Choose the highest quality
+                const sorted = levels.map((level, index) => Object.assign({ index, height: 0 }, level))
+                    .sort((a, b) => b.height - a.height);
+                currentQuality = sorted[0].index;
+            }
```
robwalch: This block was added for demo purposes on the 2016 device. 2018-19 devices support adaptive hls playback with the hls.js provider (0.12.4 fork).

src/js/environment/environment.js
```diff
@@ -175,7 +180,7 @@ Object.defineProperties(Features, {
         enumerable: true
     },
     backgroundLoading: {
-        get: memoize(() => !(OS.iOS || Browser.safari)),
+        get: memoize(() => !(OS.iOS || Browser.safari || OS.tizen)),
```
robwalch: This change is critical for ads and playlist playback. The TVs do not handle buffering media into another video element in the background while one is still playing.

src/css/controls/imports/controlbar.less
```diff
@@ -82,13 +82,23 @@
...
     width: 100%;
     padding: 0 12px;
+
+    div {
+        float: left;
+        line-height: 21px;
+    }
+
+    .jw-spacer ~ div {
+        float: right;
```
robwalch: The control bar buttons on the right will show up in reverse order in old Webkit browsers missing flex-box support, but that's better than having them jammed up against the buttons on the left.

src/js/environment/environment.js
jnatalzia: nice

src/js/environment/os-version.js
```diff
@@ -31,6 +31,8 @@ export function osVersion(osEnvironment, agent) {
         version = execResult(/OS ([._\d]+)/.exec(agent), 1);
     } else if (osEnvironment.mac) {
         version = execResult(/Mac OS X (10[._\d]+)/.exec(agent), 1);
+    } else if (osEnvironment.tizen) {
```
jnatalzia: if we wanted to add webos, i saw in video-dev it's `Web0S` (yes a zero)

src/js/providers/html5.js
```diff
@@ -476,9 +476,16 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
...
         }
         visualQuality.reason = 'initial choice';
```
jnatalzia: is it safe to fall through and hit this block? previously the reason would never get set if we found the level above

src/js/utils/browser.js
```diff
@@ -32,8 +32,13 @@ export function isMSIE() {
     return userAgentMatch(/msie/i);
 }
 
+export function isTizen() {
+    return userAgentMatch(/SMART-TV/);
```
jnatalzia: this is different than the approach above where we look for `Tizen`, can i get a quick explainer?

src/js/utils/can-autoplay.js
```diff
@@ -29,6 +29,7 @@ const URL = (global => global.URL && global.URL.createObjectURL ?
     global.URL : global.webkitURL || global.mozURL)(window);
 
 let videoBlob;
+const Blob = window.Blob;
```
jnatalzia: i assume this is what the new lint rule is for? is it worth calling it something diff? are there risks with having a var the same name as a window var?

src/js/providers/html5.js
robwalch: Yeah. That was a mistake.

src/js/utils/browser.js
robwalch: "Tizen" is not present in the pre-2017 Smart TV user-agents. `SMART-TV` indicated it is Tizen. The word _Tizen_ is found before the Tizen _version_ in later models.

src/js/utils/can-autoplay.js
robwalch: Nothing wrong with shadowing global names. We should always get globals from the global object (`window` or `self`).

src/css/controls/imports/controlbar.less
```diff
@@ -82,13 +82,23 @@
 }
 
 .jw-button-container {
+    display: block;
```
robwalch: Would be better to use autoprerixer to compile backwards compatible support:

From https://autoprefixer.github.io/
```css
    display: -webkit-box;
    display: flex;
    -webkit-box-orient: horizontal;
    -webkit-box-direction: normal;
            flex-flow: row nowrap;
    -webkit-box-flex: 1;
            flex: 1 1 auto;
    -webkit-box-align: center;
            align-items: center;
    -webkit-box-pack: center;
            justify-content: center;
    width: 100%;
    padding: 0 12px;
```

robwalch:
Test this please

src/css/controls/imports/controlbar.less
jnatalzia: sounds like you'd prefer a manual addition where needed? i tend to agree, autoprefixr could really bloat

src/css/controls/imports/settings-menu.less
```diff
@@ -135,6 +136,10 @@
     .jw-settings-close {
         margin-left: auto;
     }
+
+    div {
```
jnatalzia: lots of generic selectors, might bite us in future but seems safe enough given they are localized in these blocks

src/js/utils/browser.ts
```diff
@@ -15,12 +15,12 @@ export function isFlashSupported(): boolean {
     return !!(version && version >= __FLASH_VERSION__);
 }
 
+const isIPadOS13 = () => navigator.platform === 'MacIntel' && navigator.maxTouchPoints > 1;
```
jnatalzia: Out of curiosity what’s the reason these are functions vs set consts

src/js/utils/dom.js
```diff
@@ -49,6 +49,38 @@ export function htmlToParentElement(html) {
...
+        return parser.parseFromString(html, 'text/html').body;
+    }
+    const doc = document.implementation.createHTMLDocument('');
+    if (html.toLowerCase().indexOf('<!doctype') > -1) {
+        // eslint-disable-next-line no-unsanitized/property
+        doc.documentElement.innerHTML = html;
+    } else {
+        // eslint-disable-next-line no-unsanitized/property
+        doc.body.innerHTML = html;
+    }
```
jnatalzia: Never seen these objects/methods before! this is some _awesome_ polyfilling. 👏👏

robwalch: Yeah. Adapted from the example at https://developer.mozilla.org/en-US/docs/Web/API/DOMParser

PR #3531, opened 2019-11-15
[JW8-10688] Fix blue ring and keyboard shortcut on shortcutsToggle

mamaddox:
### This PR will...

- Make a blue ring is displayed around the shortcutsToggle across all browsers when it is tabbed to.
- Allow the shortcuts to be enabled/disabled with the enter key as opposed to the space bar.

- Update the shortcuts menu to make sure it is functioning as expected

#### Addresses Issue(s):

[JW8-10688](https://jwplayer.atlassian.net/browse/JW8-10688)

### Checklist

### DIFF
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -317,7 +317,7 @@ export default class Controls extends Events {
                     break;
                 case 13: // enter
                 case 32: // space
-                    if (document.activeElement.classList.contains('jw-switch') && evt.keyCode === 32) {
+                    if (document.activeElement.classList.contains('jw-switch') && evt.keyCode === 13) {
                         // Let event bubble up so the spacebar can control the toggle if focused on
                         return true;
                     }
@@ -382,13 +382,20 @@ export default class Controls extends Events {
         this.playerContainer.addEventListener('keydown', handleKeydown);
         this.keydownCallback = handleKeydown;
 
-        // keep controls active when navigating inside the player
         const handleKeyup = (evt) => {
-            const isTab = evt.keyCode === 9;
-            if (isTab) {
-                const insideContainer = this.playerContainer.contains(evt.target);
-                const activeTimeout = insideContainer ? 0 : ACTIVE_TIMEOUT;
-                this.userActive(activeTimeout);
+            switch (evt.keyCode) {
+                case 9: {
+                    // tab, keep controls active when navigating inside the player
+                    const insideContainer = this.playerContainer.contains(evt.target);
+                    const activeTimeout = insideContainer ? 0 : ACTIVE_TIMEOUT;
+                    this.userActive(activeTimeout);
+                    break;
+                }
+                case 32: // space
+                    evt.preventDefault();
+                    break;
+                default:
+                    break;
             }
         };
         this.playerContainer.addEventListener('keyup', handleKeyup);
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -446,7 +446,16 @@ export default class Controls extends Events {
     }
 
     disable(model) {
-        const { nextUpToolTip, settingsMenu, infoOverlay, controlbar, rightClickMenu, playerContainer, div } = this;
+        const { 
+            nextUpToolTip,
+            settingsMenu,
+            infoOverlay,
+            controlbar,
+            rightClickMenu,
+            shortcutsTooltip,
+            playerContainer,
+            div
+        } = this;
 
         clearTimeout(this.activeTimeout);
         this.activeTimeout = -1;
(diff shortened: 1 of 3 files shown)


jnatalzia:
Looks like your style changes failed the linter! https://jenkins.jwplayer.com/view/JW7/job/jw7-opensource/8092/console

```
src/css/controls/imports/switch.less
 19:8  ✖  Unexpected whitespace at end of line   no-eol-whitespace
```

jnatalzia:
test this please

src/js/view/controls/shortcuts-tooltip.js
```diff
@@ -81,7 +82,7 @@ export default function (container, api, model) {
         api.pause(settingsInteraction);
     };
     const close = () => {
-        shortcutToggle.removeEventListener('click', toggleClickHandler);
+        shortcutToggleUi.off('click tap enter', toggleClickHandler);
```
robwalch: You could just call `shortcutToggleUi.off()` here since there are no other listeners on this UI instance.

robwalch: I hope we call `close` on this when destroying the player and removing controls otherwise we'll have rogue document listeners executing code after the player is removed.

Yep:
1. Setup a player
1b. Add a break point in `documentClickHandler` on ln`if (!/jw-shortcuts|jw-switch/.test(e.target.className)) {`
2. Open the keyboard shortcuts menu
3. Run `jwplayer().remove()` in the console
4. Click the document

Result - this code executes:
```
  var documentClickHandler = function documentClickHandler(e) {
    if (!/jw-shortcuts|jw-switch/.test(e.target.className)) {
      close();
    }
  };
```

src/js/view/controls/controls.js
```diff
@@ -495,6 +511,10 @@ export default class Controls extends Events {
             infoOverlay.destroy();
         }
 
+        if (shortcutsTooltip) {
+            shortcutsTooltip.destroy();
```
mamaddox: I noticed that this shortcutsTooltip is also part of the rightClickMenu (rightclick.js) and I was wondering if it is better to destroy it with the rightClickMenu or within controls? Granted controls is where it is created and info-overlay follows the same pattern.

src/js/view/controls/shortcuts-tooltip.js
```diff
@@ -90,28 +92,34 @@ export default function (container, api, model) {
             api.play(settingsInteraction);
         }
     };
+
+    const destroy = () => {
+        close();
+    };
```
mamaddox: Do the model callbacks also need to be removed here? I noticed they are removed in info-overlay, but they are already removed in controls.js when disable is called.

jnatalzia: if there is an execution path where we could be hitting those callback after a destroy then i would say yes. if the info-overlay removal is a superset of the destroy here and will always be the entrypoint for destruction of this component i'd say we're okay not removing them here.

thoughts @robwalch ?

src/js/view/controls/controls.js
jnatalzia: i think destroying it where it is created makes the most sense

src/js/view/controls/shortcuts-tooltip.js
robwalch: I don't any listeners added to `model` in this module. Just calls to `get` and `set`...

In general, yes. If a module adds listeners to the anything, it should remove them when destroyed. So if you use `model.change/on/once` pass a context like `this` and then remove all listeners for that on destroy  `model.off(null, null, this)`

src/js/view/controls/controls.js
zetagame: There's a keyboard shortcut for opening the menu, so it isn't actually dependent on right click.

src/js/view/controls/shortcuts-tooltip.js
```diff
@@ -80,8 +81,9 @@ export default function (container, api, model) {
         isOpen = true;
         api.pause(settingsInteraction);
     };
+
     const close = () => {
-        shortcutToggle.removeEventListener('click', toggleClickHandler);
+        shortcutToggleUi.off();
```
zetagame: Why do we remove the listeners (ui.off) when it closes, and add them again when it opens? If there isn't a specific reason, we could add once during setup (in render method) and destroy when component is destroyed. If git lens says

mamaddox: Good call! No reason I know of.

mamaddox:
test this please

mamaddox:
test this please

src/js/view/controls/shortcuts-tooltip.js
```diff
@@ -90,43 +90,54 @@ export default function (container, api, model) {
             api.play(settingsInteraction);
         }
     };
+
+    const destroy = () => {
+        close();
+        shortcutToggleUi.off();
+        shortcutToggleUi.destroy();
```
robwalch: `UI.destroy` calls `off` so we're calling off twice here.

PR #3534, opened 2019-11-20
[JW8-10684] Refactor Menus for readability and performance. Make nestable. Add caption styles submenu.

zetagame:
//Todo: Update menu tests, add in @mamaddox's element focus changes, and write tickets for bugs fixed while refactoring.

### This PR will...
Refactor menus to include:
- Nestable Structure (submenus can have submenus)
- Remove Submenu Factory
- Make listener code DRYer

Fixes following bugs:
- First option is bold even if unselected.
- Down keypress skips first item.

Additional Features:
- Captions Styling Menu

Sets the groundwork for an upcoming commercial feature, and improves maintainability.

Yes

#### Addresses Issue(s):

JW8-10684
JW8-9041
JW8-9042
JW8-9044

### Checklist


src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+        this.categoryButton = null;
+        if (this.isSubmenu) {
+            if (this.parentMenu.name === 'settings') {
+                this.createCategoryButton(_localization);
+            }
+            this.parentMenu.appendMenu(this);
+        } else {
+            this.ui = addGlobalMenuKeyListener(this);
+        }
+        this.onDocumentClick = this.onDocumentClick.bind(this);
```
jnatalzia: are these binds explicitly needed with class level funcs? it should use the context of the calling obj right?

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+    createCloseButton(localization) {
+        const closeButton = button('jw-settings-close', (e) => this.close(e), localization.close, [cloneIcon('close')]);
+        this.topbar.appendChild(closeButton.element());
+        closeButton.show();
+        closeButton.ui.on('keydown', (e) => {
+            const sourceEvent = e.sourceEvent;
+            const key = sourceEvent.key.replace(/(Arrow|ape)/, '');
+            // Close settings menu when enter is pressed on the close button
+            // or when tab or right arrow key is pressed since it is the last element in topbar
+            if (key === 'Enter' || key === 'Right' || (key === 'Tab' && !sourceEvent.shiftKey)) {
```
jnatalzia: does `tab` not get covered by usual focus/blur functionality?

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+        this.close = this.close.bind(this);
+        this.toggle = this.toggle.bind(this);
+    }
+    createCloseButton(localization) {
+        const closeButton = button('jw-settings-close', (e) => this.close(e), localization.close, [cloneIcon('close')]);
+        this.topbar.appendChild(closeButton.element());
+        closeButton.show();
+        closeButton.ui.on('keydown', (e) => {
+            const sourceEvent = e.sourceEvent;
+            const key = sourceEvent.key.replace(/(Arrow|ape)/, '');
```
jnatalzia: why this over `keyCode`

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+            // or when tab or right arrow key is pressed since it is the last element in topbar
+            if (key === 'Enter' || key === 'Right' || (key === 'Tab' && !sourceEvent.shiftKey)) {
+                this.close(e);
+            }
+        }, this);
+        this.topbar.appendChild(closeButton.element());
+        return closeButton;
+    }
+    createCategoryButton(localization) {
+        const localizationKeys = {
```
jnatalzia: this feels like it's better instantiated once and placed as a top level `const` in the file. prob more of a judgement call than anything but the obj is so small that the memory footprint is negligible and we don't need to change this enum at any point

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+        prependChild(topbar, this.el);
+    }
+    createItems(genericItems, action, options = {}, Item = RadioMenuItem) {
+        const itemType = this.name;
+        const categoryButtonElement = this.categoryButton.element();
+        const menuItems = genericItems.map((item, index) => {
+            let content;
+            let argument;
+            switch (itemType) {
+                case 'quality':
```
jnatalzia: what are the benefits of this approach vs a subclass pattern where each menu type implements it's own items. i think that logical breakdown makes more sense but this is of course more concise

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+                    } else if (i !== undefined) {
+                        menuItems[i].el.focus();
+                    }
+                };
+                const evt = event.sourceEvent;
+                const nextItem = nextSibling(categoryButtonElement);
+                const prevItem = previousSibling(categoryButtonElement);
+                const nextSubItem = nextSibling(evt.target);
+                const prevSubItem = previousSibling(evt.target);
+                const key = evt.key.replace(/(Arrow|ape)/, '');
```
jnatalzia: same question here of key vs keyCode

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+        return menuItems;
+    }
+    defaultChild() {
+        return this.children.quality || this.children.captions || this.children.audioTracks || this.children.sharing;
+    }
+    setMenuItems(menuItems, initialSelectionIndex) {
+        if (!menuItems) {
+            this.removeMenu();
+        } else {
+            this.items = [];
```
jnatalzia: small thing that is probably me overoptimizing: any use of this and the below line residing in a `resetMenu` function?

jnatalzia: would make it more testable 😬 but if this is the only occurrence of this behavior, no sense in pulling it out

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+                menuItems[initialSelectionIndex].activate();
+            }
+            this.categoryButton.show();
+        }
+    }
+    appendMenu(childMenu) {
+        if (!childMenu) {
+            return;
+        }
+        // eslint-disable-next-line no-shadow
```
jnatalzia: wouldn't it be better to rename these vars instead of disabling the rule?

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+        const { el, name, categoryButton } = childMenu;
+        this.children[name] = childMenu;
+        if (categoryButton) {
```
jnatalzia: can i get a rundown on what this logic is doing? i'm a lil confused

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+        } else {
+            this.el.parentNode.classList.add('jw-settings-open');
+            this.trigger('menuVisibility', { visible: true, evt });
+            document.addEventListener('click', this.onDocumentClick);
+        }
+        this.visible = true;
+        this.el.setAttribute('aria-expanded', 'true');        
+    }
+    close(evt) {
+        // if visible, return
```
jnatalzia: conditional is correct, comment is not

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+        Object.keys(this.children).map(menuName => {
+            this.children[menuName].destroy();
+        });
+    }
+}
+
+const addGlobalMenuKeyListener = (settingsMenu) => {
+    const closeButton = settingsMenu.closeButton;
+    const settingsMenuElement = settingsMenu.el;
+    const ui = new UI(settingsMenuElement).on('keydown', function(evt) {
```
jnatalzia: where are we removing this listener on destroy?

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+                settingsMenu.close(evt);
+            }
+        };
+        const onRight = () => {
+            if (next && closeButton.element() && target !== closeButton.element()) {
+                next.focus();
+            }
+        };
+        let childMenu;
+        switch (key) {
```
jnatalzia: i'm assuming this code is straight ported from the old given the patterns it matches in the code above

jnatalzia: in that vein, my comments there apply here as well regarding style etc, but i wouldn't go too far in rewriting this stuff, esp if the structural change is more the effort of this pr

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+        const menuItemOptions = { defaultText: localization.auto };
+        changeMenuItems('quality', levels, api.setCurrentQuality, model.get('currentLevel'), menuItemOptions);
+        const childMenus = settingsMenu.children;
+        const shouldShowGear = !!childMenus.quality || !!childMenus.playbackRates || Object.keys(childMenus).length > 1;
+        controlbar.elements.settingsButton.toggle(shouldShowGear);
+    }, settingsMenu);
+    model.change('captionsList', (changedModel, captionsList) => {
+        const menuItemOptions = { defaultText: localization.off };
+        const initialIndex = model.get('captionsIndex');
+        changeMenuItems('captions', captionsList, api.setCurrentCaptions, initialIndex, menuItemOptions);
```
jnatalzia: i like this pattern a lot 🎉

src/js/view/controls/components/menu/menu.js
jnatalzia: does `this.ui.destroy` do it?

src/js/view/controls/components/simple-tooltip.js
```diff
@@ -19,6 +19,10 @@ export function SimpleTooltip(attachToElement, name, text, openCallback, closeCa
             if (instance.touchEvent) {
                 return;
             }
+            if (instance.suppress) {
+                instance.suppress = false;
```
jnatalzia: what implications does this have for other menus using this option?

src/js/view/controls/controls.js
```diff
@@ -200,34 +205,23 @@ export default class Controls extends Events {
...
-            this.userActive(activeTimeout);
-            lastState = state;
-
-            const settingsButton = this.controlbar.elements.settingsButton;
-
-            if (!visible && settingsButton) {
-                if (isKeyEvent) {
-                    settingsButton.element().focus();
-                } else if (evt) {
-                    focusFloatingElement();
```
jnatalzia: are we breaking anything by not respecting this previous `focusFloatingElement` call?

src/js/view/controls/components/menu/category-button.js
```diff
@@ -0,0 +1,39 @@
...
+        captions: 'cc-off',
+        audioTracks: 'audio-tracks',
+        quality: 'quality-100',
+        playbackRates: 'playback-rate',
+    };
+
+    // Allows override from Commercial Sharing menu
+    let icon = icons[name];
+    if (!icon) {
+        icon = icons.quality;
```
mamaddox: Is there a reason the quality icon is the default one used if one is not provided? I'm just curious.

src/js/view/controls/templates/menu/menu.js
```diff
@@ -0,0 +1,11 @@
+export const MenuTemplate = (isSubmenu, name) => {
```
mamaddox: Just a question, it seems like this MenuTemplate creates a settings menu or settings submenu. Are they interchangeable or are they a bit separate, because it seemed like SettingsMenu was more so a specific type of Menu?

src/js/view/controls/controls.js
```diff
@@ -200,34 +205,23 @@ export default class Controls extends Events {
...
             }
-        };
-        const settingsMenu = this.settingsMenu = createSettingsMenu(
-            controlbar,
-            visibilityChangeHandler,
-            model.get('localization')
-        );
-        setupSubmenuListeners(settingsMenu, controlbar, model, api);
+        });
+        settingsMenu.on('menuVisibility', (menu) => this.controlbar.trigger('menuVisibility', menu));
```
robwalch: Why does this event need to be forwarded through the `controlbar`?

It looks like you are only listening on the control bar above, so why not just listen on `settingsMenu` up there?

src/js/view/controls/controls.js
```diff
@@ -183,14 +183,19 @@ export default class Controls extends Events {
 
         this.div.appendChild(controlbar.element());
 
-        // Settings Menu
+        const localization = model.get('localization');
+        const settingsMenu = this.settingsMenu = SettingsMenu(api, model, this.controlbar, localization);
         let lastState = null;
-        const visibilityChangeHandler = (visible, evt) => {
+
+        this.controlbar.on('menuVisibility', ({ visible, evt }) => {
```
robwalch: `settingsMenu.on('menuVisibility', ({ visible, evt }) => {`?

Why not just listen on `settingsMenu` here and get rid of the event forwarding through the `controlbar` below?

src/js/view/controls/components/menu/category-button.js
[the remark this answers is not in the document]
zetagame: It's a required field of category button (those buttons on top of the menu that select what submenu it is). I'll change to return if no icon

src/js/view/controls/components/menu/menu.js
zetagame: We do a lot of custom navigation logic to conform to W3C accessibility standards. So there's a default behavior, but we want to overwrite that with the expected behavior. Refactoring the menu to have this behavior via HTML alone may be possible, but is outside of the scope of this refactor.

src/js/view/controls/components/menu/menu.js
zetagame: yep!

src/js/view/controls/components/menu/menu.js
zetagame: There goes my feeble attempt to get around line length limitations x)

src/js/view/controls/components/simple-tooltip.js
zetagame: Currently, this is only used for menus, but I can see potential for other use cases.

src/js/view/controls/controls.js
zetagame: See Todos in description

src/js/view/controls/components/menu/category-button.js
zetagame: Made it easier to test and develop, ensures there's always an icon (it's the only UI for category buttons visible without interacting directly with the button). That being said, after reading your and Joe's comments, I've decided to make it return when there isn't an icon instead.

src/js/view/controls/templates/menu/menu.js
zetagame: They're almost interchangeable after this refactor. But because a lot of people have custom implementations of our product I wanted to avoid changes the HTML class structure in any way that could affect custom skins/plugins.

src/js/view/controls/components/menu/menu-item.js
```diff
@@ -0,0 +1,33 @@
+import UI from 'utils/ui';
+import { toggleClass, createElement } from 'utils/dom';
+import { itemRadioButtonTemplate } from 'view/controls/templates/menu/menu-item';
+
+export class MenuItem {
+    constructor(_content, _action, _template = itemRadioButtonTemplate) {
+        this.el = createElement(_template(_content));
+        this.onClick = this.onClick.bind(this);
+        this.ui = new UI(this.el).on('click tap enter', (evt) => {
```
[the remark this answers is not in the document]
zetagame: Went with the latter to get rid of onclick, will update shortly

src/js/view/controls/components/menu/menu.js
zetagame: it is almost entirely ported

src/js/view/controls/components/menu/category-button.js
```diff
@@ -0,0 +1,39 @@
...
+        }, 
+        name, 
+        [(menu.icon && createElement(menu.icon)) || cloneIcon(icon)]
+    );
+
+    const buttonElement = menuCategoryButton.element();
+    buttonElement.setAttribute('role', 'menuitemradio');
+    buttonElement.setAttribute('aria-checked', 'false');
+    buttonElement.setAttribute('aria-label', localizedName);
+    if (!('ontouchstart' in window)) {
```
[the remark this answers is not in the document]
zetagame: It's a tricky one. I ported that from the old category button, but it's flawed. From MDN:

_This event model – which has since been submitted for W3C standardisation – unifies “pointer” devices (mouse, stylus, touch, etc) under a single new class of events. As this model does not, by design, include any separate ‘touch’, the feature detection for ontouchstart will naturally not work. The suggested method of detecting if a browser using Pointer Events is running on a touch-enabled device instead involves checking for the existence and return value of navigator.maxTouchPoints_

We could do something like the iPadOS detection and go off of navigator.maxTouchPoints. Let's discuss later today, I do like the idea of adding touch feature detection but I want to make sure I don't leave out a potential touch supported case.

src/js/view/controls/controls.js
zetagame: The reason I took this route is to avoid passing api and playerViewModel into the menu, and keeping things unidirectional. The innermost parent class that is responsible for both the settings menu and the menu button is controls, so it made sense to put it there.

src/js/view/controls/components/menu/menu.js
zetagame: I took it out and it no longer works. I think it's necessary here since otherwise 'this' will be the document.

zetagame:
> There will be an option to rename by localization?
We will be adding localized titles in a future update.

> Missing 50%
Not scoped in the ticket, but I'll bring this up before it's merged in.

zetagame:
test this please

zetagame:
test this please

zetagame:
test this please

src/css/controls/imports/settings-menu.less
```diff
@@ -97,17 +142,45 @@
...
     padding: 8px 20px 0 5px;
-
     .scrollbar-style();
 
     &.jw-settings-submenu-active {
         display: block;
     }
 
+    .jw-submenu-topbar {
+        box-shadow: 0 2px 9px 0 #1d1d1d;
```
jnatalzia: is this being autoprefixed? i think @robwalch had a similar comment on one of maria's prs but i don't remember the context entirely

src/js/model/storage.js
```diff
@@ -19,7 +19,8 @@ Object.assign(Storage.prototype, {
         return this.items.reduce((memo, key) => {
             const val = storage[`${this.namespace}.${key}`];
             if (val) {
-                memo[key] = serialize(val);
+                // eslint-disable-next-line no-unused-expressions
```
jnatalzia: you can abide by the lint rule by changing this to
```
memo[key] = key !== 'captions' ? serialize(val) : JSON.parse(val);
```

src/js/view/captionsrenderer.js
```diff
@@ -164,6 +169,11 @@ const CaptionsRenderer = function (viewModel) {
         }, this);
 
         _model.set('captions', _options);
+        _model.on('change:captions', (model, newOptions) => {
+            _options = newOptions;
+            styleCaptions();
```
jnatalzia: 😍

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+        closeButton.show();
+        closeButton.ui.on('keydown', (e) => {
+            const sourceEvent = e.sourceEvent;
+            const key = sourceEvent.key.replace(/(Arrow|ape)/, '');
+            // Close settings menu when enter is pressed on the close button
+            // or when tab or right arrow key is pressed since it is the last element in topbar
+            if (key === 'Enter' || key === 'Right' || (key === 'Tab' && !sourceEvent.shiftKey)) {
+                this.close(e);
+            }
+        }, this);
```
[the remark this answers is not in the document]
jnatalzia: this fix should also prevent you from having to do those `bind`s in your constructor

src/js/view/controls/components/menu/menu.js
```diff
@@ -135,52 +210,6 @@ export default class Menu extends Events {
             };
 
             const menuItem = new Item(content, menuItemClick.bind(this));
-            let topbarTarget =
```
jnatalzia: 🔥

src/js/view/controls/components/menu/menu.js
jnatalzia: bump on this question!

src/js/view/controls/components/menu/menu.js
jnatalzia: bump this comment

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -0,0 +1,345 @@
...
+        const initialIndex = model.get('captionsIndex');
+        changeMenuItems(
+            'captions', 
+            captionsList, 
+            (index) => api.setCurrentCaptions(index), 
+            initialIndex, 
+            menuItemOptions
+        );
+        const captionsMenu = settingsMenu.children.captions;
+        if (captionsMenu && !captionsMenu.children.captionsSettings) {
```
jnatalzia: lazy load, awesome stuff

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -0,0 +1,345 @@
...
+        if (colorTable[colors[i]] === value) {
+            option = colors[i];
+            break;
+        }
+    }
+    return option;
+};
+const getPercentOption = (value) => value + '%';
+const getPercentValue = (option) => parseInt(option);
+const captionStyleItems = [
```
jnatalzia: thoughts on this chunk of captions logic being at the top of the file or maybe a separate file for readability? it's a little tough to parse with it being after all the menu hookup

test/unit/menu-test.js
```diff
@@ -80,7 +81,7 @@ describe('Menu', () => {
     });
 });
 
-describe('Settings Menu', () => {
+describe.only('Settings Menu', () => {
```
jnatalzia: remove this! -- can we add a lint rule for this?

robwalch: We can. For reference see https://github.com/jwplayer/hls.js/commit/76e609f1bb5e4f1a6b2ad150be2085ce9d9245b9

src/js/view/controls/components/menu/settings-menu.js
zetagame: can put it in a utils file

src/css/controls/imports/settings-menu.less
zetagame: Yep!

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+        const childMenus = settingsMenu.children;
+        const shouldShowGear = !!childMenus.quality || !!childMenus.playbackRates || Object.keys(childMenus).length > 1;
+        controlbar.elements.settingsButton.toggle(shouldShowGear);
+    }, settingsMenu);
+    model.change('captionsList', (changedModel, captionsList) => {
+        const menuItemOptions = { defaultText: localization.off };
+        const initialIndex = model.get('captionsIndex');
+        changeMenuItems('captions', captionsList, api.setCurrentCaptions, initialIndex, menuItemOptions);
+    });
+    const setPlaybackRatesMenu = (playbackRates = model.get('playbackRates')) => {
```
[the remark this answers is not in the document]
zetagame: Removed the default assignment

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+        }
+        this.categoryButton = categoryButton(this, localizedName);
+        this.categoryButton.element().setAttribute('name', this.name);
+    }
+    createBackButton(localization) {
+        const backButton = button('jw-settings-close', this.close, localization.close, [cloneIcon('close')]);
+        return backButton;
+    }
+    createTopbar() {
+        const topbar = createElement(`<div class="jw-submenu-topbar"></div>`);
```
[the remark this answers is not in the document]
zetagame: Right now it's just for the captions submenu button's topbar

src/js/view/controls/components/menu/menu.js
```diff
@@ -0,0 +1,499 @@
...
+                const prevSubItem = previousSibling(evt.target);
+                const key = evt.key.replace(/(Arrow|ape)/, '');
+        
+                switch (key) {
+                    case 'Tab':
+                        focusElement(evt.shiftKey ? prevItem : nextItem);
+                        break;
+                    case 'Left':
+                        focusElement(prevItem || 
+                            previousSibling(document.getElementsByClassName('jw-icon-settings')[0]));
```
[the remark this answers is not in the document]
zetagame: Agreed, the listeners were ported from the previous implementation.

src/js/view/controls/components/menu/menu.js
zetagame: If the menu has a category button, append it to the main menu topbar with respect to our previous ordering logic. (Quality first, else put it before the close button)

src/js/view/controls/components/menu/menu.js
zetagame: Yep, removing these unnecessary comments.

src/js/model/storage.js
zetagame: Adding it!

zetagame:
@robwalch,@jnatalzia : I put a lot of thought into how to handle the bound listeners, and I've come to the conclusion it makes the most sense to leave them as is. As it stands, I see 4 options for how to do this:

1. Stays how it is.
2. Bind to menu anywhere function is passed (less DRY, less efficient, requires future implementers to know this has to be bound to menu)
3. Use arrow functions (have to set each as a const for reference or can’t remove event listeners, more lines of code)
4. Pass ‘this’ as context to UI (Still have to bind in non UI js cases).


Given the defensive, single operation nature of binding it in the constructor, I can't really see a downside to leaving it as is. If there's a good reason not to bind in constructor in cases like these and I'm missing it, feel free to respond here and I'll make the change.

mtoczko:
Hi @zetagame I have a suggestion:
After selecting the font color, the menu closes. And to change the font size you have to open the menu again. A better solution is to return to the subtitle settings menu

zetagame:
> Hi @zetagame I have a suggestion:
> After selecting the font color, the menu closes. And to change the font size you have to open the menu again. A better solution is to return to the subtitle settings menu

Couldn't agree more. I'll be pushing an additional PR shortly that implements this pattern.

src/js/view/controls/components/menu/category-button.js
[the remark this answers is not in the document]
robwalch: I made the comment because I want to avoid feature detection that hits browser globals like `window` in places like this. Whether or not there is pointer support is a different issue, but is related to whether or not we want to show tooltips.

robwalch:
> Given the defensive, single operation nature of binding it in the constructor, I can't really see a downside to leaving it as is.

You are adding code that is completely unnecessary.

PR #3536, opened 2019-11-22
[JW8-10549] Fix the controls not autohiding on mobile

mamaddox:
### This PR will...

- Fix the issue where controls were not automatically hiding when the user was inactive for 4 seconds on mobile.

- Bug fix

#### Addresses Issue(s):

[JW8-10549](https://jwplayer.atlassian.net/browse/JW8-10549)

### Checklist

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -98,7 +98,7 @@ function View(_api, _model) {
     }
 
     function fosMobileBehavior() {
-        return OS.mobile && !deviceIsLandscape();
+        return _isMobile && !deviceIsLandscape();
     }
 
     // Compute player size, handle DOM removal/insertion, add to views-manager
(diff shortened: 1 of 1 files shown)


jnatalzia:
@mamaddox the changes here make sense but could i get a quick explainer of the issue underlying it? was the initial `mouse` handler causing some unintentional state with regards to the controls hiding?

mamaddox:
As far as my understanding goes, this issue was occurring only if the user selected a control in the control bar after hitting play, but before the first userInactive event occurred. 
When this happened, a mousemove event was triggered inside the controlbar that set the activeTimeout to zero and didn't invoke the userInactiveTimeout. However, when the user tapped the screen, the mouseevent listeners were removed so things went back to normal. It seemed like those event listeners were not needed since they always got removed whenever the screen was tapped.

But now that I'm thinking about it, is there a case where mobile needs both mouse and tap events?

jnatalzia:
> But now that I'm thinking about it, is there a case where mobile needs both mouse and tap events?

I can't think of one, especially if we were immediately removing those listeners on tap anyway.

jnatalzia:
Hmm thinking about it, there's the possibility that older devices don't support the touch events and therefore would never hit the execution path where they are removed.

Support looks really good across mobile browsers though so I don't think this concern is justified
https://developer.mozilla.org/en-US/docs/Web/API/Touch_events

PR #3539, opened 2019-12-04
[JW8-10583] Add 'liveSyncDuration' player configuration attribute

mamaddox:
### This PR will...
- Validate and set the 'liveSyncDuration' attribute specified in the player config in the player model

This will allow users to specify how much latency they want the player to stream with through the 'liveSyncDuration' attribute. It also ensures that the 'dvrSeekLimit' and 'liveSyncDuration' attribute are the same.

[jwplayer-commercial #7304](https://github.com/jwplayer/jwplayer-commercial/pull/7304)

#### Addresses Issue(s):

JW8-10583

### Checklist

### DIFF
--- a/src/js/api/core-shim.js
+++ b/src/js/api/core-shim.js
@@ -94,7 +94,6 @@ Object.assign(CoreShim.prototype, {
             return new Providers(configuration);
         };
         model.setProvider = function() {};
-        setLiveSyncDurationAttribute(model);
 
         // Create/get click-to-play media element, and call .load() to unblock user-gesture to play requirement
         let mediaPool = MediaElementPool();
@@ -313,17 +312,6 @@ function logWarning(warning) {
         return;
     }
     console.warn(PlayerError.logMessage(warning.code));
-}
-
-function setLiveSyncDurationAttribute(model) {
-    const liveSyncDuration = model.get('liveSyncDuration');
-    if (!liveSyncDuration) {
-        model.set('liveSyncDuration', 25);
-    } else if (liveSyncDuration < 5) {
-        model.set('liveSyncDuration', 5);
-    } else if (liveSyncDuration > 30) {
-        model.set('liveSyncDuration', 30);
-    }
 }
 
 export function showView(core, viewElement) {
--- a/src/js/api/core-shim.js
+++ b/src/js/api/core-shim.js
@@ -94,6 +94,7 @@ Object.assign(CoreShim.prototype, {
             return new Providers(configuration);
         };
         model.setProvider = function() {};
+        setLiveSyncDurationAttribute(model);
 
         // Create/get click-to-play media element, and call .load() to unblock user-gesture to play requirement
         let mediaPool = MediaElementPool();
@@ -312,6 +313,17 @@ function logWarning(warning) {
         return;
     }
     console.warn(PlayerError.logMessage(warning.code));
+}
+
+function setLiveSyncDurationAttribute(model) {
+    const liveSyncDuration = model.get('liveSyncDuration');
+    if (!liveSyncDuration) {
+        model.set('liveSyncDuration', 25);
+    } else if (liveSyncDuration < 5) {
+        model.set('liveSyncDuration', 5);
+    } else if (liveSyncDuration > 30) {
+        model.set('liveSyncDuration', 30);
+    }
 }
 
 export function showView(core, viewElement) {
--- a/src/js/api/setup-steps.js
+++ b/src/js/api/setup-steps.js
@@ -33,7 +33,7 @@ export function loadPlaylist(_model) {
 
 function setPlaylistAttributes(model, playlist, feedData) {
     const attributes = model.attributes;
-    attributes.playlist = Playlist(playlist, model);
+    attributes.playlist = Playlist(playlist);
     attributes.feedData = feedData;
 }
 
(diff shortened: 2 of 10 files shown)


src/js/playlist/playlist.js
```diff
@@ -3,10 +3,12 @@ import PlaylistItem from 'playlist/item';
 import Source from 'playlist/source';
 import Providers from 'providers/providers';
 import { PlayerError, MSG_CANT_PLAY_VIDEO } from 'api/errors';
+import { isObject } from 'utils/underscore';
 
-const Playlist = function(playlist) {
+const Playlist = function(playlist, model) {
```
mamaddox: I passed the model to the Playlist because the 'liveSyncDuration' attribute is set in the model, but it needs to be copied to the playlist items.

src/js/api/core-shim.js
```diff
@@ -94,6 +94,7 @@ Object.assign(CoreShim.prototype, {
             return new Providers(configuration);
         };
         model.setProvider = function() {};
+        setLiveSyncDurationAttribute(model);
```
mamaddox: I thought this would be a good place for this because this is where the model is created and also where the other config properties are copied to the model

src/js/api/core-shim.js
```diff
@@ -314,6 +315,17 @@ function logWarning(warning) {
     console.warn(PlayerError.logMessage(warning.code));
 }
 
+function setLiveSyncDurationAttribute(model) {
```
robwalch: Can we move `getLiveSyncDuration(liveSyncDuration)` in commercial (https://github.com/jwplayer/jwplayer-commercial/pull/7304/files#diff-98c5c29fe89f7774786f064799c89076R89) to a helper in config.js here so we don't duplicate this clamping logic?

The this could look like:
```js
const liveSyncDuration = getLiveSyncDuration(model.get('liveSyncDuration'));
model.set('liveSyncDuration', liveSyncDuration);
```

But then do we have to do this here or could we just do it in config.js or playlist.js?

src/js/api/core-shim.js
```diff
@@ -314,6 +315,17 @@ function logWarning(warning) {
...
+    const liveSyncDuration = model.get('liveSyncDuration');
```
robwalch: Can we move `getLiveSyncDuration(liveSyncDuration)` in commercial (https://github.com/jwplayer/jwplayer-commercial/pull/7304/files#diff-98c5c29fe89f7774786f064799c89076R89) to a helper in config.js here so we don't duplicate this clamping logic?

The this could look like:
```js
const liveSyncDuration = getLiveSyncDuration(model.get('liveSyncDuration'));
model.set('liveSyncDuration', liveSyncDuration);
```

But then do we have to do this here or could we just do it in config.js or playlist.js?

src/js/api/setup-steps.js
```diff
@@ -33,7 +33,7 @@ export function loadPlaylist(_model) {
 
 function setPlaylistAttributes(model, playlist, feedData) {
     const attributes = model.attributes;
-    attributes.playlist = Playlist(playlist);
+    attributes.playlist = Playlist(playlist, model);
```
robwalch: I'm not sure why this is necessary. We've been getting by without passing in the model, here, and that has worked fine for other config values copied into sources.

While it's nice to see it on the item level, that doesn't justify all these changes. We should just copy it to sources with `formatSources` instead of adding all this new code.

src/js/playlist/playlist.js
robwalch: It should be passed from the config, not the model.

src/js/api/core-shim.js
mamaddox: Makes sense! Moved to config.js so it no longer needs to be set in the model. Not sure why I didn't see that before.

src/js/api/setup-steps.js
mamaddox: Done 👍 And it is much simpler.

src/js/playlist/item.js
```diff
@@ -28,11 +28,6 @@ const Item = function(config) {
         dvrSeekLimit: 25
     }, config);
 
-    if (playlistItem.dvrSeekLimit < 5) {
-        playlistItem.dvrSeekLimit = 5;
-    }
```
mamaddox: Don't think this check is needed anymore because it seems to always gets overridden by 'formatItems' in Playlist

src/js/playlist/playlist.js
```diff
@@ -45,11 +44,17 @@ export function normalizePlaylistItem(model, item, feedData) {
...
 
-    return playlistItem;
+    return formatItem(playlistItem);
 }
 
 export const fixSources = (item, model) => filterSources(formatSources(item, model), model.getProviders());
 
+function formatItem(item) {
+    const liveSyncDuration = item.sources[0].liveSyncDuration;
+    item.dvrSeekLimit = item.liveSyncDuration = liveSyncDuration;
```
robwalch: I'm concerned that this will overwrite any existing `dvrSeekLimit` set by users.

Unfortunately it doesn't look like we have any tests in commercial that do this (all the test configs set it to 25).

mamaddox: This definitely will. It will always at least overwrite it to the default 25, so that's a good thing to note.

PR #3541, opened 2019-12-10
[JW8-8919] Introducing Typescript into the player

mamaddox:
### This PR will...
- Configure the player to build, lint, and compile Typescript
- Add a type-check build task
- Convert 5 utils to Typescript 

- This PR will act as a kick-off to the introduction of Typescript into the player.
- Adding types to our code can help us have cleaner code as well as make our code easier to understand.
- It may also identify hidden issues in our code.

I added comments/questions, but this should get a general in-depth look.

#### Addresses Issue(s):

JW8-8919

### Checklist


mamaddox:
Extended ESLint Rules:
[eslint:recommend](https://github.com/eslint/eslint/blob/master/conf/eslint-recommended.js)
[typescript-eslint/recommend](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/recommended.json)

src/ts/.eslintrc
```diff
@@ -0,0 +1,32 @@
...
+    },
+    "parser": "@typescript-eslint/parser",
+    "parserOptions":{
+        "sourceType": "module",
+        "project": "./tsconfig.json"
+    },
+    "extends": [
+        "eslint:recommended",
+        "plugin:@typescript-eslint/recommended"
+    ],
```
mamaddox: List of ESLint Rules:
[eslint:recommend](https://github.com/eslint/eslint/blob/master/conf/eslint-recommended.js)
[typescript-eslint/recommend](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/src/configs/recommended.json)

src/ts/utils/random-id-generator.ts
```diff
@@ -0,0 +1,30 @@
+export const FEED_SHOWN_ID_LENGTH = 12;
```
jnatalzia: instead of a new folder, we should update the existing file - switching the suffix to `ts` and updating the webpack/babel config to recognize both types of files as worthy of bundling.

package.json
```diff
@@ -70,14 +74,16 @@
     "build": "grunt",
     "docs": "jsdoc -c docs/jsdoc.conf.json",
     "hooks": "\\cp .github/hooks/pre-push .git/hooks/pre-push",
-    "lint": "yarn lint:js && yarn lint:tests && yarn lint:styles",
+    "lint": "yarn lint:ts && yarn lint:js && yarn lint:tests && yarn lint:styles",
     "lint:js": "eslint './src/js'",
+    "lint:ts": "eslint './src/ts/**/*.ts'",
```
jnatalzia: see comment below about changing the existing files in place rather than a new folder, will remove the need for this command

src/ts/utils/random-id-generator.ts
jnatalzia: oh looks like the config has already been updated! then just the file extension swap _should_ put us in a situation to test the full build of the player

webpack.config.js
```diff
@@ -89,6 +90,17 @@ const webpackConfig = {
                         ]
                     }
                 }
+            },
```
robwalch: We can use the same loader for js and ts files. This allows us to share preset, plugin an exclude setting both script types:

https://github.com/jwplayer/jwplayer-commercial/pull/7324/files#diff-11e9f7f953edc64ba14b0cc350ae7b9dR99-R120

tsconfig.json
```diff
@@ -0,0 +1,12 @@
+{
```
robwalch: Additional configuration settings 
https://github.com/jwplayer/jwplayer-commercial/pull/7324/files#diff-e5e546dd2eb0351f813d63d1b39dbc48R2

src/js/utils/browser.ts
```diff
@@ -1,16 +1,16 @@
...
     return userAgent.match(regex) !== null;
 }
 
-function lazyUserAgentMatch(regex) {
-    return function () {
-        return userAgentMatch(regex);
-    };
+function lazyUserAgentMatch(regex: RegExp): () => boolean {
+    return () => userAgentMatch(regex);
 }
```
mamaddox: It looks like for 'Function' return types, they can be more explicitly typed like above using arrow functions or more simply typed like below:

> `function lazyUserAgentMatch(regex: RegExp): Function`

I want to note that I am not 100% as to the larger difference between the two aside from being more or less explicit with types.

src/js/utils/validator.ts
```diff
@@ -6,7 +6,7 @@ const protocol = window.location.protocol;
  * @param {any} item - The variable to test.
  * @returns {boolean} Is the value of `item` null, undefined or an empty string?
  */
```
mamaddox: Should we keep these JSDOC comments?

src/js/utils/validator.ts
```diff
@@ -47,15 +47,15 @@ export function isRtmp(file, type) {
...
+export function isYouTube(path: string, type: string): boolean {
     return (type === 'youtube') || (/^(http|\/\/).*(youtube\.com|youtu\.be)\/.+/).test(path);
 }
 
 /**
  * @param {string} value - The variable to test.
  * @returns {string} The typeof object, 'array' or 'null'.
  */
-export function typeOf(value) {
+export function typeOf(value: any): string {
```
mamaddox: Question about this function:
It looks like this method is only used for tests, but I was wondering about what this 'value' is? The comment implies it is a string, but it also seems to be checking for arrays as well?

src/js/utils/validator.ts
```diff
@@ -75,7 +75,7 @@ export function typeOf(value) {
  * @param {function} predicate - The function evaluating whether the property has a valid value and can be considered compliant. Inputs are the object and its key.
  * @returns {boolean} Does the customObj have at least the same keys as defaultObj, and do their properties also share the same keys ?
  */
-export function isDeepKeyCompliant(defaultObj, customObj, predicate) {
+export function isDeepKeyCompliant(defaultObj: {[key: string]: any}, customObj: {[key: string]: any}, predicate: (x: string, y: object) => boolean): boolean {
```
zetagame: What's the issue with using standard typing here? ie: 

`export function isDeepKeyCompliant(defaultObj: object, customObj: object, predicate: (x: string, y: object) => boolean): boolean`

src/js/utils/parser.ts
```diff
@@ -1,14 +1,16 @@
...
+import { exists } from './validator';
+import { isNaN } from './underscore';
 
 // Returns the absolute file path based on a relative filepath, and optional base path
-export function getAbsolutePath(path, base) {
-    if (!exists(base)) {
+export function getAbsolutePath(path: string): string;
+export function getAbsolutePath(path: string, base: string): string;
+export function getAbsolutePath(path: string, base?: string): string {
+    if (!base || !exists(base)) {
```
mamaddox: Overloading methods.
I know this is small, but idk if we want to space them out. Maybe it's nicer on the eyes.

test/unit/parser-test.js
```diff
@@ -69,13 +69,10 @@ describe('parser', function() {
...
 
         dimension = parser.parseDimension('35%');
-        expect(dimension, 'dimension with percentage string should be the same').to.equal('35%');
+        expect(dimension, 'dimension with percentage string should be the same').to.equal(35);
 
         dimension = parser.parseDimension('35px');
         expect(dimension, 'dimension with px string should remove px').to.equal(35);
-
-        dimension = parser.parseDimension(35);
-        expect(dimension, 'dimension with int should be itself').to.equal(35);
```
mamaddox: I removed tests like this because they are technically invalid, and it felt weird keeping type checks inside the typed functions.

src/js/utils/strings.ts
```diff
@@ -94,25 +99,32 @@ export function seconds(str, frameRate) {
 }
 
 // Convert an offset string to a number; supports conversion of percentage offsets
-export function offsetToSeconds(offset, duration, frameRate) {
-    if (isString(offset) && offset.slice(-1) === '%') {
+export function offsetToSeconds(offset: string): number;
+export function offsetToSeconds(offset: string, duration: number): number;
+export function offsetToSeconds(offset: string, duration: number, frameRate: number): number;
```
mamaddox: It looks like some methods like 
> offsetToSeconds(offset, duration, frameRate)

aren't used outside of tests. Not sure if we should keep them.

src/js/utils/underscore.d.ts
```diff
@@ -0,0 +1,6148 @@
+// Type definitions for Underscore 1.9
```
mamaddox: Since we use a custom underscore lib file, this file is the type definition for it.
I just took the type definition file for the os underscore lib and added one of our custom methods.
I'm assuming there are a lot more, but I honestly did not take the time to see how many custom functions we have.
You may need to add the custom method type definition to this file if you try to use one.
However, if we wind up typing the underscore.js file, or upgrade and use the os one, we can get rid of this.

src/js/utils/validator.ts
mamaddox: Currently, the `noImplicitAny` flag is enabled, so you need to explicitly specify what the index/key type for an object is.
I kept all the strict settings enabled at least for now (They are also enabled by default). Can definitely discuss to determine what we want to keep.

mamaddox:
test this please

mamaddox:
test this please

src/js/utils/validator.ts
robwalch: value: any

robwalch: value: any

robwalch: value: any

robwalch: value: any

src/js/utils/parser.ts
robwalch: Why?

test/unit/parser-test.js
robwalch: Why is it invalid? Any of these inputs could be supplied as width or height in a player configuration.

mamaddox: True. I hadn't thought about that. It just felt weird to have to keep type checks inside of typed functions.

src/js/utils/parser.ts
mamaddox: Makes sense. I'm still just stuck in that mindset because I like overloading. Lol

src/js/utils/strings.ts
```diff
@@ -40,23 +40,24 @@ export function extension(path) {
...
     const m = ((secondsNumber / 60) | 0) % 60;
     const s = secondsNumber % 60;
-    return pad(h, 2) + ':' + pad(m, 2) + ':' + pad(s.toFixed(3), 6);
+    return pad(h.toString(), 2) + ':' + pad(m.toString(), 2) + ':' + pad(s.toFixed(3), 6);
 }
 
 // Convert a time-representing string to a number
-export function seconds(str, frameRate) {
+export function seconds(str: string, frameRate?: number): number {
     if (!str) {
```
zetagame: We should either remove the if(!str) check, or make str nullable

src/js/utils/strings.ts
```diff
@@ -94,25 +95,29 @@ export function seconds(str, frameRate) {
 }
 
 // Convert an offset string to a number; supports conversion of percentage offsets
-export function offsetToSeconds(offset, duration, frameRate) {
+export function offsetToSeconds(offset: string, duration?: number, frameRate?: number): number | null {
```
zetagame: It looks like offset can also be a float (see line 105 of original)

src/js/utils/strings.ts
mamaddox: Made it an optional param. Just to note, I do agree with what you said where we may need to revisit our optional params in the future.

mamaddox: Made it an optional param. Just to note, I do agree with what you said where we may need to revisit our optional params in the future.

src/js/utils/validator.ts
mamaddox: To note, there is the '[suppressImplicitAnyIndexErrors'](https://www.typescriptlang.org/docs/handbook/compiler-options.html) tsconfig compiler option

src/js/utils/strings.ts
```diff
@@ -1,12 +1,12 @@
-import { isValidNumber, isString } from 'utils/underscore';
+import { isValidNumber, isString } from './underscore';
```
jnatalzia: did we figure out the symlinking here? where `utils` points to the proper dir (especially across jwplayer and commercial)

src/js/utils/strings.ts
```diff
@@ -16,7 +16,7 @@ export function pad(str, length, padder) {
 }
 
 // Get the value of a case-insensitive attribute in an XML node
-export function xmlAttribute(xml, attribute) {
+export function xmlAttribute(xml: Element, attribute: string): string {
```
jnatalzia: what is the difference between `Element` for the xml here and `Document` above?

src/js/utils/strings.ts
```diff
@@ -40,23 +40,24 @@ export function extension(path) {
...
 
 // Convert a time-representing string to a number
-export function seconds(str, frameRate) {
+export function seconds(str?: string | number, frameRate?: number): number {
     if (!str) {
         return 0;
     }
     if (isValidNumber(str)) {
-        return str;
+        return parseFloat(str.toString());
```
jnatalzia: hmmm, if we type the return value of `isValidNumber` do we still need this `parse/toString` set of calls?

src/js/utils/underscore.d.ts
jnatalzia: is there a way to include this as a dependency rather than committing it to the repo?

src/js/utils/strings.ts
```diff
@@ -94,25 +95,29 @@ export function seconds(str, frameRate) {
...
     if (isString(offset) && offset.slice(-1) === '%') {
         const percent = parseFloat(offset);
         if (!duration || !isValidNumber(duration) || !isValidNumber(percent)) {
             return null;
         }
         return duration * percent / 100;
     }
-    return seconds(offset, frameRate);
+
+    if (frameRate) {
```
jnatalzia: what is the type error we get without this conditional compared to the above `seconds(offset, frameRate)`

src/js/utils/validator.ts
jnatalzia: i'm fine with punting this as long as the comments match the type definitions

src/js/utils/validator.ts
jnatalzia: we should create an interface or type for the generic object `{[key: string]: any},` as i can see it being used quite a lot

src/js/utils/strings.ts
mamaddox: Nice catch! There isn't one. I forgot to remove this when I removed the method overloads.

src/js/utils/strings.ts
mamaddox: Nope! Not needed. Not sure why I did that.

src/js/utils/strings.ts
mamaddox: Relooking at this, it looks like above should be returning an `XMLDocument` instead. This `Element` is just an XML node. Apparently `Element` inherits from `Node`, `ChildNode`, and `ParentNode`.
As far as I know the `XMLDocument` can contain these elements.

src/js/utils/parser.ts
```diff
@@ -81,21 +81,22 @@ export function serialize(val) {
...
     }
-    return dimension;
+    return parseInt(dimension.replace('px', ''), 10);
 }
 
 // Returns a formatted time string from "mm:ss" to "hh:mm:ss" for the given number of seconds
-export function timeFormat(sec, allowNegative) {
-    if ((sec <= 0 && !allowNegative) || isNaN(parseInt(sec))) {
+export function timeFormat(sec: number, allowNegative?: boolean): string {
+    if (isNaN(sec) || !isFinite(sec) || (sec <= 0 && !allowNegative)) {
```
jnatalzia: are we positive `sec` is only a string when we want to throw it out? previously we would try to parse int before getting rid of iit

src/js/utils/validator.ts
jnatalzia: bump this comment

src/js/utils/parser.ts
mamaddox: Somehow I glanced over that. Good catch!

src/js/utils/strings.ts
mamaddox: Looks like this 'baseUrl' compiler option in the tsconfig will help
https://www.typescriptlang.org/docs/handbook/module-resolution.html#base-url

mamaddox:
test this please

src/js/types/generic.type.ts
```diff
@@ -0,0 +1 @@
+export type GenericObjectType = { [key: string]: any };
```
jnatalzia: makes sense to have something like this in a types folder :+1: setting a note here for other defined types as they should be present in the file where the logic is declared

src/js/utils/browser.ts
```diff
@@ -81,7 +81,7 @@ export function flashVersion() {
     let flash;
 
     if (plugins) {
-        flash = plugins['Shockwave Flash'];
+        flash = plugins.namedItem('Shockwave Flash');
```
jnatalzia: is there a lint rule that's yelling about the old way?

test/unit/parser-test.js
```diff
@@ -69,7 +69,7 @@ describe('parser', function() {
         expect(dimension, 'dimension with empty string should be 0').to.equal(0);
 
         dimension = parser.parseDimension('35%');
-        expect(dimension, 'dimension with percentage string should be the same').to.equal('35%');
+        expect(dimension, 'dimension with percentage string should be the same').to.equal(35);
```
jnatalzia: hm this seems like a breaking change based on your parse int in the parent function. the test should not need to be changed to pass unless it's something like the above where we are swapping two falsy valuess

test/unit/strings-test.js
```diff
@@ -81,7 +81,7 @@ describe('strings', function() {
     });
 
     function timeConversionTest(converter) {
-        let sec = converter(5);
```
jnatalzia: same comment here - the test is testing real applications for this functionality, we should still be able to pass a number in here, maybe we need a union type in the function?

mamaddox: makes sense. It's funny because I had added a union type for some other reason, but didn't revert this change.

test/unit/parser-test.js
mamaddox: done

src/js/utils/browser.ts
mamaddox: This was actually throwing a type error. The index type for PluginArrays is 'number', but since we are using a 'string', it was throwing an 'implicitAny' type error

mamaddox:
test this please

mamaddox:
test this please

mamaddox:
test this please

webpack.config.js
```diff
@@ -69,13 +70,16 @@ const webpackConfig = {
...
-                test: /\.js$/,
+                test: /\.(?:ts|js)$/,
                 exclude: /\/node_modules\//,
                 use: {
                     loader: 'babel-loader',
                     options: {
                         babelrc: false,
-                        presets: ['@babel/preset-env'],
+                        presets: [
+                            '@babel/preset-env',
```
jnatalzia: same q as in commercial 

> does the ordering here matter? like should we be running the TS typescript before the env preset?

i'm curious what the babel output looks like with it in this order

jnatalzia: Verified output is properly transpiled and plays in IE11

src/js/types/generic.type.ts
zetagame: Code looks looks great! Giving the PR an approve, but this line gives me some pause. Is this for the no implicit any rule? Def agree with your comment below @mamaddox, and if we're already creating workarounds I think it is best to remove the rule altogether and replace it with no explicit any. If the issue is to avoid typing values and returns as `any`, no explicit any will accomplish this without the weird side-effects.

mamaddox: Yeah. This is for that `noImplicitAny` rule.

I just want to bump this comment: https://github.com/jwplayer/jwplayer/pull/3541#discussion_r359961348

> To note, there is the '[suppressImplicitAnyIndexErrors'](https://www.typescriptlang.org/docs/handbook/compiler-options.html) tsconfig compiler option

This setting would suppress these `noImplicitAny` errors when indexing objects (which we have already pointed out twice). Something to keep in mind.

src/js/utils/parser.ts
```diff
@@ -1,14 +1,14 @@
 import { exists } from 'utils/validator';
-import { isNaN } from 'utils/underscore';
+import { isNaN, isValidNumber } from 'utils/underscore';
 
 // Returns the absolute file path based on a relative filepath, and optional base path
-export function getAbsolutePath(path, base) {
-    if (!exists(base)) {
+export function getAbsolutePath(path: string, base?: string): string {
+    if (!base || !exists(base)) {
```
zetagame: `!base` here might be a redundant check. If `exists(base)===true` I don't think `!!base` can possibly be false in application (theoretically if base is 0 it could be, but I can't think of a case where a base path would be `0`).

src/js/utils/parser.ts
```diff
@@ -81,21 +81,26 @@ export function serialize(val) {
...
-export function parseDimension(dimension) {
-    if (typeof dimension === 'string') {
-        if (dimension === '') {
-            return 0;
-        } else if (dimension.lastIndexOf('%') > -1) {
-            return dimension;
-        }
-        return parseInt(dimension.replace('px', ''), 10);
+export function parseDimension(dimension: string): number | string {
+    if (isValidNumber(dimension)) {
```
zetagame: CSS can have non string dimension values. ie: `margin: 0`. This function isn't actually used in player or commercial either. @robwalch my intuition tells me we shouldn't remove it in case it's being used in custom implementations. Let me know if I'm wrong and we can remove the function.

src/js/utils/parser.ts
mamaddox: It is redundant. However, because `base` is an optional parameter and `getAbsoultePath` doesn't know that `exists` checks if the input is undefined, when `base` is used later in this method, it raises an error since it "could be undefined". Not sure if there is a workaround for that.

jnatalzia: ahhh so if i'm reading this right would we potentially be able to get rid of this extra check when `exists` is typed or no? either way let's discuss in standup and come to an agreement on something - so close to getting this merged 👍

src/js/types/generic.type.ts
jnatalzia: Keeping for now 👍

src/js/utils/validator.ts
```diff
@@ -1,3 +1,5 @@
+import { GenericObjectType } from 'types/generic.type';
```
jnatalzia: `GenericObject` @zetagame 's contribution of the century

jnatalzia:
test this please

PR #3544, opened 2019-12-12
[JW8-10769] Add bidding error code.

vseventer:
### This PR will...
Add a new code for when the bidding.js file failed to load.

Header bidding will be loaded as player plugin going forward, and therefore deserves its own error code.

N/A

jwplayer/jwplayer-commercial#7322
jwplayer/jwplayer-ads-header-bidding#137

#### Addresses Issue(s):
JW8-10769

### Checklist

### DIFF
--- a/src/js/plugins/utils.js
+++ b/src/js/plugins/utils.js
@@ -29,6 +29,9 @@ export function mapPluginToCode(pluginUrl) {
         case 'gapro':
             code = 305006;
             break;
+        case 'bidding':
+            code = 305007;
+            break;
         default:
             break;
     }
--- a/src/js/plugins/utils.js
+++ b/src/js/plugins/utils.js
@@ -4,39 +4,8 @@ export const getPluginName = function(url) {
     return url.replace(/^(.*\/)?([^-]*)-?.*\.(js)$/, '$2');
 };
 
-export function mapPluginToCode(pluginUrl) {
-    let code = 305000;
-    if (!pluginUrl) {
-        return code;
-    }
-
-    switch (getPluginName(pluginUrl)) {
-        case 'jwpsrv':
-            code = 305001;
-            break;
-        case 'googima':
-            code = 305002;
-            break;
-        case 'vast':
-            code = 305003;
-            break;
-        case 'freewheel':
-            code = 305004;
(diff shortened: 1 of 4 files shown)


src/js/plugins/utils.js
```diff
@@ -29,6 +29,9 @@ export function mapPluginToCode(pluginUrl) {
         case 'gapro':
             code = 305006;
             break;
+        case 'bidding':
+            code = 305007;
+            break;
```
robwalch: This method should be stubbed in `jwplayer` and extended in jwplayer-commercial with this implementation containing commercial plugin error codes.

To extend this module, create src/js/plugins/utils.js in jwplayer-commmercial and import mapPluginToCode from os/plugins/utils.js.

vseventer: @robwalch All plugins in this `switch` are commercial only correct?

robwalch: Correct.

vseventer:
Test this please

src/js/plugins/utils.js
```diff
@@ -4,36 +4,8 @@ export const getPluginName = function(url) {
...
-        case 'gapro':
-            code = 305006;
-            break;
-        default:
-            break;
-    }
-
-    return code;
+export function mapPluginToCode(/* pluginUrl */) {
+    return 305000;
```
robwalch: Did this get moved to commercial?

vseventer: Yes - see https://github.com/jwplayer/jwplayer-commercial/pull/7322/files#diff-5b0f8381f2c9891927f1d9948f531671R10

src/js/plugins/utils.js
```diff
@@ -4,36 +4,8 @@ export const getPluginName = function(url) {
...
-            break;
-        case 'gapro':
-            code = 305006;
-            break;
-        default:
-            break;
-    }
-
-    return code;
+export function mapPluginToCode(/* pluginUrl */) {
```
robwalch: It might make sense to rename this to `getPluginErrorCode` since the mapping is no longer happening here, and we shouldn't need to comment on commercial features here.

PR #3547, opened 2019-12-16
[JW8-10791] Bugfix/menu accessability order fix

zetagame:
### This PR will...
Re-implement the old accessibility order.

WCAG/ARIA spec is to select first menu item on open, however this would prevent accessibility users from knowing about other submenus. As a result the category button row can be seen as a menu itself, so focus should go to the default menu's category button when main menu is opened. This was the behavior prior to refactor, so this PR simply reverts it to how it was before.

#### Addresses Issue(s):

JW8-10791

### Checklist

### DIFF
--- a/src/js/view/controls/components/menu/menu.js
+++ b/src/js/view/controls/components/menu/menu.js
@@ -51,9 +51,10 @@ export default class Menu extends Events {
         return quality || captions || audioTracks || sharing || playbackRates;
     }
     createItemsContainer() {
+        const close = this.close;
         const itemsContainerElement = this.el.querySelector('.jw-settings-submenu-items');
         const getTopbar = () => this.topbar;
-        const onKeydown = function(event) {
+        const onKeydown = function(evt) {
             if (event.target.parentNode !== itemsContainerElement) {
                 return;
             }
@@ -64,8 +65,7 @@ export default class Menu extends Events {
                     itemsContainerElement.childNodes[i].focus();
                 }
             };
-            const evt = event.sourceEvent;
-            const target = evt.target;
+            const { sourceEvent, target } = evt;
             const isFirstChild = itemsContainerElement.firstChild === target;
             const isLastChild = itemsContainerElement.lastChild === target;
             const topbar = getTopbar();
@@ -73,10 +73,10 @@ export default class Menu extends Events {
             const prevItem = backButtonElement || previousSibling(topbarTarget);
             const nextSubItem = nextSibling(evt.target);
             const prevSubItem = previousSibling(evt.target);
-            const key = evt.key.replace(/(Arrow|ape)/, '');
+            const key = sourceEvent.key.replace(/(Arrow|ape)/, '');
             switch (key) {
                 case 'Tab':
-                    focusElement(evt.shiftKey ? prevItem : nextItem);
+                    focusElement(sourceEvent.shiftKey ? prevItem : nextItem);
                     break;
                 case 'Left':
                     focusElement(prevItem || 
@@ -99,13 +99,16 @@ export default class Menu extends Events {
                         focusElement(nextSubItem, 0);
                     }
                     break;
+                case 'Escape':
+                    close(event);
+                    break;
                 default:
                     break;
             }
-            evt.preventDefault();
+            sourceEvent.preventDefault();
             if (key !== 'Esc') {
                 // only bubble event if esc key was pressed
-                evt.stopPropagation();
+                sourceEvent.stopPropagation();
             }
             
         };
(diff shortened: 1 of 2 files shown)


src/js/view/controls/components/menu/menu.js
```diff
@@ -51,9 +51,10 @@ export default class Menu extends Events {
         return quality || captions || audioTracks || sharing || playbackRates;
     }
     createItemsContainer() {
+        const close = this.close;
         const itemsContainerElement = this.el.querySelector('.jw-settings-submenu-items');
         const getTopbar = () => this.topbar;
-        const onKeydown = function(event) {
+        const onKeydown = function(evt) {
```
jnatalzia: changing this to an arrow func gets rid of any need to cache `this.close` above. given it's the only instance of this, that seems safe

src/js/view/controls/components/menu/menu.js
```diff
@@ -173,6 +176,51 @@ export default class Menu extends Events {
     }
     createTopbar() {
         const topbar = createElement(`<div class="jw-submenu-topbar"></div>`);
+        const itemsContainer = this.itemsContainer.el;
+        const settingsMenu = this.mainMenu;
+        const categoryButton = this.categoryButton;
+        this.topbarUI = new UI(topbar).on('keydown', (evt) => {
```
jnatalzia: should we also `stopPropagation` here to avoid any conflicts with the internal menu listeners for focus or is that not needed?

src/js/view/controls/components/menu/menu.js
```diff
@@ -294,51 +342,56 @@ export default class Menu extends Events {
...
             return;
         }
-        backButtonTarget = null;
-        let focusEl;
         if (this.isSubmenu) {
+            const sourceEvent = evt.sourceEvent || event;
+            const firstItem = this.topbar ? this.topbar.firstChild : this.items[0].el;
+            const lastItem = this.items[this.items.length - 1].el;
+            const isKeydown = sourceEvent.type === 'keydown';
+            const key = isKeydown && sourceEvent.key.replace(/(Arrow|ape)/, '');
```
jnatalzia: this logic feels very similar to the key normalization done elsewhere. can we centralize somehow?

src/js/view/controls/components/menu/menu.js
```diff
@@ -294,51 +342,56 @@ export default class Menu extends Events {
...
+            const itemTarget = isKeydown && key === 'Up' ? lastItem : firstItem;
```
jnatalzia: can't we check `key === 'Up` here instead since we're already doing some conditional blocking on the `key` var

src/js/view/controls/components/menu/menu.js
```diff
@@ -294,51 +342,56 @@ export default class Menu extends Events {
...
+            if (this.visible && !this.openMenus.length) {
+                if (this.items.length && isKeydown) {
```
jnatalzia: what does this logic enable within the menu?

src/js/view/controls/components/menu/menu.js
```diff
@@ -294,51 +342,56 @@ export default class Menu extends Events {
...
+                    itemTarget.focus();
+                }
+                return;
+            }
             const { mainMenu, parentMenu, categoryButton } = this;
-            if (parentMenu.openMenus.length) {
-                parentMenu.closeChildren();
+            parentMenu.openMenus.push(this.name);
```
jnatalzia: if you push this after do you still need the conditional or do you need `closeChildren` to take the newest pushed menu into account here?

jnatalzia: yeah seems like you could also avoid this new parameter by changing the order of these two lines

src/js/view/controls/components/menu/menu.js
```diff
@@ -490,7 +535,8 @@ const addGlobalMenuKeyListener = (settingsMenu) => {
                 break;
             case 'Up':
             case 'Down':
-                openMenu(key);
+            case 'Enter':
+                menu().open(evt);
```
jnatalzia: is there a chance for this function to not return what is expected? potential breakage here if so

src/js/view/controls/components/menu/menu.js
zetagame: Good call

src/js/view/controls/components/menu/menu.js
zetagame: Nope, but I should rename that function to make it clear, so good comment either way :)

src/js/view/controls/components/menu/menu.js
zetagame: I can make a util for normalizeKey that does the replace. Alternatively, can switch to keycodes (I usually prefer keyCodes but the old code uses this form of normalized key everywhere, and it does have the advantage of people not having to look them up so I opted to keep it).

src/js/view/controls/components/menu/menu.js
zetagame: #1 comment

src/js/view/controls/components/menu/menu.js
zetagame: A different focus target depending on if the menu is already visible and it's clicked again.

src/js/view/controls/components/menu/menu.js
zetagame: This pattern fixes an issue. Basically, when you close all of a menu's child menus (like the settings menu) it automatically closes that menu. This is to prevent that reflow, by making sure there's always an open menu in the array when switching from one top level to another.

src/js/view/controls/components/menu/menu.js
jnatalzia: ah yeah, if we had more time the controller idea you had would be killer here, this logic can get v tough to follow with all thee branchiing paths 👍

src/js/view/controls/components/menu/menu.js
jnatalzia: oh neat, good stuff

PR #3548, opened 2019-12-16
[JW8-10801][JW8-10800][JW8-10798] Menus: Edge Case Rendering Issues

zetagame:
### This PR will...
Fix the following issues:
- Playback rates menu doesn't always render when available.
- Quality menu re-renders when setting quality on a track with multiple quality src mp4s.
- Quality appears to not set (Labeling issue, active item looks inactive, so when it's clicked nothing happens)

In addition, several other improvements were made:
- Fixed cases that could result in orphan nodes
- Ensured all event listeners destroyed on destroy
- Re-commented the model setup in settings-menu.js with the original annotations for clarity.
- Fixes toggling issues when non default menu open.
- Fixes tooltip suppress error on mobile

To fix the aforementioned issues and implement improvements.

#### Addresses Issue(s):

JW8-10801
JW8-10800
JW8-10798

### Checklist

### DIFF
--- a/src/js/view/controls/components/menu/menu.js
+++ b/src/js/view/controls/components/menu/menu.js
@@ -2,7 +2,6 @@ import UI from 'utils/ui';
 import Events from 'utils/backbone.events';
 import { 
     createElement, 
-    emptyElement, 
     prependChild, 
     nextSibling, 
     previousSibling
@@ -37,7 +36,9 @@ export default class Menu extends Events {
         this.categoryButton = null;
         this.closeButton = (this.parentMenu && this.parentMenu.closeButton) || this.createCloseButton(_localization);
         if (this.isSubmenu) {
-            this.categoryButton = this.parentMenu.categoryButton || this.createCategoryButton(_localization);
+            if (this.parentMenu.name === this.mainMenu.name) {
+                this.categoryButton = this.createCategoryButton(_localization);
+            }
             if (this.parentMenu.parentMenu && !this.mainMenu.backButton) {
                 this.mainMenu.backButton = this.createBackButton(_localization);
             }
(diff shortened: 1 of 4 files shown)


src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -27,6 +27,7 @@ const SettingsMenu = (api, model, controlbar, localization) => {
             controlBarButton.show();
         }
     };
+    let levelChangedManually = false;
```
jnatalzia: is this leftover code? i don't see it in use beyond being set tot `true/false`

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -54,17 +69,52 @@ const SettingsMenu = (api, model, controlbar, localization) => {
...
+    }, settingsMenu);
+    model.on('change:currentAudioTrack', (changedModel, currentAudioTrack) => {
+        onMenuItemSelected(settingsMenu.children.audioTracks, currentAudioTrack);
+    }, settingsMenu);
+
+    // Captions
+    model.on('change:playlistItem', () => {
+        // captions.js silently clears captions when the playlist item changes. The reason it silently clear captions
+        // instead of dispatching an event is because we don't want to emit 'captionsList' if the new list is empty.
+        settingsMenu.removeMenu('captions');
```
jnatalzia: so if there are new captions, the captions code adds them i assume?

src/js/view/controls/components/menu/menu.js
```diff
@@ -447,7 +450,10 @@ export default class Menu extends Events {
             }
         });
     }
-    toggle(evt) {
+    toggle(evt, isDefault) {
```
jnatalzia: can i get an overview on what this `isDefault` is doing?

zetagame: It's the param for if it's opened via the default action (clicking the settings icon). Passing this through fixes a toggle issue

src/js/view/controls/components/menu/settings-menu.js
zetagame: Yep! Same pattern we used pre-refactor for that.

src/js/view/controls/components/menu/settings-menu.js
zetagame: When you change the level manually, the api is set to dispatch a 'levels' event, which in this case will cause an unnecessary re-render of the quality menu (items didn't change, simply selected one, so better not to re-render). Rob walked me through why that event dispatch was necessary, so this was my way of preventing the additional reflow when that happens.

src/js/view/controls/components/menu/menu.js
```diff
@@ -470,11 +476,17 @@ export default class Menu extends Events {
                 this.categoryButton.ui.destroy();
             }
             if (this.itemsContainer) {
+                emptyElement(this.itemsContainer.el);
                 this.itemsContainer.destroy();
             }
             if (this.topbarUI) {
                 this.topbarUI.destroy();
             } 
+            if (this.items) {
```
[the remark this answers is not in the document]
zetagame: Destroy just removes event listeners/references on an instance of ui.js. It doesn't remove the listeners on the containing element. Order doesn't really matter here, we just have to do both.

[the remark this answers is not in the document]
robwalch: Not sure you need the `if` here unless you set `this.items` to `null` in this block.

It looks like `this.items` is always an array, so if you want to be safe empty (`items.length = 0`) or reset (`items = []`) here. 

Might want to sub-function `this.items.forEach(item => item.destroy())` too since you do it in two places.

src/js/view/controls/components/menu/settings-menu.js
robwalch: Only with the html5 provider and multiple mo4 sources which is a really small percentages of embeds. Any provider could dispatch "levels" for it's own reasons. I don't think fitting this flag to that one edge case makes sense. This optimization for such a small usecase (mp4 levels) is not worth the extra code. It would be better just always redraw the menu.

src/js/view/controls/components/menu/menu.js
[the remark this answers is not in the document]
zetagame: Good call, I'll remove this/make it more DRY

src/js/view/controls/components/menu/settings-menu.js
zetagame: This was the least lines of code I could come up with to achieve the desired result. Otherwise we run into this issue:

I click the item -> It sets the level -> Before anything else executes, the 'levels' event is dispatched -> The menu is recreated and the items are replaced with the new items -> the old items function continues, and garbage isn't collected until the function is finished -> any item transformations are done on the orphaned nodes before they're garbage collected

Now I can add state to the items to fix this, but this was the simplest way I came up with to prevent this in an OOP pattern.

zetagame: Actually, just came up with something. I'll try it out and report back.

zetagame:
test this please

src/js/view/controls/components/menu/menu.js
```diff
@@ -53,58 +54,49 @@ export default class Menu extends Events {
...
+
             const { sourceEvent, target } = evt;
-            const isFirstChild = itemsContainerElement.firstChild === target;
-            const isLastChild = itemsContainerElement.lastChild === target;
             const topbar = getTopbar();
-            const nextItem = closeButtonElement || nextSibling(topbarTarget);
-            const prevItem = backButtonElement || previousSibling(topbarTarget);
-            const nextSubItem = nextSibling(evt.target);
-            const prevSubItem = previousSibling(evt.target);
+            const settingsElement = document.getElementsByClassName('jw-icon-settings')[0];
```
mtoczko: Will it work if we have 2 players on the site?

zetagame: Not always, but this was ported directly from the original implementation, so any bugs surfaced from this already exist in prod. I agree there should be some form of linking of the toggle button and the main menu, but that's beyond the scope of the refactor.

mtoczko: I tested and after closing the menu in #2 player focus is on #1 player.
#2 Player
Open menu > keyDown > ESC 
The problem only occurs when I use the up or down key.

PR #3550, opened 2019-12-16
[JW8-10797] Fix liveSyncDuration not present on each playlist item

mamaddox:
### This PR will...
Fix the issue that the liveSyncDuration attribute was not appearing on each playlist item

bugfix

#### Addresses Issue(s):

JW8-10797

### Checklist

### DIFF
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -36,6 +36,10 @@ class Model extends SimpleModel {
         config.instreamMode = !!config.instream;
         delete config.instream;
         delete config.mediaModel;
+
+        Object.keys(config.setupConfig).forEach(key => {
+            config.setupConfig[key] = config[key];
+        });
         return config;
     }
 
--- a/src/js/controller/model.js
+++ b/src/js/controller/model.js
@@ -37,9 +37,11 @@ class Model extends SimpleModel {
         delete config.instream;
         delete config.mediaModel;
 
-        Object.keys(config.setupConfig).forEach(key => {
-            config.setupConfig[key] = config[key];
-        });
+        if (config.setupConfig) {
+            Object.keys(config.setupConfig).forEach(key => {
+                config.setupConfig[key] = config[key];
+            });
+        }
         return config;
     }
 


src/js/controller/model.js
```diff
@@ -36,6 +36,12 @@ class Model extends SimpleModel {
         config.instreamMode = !!config.instream;
         delete config.instream;
         delete config.mediaModel;
+
+        if (config.setupConfig) {
+            Object.keys(config.setupConfig).forEach(key => {
+                config.setupConfig[key] = config[key];
+            });
+        }
```
mamaddox: The values in setupConfig were out of sync with the actual configuration values making it seem that certain attributes were missing. This just makes sure that both are in sync.

zetagame: Not sure if we want to override setup config in this way, as jwplayer().getConfig().setupConfig is used for getting the config of setup block. If any of those values are re-written in config before this step (like by an A/B test), it may break that functionality. @robwalch, thoughts?

jnatalzia: yeah we'll want to avoid changing `setupConfig` in this way. `setupConfig` should contain the _original_ config used to instantiate the player, it should not update as things change. 

are you trying to take the top level `liveSyncDuration` and put it on each individual item? if that's the case those items should not be coming from `setupConfig`.

mamaddox: That makes sense.
I did notice that this got mutated, which was why some attributes were only present on some playlist items. It seemed to be in a half updated state, so I assumed that it was supposed to be updated and should mirror the updated config (I shouldn't have made that assumption).
It looks like we just need to deep copy the original setup then so that we can ensure it doesn't get changed.

robwalch:
https://jwplayer.atlassian.net/browse/JW8-10797 is invalid

src/js/controller/model.js
robwalch: We don't need a deep copy. TEs should not be testing that `liveSyncDuration` gets added anywhere. They should be testing the behavior of DVR streams when this is added to the config and when it is not.

mamaddox: The only follow up question I have then (outside of this scope) is:
Is it fine that the setupConfig changes? Because calling jwplayer().getConfig().setupConfig will not give you the original config, but will give you a semi-updated one.

robwalch: @mamaddox for the mutation issue, you might want to look at where Playlist and Source objects are created and if they are or are not copied by reference from the config.

item.js creates a new playlist item object for each item, and source.js does the same for each source. So if these are mutated it's because we're performing changes before calling `normalizePlaylistItem(_model, new Item(item), item.feedData || {});` on each item in the playlist passed to `player.setup()`.

robwalch: > Is it fine that the setupConfig changes?

See above. Changes when and how? We do shallow copies of the config into the player model. `setupConfig` should be a copy by reference of the original config, not a deep copy. If it's mutated by us or the developer it should reflect that. If anything we should be careful not to be the ones that modify it.

mamaddox: Hmmm. Okay. I'm not sure when or how, but I guess I was just a bit confused because it does sometimes contain new attributes that were not specified in the originalConfig. I think this was why this bug was identified. But it does look like playlist, item, and source never modify the config and only copy/reference it, so idk.

robwalch: >  it does sometimes contain new attributes

When? Please file an issue if you can provide steps to reproduce.

mamaddox: Will do 👍

PR #3553, opened 2019-12-18
Simplify Menu.js, remove redundant expressions.

zetagame:


src/js/view/controls/components/menu/menu.js
```diff
@@ -465,6 +446,15 @@ export default class Menu extends Events {
             this.close();
         }
     }
+    destroyItems() {
+        if (this.name === 'captionsSettings') {
+            debugger;
```
jnatalzia: need to remove this before things will build

src/js/view/controls/components/menu/menu.js
```diff
@@ -270,7 +251,9 @@ export default class Menu extends Events {
                     return;
                 }
                 action(argument || index);
-                if (menuItem.deactivate) {  
+                // Re-assignment from index prevents side effect if action re-renders menu
+                const targetItem = this.items[index];
+                if (targetItem.deactivate) {
```
jnatalzia: if this is the case, shouldn't we also replace `menuItem.active` above?

also i'm not clear on why a re-render would break these variables - it feels like the problem lies elsewhere

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -132,6 +126,7 @@ const SettingsMenu = (api, model, controlbar, localization) => {
             captionsMenu.topbar = captionsMenu.topbar || captionsMenu.createTopbar();
             const captionsSettingsMenu = new Menu('captionsSettings', captionsMenu, localization);
             captionsSettingsMenu.title = 'Subtitle Settings';
+            window.captionsSettingsMenu = captionsSettingsMenu;
```
jnatalzia: remove thiis

src/js/view/controls/components/menu/menu.js
zetagame: I click the item -> It sets the level -> Before anything else executes, the 'levels' event is dispatched -> The menu is recreated and the items are replaced with the new items -> the old items function continues, and garbage isn't collected until the function is finished -> any item transformations are done on the orphaned nodes before they're garbage collected

zetagame: That's why it's assigned after action()

zetagame: Oh, just had an idea, one sec.

PR #3555, opened 2019-12-20
[JW8-10816] Reset button re-render issue fix

zetagame:
### This PR will...
Resolve an exception caused by an issue with the captions settings menu reset button.

This exception would be triggered whenever reset was used after the menu was destroyed, and would block several other functionalities as a result.

#### Addresses Issue(s):

JW8-###

### Checklist

### DIFF
--- a/src/js/view/controls/components/menu/settings-menu.js
+++ b/src/js/view/controls/components/menu/settings-menu.js
@@ -1,5 +1,5 @@
 import Menu from 'view/controls/components/menu/menu';
-import { MenuItem, RadioMenuItem } from 'view/controls/components/menu/menu-item';
+import { MenuItem } from 'view/controls/components/menu/menu-item';
 import { itemMenuTemplate } from 'view/controls/templates/menu/menu-item';
 import { _defaults as CaptionsDefaults } from 'view/captionsrenderer';
 import { captionStyleItems } from './utils';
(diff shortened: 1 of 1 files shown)


src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -138,13 +138,13 @@ const SettingsMenu = (api, model, controlbar, localization) => {
...
                 model.set('captions', newStyles);
             };
-            const resetItem = new RadioMenuItem('Reset', () => {
-                model.set('captions', CaptionsDefaults);
-                renderCaptionsSettings();
-            });
-            resetItem.el.classList.add('jw-settings-reset');
             const persistedOptions = model.get('captions');
             const renderCaptionsSettings = () => {
+                const resetItem = new MenuItem('Reset', () => {
```
jnatalzia: can you explain the difference between `RadioMenuItem` and `MenuItem` and how that helps with this exception?

also are we sure that on render, the previous `resetItem` is garbage collected and we aren't creating a memory leak here?

zetagame: Yep, it is because it's added when we call setItem, so it's destroyed whenever it should be. RadioMenuItems will also deactivate any active menu items on activation, as well as add an attribute. This isn't a radio button option, so these additional methods were unnecessary and did nothing.

PR #3563, opened 2019-12-26
Focus the container explicitly on display play

jnatalzia:
### This PR will...
Change behavior on display (non-play button) click to focus the player element explicitly.

In firefox, the default behavior focuses the `video` tag which stops any aria attributes from being read and as a result has a degraded experience for screen reads

Nope!

Nah

#### Addresses Issue(s):

JW8-5617

### Checklist

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -448,6 +448,8 @@ function View(_api, _model) {
         clickHandler.on({
             click: () => {
                 _this.trigger(DISPLAY_CLICK);
+                // Ensures that Firefox focuses the container not the video tag for aria compatibility
+                _getCurrentElement().focus();
 
                 if (_controls) {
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -448,6 +448,8 @@ function View(_api, _model) {
         clickHandler.on({
             click: () => {
                 _this.trigger(DISPLAY_CLICK);
+                // Ensures that Firefox focuses the container not the video tag for aria compatibility
+                _getCurrentElement().focus();
```
zetagame: Does this still work when settingsMenuVisible() returns true?

jnatalzia: Great question! I opened the settings menu a few times, clicked the background and it does in fact keep the player container focused, should be all set there

PR #3564, opened 2019-12-27
[JW8-10824][JW8-10822] Menus: Fix gear icon visibility edge-case, remove caption settings menu when casting

zetagame:
### This PR will...
Fix a bug where the gear icon is visible when playBackRateControls is set to true on a live stream.

Also makes the following improvements:
- Separates controlbar logic from settings menu.
- Consolidates controlbar button updating based on menu availability to a single function.
- Removes listeners from settings menu when destroyed via model.off
- Removes the redundancy from menuVisibility listener pattern in controls.js, as per @robwalch's prior suggestion.
- Removes Subtitle Settings when casting.

Fixes a bug, makes a memory optimization.

#### Addresses Issue(s):

JW8-10824
JW8-10822

### Checklist

### DIFF
--- a/src/js/view/controls/components/menu/settings-menu.js
+++ b/src/js/view/controls/components/menu/settings-menu.js
@@ -4,7 +4,7 @@ import { itemMenuTemplate } from 'view/controls/templates/menu/menu-item';
 import { _defaults as CaptionsDefaults } from 'view/captionsrenderer';
 import { captionStyleItems } from './utils';
 
-const SettingsMenu = (api, model, controlbar, localization) => {
+const SettingsMenu = (api, model, localization) => {
     const settingsMenu = new Menu('settings', null, localization);
     const changeMenuItems = (menuName, items, onItemSelect, defaultItemIndex, itemOptions) => {
         if (!items || items.length <= 1) {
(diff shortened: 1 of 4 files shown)


src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -52,7 +42,7 @@ const SettingsMenu = (api, model, controlbar, localization) => {
     // Quality Tracks
     model.change('levels', (changedModel, levels) => {
         setLevelsMenu(levels);
-    }, settingsMenu);
+    });
```
robwalch: It seems like all of these model listeners are leaky. There is nothing removing these listeners when the menu is removed and destroyed - is there? So the references to the menus live on in the model listeners and when new menus are created new listeners are added.

The way to test this is set up a player with a playlist that has items with different menus (one playlist item has levels, one does not.) Then go through the playlist a few times to see if this model's listeners increase in number over time or level out.

robwalch: Usually we would pass the model into the menu instance, and have it register listeners with a context of `this`. Then in `menu.destroy` all listeners bound to the instance are removed:

```
class ViewItem {

  constructor(model) {
     this.model = model;
     model.on('change:x', this.onChangeX, this);
     model.on('change:y', this.onChangeY, this);
  }

  destroy () {
    if (this.model) {
       this.model.off(null, null, this);
   }
}
```

zetagame: Couldn't agree more, here's my thoughts on that: https://github.com/jwplayer/jwplayer/pull/3565

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -256,6 +251,13 @@ const SettingsMenu = (api, model, controlbar, localization) => {
         setPlaybackRatesMenu(model.get('playbackRates'));
     }, settingsMenu);
 
+    const destroy = settingsMenu.destroy;
```
zetagame: Temporary performance fix until 8.12.2 changes go in

PR #3570, opened 2020-01-08
Only disable track on detach for HTML with sideload

jnatalzia:
### This PR will...
Update our html 5 provider so that when the media is detached it only disables the running track if it has been sideloaded.

There is an issue with safari that is potentially related to this [webkit bug report](https://bugs.webkit.org/show_bug.cgi?id=184443) in which after the media is detached and reattached the disabled track (despite being reenabled) never shows in the viewport.

It appears to be a Safari issue but this change works around it

Nah

None 

#### Addresses Issue(s):

JW8-10239

### Checklist

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -265,8 +265,12 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
             clearTimeouts();
             // Stop listening to track changes so disabling the current track doesn't update the model
             this.removeTracksListener(_videotag.textTracks, 'change', this.textTrackChangeHandler);
-            // Prevent tracks from showing during ad playback
-            this.disableTextTrack();
+
+            const currTrack = this.getCurrentTextTrack();
+            // Prevent sideloaded tracks from showing during ad playback
+            if (currTrack && currTrack.sideloaded) {
+                this.disableTextTrack();
+            }
         },
         attachMedia() {
             VideoAttached.attachMedia.call(_this);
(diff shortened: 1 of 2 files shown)


src/js/providers/tracks-mixin.js
```diff
@@ -423,6 +425,10 @@ function disableTextTrack() {
     }
 }
 
+function getCurrentTextTrack() {
+    return this._textTracks && this._textTracks[this._currentTextTrackIndex];
```
robwalch: `this.getCurrentTextTrack()` should replace `this._textTracks[this._currentTextTrackIndex]` everywhere else in this file.

src/js/providers/html5.js
```diff
@@ -265,8 +265,12 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
             clearTimeouts();
             // Stop listening to track changes so disabling the current track doesn't update the model
             this.removeTracksListener(_videotag.textTracks, 'change', this.textTrackChangeHandler);
-            // Prevent tracks from showing during ad playback
-            this.disableTextTrack();
+
+            const currTrack = this.getCurrentTextTrack();
+            // Prevent sideloaded tracks from showing during ad playback
+            if (currTrack && currTrack.sideloaded) {
+                this.disableTextTrack();
```
robwalch: This assumes that:
1. embedded captions will be removed automatically when the current video src is unloaded
2. that detachMedia does not require any active track to be disabled for other reasons than ads playback (airplay?)

jnatalzia: Updated to ensure that Safari is the only affected browser.

Tested in Safari (HLS - sideloaded and embedded)
Firefox (MP4 - Sideloaded)
Chrome (MP4 - Sideloaded)

Re : airplay, I couldn't get any midrolls to play in airplay mode in latest safari on release, not sure if there's something else to do here but does seem like these executioin paths are even currently hit

src/js/providers/tracks-mixin.js
jnatalzia: great call

PR #3571, opened 2020-01-08
[JW8-10764] Fix long captions not wrapping

mamaddox:
### This PR will...
Wrap captions that extend longer than the width of the player and correctly position them

Bugfix

No

Negative

#### Addresses Issue(s):

[JW8-10764](https://jwplayer.atlassian.net/browse/JW8-10764)

### Checklist

### DIFF
--- a/src/js/polyfills/webvtt.js
+++ b/src/js/polyfills/webvtt.js
@@ -420,7 +420,7 @@ function CueStyleBox(window, cue) {
         top: 0,
         bottom: 0,
         display: 'inline',
-        'white-space': 'pre',
+        whiteSpace: 'pre',
         writingMode,
         unicodeBidi: 'plaintext',
     };
@@ -502,11 +502,17 @@ function CueStyleBox(window, cue) {
         this.applyStyles({
             top: this.formatStyle(box.top, 'px'),
             bottom: this.formatStyle(box.bottom, 'px'),
-            left: this.formatStyle(box.left, 'px'),
             paddingRight: this.formatStyle(box.right, 'px'),
             height: this.formatStyle(box.height, 'px'),
             transform: ''
         });
+
+        // Apply the left styling if the cue has not been wrapped.
+        if (this.cueDiv.style['white-space'] === 'pre') {
+            this.applyStyles({
+                left: this.formatStyle(box.left, 'px'),
+            });
+        }
     };
 
 }
@@ -526,7 +532,7 @@ function BoxPosition(obj) {
     let width;
     let top;
 
-    const { div } = obj;
+    const { div, cueDiv } = obj;
     if (div) {
         height = div.offsetHeight;
         width = div.offsetWidth;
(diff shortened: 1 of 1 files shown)


src/js/polyfills/webvtt.js
```diff
@@ -551,6 +557,11 @@ function BoxPosition(obj) {
     this.width = obj.width || width;
     this.lineHeight = lh !== undefined ? lh : obj.lineHeight;
 
+    if (cueDiv && div && this.width > div.parentElement.offsetWidth) {
+        cueDiv.style.whiteSpace = 'pre-wrap';
+        div.style.left = '0px';
```
mamaddox: Some strange behavior was happening where the caption would get positioned, while extending off screen.
Then the wrap would occur, but the left style would get updated based on the first position, so the caption would still be off screen, but correctly wrapped.
Setting this left style here moved the caption to the correct spot.

src/js/polyfills/webvtt.js
```diff
@@ -502,11 +502,17 @@ function CueStyleBox(window, cue) {
...
             top: this.formatStyle(box.top, 'px'),
             bottom: this.formatStyle(box.bottom, 'px'),
-            left: this.formatStyle(box.left, 'px'),
             paddingRight: this.formatStyle(box.right, 'px'),
             height: this.formatStyle(box.height, 'px'),
             transform: ''
         });
+
+        // Apply the left styling if the cue has not been wrapped.
+        if (this.cueDiv.style['white-space'] === 'pre') {
```
zetagame: I think this is supposed to be this.cueDiv.style.whiteSpace now. Might want to see if there's any other references to the old property name hanging around as well.

zetagame: I think this is supposed to be this.cueDiv.style.whiteSpace now. Might want to see if there's any other references to the old property name hanging around as well.

mamaddox: Good catch!

src/js/polyfills/webvtt.js
```diff
@@ -551,6 +557,11 @@ function BoxPosition(obj) {
     this.width = obj.width || width;
     this.lineHeight = lh !== undefined ? lh : obj.lineHeight;
 
+    if (cueDiv && div && this.width > div.parentElement.offsetWidth) {
```
jnatalzia: since we are wrapping the text, shouldn't we also set the width to the parent width?

src/js/polyfills/webvtt.js
jnatalzia: iirc, `boxposition` is created when we are attempting to find the proper placement for the captions box which is after the `CueStyleBox`. i would like to find a solution where we patch it in CueStyleBox above and then box positions continue to just work ™️  when given any shape of container. 

i know we've gone a bit in circles on this one so i could be misunderstanding the root of the problem but have we tried pulling this logic into the stylle box constructor and seeing what positioning is chosen? is it the same weird behavior we were seeing with the smaller box and/or mispositioning?

mamaddox: The only issue we could have with trying to patch it in the CueStyleBox is that the width of the elements are unknown at that point, since they haven't been appended to the page yet. So either the wrapping needs to be determined by something other than the width, or the logic can be performed elsewhere such as processCues?

src/js/polyfills/webvtt.js
mamaddox: moved this logic out of BoxPosition

PR #3573, opened 2020-01-09
Remove the focus ring that exists on the video element in android chrome

jnatalzia:
### This PR will...
Add additional styling rules to the video tag which disable an orange ring around that tag.

When you enter fullscreen in android chrome the video tag focuses and creates an orange focus ring

None I can think of

Nah 

#### Addresses Issue(s):

JW8-10814

### Checklist

### DIFF
--- a/src/css/jwplayer/imports/video.less
+++ b/src/css/jwplayer/imports/video.less
@@ -9,6 +9,9 @@
         height: 100%;
         margin: auto;
         background: transparent;
+        -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
+        -webkit-focus-ring-color: rgba(0, 0, 0, 0);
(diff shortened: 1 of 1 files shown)


src/css/jwplayer/imports/video.less
```diff
@@ -9,6 +9,9 @@
         height: 100%;
         margin: auto;
         background: transparent;
+        -webkit-tap-highlight-color: rgba(0, 0, 0, 0);
+        -webkit-focus-ring-color: rgba(0, 0, 0, 0);
```
zetagame: `-webkit-focus-ring-color` Can be removed. It isn't a CSS property, but rather the macOS/Chrome vendor prefix for the default color's value.

PR #3575, opened 2020-01-13
[JW8-10804] Accessibility Enhancements

zetagame:
### This PR will...
-Implement localized caption settings menu names
-Re-implement localized names for category buttons
-Correct menu button aria attributes
-Remove potential duplicate settings button in captions menu
-Fixes Visual Styles on Selected AudioTrack
-Updates tests

Improves accessibility and removes bugs.

Is it cool that I renamed `dropshadow` to `dropShadow`? Allows the potential for more programatic setting of style properties.

https://github.com/jwplayer/jwplayer-commercial/pull/7373

#### Addresses Issue(s):

JW8-10804
JW8-10848
JW8-10849

### Checklist


zetagame:
test this please

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -79,18 +81,26 @@ class SettingsMenu extends Menu {
     onAudioTracks(model, audioTracks) {
         this.setupMenu(
             'audioTracks', 
+            this.localization.audioTracks,
             audioTracks, 
             (index) => this.api.setCurrentAudioTrack(index), 
             model.get('currentAudioTrack')
         );
     }
+    onAudioTrackIndex(model, trackIndex) {
```
jnatalzia: new functionality?

test/unit/menu-test.js
```diff
@@ -161,22 +157,19 @@ describe('Menu', () => {
...
 
     it('Emits an event when a submenu is added', () => {
-        settingsMenu = new SettingsMenu(api, viewModel, controlbar, localization);
         settingsMenu.trigger = sinon.spy();
         viewModel.set('levels', [{ label: 'Auto' }, { label: '1080p' }]);
         expect(settingsMenu.trigger.calledWith('menuAppended')).to.equal(true);
     });
 
     it('Emits an event when a submenu is removed', () => {
-        settingsMenu = new SettingsMenu(api, viewModel, controlbar, localization);
```
jnatalzia: is there a danger in this test setup? reusing the same settings menu across all these tests could be problematic since it requires order to remain the same which might not always be the case. 

why did we remove the instantiations?

zetagame: It's already in the beforeEach block

src/js/view/controls/components/menu/settings-menu.js
zetagame: I noticed while working on this the default audio track isn't selected on load. Added a listener for when it gets set to decorate the proper item. It now follows the same pattern as the rest of the menus. Created a ticket for it here, adding it to the PR description shortly: https://jwplayer.atlassian.net/secure/RapidBoard.jspa?rapidView=306&modal=detail&selectedIssue=JW8-10849

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -79,18 +81,26 @@ class SettingsMenu extends Menu {
...
+        if (!this.children.audioTracks) {
+            return;
+        }
+        this.children.audioTracks.items[trackIndex].activate();
```
jnatalzia: are we sure this will always have a value? also how does the highlight currently work without this logic (when selected an item, i know this bug fix is for the first default selection). i worry if we're spreading our logic for highlighting throughout both modules even though this shells the call out to the submenu. some food for thought

zetagame: Ah, just saw this. It appears to always have a value. From ProgramController:
```
if (currentTrack > -1 && tracks.length > 0 && currentTrack < tracks.length) {
  mediaModel.set('currentAudioTrack', parseInt(currentTrack));
}
```

All listeners for external selection like `onAudioTrackIndex` are in settings-menu.js. There is a function that adds listeners to menu items for click, where it selects it and activates it if it isn't already active. Also the base activate function checks if active before activating as well.

jnatalzia: i guess my question is more does this logic make more sense in the child menu? is that possible to do or does the generic factory make that more trouble than it's worth? just feels a little strange to have a listener in the parent that then shells the calls out to the child

test/unit/menu-test.js
jnatalzia: ahhh nice!

src/js/view/controls/components/menu/settings-menu.js
zetagame: Having all of the data-driven aspects of state under a single container makes it trivial to remove event listeners on destroy, and to debug if there are issues. Also, since the event lifecycle isn't normalized between various events (for example change:captions doesn't fire if the new array is empty, other change events do). Settings menu is our main menu, so it makes sense to treat it as our state container, with the data returned from those model listeners as the source of truth.

PR #3580, opened 2020-01-17
[JW8-10779] A/B Test Zooming Thumbnails

mamaddox:
### This PR will...
Add a configuration to the player to create a zooming in and out effect for thumbnails.
A/B Test
No
[jwplayer-commercial #7382](https://github.com/jwplayer/jwplayer-commercial/pull/7382)
#### Addresses Issue(s):

[JW8-10779](https://jwplayer.atlassian.net/browse/JW8-10779)

### Checklist

### DIFF
--- a/src/js/view/preview.js
+++ b/src/js/view/preview.js
@@ -12,6 +12,11 @@ function validState(state) {
 Object.assign(Preview.prototype, {
     setup: function(element) {
         this.el = element;
+        this.hasZoomThumbnail = !this.model.get('autostart') && this.model.get('_abZoomThumbnail');
+        
+        if (this.hasZoomThumbnail) {
+            this.enableZoom();
+        }
     },
     setImage: function(img) {
         // Remove onload function from previous image
@@ -29,6 +34,25 @@ Object.assign(Preview.prototype, {
         style(this.el, {
             backgroundImage: backgroundImage
         });
+    },
+    enableZoom: function() {
+        if (this.model.get('state' !== 'idle')) {
+            return;
+        }
+        
+        setTimeout(() => {
+            const originX = Math.ceil(Math.random() * 100) + '%';
+            const originY = Math.ceil(Math.random() * 100) + '%';
+            
+            this.el.classList.add('jw-ab-zoom-thumbnail');
+            this.el.style.transformOrigin = originX + ' ' + originY;
+        }, 2000);
+        
+        this.model.once('change:state', this.destroy, this);
+        this.model.on('change:viewable', this.pauseZoom, this);
+    },
+    pauseZoom: function() {
+        this.el.style.animationPlayState = this.model.get('viewable') ? 'running' : 'paused';
     },
     resize: function(width, height, stretching) {
         if (stretching === 'uniform') {
@@ -63,6 +87,13 @@ Object.assign(Preview.prototype, {
     },
     element: function() {
         return this.el;
+    },
+    destroy: function() {
+        if (this.hasZoomThumbnail) {
+            this.el.classList.remove('jw-ab-zoom-thumbnail');
+            this.model.off('change:state', this.destroy, this);
+            this.model.off('change:viewable', this.pauseZoom, this);
+        }
     }
 });
 
--- a/src/js/view/preview.js
+++ b/src/js/view/preview.js
@@ -17,7 +17,7 @@ Object.assign(Preview.prototype, {
         if (this.hasZoomThumbnail) {
             this.zoomOriginX = Math.ceil(Math.random() * 100) + '%';
             this.zoomOriginY = Math.ceil(Math.random() * 100) + '%';
-            
+
             this.model.on('change:viewable', this.pauseZoomThumbnail, this);
             this.model.on('change:isFloating', this.enableZoomThumbnail, this);
         }
@@ -48,15 +48,18 @@ Object.assign(Preview.prototype, {
             return;
         }
         
+        clearTimeout(this.zoomThumbnailTimeout);
         this.zoomThumbnailTimeout = setTimeout(() => {
             this.el.classList.add('jw-ab-zoom-thumbnail');
             this.el.style.transformOrigin = this.zoomOriginX + ' ' + this.zoomOriginY;
         }, 2000);
     },
     pauseZoomThumbnail: function() {
+        clearTimeout(this.zoomThumbnailTimeout);
         this.el.style.animationPlayState = this.model.get('viewable') ? 'running' : 'paused';
     },
     removeZoomThumbnail: function() {
+        clearTimeout(this.zoomThumbnailTimeout);
         this.el.classList.remove('jw-ab-zoom-thumbnail');
     },
     resize: function(width, height, stretching) {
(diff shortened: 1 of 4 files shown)


src/js/view/preview.js
```diff
@@ -30,6 +35,25 @@ Object.assign(Preview.prototype, {
...
+        if (this.model.get('state' !== 'idle')) {
+            return;
+        }
+        
+        setTimeout(() => {
+            const originX = Math.ceil(Math.random() * 100) + '%';
+            const originY = Math.ceil(Math.random() * 100) + '%';
+            
+            this.el.classList.add('jw-ab-zoom-thumbnail');
+            this.el.style.transformOrigin = originX + ' ' + originY;
```
jnatalzia: So we’re changing origin every 2 s and zooming over 10, just want to be sure that’s expected (for the origin to change mid animation)

mamaddox: Hmmm. That shouldn't happen. The origin should only be set once, but there is a 2s delay before the animation begins.

jnatalzia: Ah! It's a timeout, not an interval - my bad.

jnatalzia: we should set the result of this timeout to a variable and `clearTimeout` in the `destroy` function to 4ensure this doesn't get run after destroy

src/js/view/preview.js
```diff
@@ -30,6 +35,25 @@ Object.assign(Preview.prototype, {
...
+        }, 2000);
+        
+        this.model.once('change:state', this.destroy, this);
```
jnatalzia: why do we destroy the preview on state change? i understand we're moving from idle to playing but unclear on why this wasn't needed before

mamaddox: I did it to ensure the animation stopped since the element remains and because there was no longer a need for the event listeners. However, I can also just remove the zoomThumbnail class.

src/js/view/preview.js
```diff
@@ -63,6 +87,17 @@ Object.assign(Preview.prototype, {
...
         return this.el;
+    },
+    destroy: function() {
+        if (this.hasZoomThumbnail) {
+            this.el.classList.remove('jw-ab-zoom-thumbnail');
+            this.model.off('change:state', this.destroy, this);
+            this.model.off('change:viewable', this.pauseZoom, this);
+        }
+
+        if (this.zoomThumbnailTimeout) {
```
jnatalzia: you can remove the `if` condition here, `clearTimeout` can safely handle a falsy timeout

src/js/view/preview.js
jnatalzia: hmmm yeah i might opt for that - when a use selects `jwplayer().stop()` it's going to put the player back into the idle state so we may want to keep it around for that?

thoughts @robwalch

mamaddox: I agree. Because they will also get removed when the preview is destroyed

src/js/view/preview.js
```diff
@@ -63,6 +87,17 @@ Object.assign(Preview.prototype, {
     },
     element: function() {
         return this.el;
+    },
+    destroy: function() {
+        if (this.hasZoomThumbnail) {
+            this.el.classList.remove('jw-ab-zoom-thumbnail');
+            this.model.off('change:state', this.destroy, this);
+            this.model.off('change:viewable', this.pauseZoom, this);
```
robwalch: This will remove all model listeners for this instance.
`this.model.off(null, null, this);`

src/js/view/preview.js
robwalch: Adding a destroy method is good, but I would prefer that the `once('change:state'` callback run something like `this.stopZoom` or `this. pauseZoom` and you just call that method in `destroy`.

And, since the state can go from IDLE to something else, and then back to IDLE, this could be an `on` handler that checks state. ⚠️  However! it is not efficient to handle all state changes in UI elements. The view is already doing that and debouncing things that would change on buffering, so call these methods from the view for optimal performance.  See  `function _stateUpdate(state) {` in view.js

src/js/view/preview.js
```diff
@@ -12,6 +12,11 @@ function validState(state) {
 Object.assign(Preview.prototype, {
     setup: function(element) {
         this.el = element;
+        this.hasZoomThumbnail = !this.model.get('autostart') && this.model.get('_abZoomThumbnail');
```
robwalch: `model.get('autostart')` doesn't really mean we will or won't autostart. I would ignore that and just let state changes and `setImage` (plus the ab flag) determine if we should animate-zoom the image or not.

src/js/view/view.js
```diff
@@ -804,6 +804,9 @@ function View(_api, _model) {
                 if (_captionsRenderer) {
                     _captionsRenderer.hide();
                 }
+                if (_preview && !_model.get('isFloating')) {
```
jnatalzia: youu're checking floating in this function as well - only need to do it in one spot - i'd suggest in `enableZoomThumbnail`

src/js/view/preview.js
```diff
@@ -29,6 +38,26 @@ Object.assign(Preview.prototype, {
...
+            return;
+        }
+        
+        this.zoomThumbnailTimeout = setTimeout(() => {
+            this.el.classList.add('jw-ab-zoom-thumbnail');
+            this.el.style.transformOrigin = this.zoomOriginX + ' ' + this.zoomOriginY;
+        }, 2000);
+    },
+    pauseZoomThumbnail: function() {
+        this.el.style.animationPlayState = this.model.get('viewable') ? 'running' : 'paused';
```
robwalch: `clearTimeout(this.zoomThumbnailTimeout);` here incase `pauseZoomThumbnail` runs before the `enableZoomThumbnail` timeout.

src/js/view/preview.js
```diff
@@ -29,6 +38,26 @@ Object.assign(Preview.prototype, {
...
+       
+        if (this.hasZoomThumbnail) {
+            this.enableZoomThumbnail();     
+        }
+    },
+    enableZoomThumbnail: function() {
+        if (this.model.get('isFloating')) {
+            return;
+        }
+
```
robwalch: You could also run `clearTimeout(this.zoomThumbnailTimeout);`  so there can only ever be one active `zoomThumbnailTimeout`.

src/js/view/preview.js
```diff
@@ -29,6 +38,26 @@ Object.assign(Preview.prototype, {
...
+        
+        this.zoomThumbnailTimeout = setTimeout(() => {
+            this.el.classList.add('jw-ab-zoom-thumbnail');
+            this.el.style.transformOrigin = this.zoomOriginX + ' ' + this.zoomOriginY;
+        }, 2000);
+    },
+    pauseZoomThumbnail: function() {
+        this.el.style.animationPlayState = this.model.get('viewable') ? 'running' : 'paused';
+    },
+    removeZoomThumbnail: function() {
```
robwalch: `clearTimeout(this.zoomThumbnailTimeout);`

src/js/view/preview.js
```diff
@@ -63,6 +92,14 @@ Object.assign(Preview.prototype, {
...
     element: function() {
         return this.el;
+    },
+    destroy: function() {
+        if (this.hasZoomThumbnail) {
+            this.removeZoomThumbnail();
+            this.model.off(null, null, this);
+        }
+
+        clearTimeout(this.zoomThumbnailTimeout);
```
robwalch: you can remove it here if we call `clearTimeout` in `removeZoomThumbnail`.

PR #3588, opened 2020-01-28
[JW8-10896] Reduce Player Asset Size

robwalch:
Change babel settings so that we use loose mode and enable treeshaking by disabling the default conversion of modules to commonjs.

Reference https://medium.com/@bluepnume/javascript-tree-shaking-like-a-pro-7bf96e139eb7

As a result the web player release assets will have a smaller footprint which is better for everybody. (Note these sizes are from the commercial project bin-release assets)

```
189K > 170K (-19K) jwplayer.controls.js
339K > 312K (-27K) jwplayer.core.controls.html5.js
310K > 283K (-27K) jwplayer.core.controls.js
346K > 319K (-27K) jwplayer.core.controls.polyfills.html5.js
317K > 290K (-27K) jwplayer.core.controls.polyfills.js
131K > 123K (- 8K) jwplayer.core.js
113K > 108K (- 5K) jwplayer.js
 36K >  36K (   0) provider.html5.js
```

Loose mode differences:

1. Uses `+` operator for string concatenation in transpiled string templates (smaller) 
```js
  var playerId = element.id || "player-".concat(uniqueId);
  // loose:
  var playerId = element.id || "player-" + uniqueId;
```
2. Uses `undefined` check for assigning argument defaults (smaller, and skips assignment when the argument is undefined ). A lot of times I use || as an optimization to avoid this boiler plate, so loose mode is much better.
```js
getSafeRegion: function getSafeRegion() {
      var excludeControlbar = arguments.length > 0 && arguments[0] !== undefined ? arguments[0] : true;
// loose:
getSafeRegion: function getSafeRegion(excludeControlbar) {
      if (excludeControlbar === void 0) {
        excludeControlbar = true;
```

3. Removes `_classCallCheck` from all modules with classes and their constructors (not needed and is a huge saving in size and execution of code that we do not need).  You can call Class() as a function without the new keyword in loose, but we’d never do that because the result would not be an instance and would not work.
```js
// Every module with a class (before):
function _classCallCheck(instance, Constructor) { if (!(instance instanceof Constructor)) { throw new TypeError("Cannot call a class as a function"); } }
function _defineProperties(target, props) { for (var i = 0; i < props.length; i++) { var descriptor = props[i]; descriptor.enumerable = descriptor.enumerable || false; descriptor.configurable = true; if ("value" in descriptor) descriptor.writable = true; Object.defineProperty(target, descriptor.key, descriptor); } }
function _createClass(Constructor, protoProps, staticProps) { if (protoProps) _defineProperties(Constructor.prototype, protoProps); if (staticProps) _defineProperties(Constructor, staticProps); return Constructor; }
// And if necessry: _possibleConstructorReturn, _assertThisInitialized, _get, _superPropBase, _getPrototypeOf, _inherits, _setPrototypeOf
  function ClassName() {
    _classCallCheck(this, ClassName);
...
  _createClass(ClassName, null, [{
    key: "methodName",
    value: function...
    
 // loose:
   function ClassName() {
...
var _proto = ClassName.prototype;
_proto.methodName = function ...
```

The overhead required to parse and define each class as well as create instances is really concerning. By removing this the player will be more performant. These settings are already being used in hls.js.

4. Use typeof operator instead of ES6 Symbol compatible polyfill in each module using typeof
```js
function _typeof(obj) { if (typeof Symbol === "function" && typeof Symbol.iterator === "symbol") { _typeof = function _typeof(obj) { return typeof obj; }; } else { _typeof = function _typeof(obj) { return obj && typeof Symbol === "function" && obj.constructor === Symbol && obj !== Symbol.prototype ? "symbol" : typeof obj; }; } return _typeof(obj); }
...
if (_typeof(pluginsConfig) !== 'object') {

// loose (the code is left as we wrote it):
if (typeof pluginsConfig !== 'object') {
```

https://github.com/jwplayer/jwplayer-commercial/pull/7395

#### Addresses Issue(s):
JW8-10896

### Checklist

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -77,7 +77,7 @@ const webpackConfig = {
                     options: {
                         babelrc: false,
                         presets: [
-                            '@babel/preset-env',
+                            ['@babel/preset-env', { loose: true, modules: false }],
                             '@babel/preset-typescript'
                         ],
                         plugins: [
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -120,8 +120,7 @@ module.exports = function(grunt) {
                 junitReporter: {
                     suite: '<%= grunt.task.current.target %>',
                     outputDir: 'reports/junit'
-                },
-                concurrency: 1
+                }
             },
             headless: {
                 browsers: ['ChromeHeadless']


robwalch:
Added @vseventer and @nebnitt because I think we should make the same babel preset changes to ads and analytics plugins to reduce the size of those assets as well.

webpack.config.js
```diff
@@ -75,7 +75,9 @@ const webpackConfig = {
                     loader: 'babel-loader',
                     options: {
                         babelrc: false,
-                        presets: ['@babel/preset-env'],
+                        presets: [
+                            ['@babel/preset-env', { loose: true, modules: false }]
```
jnatalzia: @mamaddox made a similar change i think with the typescript that caused some issues but i don't remember if it was these exact options.

jnatalzia: yeah i think the `class check` specifically caused issues with the way we bundle related but i'll let maria confirm

mamaddox: These were the options I removed. But, there were some tsconfig settings that were definitely causing some issues. I think the config settings were the root of the cause tbh, but I need to verify first.

robwalch: `modules: false` is probably safe. It's `loose` mode that really saves space by cutting out a boiler plate code.

The test failure we're seeing in commercial (not this project) is that the `get displayMode` override in `PlaylistController`  overrides the getter in `RecsController` but falls back to calling the getter on the base class with `super.displayMode` < that's where it breaks. Updating webpack and babel doesn't resolve it, but the fix is "super" simple:

https://github.com/jwplayer/jwplayer-commercial/pull/7395/commits/fd1734f268e99be3802bf400747d99dd49062877

I tried but could not reproduce the issue in Babel repl or find any open issues.
https://babeljs.io/en/repl#?browsers=&build=&builtIns=false&spec=false&loose=true&code_lz=MYGwhgzhAEBCkFNoG8BQ1oHMEBdo4QhwAoBKFdDaAJ1wFdqA7aARgG5KBfVb1USGAHkARgCtoCAB4FGAExjwISNBgD0q6IIBuCatQCWspNhwFq0MHOh0l0CHQAOuyifyES5FVX0AzaMR8wECVyWhwGZgAmDioaeiY7R10AOgIiGOhuXmAAe0YiaByxaABeaEYEAHdNMTIOXPyckARkkBzMYiLRVPdSDiA&debug=false&forceAllTransforms=false&shippedProposals=false&circleciRepo=&evaluate=true&fileSize=true&timeTravel=false&sourceType=module&lineWrap=true&presets=es2015%2Creact%2Cstage-2%2Cenv&prettier=false&targets=&version=7.8.3&externalPlugins=

nicolo-ribaudo: `modules: false` isn't probably needed, since it's already the default behavior when using Webpack.

Issue #3519, opened 2019-11-04
HTML5 Povider - live + pdt 


mtoczko:
### Expected Behavior

No errors in metadata handling (PROGRAM-DATE-TIME)

### Actual Behavior
Live stream + pdt (playlist duration 30s)
```
#EXT-X-VERSION:4
#EXT-X-TARGETDURATION:10
#EXT-X-MEDIA-SEQUENCE:173
#EXT-X-PROGRAM-DATE-TIME:2019-11-04T14:01:55.565+00:00
#EXTINF:10.0,
media-u0gia6ek1_b2228224_slpl_173.ts
#EXTINF:10.0,
media-u0gia6ek1_b2228224_slpl_174.ts
#EXTINF:10.0,
media-u0gia6ek1_b2228224_slpl_175.ts
```
**console**
```
TypeError: The provided value is non-finite

```
**function**
```
createCue: function(e, t, i) {
                var n = window.VTTCue || window.TextTrackCue,
                    o = Math.max(t || 0, e + .25);
                return new n(e, o, i)
```
```
e: 0
i: "{\"metadataType\":\"program-date-time\",\"programDateTime\":\"2019-11-04T14:55:46.345Z\",\"start\":0,\"end\":null}"
n: function()
o: Infinity
t: Infinity
```
When the playlist duration is longer than 30s, there is no error in the console.
```
#EXT-X-VERSION:4
#EXT-X-TARGETDURATION:10
#EXT-X-MEDIA-SEQUENCE:173
#EXT-X-PROGRAM-DATE-TIME:2019-11-04T14:01:55.565+00:00
#EXTINF:10.0,
media-u0gia6ek1_b2228224_slpl_173.ts
#EXTINF:10.0,
media-u0gia6ek1_b2228224_slpl_174.ts
#EXTINF:10.0,
media-u0gia6ek1_b2228224_slpl_175.ts
#EXTINF:10.0,
media-u0gia6ek1_b2228224_slpl_176.ts
```
**Proposed solution**
https://github.com/jwplayer/jwplayer/blob/4bf733f143537f0c6c3ff5f7fad34a572cded015/src/js/providers/tracks-mixin.js#L47

```
const cueEnd = Math.max( isFinite(end) ? end : 0, start + 0.25); 
```

### Environment

- *operating system | version* macOS 10.14.6
- *browser | version* Safari  13.0.1
- *JW Player version | url* 8.11.5

karimMourra:
[...]

PR #3592, opened 2020-02-04
[JW8-10829] Expose "metadataCueParsed" event in Safari for ID3 text track cues

robwalch:
### This PR will...
- [JW8-10829] Expose "metadataCueParsed" event in Safari for ID3 text track cues
- [JW8-10898] Add public `setStartDateTime` method to the html5 provider

API event parity with hls.js provider playback. When a new cue found, we trigger a "metadataCueParsed" event.

`setStartDateTime` is used in jwplayer-commercial to enable commercial features that depend on PROGRAM-DATE-TIME availability in Safari (https://github.com/jwplayer/jwplayer-commercial/pull/7411).

#### Addresses Issue(s):
JW8-10829
JW8-10898

### Checklist

### DIFF
--- a/src/js/providers/html5.js
+++ b/src/js/providers/html5.js
@@ -2,7 +2,7 @@ import { qualityLevel } from 'providers/data-normalizer';
 import { Browser, OS } from 'environment/environment';
 import { isAndroidHls } from 'providers/html5-android-hls';
 import {
-    STATE_IDLE, STATE_PLAYING, STATE_STALLED, MEDIA_META_CUE_PARSED, MEDIA_META, MEDIA_ERROR, WARNING,
+    STATE_IDLE, STATE_PLAYING, STATE_STALLED, MEDIA_META, MEDIA_ERROR, WARNING,
     MEDIA_VISUAL_QUALITY, MEDIA_TYPE, MEDIA_LEVELS, MEDIA_LEVEL_CHANGED, MEDIA_SEEK, NATIVE_FULLSCREEN, STATE_LOADING
 } from 'events/events';
 import VideoEvents from 'providers/video-listener-mixin';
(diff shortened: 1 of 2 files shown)


src/js/providers/html5.js
```diff
@@ -57,18 +57,22 @@ function _removeListeners(eventsHash, videoTag) {
...
-    this.currentTime = -1;
+    _this.currentTime = -1;
 
     // Attempt to reload video on error
-    this.retries = 0;
-    this.maxRetries = 3;
+    _this.retries = 0;
+    _this.maxRetries = 3;
+
+    _this.loadAndParseHlsMetadata = _playerConfig.loadAndParseHlsMetadata || true;
```
jnatalzia: wouldn't this always return true even if `loadAndParseHlsMetadata` is false?

src/js/providers/tracks-mixin.js
```diff
@@ -60,19 +63,20 @@ function setTextTracks(tracks) {
...
-        this._textTracks = this._textTracks.filter(function(track) {
+        this._textTracks = this._textTracks.filter((track) => {
             const trackId = track._id;
             if (this.renderNatively && trackId && trackId.indexOf('nativecaptions') === 0) {
                 delete this._tracksById[trackId];
                 return false;
             } else if (track.name && track.name.indexOf('Unknown') === 0) {
                 this._unknownCount++;
             }
+            if (trackId.indexOf('nativemetadata') === 0 && track.inBandMetadataTrackDispatchType === 'com.apple.streaming') {
```
jnatalzia: can i get a quick explanation of this?

src/js/providers/tracks-mixin.js
```diff
@@ -667,62 +688,127 @@ function _clearSideloadedTextTracks() {
 }
 
 function _cueChangeHandler(e) {
```
jnatalzia: i'm not 100% on what is happening here? is the additional logic exposing id3 information that previously we left up to hls.js?

src/js/providers/html5.js
robwalch: woops! Thanks.

src/js/providers/tracks-mixin.js
robwalch: Metadata TextTracks created by Safari have an `id` that starts with 'nativemetadata' (if they create a second track it's called 'nativemetadata1') and an `inBandMetadataTrackDispatchType` of 'com.apple.streaming'.

It was incorrect to assume there would only be one `nativemetadata` track, and that it would be the one added by Safari to handle ID3. We insert our own `nativemetadata` track before Safari, so the track it creates has an id of 'nativemetadata1'.

src/js/providers/tracks-mixin.js
robwalch: No. This is what triggers `"metadataCueParsed"` events when new cues are found in metadata track.cues, and `"meta"` events when new active cues in metadata track.activeCues. Previously we ignored `track.cues` and only fired meta events for active cues. A cue is active when the currentTime enters its range.

src/js/providers/html5.js
```diff
@@ -349,27 +352,31 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
...
-                delete metadata.metadataType;
-                _this.trigger(MEDIA_META_CUE_PARSED, { metadataType, metadata });
+                _this.setStartDateTime(startDateTime);
             }
         }
     }
 
+    _this.setStartDateTime = function(startDateTime) {
+        _this.startDateTime = startDateTime;
+        const programDateTime = new Date(startDateTime).toISOString();
```
jnatalzia: assuming the start date time contains timezone?

src/js/providers/tracks-mixin.js
jnatalzia: aha! makes sense 👍

mtoczko:
Hi @robwalch, 
Could you fix the bug in provider_html5? #3519

robwalch:
@mtoczko Make a comment in #3519 requesting to reopen the issue. I don't know the history or why it was closed. Provide a live stream and player config if you want the issue reproduced quickly by our team.

src/js/providers/html5.js
robwalch: @jnatalzia https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Date/toISOString

PR #3595, opened 2020-02-06
[JW8-10727] Async Playback Control API

robwalch:
### This PR will...
Implement the async playlist item API.

The goals of this API are to:
1. Specify a promise for each playlist item that allows
    - navigation to and playback of that item (automatic or manual) to be blocked indefinitely
    - replacement or mutation of that item (with the value returned by the promise)
    - navigation and playback of the item to be skipped (by rejecting the promise)
2. Maintain current plugin functionality by not triggering "playlistItem" until the playlist item promise is resolved (allows for ad schedule insertion, and proper analytics behavior). "ready" is also deferred when the first item is blocked. While this impacts setup time, it does not impact time to first frame. Async operations can be performed prior to setup to avoid this.

The current API methods being added are:

1. `setPlaylistItemCallback(callback)`
    - Supply one function that is run before each item which can return a promise for that item (item promise persists for the life of the playlist unless the callback is replaced.
    - There is only one `playlistItemCallback` stored in the model, so you could also specify this callback in the config at setup.
2. `removePlaylistItemCallback()` (same as `setPlaylistItemCallback(null)`)
3. `getPlaylistItemPromise(index)`
    - Returns the promise of the playlist item at `playlist[index]` or `"nextUp"` when `index = -1`

https://github.com/jwplayer/jwplayer-commercial/pull/7339
https://github.com/jwplayer/jwplayer-ads-vast/pull/637

#### Addresses Issue(s):
[JW8-10727](https://jwplayer.atlassian.net/browse/JW8-10727)

### Checklist


src/js/controller/controller.js
```diff
@@ -955,6 +955,14 @@ Object.assign(Controller.prototype, {
         this.getWidth = () => _model.get('containerWidth');
         this.getHeight = () => _model.get('containerHeight');
         this.getItemQoe = () => _model._qoeItem;
+
+        this.setItemCallback = function (callback) {
+            _programController.itemCallback = callback;
+        };
+        this.getItemPromise = function (index = (_model.get('item') + (_model.get('itemReady') ? 1 : 0))) {
```
jnatalzia: this is a style thing so no need to change but i think it's a bit tough to parse having the default be so complex but inlined here. just a thought!

src/js/controller/controller.js
```diff
@@ -1036,15 +1044,14 @@ Object.assign(Controller.prototype, {
...
-                const playlist = _model.get('playlist');
-                playlist[index] = item;
-
-                if (index === _model.get('item') && _model.get('state') === 'idle') {
-                    this.setItemIndex(index);
+            const newItem = _programController.setPlaylistItem(index, item);
+            if (newItem && index === _model.get('item') && _model.get('state') === 'idle') {
+                const itemPromise = this.itemPromises[index];
+                if (itemPromise) {
+                    itemPromise.reject(new Error('Item replaced'));
```
jnatalzia: nice 👍

src/js/controller/controller.js
```diff
@@ -1036,15 +1044,14 @@ Object.assign(Controller.prototype, {
...
-
-            if (item) {
-                const playlist = _model.get('playlist');
-                playlist[index] = item;
-
-                if (index === _model.get('item') && _model.get('state') === 'idle') {
-                    this.setItemIndex(index);
+            const newItem = _programController.setPlaylistItem(index, item);
+            if (newItem && index === _model.get('item') && _model.get('state') === 'idle') {
+                const itemPromise = this.itemPromises[index];
```
jnatalzia: it looks like `itemPromises` only exists on `programController` - is there some magic i'm missing that's making `this` share context here? otherwise i don't see how this would get populated

src/js/controller/next-item-promise.js
```diff
@@ -0,0 +1,60 @@
+
+export class ItemPromise {
+
+    constructor (index, model, api) {
+        this.index = index;
+        this.model = model;
+        this.api = api;
+        this.promise = new Promise((resolve, reject) => {
+            this.resolve = resolve;
```
jnatalzia: i _love_ this approach to encapsulating the promise without a single overcomplex function

src/js/controller/next-item-promise.js
```diff
@@ -0,0 +1,60 @@
...
+        return promise;
+    }
+
+    getItem(index) {
+        const { model } = this;
+        if (index === -1) {
+            return model.get('nextUp');
+        }
+        const playlist = model.get('playlist');
+        return playlist[index];
```
jnatalzia: i'm assuming we're confident that playlist will always exist otherwise we'll have bigger problems

src/js/controller/next-item-promise.js
```diff
@@ -0,0 +1,60 @@
...
+        this.async = null;
+    }
+
+    preload () {
+        return this.run();
+    }
+
+    setItem (index) {
+        if (index !== this.index) {
+            this.reject(new Error(`Item ${this.index} was skipped.`));
```
jnatalzia: this is probably a result of me not fully groking the execution pathing but is `setItem` in this promise controller another method to change the item vs an explicit call to `setPlaylistItem` on the global api?

src/js/program/ad-program-controller.js
```diff
@@ -18,7 +18,7 @@ export default class AdProgramController extends ProgramController {
         // (i.e. prerolls, midrolls, and postrolls must use the same tag)
         let mediaElement;
         if (this.backgroundLoading) {
-            // The media pool has reserves an element for ads to use. It is reserved on setup and is not used by other media
+            // The media pool has reserved an element for ads to use. It is reserved on setup and is not used by other media
```
jnatalzia: it might be `The media pool reserves an element for ads to use`?

src/js/program/program-controller.js
```diff
@@ -4,34 +4,61 @@ import cancelable from 'utils/cancelable';
...
+        });
+        return this.getItemPromise(index, api).setItem(index).then((playlistItem) => {
+            clearTimeout(deferBufferingState);
+            if (playlistItem && playlistItem !== item) {
+                this.setPlaylistItem(index, playlistItem);
+                return playlistItem;
+            }
+            return item;
+        }).catch((/* itemPromiseError */) => {
+            clearTimeout(deferBufferingState);
```
jnatalzia: smart use of timeout here

src/js/program/program-controller.js
```diff
@@ -44,58 +71,64 @@ class ProgramController extends Events {
...
                 // Don't do anything if we've tried to load another provider while this promise was resolving
                 // We check using the mediaModel because it is unique per item, and per instance of that item
                 if (mediaModelContext === model.mediaModel) {
                     nextMediaController.activeItem = item;
                     this._setActiveMedia(nextMediaController);
                     return nextMediaController;
                 }
-            })
-            .catch(err => {
+            }).catch(err => {
```
jnatalzia: this is my lack of understanding of bg loading alongside some difficult to parse changes. can we maybe walk through this file in person or if you feel like it's explainable by text feel free to just reply

src/js/program/program-controller.js
```diff
@@ -313,23 +346,32 @@ class ProgramController extends Events {
...
             .then(nextMediaController => {
                 nextMediaController.activeItem = item;
                 nextMediaController.preload();
                 return nextMediaController;
             })
             .catch(() => {
                 background.clearNext();
-            })
-        );
+            });
```
jnatalzia: mismatched tabs?

src/js/program/program-controller.js
```diff
@@ -709,6 +751,41 @@ class ProgramController extends Events {
...
+            }
+        });
+        this.itemPromises.length = 0;
+    }
+
+    setPlaylistItem(index, item) {
+        const { model } = this;
+        const newItem = normalizePlaylistItem(model, new Item(item), item.feedData || {});
+        if (newItem) {
+            model.get('playlist')[index] = newItem;
```
jnatalzia: does this get around the `change` handlers or something?

src/js/controller/controller.js
robwalch: We could just require an index. What do you think?

src/js/controller/controller.js
robwalch: No magic. Good catch. Fixed in 46c3e77 ...and the commit after that.

If the user calls `jwplayer().setPlaylistItem(currentIndex, { file: 'video' })` while the player is idle, we will reload the current item. We should always clear the item promise when items are replaced in the api>controller.

src/js/controller/next-item-promise.js
robwalch: hoisting those is not great, but one promise per class instance keeps it under control...

src/js/controller/next-item-promise.js
robwalch: The player won't setup without one.

src/js/controller/next-item-promise.js
robwalch: No. `ItemPromise.setItem(index)` is called to run the itemPromise as the program-controller attempts to change the item index. The index check was added to reject skipped items, which looks like it needs more work and testing (we'd need to replace the promise for the next time it's used...)

The idea is that there are two places where run the async callback, when attempting to preload or load the item. So `ItemPromise` has two methods `preload` and `setItem` and they just call run. If we want to extend the arguments we pass to the callback to include which event context we're in we could pass that into `run` from these methods.
```js
    preload () {
        return this.run();
    }

    setItem (index) {

        return this.run();
    }
```

src/js/program/ad-program-controller.js
robwalch: You're right. I don't like that comment so I removed it. Read the source.

src/js/program/program-controller.js
robwalch: Let's talk through it. It wasn't easy for me either. This code and it's execution is complex and I don't intend to commit it to memory.

src/js/program/program-controller.js
robwalch: No. I moved the `loadPromise` promise chain out of `background.setNext(item, loadPromise);`

src/js/program/program-controller.js
robwalch: Setting model attributes directly on `model.attributes` gets around the change event fire when using `model.set`.

In this case we're just replacing an item in the playlist. We're not setting or replacing the playlist array itself. `.setPlaylistItem()` should be silent except for https://github.com/jwplayer/jwplayer/pull/3595#discussion_r376654231

src/js/program/program-controller.js
```diff
@@ -4,34 +4,61 @@ import cancelable from 'utils/cancelable';
...
+    asyncItem(index, item) {
+        const { api, model } = this;
+        // Set the player state to buffering if there is a playlist item callback
+        const deferBufferingState = setTimeout(() => {
+            model.set(PLAYER_STATE, STATE_BUFFERING);
+        });
+        return this.getItemPromise(index, api).run().then((playlistItem) => {
+            clearTimeout(deferBufferingState);
+            if (playlistItem && playlistItem !== item) {
+                this.setPlaylistItem(index, playlistItem);
```
pajong: `this.setPlaylistItem(index, playlistItem)` creates a new instance of `Item` to set it to `playlist[index]`, but `playlistItem` is being used to set to `this.item` in `media-controller`.

DAI plugin uses `getPlaylistItem` to get the current playlistItem and modify the source, but hls uses the item in `media-controller`.

Can we change this to `return this.setPlaylistItem(index, playlistItem);` so that the active `item` in `media-controller` matches the result of `getPlaylistItem`?
If not, is there another API that DAI plugin can use to get the active item in `media-controller`?

src/js/controller/controller.js
jnatalzia: i think i'd prefer that as long as that work doesn't grow in complexity being outside the scope of the parent function

src/js/program/program-controller.js
jnatalzia: scheduled some time for tuesday

src/js/controller/controller.js
```diff
@@ -88,7 +87,7 @@ Object.assign(Controller.prototype, {
             _trigger(type, event);
         }, _this);
 
-        const _programController = this._programController = new ProgramController(_model, mediaPool);
+        const _programController = this._programController = new ProgramController(_model, mediaPool, _api);
```
robwalch: item callback context should be public api

src/js/controller/controller.js
robwalch: Will make this require `index`

src/js/program/program-controller.js
```diff
@@ -44,58 +71,64 @@ class ProgramController extends Events {
...
+        if (!background.isNext(item)) {
+            // Loading a new item invalidates all background loading media
+            this._destroyBackgroundMedia();
         }
-        // Loading a new item invalidates all background loading media
-        this._destroyBackgroundMedia();
+        return (this.loadPromise = this.asyncItem(index, item).then((playlistItem) => {
+            if (!playlistItem) {
+                return;
+            }
```
robwalch: Double check if it's possible to not have an item here.

src/js/program/program-controller.js
robwalch: @pajong 
Once the media-controller is created for an item and active (on `"playlistItem"` event), `mediaController.item` is the active item loaded into the provider. You can change the item in the playlist, but it will not change the active item. That hasn't changed.

The `"playlistItem"`  event should contain the active item. The `getPlaylistItem(index)` method returns what is in the playlist.

If you change what is in the playlist after "playlistItem" with `setPlaylistItem`, that change is not active until navigating back to that item index.

Please describe what if any test case is affected. Include a link, repro steps, and expected/actual behavior. It would help if you also include links to the implementation in DAI for clarity.

If there isn't a new bug, and you are just concerned with how the async API could impact DAI functionality, please provide examples with links to DAI code.

src/js/controller/controller.js
robwalch: item callback context should be public api

src/js/controller/controller.js
robwalch: `getItemPromise` should require an `index`.

src/js/program/program-controller.js
pajong: The problem that I am seeing with DAI is that `this.setPlaylistItem(index, playlistItem)` creates a new instance, then sets to `playlist`, but `playlistItem` is used to set in the `media-controller` after this point. My question is simply why we are not setting it to the newly created instance if we are setting `mediaController.item` anyways.

There is no current test pages broken, but the problem only surfaces if we have the `playlistItemPromise` -  I can put up a test page to reflect that if that helps

src/js/program/program-controller.js
```diff
@@ -44,58 +71,64 @@ class ProgramController extends Events {
...
-            return this.loadPromise;
+        // Destroy background loading item early so we can reuse the media elements in asyncItem
+        if (!background.isNext(item)) {
+            // Loading a new item invalidates all background loading media
+            this._destroyBackgroundMedia();
         }
-        // Loading a new item invalidates all background loading media
-        this._destroyBackgroundMedia();
+        return (this.loadPromise = this.asyncItem(index, item).then((playlistItem) => {
+            if (!playlistItem) {
```
robwalch: Double check if it's possible to not have an item here.

src/js/program/program-controller.js
robwalch: Start with linking to the code in DAI. I think your point is more about when this code reads the item, and that `api.setPlaylistItem` can be called externally while one or more of these promises is unresolved.

src/js/program/program-controller.js
```diff
@@ -44,58 +71,64 @@ class ProgramController extends Events {
...
-            this.loadPromise = this._activateBackgroundMedia();
-            return this.loadPromise;
+        // Destroy background loading item early so we can reuse the media elements in asyncItem
+        if (!background.isNext(item)) {
+            // Loading a new item invalidates all background loading media
+            this._destroyBackgroundMedia();
         }
-        // Loading a new item invalidates all background loading media
-        this._destroyBackgroundMedia();
+        return (this.loadPromise = this.asyncItem(index, item).then((playlistItem) => {
```
robwalch: @pajong Are you saying `api.setPlaylistItem` is called by DAI while `this.asyncItem` is running?

src/js/program/program-controller.js
```diff
@@ -4,34 +4,61 @@ import cancelable from 'utils/cancelable';
...
 
+    asyncItem(index, item) {
+        const { api, model } = this;
+        // Set the player state to buffering if there is a playlist item callback
+        const deferBufferingState = setTimeout(() => {
+            model.set(PLAYER_STATE, STATE_BUFFERING);
+        });
+        return this.getItemPromise(index, api).run().then((playlistItem) => {
+            clearTimeout(deferBufferingState);
+            if (playlistItem && playlistItem !== item) {
```
robwalch: @pajong what if this line is changed to 
`if (playlistItem && playlistItem !== this.model.get('playlist')[index]) {`?

src/js/program/program-controller.js
```diff
@@ -4,34 +4,61 @@ import cancelable from 'utils/cancelable';
...
+                this.setPlaylistItem(index, playlistItem);
+                return playlistItem;
+            }
+            return item;
```
robwalch: `return playlist[index];`

src/js/program/program-controller.js
pajong: No, this `playlistItem` being returned from `asyncItem` is not normalized, but `this.setPlaylistItem` normalizes the same `playlistItem` and sets it - contains the same thing but a different instance

src/js/program/program-controller.js
pajong: The issue is not the check here - the issue is the two lines following this line.
`this.setPlaylistItem(index, playlistItem)` does:
`const newItem = normalizePlaylistItem(model, new Item(item), item.feedData || {});
        if (newItem) {
            model.get('playlist')[index] = newItem;
            return newItem;
        }
}`
This returned `newItem` is but thrown away, and `playlistItem` (without the normalize) is being used to set activeItem in `media-controller`

robwalch: Thanks for catching this @pajong. I've made the following changes and added some tests to cover item replacements and background loading transition https://github.com/jwplayer/jwplayer/pull/3595/commits/8a2756844342a79fbe14cc96e26ff93576ac19d4

src/js/controller/controller.js
robwalch: @jnatalzia I made the change, and made sure to also support `getItemPromise(-1)` for the `"nextUp"` item. I'm a little worried that if someone uses this API they won't want to thing about if they are in a playlist or recs, but we can always make index optional and work out those details.

src/js/program/program-controller.js
```diff
@@ -4,34 +4,60 @@ import cancelable from 'utils/cancelable';
...
 
+    asyncItem(index, item) {
+        const { model } = this;
+        // Set the player state to buffering if there is a playlist item callback
+        const deferBufferingState = setTimeout(() => {
+            model.set(PLAYER_STATE, STATE_BUFFERING);
+        });
+        return this.getItemPromise(index).run().then((playlistItem) => {
+            clearTimeout(deferBufferingState);
+            if (playlistItem && playlistItem !== item) {
```
pajong: This block of code runs even if we have already called `replaceItem` in `backgroundLoad` function, because we are checking the result of the `getItemPromise` (not normalized) to `model.get('playlist')[index]` (normalized in `replaceItem`). I will try to come up with a solution before we sync up tomorrow.

robwalch: Thanks @pajong! I moved item replacement in the `ItemPromise` module: https://github.com/jwplayer/jwplayer/pull/3595/commits/dfcf4f3beafc5b084657776af41b3a4625eb9cec

Since it's a new module I will convert it to TypeScript tomorrow and come up with a better name for it.

Now the `ItemPromise.run()` will always return the item that will be used, and if it needed to be normalized and replaced that will already be done. If the callback does not return a promise we'll also resolve the item promise faster.

pajong: Verified that the preloaded preroll is not destroyed anymore on the VAST side. Thank you! I will run additional tests tomorrow to see if we have anything to refine

pajong:
test this please

src/js/controller/controller.js
```diff
@@ -1036,14 +1053,21 @@ Object.assign(Controller.prototype, {
...
-                const playlist = _model.get('playlist');
-                playlist[index] = item;
-
-                if (index === _model.get('item') && _model.get('state') === 'idle') {
-                    this.setItemIndex(index);
+            const playlist = _model.get('playlist');
+            const playlistItem = playlist[index];
+            if (item && item !== playlistItem) {
+                const newItem = normalizePlaylistItem(_model, new Item(item), item.feedData || {});
+                if (newItem) {
```
pajong: Correct me if I am wrong, but shouldn't we be replacing the playlistItem with this `newItem` somewhere?
`playlist[index] = newItem` is missing

src/js/controller/controller.js
```diff
@@ -1056,13 +1056,13 @@ Object.assign(Controller.prototype, {
...
+                const asyncItemControllerToReject = _programController.asyncItems[index];
+                const asyncItemController = _programController.getAsyncItem(index);
+                const newItem = asyncItemController.replace(item);
+                _programController.asyncItems[index] = null;
                 if (newItem) {
-                    // If the item is replaced using the api here outside of program-controller, reset any item promise
-                    const asyncItemController = _programController.asyncItems[index];
-                    if (asyncItemController) {
+                    if (asyncItemControllerToReject) {
                         asyncItemController.reject(new Error('Item replaced'));
```
pajong: Should this be `asyncItemControllerToReject.reject(new Error('Item replaced'));`?

PR #3596, opened 2020-02-06
Typescript 8 things

jnatalzia:
### This PR will...
Type 8 different files

We are typescripting things

All of it, this is new

Nope

#### Addresses Issue(s):

JW8-10909

### Checklist


src/js/types/generic.type.d.ts
```diff
@@ -0,0 +1,11 @@
+declare module '*';
```
jnatalzia: this is _supposed_ to make any module typed as `any` but i can't get it work

jnatalzia:
test this please

src/js/parsers/rssparser.ts
```diff
@@ -31,17 +38,25 @@ export default function parseRss(dat) {
...
+        title?: string;
+        mediaid?: string;
+        date?: string;
+        description?: string;
+        link?: string;
+        tags?: string;
+    } = {};
     for (let i = 0; i < obj.childNodes.length; i++) {
-        const node = obj.childNodes[i];
+        const node = obj.childNodes[i] as PageNode;
```
zetagame: What does as PageNode do here?

jnatalzia: TS knows that `obj` is a dom node and `childNodes` is a pre-typed `ChildNode` unfortunately for what I presume are compatibility reasons with older browsers, we look for information on these nodes that TS doesn't think is present on a `ChildNode` type (such as `.text`) - so what I'm doing here is telling TS to treat these `ChildNodes` all as `PageNode` since `PageNode` is a union type of the `ChildNode` and some of those additional compat properties

jnatalzia: I'm unclear on how safe or unsafe this type of operation is but i assume we'll be coming across it a lot, so would be good to discuss as a team

zetagame: It could lead to a lot of transgressions of open/close principle, but I'm down to discuss.

src/js/parsers/captions/captions.d.ts
```diff
@@ -0,0 +1,5 @@
+export type CaptionEntryData = {
```
jnatalzia: made this shared file for all types captions (given this is shared between a bunch of parsers)

src/js/events/change-state-event.ts
```diff
@@ -26,7 +27,14 @@ export default function ChangeStateEvent(model, newstate, oldstate) {
         const type = newstate.replace(/(?:ing|d)$/, '');
         const reason = normalizeReason(newstate, model.mediaModel.get('mediaState'));
 
-        const evt = {
+        const evt: {
+            type: string;
```
zetagame: maybe interface for newstate, oldstate, reason if every event has those fields.

src/js/parsers/rssparser.ts
zetagame: Also lets you totally bypass required properties:
image

Usually in TS union types can be done with this pattern safely:

```
interface IStudent {
    id: string;
    age: number;
}

interface IWorker {
    companyId: string;
}

type IUnionType = IStudent | IWorker;
```

src/js/types/generic.type.ts
```diff
@@ -1 +1,11 @@
...
+
+export type CoreModel = any;
+
+export type PageNode = (Element | ChildNode) & {
+    baseName: string;
+    localName: string;
+    text: string;
+};
+
+export type PlaylistItemType = any;
```
zetagame: Would like to explicitly type PlaylistItem at some point. It's used in a lot of places, could be good to standardize the item types of our events (playlistItem, captionsItem, quality, audioTrack etc.)

jnatalzia: 100% agree! this is a placeholder until we type the actual playlist/item.js file

src/js/events/change-state-event.ts
jnatalzia: i like this idea

src/js/parsers/rssparser.ts
```diff
@@ -31,17 +38,25 @@ export default function parseRss(dat) {
...
-        const node = obj.childNodes[i];
+        const node: PageNode = obj.childNodes[i];
         const name = localName(node);
         if (!name) {
             continue;
         }
         switch (name.toLowerCase()) {
             case 'enclosure':
-                item.file = xmlAttribute(node, 'url');
+                item.file = xmlAttribute(node as Element, 'url');
```
jnatalzia: @zetagame can't figure out a way around this one, i tried redefining `const n:Element = node;` above this and it errors out. reasoning is `obj.childNodes` returns a `ChildNode` and since the `PageNode` type is a union type of `(Element | ChildNode)` that decision is made higher and it sticks to this being explicitly a `ChildNode` - I wonder if there's a built in type that works for this however I tried a few to no avail so I'm willing to commit this as is. Thoughts?

zetagame: Good catch! I have a PR up now for a proposed solution.

PR #3599, opened 2020-02-10
Activate no-unsafe-any lint rule

mamaddox:
### This PR will...
Activate the `no-unsafe-any` linting rule

** PR to remove `no-unsafe-any` rule for now. https://github.com/jwplayer/jwplayer/pull/3600 **
#### Addresses Issue(s):

JW8-###

### Checklist

### DIFF
--- a/src/js/utils/browser.ts
+++ b/src/js/utils/browser.ts
@@ -88,45 +88,17 @@ export function flashVersion(): number {
 
     if (typeof window.ActiveXObject !== 'undefined') {
         try {
-<<<<<<< HEAD
-<<<<<<< HEAD
             const flashObj = new window.ActiveXObject('ShockwaveFlash.ShockwaveFlash') as FlashObject;
             if (flashObj) {
                 return parseFloat(flashObj.GetVariable('$version').split(' ')[1].replace(/\s*,\s*/, '.'));
-=======
-            const flashObj = new window.ActiveXObject('ShockwaveFlash.ShockwaveFlash');
-            if (flashObj) {
-                return parseFloat((flashObj as FlashObject).GetVariable('$version').split(' ')[1].replace(/\s*,\s*/, '.'));
->>>>>>> Activate no-unsafe-any lint rule
-=======
-            const flashObj = new window.ActiveXObject('ShockwaveFlash.ShockwaveFlash') as FlashObject;
-            if (flashObj) {
-                return parseFloat(flashObj.GetVariable('$version').split(' ')[1].replace(/\s*,\s*/, '.'));
->>>>>>> Return flash plugin in flashVersion and cast when vars are defined
             }
         } catch (e) {
             return 0;
         }
-<<<<<<< HEAD
-<<<<<<< HEAD
-=======
-
-        return 0;
->>>>>>> Activate no-unsafe-any lint rule
-=======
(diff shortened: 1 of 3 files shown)


src/js/utils/browser.ts
```diff
@@ -78,26 +78,29 @@ export function flashVersion(): number {
...
+            const flashObj = new window.ActiveXObject('ShockwaveFlash.ShockwaveFlash');
+            if (flashObj) {
+                return parseFloat((flashObj as FlashObject).GetVariable('$version').split(' ')[1].replace(/\s*,\s*/, '.'));
             }
         } catch (e) {
             return 0;
         }
 
-        return flash;
+        return 0;
```
mamaddox: Correct me if I'm wrong, but it seems like `flash` at this point would either be `undefined` or the `flashPlugin` if it didn't have a description. Those should probably return as `0`

mamaddox:
My only concern with this is it might lead to the need for additional castings.

src/js/utils/validator.ts
```diff
@@ -85,7 +85,7 @@ export function isDeepKeyCompliant(defaultObj: GenericObject, customObj: Generic
             const customValue = customObj[key];
```
jnatalzia: does it not work if we do `const customValue: GenericObject = ...` here?

mamaddox: Not exactly. It causes an `unsafe-any` on `customObj[key]` instead because it would return an `any` type.

But it could be changed to:
`const customValue = customObj[key] as GenericObject`

jnatalzia: oof

src/js/utils/browser.ts
robwalch: `flash` could have been set to `plugins.namedItem('Shockwave Flash')` and not have a `description` property so we return that value here.

mamaddox:
Since we have been using the `any` type as a placeholder for modules that have not been typed yet, enabling this rule will throw an error whenever we try to access something from those placeholder types.

For example, since our `model` has an `any` type for now, calling `model.get()` will throw this error.

For now, we can disable it. https://github.com/jwplayer/jwplayer/pull/3600

src/js/utils/validator.ts
```diff
@@ -81,8 +81,8 @@ export function isDeepKeyCompliant(defaultObj: GenericObject, customObj: Generic
     const defaultKeys = Object.keys(defaultObj);
     return Object.keys(customObj).length >= defaultKeys.length &&
         defaultKeys.every(key => {
-            const defaultValue = defaultObj[key];
-            const customValue = customObj[key];
+            const defaultValue = defaultObj[key] as GenericObject;
```
zetagame: We're casting this to an object before checking if it's an object. This should be able to consume any value we can have for an object property, so this is a place implicit any should be fine.

mamaddox: That makes sense. Knew there was a reason the cast was after this check before.

src/js/utils/validator.ts
zetagame: typeof guards fix this:
https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-guards-and-differentiating-types

jnatalzia:
Closing for now

PR #3600, opened 2020-02-10
Remove no-unsafe-any linting rule

mamaddox:
### This PR will...
Remove the `no-unsafe-any` linting rule

Related to: https://github.com/jwplayer/jwplayer/pull/3599#issuecomment-584289214
#### Addresses Issue(s):

JW8-###

### Checklist

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -189,7 +189,6 @@
                         ]
                     }
                 }],
-                "@typescript-eslint/no-unsafe-any": true,
                 "@typescript-eslint/no-explicit-any": 0,
                 "@typescript-eslint/adjacent-overload-signatures": "error",
                 "@typescript-eslint/ban-ts-ignore": "error",


.eslintrc
```diff
@@ -189,7 +189,6 @@
                         ]
                     }
                 }],
-                "@typescript-eslint/no-unsafe-any": true,
```
zetagame: if something is tripping this rule on type-check it isn't type safe. Are we removing this permanently, or is this a temporary work around for something?

mamaddox: Currently, this rule is not actually enabled (my fault), so nothing is tripping it. I want to remove it for now because it will error while we are converting files.

https://github.com/jwplayer/jwplayer/pull/3599#issuecomment-584289214

> Since we have been using the `any` type as a placeholder for modules that have not been typed yet, enabling this rule will throw an error whenever we try to access something from those placeholder types.
>
> For example, since our `model` has an `any` type for now, calling `model.get()` will throw this error.
>
> For now, we can disable it. #3600

PR #3602, opened 2020-02-12
Type 5 more files

jnatalzia:
### This PR will...
Type the following files:

* src/js/playlist/track.js
* src/js/plugins/plugins.js
* src/js/program/program-constants.js
* src/js/program/shared-media-pool.js
* src/js/program/media-element-pool.js

TS'ing the player

Nothing I can think of

Nope

#### Addresses Issue(s):

JW8-10924

### Checklist


src/js/types/generic.type.ts
```diff
@@ -9,3 +9,11 @@ export type PageNode = (Element | ChildNode) & {
 };
 
 export type PlaylistItemType = any;
+
+export type PlaylistItemTrack = GenericObject & {
```
jnatalzia: can fix this up when we fully type the track obj

src/js/playlist/track.ts
```diff
@@ -6,7 +8,7 @@
  * @property {boolean} default - Enable the track by default.
  */
 
-const Track = function(config) {
+const Track = function(config: PlaylistItemTrack): PlaylistItemTrack | undefined {
```
jnatalzia: this seems like the file we should type the track in, should i just grab the options from the current config schema and port them? cc @robwalch

jnatalzia:
test this please

src/js/program/shared-media-pool.ts
```diff
@@ -0,0 +1,20 @@
+import { MediaElementPoolInt, OptionalUndefVideo } from './media-element-pool';
```
dYale: Can this be an absolute path?

jnatalzia: ugh vs code autoimport killin me

src/js/playlist/track.ts
```diff
@@ -1,3 +1,12 @@
+type PlaylistItemTrack = {
+    file?: string;
+    kind: string;
```
robwalch: You can enum or `'valueA' | 'valueB'` the kinds. Will help avoid typos.

`"captions" | "metadata" | "thumbnails" | "chapters"`

I'm not sure if `"subtitles"` is also valid (alias for `"captions"`). I vaguely remember an issue like this but don't see support for it anywhere.

src/js/playlist/track.ts
```diff
@@ -1,3 +1,12 @@
...
+    default: boolean;
+    name: string;
+    label: string;
+    language: string;
```
robwalch: Everything except for kind is optional.

src/js/plugins/plugins.ts
```diff
@@ -1,16 +1,21 @@
...
+import { CoreModel, PlayerAPI, GenericObject } from 'types/generic.type';
+
+declare global {
+    interface Window { jwplayerPluginJsonp: Function }
+}
 
 const pluginsModel = new PluginsModel();
 
-export const registerPlugin = function(name, minimumVersion, pluginClass) {
+export const registerPlugin = function(name: string, minimumVersion: string, pluginClass: Function): void {
```
robwalch: `pluginClass: (api: PlayerAPI, config: GenericObject, div: HTMLDivElement) => void`

src/js/program/media-element-pool.ts
```diff
@@ -20,30 +36,30 @@ export default function MediaElementPool() {
...
         },
-        played() {
+        played(): void {
             primed = true;
         },
-        getPrimedElement() {
+        getPrimedElement(): OptionalUndefVideo | null {
             if (pool.length) {
                 // Shift over pop so that we cycle through elements instead of reusing the same one
                 return pool.shift();
```
robwalch: This would never be undefined

src/js/program/media-element-pool.ts
```diff
@@ -57,34 +73,34 @@ export default function MediaElementPool() {
...
-function primeMediaElementForPlayback(mediaElement) {
+function primeMediaElementForPlayback(mediaElement: HTMLVideoElement): void {
     // If we're in a user-gesture event call load() on video to allow async playback
     if (!mediaElement.src) {
         mediaElement.load();
     }
 }
 
-export function createMediaElement(options) {
+export function createMediaElement(options: GenericObject | void): HTMLVideoElement {
```
robwalch: `options?: GenericObject` ?

src/js/program/shared-media-pool.ts
```diff
@@ -0,0 +1,20 @@
...
+
+export default function SharedMediaPool(sharedElement: HTMLVideoElement, mediaPool: MediaElementPoolInt): MediaElementPoolInt {
+    return Object.assign({}, mediaPool, {
+        prime(): void {
+            if (!sharedElement.src) {
+                sharedElement.load();
+            }
+        },
+        getPrimedElement(): OptionalUndefVideo {
+            return sharedElement;
```
robwalch: `sharedElement` is always and `HTMLVideoElement`.

Is the return type here just used to match `MediaElementPool .getPrimedElement` ? Why?

src/js/playlist/track.ts
jnatalzia: :+1:

src/js/program/media-element-pool.ts
jnatalzia: interestingly TS does not recognize the `.length` check as a safeguard here - so it's expecting me to cover the case where `shift` returns a non-element (length 0). i have an idea of how to change the func to fix.

src/js/program/shared-media-pool.ts
jnatalzia: yeah i was matching - was unclear on use case for the func, updated!

src/js/program/media-element-pool.ts
```diff
@@ -20,30 +36,31 @@ export default function MediaElementPool() {
...
             primed = true;
         },
-        played() {
+        played(): void {
             primed = true;
         },
-        getPrimedElement() {
-            if (pool.length) {
+        getPrimedElement(): HTMLVideoElement | null {
+            const el = pool.shift();
```
jnatalzia: interestingly TS does not recognize the .length check as a safeguard here - so it's expecting me to cover the case where shift returns a non-element (length 0).

this was my idea of how to fix. based on MDN, this should always return undef for len 0

src/js/playlist/track.ts
robwalch: Should there be a `PlaylistItemTrack` type and a `PlaylistItemTrackOptions` type?

jnatalzia: i think they are functionally the same in this context right?

robwalch: The same if the optional properties are. Maybe Track always assigns a value to certain props?

src/js/playlist/track.ts
```diff
@@ -6,7 +15,7 @@
  * @property {boolean} default - Enable the track by default.
  */
 
-const Track = function(config) {
+const Track = function(config?: PlaylistItemTrack): PlaylistItemTrack | undefined {
```
robwalch: So literally anything can be passed in for `config`. I really is `any` or generic | undefined.

Should we use `PlaylistItemTrack | void` for the return value?

src/js/playlist/track.ts
```diff
@@ -1,3 +1,12 @@
+type PlaylistItemTrack = {
+    file?: string;
+    kind: 'captions' | 'metadata' | 'thumbnails' | 'chapters';
+    default?: boolean;
```
robwalch: `default` is no longer optional.

src/js/playlist/track.ts
```diff
@@ -1,3 +1,12 @@
+type PlaylistItemTrack = {
+    file?: string;
+    kind: 'captions' | 'metadata' | 'thumbnails' | 'chapters';
```
robwalch: We might want to enforce the value for `kind` and boolean type for `default` since we could end up with anything thanks to `Object.assign`.

src/js/program/media-element-pool.ts
robwalch: Would it accept `return pool.shift() || null;` ?

src/js/program/media-element-pool.ts
```diff
@@ -20,30 +36,31 @@ export default function MediaElementPool() {
...
+            if (el) {
                 // Shift over pop so that we cycle through elements instead of reusing the same one
-                return pool.shift();
+                return el;
             }
             return null;
         },
-        getAdElement() {
+        getAdElement(): OptionalUndefVideo {
```
robwalch: I think adElement is always defined.

src/js/program/media-element-pool.ts
```diff
@@ -20,30 +36,31 @@ export default function MediaElementPool() {
...
             return adElement;
         },
-        getTestElement() {
+        getTestElement(): OptionalUndefVideo {
```
robwalch: testElement must always be defined.

src/js/program/shared-media-pool.ts
```diff
@@ -0,0 +1,20 @@
+import { MediaElementPoolInt } from 'program/media-element-pool';
+
+export default function SharedMediaPool(sharedElement: HTMLVideoElement, mediaPool: MediaElementPoolInt): MediaElementPoolInt {
```
robwalch: It make make more sense to convert to classes. It looks like `MediaElementPoolInt` is making compromises (`OptionalUndefVideo` for a few methods that match across "classes") to fit these different modules.

robwalch: We don't have to go as far as making these classes, but it's worth considering what is cleaner. I _think_ we have good test coverage for these modules.

src/js/playlist/track.ts
jnatalzia: hmm yeah there's some danger here. with the object assign, the types yell about passing in `GenericObject` when we are using object.assign on it (extra props). i guess I could use an interface here for `PlaylistItemTrack` but that also feels incorrect

src/js/program/media-element-pool.ts
jnatalzia: same problem with `pool.shift` the possibility of it being undefined causes a need for the optional undef. we should think more holistically about this, i think we may uncover a smarter approach when dependencies that call this function are type

src/js/program/media-element-pool.ts
jnatalzia: same comment re: `pool.shift`

src/js/program/shared-media-pool.ts
jnatalzia: agree that a class would be cleaner, i'm wishy-washy on scope creep for these changes. in general i think my approach is the less change to the underlying code the better given we aren't really expecting to QA these changes -- it might add a lot of overhead with a little return but i'm not sold on an opinion here

src/js/playlist/track.ts
jnatalzia: Not sure exactly what you mean by `enforce`?

robwalch: > enforce?
After `Object.assign` replace invalid values with their defaults.

src/js/program/media-element-pool.ts
robwalch: I thought `adElement` was a fixed value. It's assigned using `shift()`?
Either break out the media element creating function to assign `adElement` or assign `adElement as HTMLMediaElemt` because we know better (we're always filling the pool and shifting one element out for ads).

src/js/program/media-element-pool.ts
robwalch: Use `as` to assign or if we don't need the element to come from the pool, then break out the element factory.

robwalch: Just make sure we don't do anything with the pool before pulling out test and ad elements (like "priming" the tags).

src/js/program/shared-media-pool.ts
robwalch: OK let's not refactor, but try to lock down the types using `as` to assign so return types aren't wishy-washy. That will hurt us when we type the code using this module.

src/js/playlist/track.ts
robwalch: `PlaylistItemTrack` makes sense as an interface that extends `GenericObject`.

src/js/playlist/track.ts
jnatalzia: ah, updating

src/js/program/media-element-pool.ts
jnatalzia: updating!

src/js/program/shared-media-pool.ts
jnatalzia: removed the `OptionalUndef` in favor of `as`

src/js/playlist/track.ts
```diff
@@ -15,7 +26,10 @@ const Track = function(config) {
     return Object.assign({}, {
         kind: 'captions',
         'default': false
-    }, config);
+    }, config, {
+        default: false,
+        kind: 'captions'
```
jnatalzia: @robwalch what should the default be here?

PR #3606, opened 2020-02-14
[No Ticket] Add Grunt task to quickly lint and typecheck ts files

zetagame:
### This PR will...
Add a grunt task to check TS files.

#### Addresses Issue(s):

JW8-###

### Checklist

### DIFF
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -264,6 +264,11 @@ module.exports = function(grunt) {
         'karma:local'
     ]);
 
+    grunt.registerTask('ts', [
+        'lint',
+        'type-check'
+    ]);
+
     grunt.registerTask('serve', [
         'connect:livereload',
         'watch'
--- a/Gruntfile.js
+++ b/Gruntfile.js
@@ -264,11 +264,6 @@ module.exports = function(grunt) {
         'karma:local'
     ]);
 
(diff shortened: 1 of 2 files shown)


Gruntfile.js
```diff
@@ -264,6 +264,11 @@ module.exports = function(grunt) {
         'karma:local'
     ]);
 
+    grunt.registerTask('ts', [
```
robwalch: I would prefer that this be added as a script task in `project.json`.

- `yarn ts` or `npm run ts` to run
- Less reliance on grunt
- We can still run script tasks in grunt by calling to `yarn` ci

jnatalzia: i would be down to add to `yarn type-check`

package.json
```diff
@@ -84,7 +84,8 @@
     "squash": "node parse-squashed-files.js",
     "test": "karma start",
     "karma:hook": "karma start --browsers=ChromeHeadless --reporters=mocha",
-    "type-check": "tsc"
+    "type-check": "tsc",
+    "ts-check": "tsc && eslint './src/js' --ext .ts --ext .js"
```
jnatalzia: `--ext .ts` only?

zetagame: good call

PR #3607, opened 2020-02-15
Fix html5 instream attach/detach provider cleanup in Safari

robwalch:
### This PR will...
Properly destroy providers used for instream playback by ad-program-controller.
Removes all `video.textTracks` and `textTracks[track]` event listeners added by the provider.

The instream provider was handling textTrack cue change events, and due to a shared object being used to track triggered cues, some "meta"  were never triggered by the main provider.

This issue probably went unnoticed previously because there were fewer overlapping metadata cues in Safari tests.

Test updates and some small but important fixes to Safari side hls playlist loading: https://github.com/jwplayer/jwplayer-commercial/pull/7435

#### Addresses Issue(s):
JW8-10898

### Checklist

### DIFF
--- a/src/js/program/media-controller.js
+++ b/src/js/program/media-controller.js
@@ -21,7 +21,7 @@ export default class MediaController extends Events {
         this.provider = provider;
         this.providerListener = new ProviderListener(this);
         this.thenPlayPromise = cancelable(() => {});
-        addProviderListeners(this);
+        this.addProviderListeners();
         this.eventQueue = new ApiQueueDecorator(this, ['trigger'],
             () => !this.attached || this.background);
     }
@@ -316,14 +316,15 @@ export default class MediaController extends Events {
     set volume(volume) {
         this.provider.volume(volume);
     }
+
+    addProviderListeners() {
+        this.provider.off();
+        this.provider.on('all', this.providerListener, this);
+    }
 }
 
 function syncPlayerWithMediaModel(mediaModel) {
     // Sync player state with mediaModel state
     const mediaState = mediaModel.get('mediaState');
     mediaModel.trigger('change:mediaState', mediaModel, mediaState, mediaState);
-}
-
-function addProviderListeners(mediaController) {
-    mediaController.provider.on('all', mediaController.providerListener, mediaController);
 }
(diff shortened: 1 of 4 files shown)


src/js/providers/tracks-mixin.js
```diff
@@ -15,8 +15,8 @@ const Tracks = {
     _metaCuesByTextTime: null,
     _currentTextTrackIndex: -1,
     _unknownCount: 0,
-    _activeCues: {},
-    _cues: {},
```
robwalch: These object values were being applied across provider instances :(

jnatalzia: ah! great catch

src/js/program/media-controller.js
```diff
@@ -316,14 +316,15 @@ export default class MediaController extends Events {
...
 }
 
 function syncPlayerWithMediaModel(mediaModel) {
     // Sync player state with mediaModel state
     const mediaState = mediaModel.get('mediaState');
     mediaModel.trigger('change:mediaState', mediaModel, mediaState, mediaState);
 }
-
-function addProviderListeners(mediaController) {
-    mediaController.provider.on('all', mediaController.providerListener, mediaController);
```
jnatalzia: assuming `mediaController.provider` is not always the provider instantiated with _this_ media controller? or is this just to make things safer?

src/js/providers/tracks-mixin.js
```diff
@@ -701,8 +703,8 @@ function _cueChangeHandler(e) {
 
     if (activeCues && activeCues.length) {
         const previousActiveCues = this._activeCues[_id];
-        this._activeCues[_id] = Array.prototype.slice.call(activeCues);
-        this.triggerActiveCues(activeCues, previousActiveCues);
+        const currentActiveCues = this._activeCues[_id] = Array.prototype.slice.call(activeCues);
+        this.triggerActiveCues(currentActiveCues, previousActiveCues);
```
jnatalzia: oof

src/js/program/media-controller.js
robwalch: yeah I think this was just a helper. I didn't need to add it to the interface and can move this code into the constructor. I only moved it because I considered calling it again.

src/js/providers/tracks-mixin.js
robwalch: Saves us from slicing twice.

PR #3610, opened 2020-02-19
[JW8-10949] Fix event order when replaying

pajong:
### This PR will...
- Wait for `setItemIndex(0)` promise before triggering `beforePlay` event

- `setItemIndex(0)` fires `playlistItem` event
- If we do not wait for the promise, `beforePlay` event fires before we trigger `playlistItem` event
- Wrong event order breaks ad plugins

#### Addresses Issue(s):

JW8-10949

### Checklist

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -500,43 +500,46 @@ Object.assign(Controller.prototype, {
                 return Promise.resolve();
             }
 
+            let setItemPromise = Promise.resolve();
             if (_model.get('state') === STATE_COMPLETE) {
                 _stop(true);
-                _this.setItemIndex(0);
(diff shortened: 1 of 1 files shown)


pajong:
> making play always async is not a good idea. Instead consider:
> 
> ```js
> if (_model.get('state') === STATE_COMPLETE) {
>     _stop(true);
>    this.setItemIndex(0).then(() => {
>        _play(meta));
>   }
>  }
> ```
> 
> That should keep this sync in most cases. The recursion above could be bad so maybe we break out play into two methods.

@robwalch Thank you!
The recursion seems fine since we are just repeating error check, playReason update, and adState check. If we still don't want the repetition, I can break out the method into two.

PR #3613, opened 2020-02-21
[JW8-10934] Typescript 4 files

zetagame:
### This PR will...
TS

TS

TS

#### Addresses Issue(s):

JW8-10934

### Checklist


src/js/providers/utils/stream-type.ts
```diff
@@ -1,34 +1,35 @@
 /** @module */
+import { StreamType } from 'types/generic.type';
```
jnatalzia: given it's part of this file, I think it makes more sense to export it from this file. the generic is mostly used for placeholder types we hope to fill in later

jnatalzia:
also looks like the build is failing on this one

PR #3614, opened 2020-02-21
Preroll is not preloaded for item when using recommendations with multiitem playlist

jnatalzia:
### This PR will...
Fix a bug where the ads plugin would not preload ads if there was a multi-item playlist using recs

Bugfixing async api

Left comments inline

#### Tested
* Config in the relevant ticket
    * BG Load occurs for 2nd playlist item now
    * BG Load occurs for recs still once the original 2 item playlist is complete

#### To Test
* Recs only playlist w/ autoplay and 10 second countdown
* Playlist only w/ autoplay and 10 second countdown

Nope

#### Addresses Issue(s):

JW8-10958

### Checklist

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -191,7 +191,7 @@ Object.assign(Controller.prototype, {
                     if (allowPreload && position && duration > 0 && position >= duration - BACKGROUND_LOAD_OFFSET) {
                         mediaModel.off('change:position', onPosition, this);
                         _programController.backgroundLoad(item, index);
-                    } else if (recsAuto) {
+                    } else if (recsAuto && !item) {
                         index = -1;
                         item = _model.get('nextUp');
                     }


src/js/controller/controller.js
```diff
@@ -191,7 +191,7 @@ Object.assign(Controller.prototype, {
                     if (allowPreload && position && duration > 0 && position >= duration - BACKGROUND_LOAD_OFFSET) {
                         mediaModel.off('change:position', onPosition, this);
                         _programController.backgroundLoad(item, index);
-                    } else if (recsAuto) {
+                    } else if (recsAuto && !item) {
```
jnatalzia: my one concern with this change is i'm not 100% on the behavior of recs that are on the item not the playlist (in the `recommendations` key) - we should test this when reviewing these changes

jnatalzia: also this makes me think this has always been broken?

pajong: I think this is broken after the async changes since we are passing in `index` to `_programController.backgroundLoad`, and that passed in `index` is used to get the `asyncItem`.
Before the async changes, it did not take in `index`, and we use the following to get the next item (line 184 in this PR):
```
let index = model.get('item') + 1;
let item = model.get('playlist')[index];
```

jnatalzia:
@aizaztoppa noted that this fix does not work on safari. we don't background load on safari in the player but we do in the ads plugin, regardless of browser (@pajong can you confirm this? just what I'm seeing in the code). So the ads plugin gets the promise way before the player attempts to run the playlist item code based on that. We may need to stub the backgroundLoad func for Safari to ensure the async callback runs for that ads case but @robwalch probably will have more ideas on how to fix this.

PR #3615, opened 2020-02-21
[JW8-10957] Use attached flag in setActiveItem while waiting for asyncActiveItem

pajong:
### This PR will...
- Use `attached` flag in `mediaController` while waiting for the `asyncActiveItem` to block events

- Calling `mediaController.off` results all the events until `this._setActiveMedia(mediaController);` call, especially when we are re-using the same `mediaController` from the previous item
- `mediaType` event is fired when setting a new activeItem to `mediaController`
`mediaController.activeItem = playlistItem;`
-- `mediaType` event is lost if we use `mediaController.off`
-- Using `detach` to queue up the events and later flushing with `attach` solves the issue

- Tried using `mediaController.attach()` and `mediaController.detach()`, but it was causing other regressions since they have other function calls (`setPlaybackRate` and ` provider.attachMedia()`)

#### Addresses Issue(s):

JW8-10957

### Checklist

### DIFF
--- a/src/js/program/program-controller.js
+++ b/src/js/program/program-controller.js
@@ -78,7 +78,7 @@ class ProgramController extends Events {
         // Reset the mediaModel now to synchronously "cancel" play promise callbacks
         // that check `mediaModel === model.mediaModel`
         if (mediaController) {
-            mediaController.off();
+            mediaController.detach();
             mediaController.mediaModel.off();
             model.attributes.mediaModel = new MediaModel();
             model.mediaModel.attributes = mediaController.mediaModel.clone();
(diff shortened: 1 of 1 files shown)


src/js/program/program-controller.js
```diff
@@ -122,6 +122,8 @@ class ProgramController extends Events {
                     // We can synchronously reuse the current mediaController
                     // Reinitialize the mediaController with the new item, allowing a new playback session
                     mediaController.activeItem = playlistItem;
+                    mediaController.attached = true;
+                    mediaController.eventQueue.flush();
```
robwalch: Can these go in `_setActiveMedia`? It's called from a couple places and will always need to be attached.

pajong:
test this please

src/js/program/program-controller.js
```diff
@@ -78,7 +78,7 @@ class ProgramController extends Events {
         // Reset the mediaModel now to synchronously "cancel" play promise callbacks
         // that check `mediaModel === model.mediaModel`
         if (mediaController) {
-            mediaController.off();
+            mediaController.attached = false;
```
robwalch: Just remove off, and don't use `attached`. We can put it back and add listeners earlier if we need to.

PR #3616, opened 2020-02-21
[JW8-10964] Typescript 5 files

zetagame:
Typescript 5 files

#### Addresses Issue(s):

JW8-10964

### Checklist

### DIFF
--- a/src/js/utils/helpers.ts
+++ b/src/js/utils/helpers.ts
@@ -1,3 +1,4 @@
+import { GenericObject } from 'types/generic.type';
 import * as playerutils from 'utils/playerutils';
 import * as validator from 'utils/validator';
 import * as parser from 'utils/parser';
@@ -42,7 +43,7 @@ import { log } from 'utils/log';
 import { genId } from 'utils/random-id-generator';
 
 // TODO: deprecate (jwplayer-ads-vast uses utils.crossdomain(url))
-function crossdomain(uri) {
+function crossdomain(uri: string): boolean {
     const a = document.createElement('a');
     const b = document.createElement('a');
     a.href = location.href;
@@ -55,7 +56,7 @@ function crossdomain(uri) {
 }
 
 // The predicate received the arguments (key, value) instead of (value, key)
-const foreach = function (aData, fnEach) {
+const foreach = function (aData: GenericObject, fnEach: (key: string, value: any) => void): void {
     for (let key in aData) {
         if (Object.prototype.hasOwnProperty.call(aData, key)) {
             fnEach(key, aData[key]);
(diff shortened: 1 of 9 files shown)


src/js/utils/helpers.ts
```diff
@@ -55,17 +56,19 @@ function crossdomain(uri) {
...
     }
 };
 
-const noop = function () {};
+const noop = () => {
+    // noop
+};
 
-const helpers = Object.assign({}, parser, validator, playerutils, {
+const helpers: GenericObject = Object.assign({}, parser, validator, playerutils, {
```
robwalch: It might be interesting to type this as it goes on the public api as `jwplayer().utils` and in commercial for backward compatibility `jwplayer.utils`. See https://github.com/jwplayer/jwplayer/blob/master/test/unit/utils-defs-test.js

src/js/utils/in-interaction-event.ts
```diff
@@ -0,0 +1,6 @@
+import { GenericObject } from 'types/generic.type';
+
+export function inInteraction(event: GenericObject | undefined): boolean {
```
robwalch: `event` is optional. When it is passed in, the expected type is `Event`.

```suggestion
export function inInteraction(event?: Event): boolean {
```

src/js/utils/in-interaction-event.ts
```diff
@@ -0,0 +1,6 @@
...
+    event = event || window.event;
+    return !!event && /^(?:mouse|pointer|touch|gesture|click|key)/.test(event.type);
```
robwalch: You can leave in the  ` | undefined` above to match the type of `window.event` (`Event | undefined`) but it would be better to keep the function definition strict (optional `Event`) and conform to `window.event` here with a new const.

 ```suggestion
    const interaction: Event | undefined = event || window.event;
    return !!interaction && /^(?:mouse|pointer|touch|gesture|click|key)/.test(interaction.type);
```

File: src/js/utils/log.ts [hunk 4]
```diff
@@ -0,0 +1,7 @@
+/* eslint-disable no-console */
+
+const noop = () => {
+    // noop
+};
```
robwalch: Why define `noop` upfront?

Inline avoided early assignment and scope polution.

src/js/utils/helpers.ts
zetagame: Good call, will update

src/js/utils/in-interaction-event.ts
zetagame: That is what this means in the above context (as an argument as apposed to variable assignment). Event or undefined.

src/js/utils/in-interaction-event.ts
zetagame: We're re-assigning so type is already enforced

image

src/js/utils/in-interaction-event.ts
zetagame: Agreed, making this change now

src/js/utils/log.ts
zetagame: Good call, hoisting the func sails cap'n.

src/js/utils/helpers.ts
zetagame: As per earlier discussion, I'd like to wait on this until the dependencies are typed, since that will have a similar result.

PR #3619, opened 2020-02-25
Typescript more things!

jnatalzia:
### This PR will...
Type the following files:

* src/js/templates/logo.js
* src/js/templates/player.js
* src/js/utils/active-tab.js
* src/js/utils/aria.js

TS

Nah
Nah
#### Addresses Issue(s):

JW8-10728

### Checklist

### DIFF
--- a/src/js/templates/logo.ts
+++ b/src/js/templates/logo.ts
@@ -1,4 +1,6 @@
-export default (position: string, hide: boolean): string => {
+import { HTMLTemplateString } from '../types/generic.type';
+
+export default (position: string, hide: boolean): HTMLTemplateString => {
     const jwhide = hide ? ' jw-hide' : '';
     return `<div class="jw-logo jw-logo-${position}${jwhide} jw-reset"></div>`;
 };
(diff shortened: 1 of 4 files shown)


src/js/templates/logo.ts
```diff
@@ -1,4 +1,4 @@
-export default (position, hide) => {
+export default (position: string, hide: boolean): string => {
```
zetagame: I added an HTMLTemplateString type so we can better identify template usage, and add in actual string template detection when it becomes a feature. We can take it out if it's not needed, but if you like it we can use it here for return type (and a few others)

zetagame: I added an [HTMLTemplateString](https://github.com/jwplayer/jwplayer/pull/3613/files) type so we can better identify template usage, and add in actual string template detection when it becomes a feature. We can take it out if it's not needed, but if you like it we can use it here for return type (and a few others)

jnatalzia: oh YES!

PR #3622, opened 2020-02-26
[JW8-10936] TypeScript 4 more utils

robwalch:
### This PR will type:
- src/js/utils/cancelable.js
- src/js/utils/clock.js
- src/js/utils/date.js
- src/js/utils/defaults.js

#### Addresses Issue(s):
JW8-10936

### Checklist

### DIFF
--- a/src/js/utils/cancelable.ts
+++ b/src/js/utils/cancelable.ts
@@ -1,9 +1,14 @@
-export default function cancelable(callback) {
+type Cancelable = {
+    async: (...args: [any]) => Promise<any>;
+    cancel: () => void;
+    cancelled: () => boolean;
+}
+
+export default function cancelable(callback: (result?: any) => any): Cancelable {
     let cancelled = false;
 
     return {
-        async: function() {
-            const args = arguments;
+        async: function(...args: [any]): Promise<any> {
             return Promise.resolve().then(() => {
                 if (cancelled) {
                     return;
--- a/src/js/utils/clock.ts
+++ b/src/js/utils/clock.ts
@@ -1,6 +1,7 @@
 import { now as getTime } from 'utils/date';
+import { GenericObject } from '../types/generic.type';
 
-const performance = window.performance || {
+const performance: Performance | GenericObject = window.performance || {
     timing: {}
 };
 const startDate = performance.timing.navigationStart || getTime();
@@ -9,10 +10,10 @@ if (!('now' in performance)) {
     performance.now = () => (getTime() - startDate);
 }
 
-export function now() {
+export function now(): number {
     return performance.now();
 }
 
-export function dateTime() {
+export function dateTime(): number {
     return startDate + performance.now();
 }
--- a/src/js/utils/defaults.ts
+++ b/src/js/utils/defaults.ts
@@ -1,8 +1,9 @@
+import { GenericObject } from '../types/generic.type';
+
 // Adds properties to the first object from the rest
 // Does not add properties which exist anywhere in the object or it's prototype chain (no shadowing, no overriding)
-
-export default function defaults(obj) {
-    Array.prototype.slice.call(arguments, 1).forEach(function(source) {
+export default function defaults(obj: GenericObject, ...rest: any[]): GenericObject {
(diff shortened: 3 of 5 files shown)


src/js/utils/cancelable.ts
```diff
@@ -1,9 +1,14 @@
-export default function cancelable(callback) {
+type Cancelable = {
+    async: (...args: [any]) => Promise<any>;
+    cancel: () => void;
+    cancelled: () => boolean;
+}
+
+export default function cancelable(callback: (result?: any) => any): Cancelable {
```
zetagame: Do cancelable callbacks always return something? Might need to be => any | void

src/js/utils/clock.ts
```diff
@@ -1,6 +1,6 @@
 import { now as getTime } from 'utils/date';
 
-const performance = window.performance || {
+const performance: Performance | any = window.performance || {
```
zetagame: Semantically the same as `performance: any`.

src/js/utils/defaults.ts
```diff
@@ -1,8 +1,9 @@
+import { GenericObject } from '../types/generic.type';
+
 // Adds properties to the first object from the rest
 // Does not add properties which exist anywhere in the object or it's prototype chain (no shadowing, no overriding)
-
-export default function defaults(obj) {
-    Array.prototype.slice.call(arguments, 1).forEach(function(source) {
+export default function defaults(obj: GenericObject, ...rest: any[]): GenericObject {
```
zetagame: are we polyfilling rest spread?

src/js/utils/cancelable.ts
robwalch: I thought that would be covered by `any` since it disables type checks.

src/js/utils/clock.ts
robwalch: I can change to `Performance | GenericObject`.

src/js/utils/defaults.ts
robwalch: Babel handles it. Recent eslint rule changes don't give me a choice:
https://github.com/jwplayer/jwplayer/commit/befe9f10da2f171b3d51473887633968eb8ff65f
https://github.com/jwplayer/jwplayer/commit/7404a3e284b094e8a70b870826cf46ed48abf980
```
"prefer-rest-params": "error",
"prefer-spread": "error",
```

src/js/utils/cancelable.ts
mamaddox: I know @zetagame and I had briefly chatted about `any` and `void` and after looking a bit more into things, it seems I was wrong in assuming `void` was not of type `any`.

[Void Type](https://github.com/Microsoft/TypeScript/blob/master/doc/spec.md#325-the-void-type)

> The Void type is a subtype of the Any type and a supertype of the Null and Undefined types, but otherwise Void is unrelated to all other types.

So this would be fine.

src/js/utils/defaults.ts
mamaddox: Can remove those if wanted. I took those rules from the recommended typescript-eslint rules list.

robwalch: Rest spreads are great! Doesn't this looklbetter?

mamaddox: It is beautiful

src/js/utils/clock.ts
```diff
@@ -1,6 +1,7 @@
 import { now as getTime } from 'utils/date';
+import { GenericObject } from '../types/generic.type';
 
-const performance = window.performance || {
+const performance: Performance | GenericObject = window.performance || {
```
jnatalzia: should we type `{ timing: GenericObject}` instead for the second half of the union?

src/js/utils/clock.ts
robwalch: done

src/js/utils/clock.ts
robwalch: Explained earlier https://github.com/jwplayer/jwplayer/pull/3622#discussion_r384792807

PR #3623, opened 2020-02-27
[JW8-10933] Typescript 4 provider files

mamaddox:
### This PR will...
Type:

- src/js/providers/bandwidth-monitor.js
- src/js/providers/constants.js
- src/js/providers/data-normalizer.js
- src/js/providers/html5-android-hls.js

See comments
#### Addresses Issue(s):

JW8-10933

### Checklist


src/js/providers/data-normalizer.ts
```diff
@@ -0,0 +1,15 @@
+export function qualityLevel(level: any): QualityLevel {
```
mamaddox: Assumed this was `any` because it looks like it can be a PlaylistItemSource or a flash/shaka/hls manifest level.

src/js/providers/bandwidth-monitor.ts
```diff
@@ -1,11 +1,12 @@
 import { BANDWIDTH_ESTIMATE } from 'events/events';
 import { isValidNumber } from 'utils/underscore';
+import { DefaultProvider } from 'types/generic.type';
 
-export default function BandwidthMonitor(provider, initialEstimate) {
-    let bandwidthMonitorInterval = null;
+export default function BandwidthMonitor(provider: DefaultProvider, initialEstimate: number): BandwidthMonitor {
```
mamaddox: Added this `DefaultProvider` type because it looks like only shaka and hlsjs providers use this.

mamaddox:
test this please

src/js/providers/bandwidth-monitor.ts
jnatalzia: we should make a note in the TS confluence doc about moving `DefaultProvider` to `providers/default` once that gets typed

src/js/providers/bandwidth-monitor.ts
```diff
@@ -18,11 +19,17 @@ export default function BandwidthMonitor(provider, initialEstimate) {
...
             clearInterval(bandwidthMonitorInterval);
         },
-        getEstimate() {
+        getEstimate(): number {
             return bandwidthEstimate;
         }
     };
 }
+
+export type BandwidthMonitor = {
```
jnatalzia: this is a personal style thing so up to you, but i would hoist these to the top of the file for ease of access

src/js/providers/data-normalizer.ts
jnatalzia: I might consider doing a union type between generic object and Quality level for this, that way we get type safety on the attrs that are there but allow others in as well.

example: https://github.com/jwplayer/jwplayer/pull/3602/files#diff-c1229db6acf267fd235b1924faf60e62R3

src/js/providers/html5-android-hls.ts
```diff
@@ -1,6 +1,7 @@
 import { OS, Browser } from 'environment/environment';
+import { PlaylistItemSource } from 'types/generic.type';
 
-export function isAndroidHls(source) {
+export function isAndroidHls(source: PlaylistItemSource): boolean | null {
```
jnatalzia: same comment about updating the known tech debt in that confluence page

src/js/providers/bandwidth-monitor.ts
mamaddox: 👍 I also added a note for these in the TS spreadsheet by the files

src/js/providers/bandwidth-monitor.ts
```diff
@@ -1,11 +1,12 @@
...
-export default function BandwidthMonitor(provider, initialEstimate) {
-    let bandwidthMonitorInterval = null;
+export default function BandwidthMonitor(provider: DefaultProvider, initialEstimate: number): BandwidthMonitor {
+    let bandwidthMonitorInterval: number | undefined;
     let bandwidthEstimate = initialEstimate;
     return {
-        start() {
+        start(): void {
             this.stop();
             setInterval(() => {
```
robwalch: Is it just me or is `bandwidthMonitorInterval` never being set?

If so this is a bug and it should be set here `bandwidthMonitorInterval = setInterval `

src/js/providers/data-normalizer.ts
```diff
@@ -0,0 +1,17 @@
+import { GenericObject } from 'types/generic.type';
+
+export type QualityLevel = {
+    label: string;
+    bitrate?: number;
+    width?: number;
+    height?: number;
+};
+
+export function qualityLevel(level: QualityLevel | GenericObject): QualityLevel {
```
jnatalzia: `QualityLevel & GenericObject` ensures that it at least following `QualityLevel` and allows more props, doing the or means a blank object could be passed in and pass

mamaddox: That makes more sense

src/js/providers/bandwidth-monitor.ts
```diff
@@ -1,13 +1,20 @@
...
+
+export default function BandwidthMonitor(provider: DefaultProvider, initialEstimate: number): BandwidthMonitor {
+    let bandwidthMonitorInterval: number | undefined;
     let bandwidthEstimate = initialEstimate;
     return {
-        start() {
+        start(): void {
             this.stop();
-            setInterval(() => {
+            bandwidthMonitorInterval = window.setInterval(() => {
```
mamaddox: It's defaulting to using NodeJS typings so this was needed to explicitly use the window.setInterval

robwalch: We do the same thing in hls.js (`window.setInterval` or `self.setInterval` and same for `setTimeout`).

PR #3632, opened 2020-03-06
[JW8-10993] Typescript 7 files

mamaddox:
### This PR will...
Typescript the following files

- src/js/view/error-container.js
- src/js/view/controls/next-display-icon.js
- src/js/view/controls/templates/display-container.js
- src/js/view/controls/templates/display-icon.js
- src/js/view/controls/templates/info-overlay.js
- src/js/view/controls/templates/nextup.js
- src/js/api/errors.js

Had to add some additional attributes to `model` and `simplemodel` so please check
#### Addresses Issue(s):

JW8-10993

### Checklist


src/js/api/errors.ts
```diff
@@ -161,18 +165,23 @@ export const MSG_TECHNICAL_ERROR = 'technicalError';
  * @param {sourceError} [Error] - The lower level error, caught by the player, which resulted in this error.
  */
 export class PlayerError {
```
jnatalzia: can we use this class as the type when implementing or should we create a separate interface?

src/js/api/errors.ts
```diff
@@ -161,18 +165,23 @@ export const MSG_TECHNICAL_ERROR = 'technicalError';
...
-    constructor(key, code, sourceError = null) {
+    code: ErrorCode;
+    sourceError: Error | PlayerError | null;
```
jnatalzia: does `sourceError?: Error | PlayerError` work here?

src/js/api/errors.ts
```diff
@@ -161,18 +165,23 @@ export const MSG_TECHNICAL_ERROR = 'technicalError';
...
+    key: ErrorKey | null = null;
+    message: string | null = null;
+
+    constructor(key: ErrorKey | null, code: ErrorCode, sourceError: Error | PlayerError | null = null) {
```
jnatalzia: `key?: ErrorKey, code:...`

src/js/api/errors.ts
```diff
@@ -184,21 +193,21 @@ export class PlayerError {
     }
 }
 
-export function convertToPlayerError(key, code, error) {
+export function convertToPlayerError(key: ErrorKey | null, code: ErrorCode, error: Error | PlayerError): PlayerError {
```
jnatalzia: `key?:...` for optional params here

src/js/templates/error.ts
```diff
@@ -1,6 +1,6 @@
 import type { HTMLTemplateString } from 'types/generic.type';
 
-export default (id: string, message: string, errorCode: string, code: string): HTMLTemplateString => {
+export default (id: string, message: string | null, errorCode: string, code: string): HTMLTemplateString => {
```
jnatalzia: hmm this is an interesting case - can you use the optional `?` and then have required params after? prob not right? if so good as is

src/js/view/controls/next-display-icon.ts
```diff
@@ -1,21 +1,25 @@
 import UI from 'utils/ui';
+import type { CoreModel, PlayerAPI, PlaylistItemType } from 'types/generic.type';
 
 export default class NextDisplayIcon {
```
jnatalzia: same q about using this class as a type vs creating an interface and how that works because i truly don't know

src/js/api/errors.ts
zetagame: That would be equivalent to Error | PlayerError | undefined.

src/js/templates/error.ts
mamaddox: yeah. can't have optional params before required params.

src/js/api/errors.ts
mamaddox: since code is required, can't have optional params before. seems like we usually pass in null for our usages.

src/js/api/errors.ts
mamaddox: same as above, unless `code` is not required? though then that would allow for empty errors.

src/js/api/errors.ts
mamaddox: I'm pretty sure we can use it as the type: [#TS Class - Advanced Techniques](https://www.typescriptlang.org/docs/handbook/classes.html#advanced-techniques)

> When you declare a class in TypeScript, you are actually creating multiple declarations at the same time. The first is the type of the instance of the class.

They also have an example on that page.

yarn.lock
```diff
@@ -8203,10 +8203,10 @@ typedarray@^0.0.6:
   version "0.0.6"
   resolved "https://registry.yarnpkg.com/typedarray/-/typedarray-0.0.6.tgz#867ac74e3864187b1d3d47d996a78ec5c8830777"
 
-typescript@3.7.3:
-  version "3.7.3"
-  resolved "https://registry.yarnpkg.com/typescript/-/typescript-3.7.3.tgz#b36840668a16458a7025b9eabfad11b66ab85c69"
-  integrity sha512-Mcr/Qk7hXqFBXMN7p7Lusj1ktCBydylfQM/FZCk5glCNQJrCUKPkMHdo9R0MTFWsC/4kPFvDS0fDPvukfCkFsw==
+typescript@3.8.3:
```
jnatalzia: prob worth rebase-ing against master and seeing if these lock changes go away

mamaddox:
test this please

PR #3633, opened 2020-03-06
[JW8-10991] Typescripting files

zetagame:
### This PR will...
TS
SVG Module Type Definition
Slightly stricter typing for helpers for readability.

TS

TS

#### Addresses Issue(s):

JW8-10991

### Checklist

### DIFF
--- a/src/js/types/custom.d.ts
+++ b/src/js/types/custom.d.ts
@@ -1,0 +1,4 @@
+declare module '*.svg' {
+  const content: any;
+  export default content;
+}
--- a/src/js/view/controls/components/button.ts
+++ b/src/js/view/controls/components/button.ts
@@ -1,7 +1,24 @@
 import UI from 'utils/ui';
 import svgParse from 'utils/svgParser';
+import helpers from 'utils/helpers';
+import { PlayerAPI } from 'types/generic.type';
 
-export default function (icon, apiAction, ariaText, svgIcons) {
+export type Button = {
+    ui: UI;
+    element: () => HTMLElement;
+    toggle: (condition: any) => void;
+    show: () => void;
+    hide: () => void;
+}
+
+export type ButtonName = string;
+
+export default function (
+    icon: string, 
+    apiAction: (value: unknown) => PlayerAPI, 
+    ariaText: string, 
+    svgIcons: Array<SVGElement|string>
+): Button {
     const element = document.createElement('div');
     element.className = 'jw-icon jw-icon-inline jw-button-color jw-reset ' + icon;
     element.setAttribute('role', 'button');
@@ -13,7 +30,7 @@ export default function (icon, apiAction, ariaText, svgIcons) {
 
     element.style.display = 'none';
 
-    const ui = new UI(element).on('click tap enter', apiAction || function() {});
+    const ui: UI = new UI(element).on('click tap enter', apiAction || helpers.noop);
 
     if (svgIcons) {
         Array.prototype.forEach.call(svgIcons, svgIcon => {
@@ -27,21 +44,22 @@ export default function (icon, apiAction, ariaText, svgIcons) {
 
     return {
         ui,
-        element: function() {
+        element: function(): HTMLElement {
             return element;
         },
-        toggle: function(m) {
+        toggle: function(this: Button, m: any): void {
             if (m) {
                 this.show();
             } else {
                 this.hide();
             }
         },
-        show: function() {
+        show: function(): void {
             element.style.display = '';
         },
-        hide: function() {
+        hide: function(): void {
             element.style.display = 'none';
         }
     };
 }
+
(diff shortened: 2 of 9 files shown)


src/js/types/custom.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module '*.svg' {
```
jnatalzia: nice

src/js/types/custom.d.ts
```diff
@@ -0,0 +1,4 @@
...
+  const content: any;
```
jnatalzia: should content be type `string` since that's what all the svgs should be?

src/js/view/controls/components/button.ts
```diff
@@ -13,7 +30,7 @@ export default function (icon, apiAction, ariaText, svgIcons) {
 
     element.style.display = 'none';
 
-    const ui = new UI(element).on('click tap enter', apiAction || function() {});
+    const ui: UI = new UI(element).on('click tap enter', apiAction || helpers.noop);
```
jnatalzia: nice

src/js/view/controls/components/menu/category-button.ts
```diff
@@ -1,9 +1,14 @@
-import button from 'view/controls/components/button';
+import button, { Button } from 'view/controls/components/button';
 import { cloneIcon } from 'view/controls/icons';
 import { SimpleTooltip } from 'view/controls/components/simple-tooltip';
 import { createElement } from 'utils/dom';
+import { Tooltip, Menu } from 'types/generic.type';
 
-const categoryButton = (menu) => {
+type CategoryButton = Button & {
```
jnatalzia: ugh i keep waffling on interface vs type. here you _could_ do `interface CategoryButton extends Button { tooltip: Tooltip; }` but these are _stateless_ components right? so type makes more sense?

either way i'm fine with just thinking out loud

src/js/view/controls/components/menu/category-button.ts
```diff
@@ -32,9 +37,11 @@ const categoryButton = (menu) => {
     buttonElement.setAttribute('aria-expanded', 'false');
     buttonElement.setAttribute('aria-haspopup', 'true');
     buttonElement.setAttribute('aria-controls', menu.el.id);
+
     if (!('ontouchstart' in window)) {
-        menuCategoryButton.tooltip = SimpleTooltip(buttonElement, name, title);
+        (menuCategoryButton as CategoryButton).tooltip = SimpleTooltip(buttonElement, name, title);
```
jnatalzia: can you do `const menuCategoryButton: CategoryButton =...` above instead?

src/js/view/controls/components/menu/close-button.ts
```diff
@@ -1,8 +1,9 @@
-import button from 'view/controls/components/button';
+import button, { Button } from 'view/controls/components/button';
 import { cloneIcon } from 'view/controls/icons';
+import { GenericObject, Menu } from 'types/generic.type';
 
-export default (settingsMenu, localization) => {
-    const closeButton = button(
+export default (settingsMenu: Menu, localization: GenericObject): Button => {
```
jnatalzia: is it worth making a placeholder type for the localization stuff?

src/js/view/controls/components/menu/utils.ts
```diff
@@ -1,5 +1,32 @@
-const colorValues = ['#ffffff', '#000000', '#ff0000', '#00ff00', '#0000ff', '#ffff00', 'ff00ff', '#00ffff'];
-const fontValues = [
+export interface CaptionsLocalization {
```
jnatalzia: 👏

src/js/view/controls/templates/menu/menu-item.ts
```diff
@@ -1,7 +1,17 @@
-import ARROW_RIGHT_ICON from 'assets/SVG/arrow-right.svg';
-export const itemTemplate = (content) => `<button type="button" class="jw-reset-text jw-settings-content-item" aria-label="${content}" dir="auto">${content}</button>`;
+import * as ARROW_RIGHT_ICON from 'assets/SVG/arrow-right.svg';
```
jnatalzia: have to change due to the declared module setup?

src/js/types/custom.d.ts
zetagame: Could be a string, could be an image, maybe others? I'm just using the suggested definition webpack's docs provided.

src/js/view/controls/components/menu/category-button.ts
zetagame: I liked type because if I hover over Button in vscode as a type it shows all of the required properties and their types/call sigs, but as an interface I would have to click into the file. I prefer interface organizationally.

src/js/view/controls/components/menu/category-button.ts
zetagame: ButtonCategoryButton is a member of Button, so this is a safe cast. That being said, I like your suggestion more because the return type will always be CategoryButton

src/js/view/controls/templates/menu/menu-item.ts
zetagame: Good call. Left in from when I was trying various fixes. Doesn't change much but I'd prefer not to make unneeded changes. Reverted!

zetagame:
test this please

PR #3631, opened 2020-03-06
JW8-11009 Upgrade TS to 3.8 - Add import type

jnatalzia:
~Blocked until https://github.com/babel/babel/pull/11171 is merged~

### This PR will...
Upgrade typescript to 3.8

### Why is this Pull Request needed?
Lots of cool features to take advantage of

### Are there any points in the code the reviewer needs to double check?
Did I miss any `import type`s

### Are there any Pull Requests open in other repos which need to be merged with this?
https://github.com/jwplayer/jwplayer-commercial/pull/7467

#### Addresses Issue(s):

JW8-11009

### Checklist
- [ ] Jenkins builds and unit tests are passing
- [ ] I have reviewed the automated results

PR #3634, opened 2020-03-06
[JW8-10992] TypeScript Model, SimpleModel and Events

robwalch:
### This PR will...
Convert `Model`, `SimpleModel` and `Events` to TypeScript.

#### Addresses Issue(s):
JW8-10992

### Checklist

### DIFF
--- a/src/js/controller/events-middleware.ts
+++ b/src/js/controller/events-middleware.ts
@@ -1,6 +1,7 @@
-import type { GenericObject, CoreModel } from 'types/generic.type';
+import type { GenericObject } from 'types/generic.type';
+import type Model from './model';
 
-export default function middleware(model: CoreModel, type: string, currentState: GenericObject): GenericObject {
+export default function middleware(model: Model, type: string, currentState: GenericObject): GenericObject {
     let newState: GenericObject = currentState;
 
     switch (type) {
--- a/src/js/controller/model.ts
+++ b/src/js/controller/model.ts
@@ -1,13 +1,61 @@
 import { OS } from 'environment/environment';
 import SimpleModel from 'model/simplemodel';
 import { INITIAL_PLAYER_STATE, INITIAL_MEDIA_STATE } from 'model/player-model';
-import { STATE_IDLE } from 'events/events';
+import { InternalPlayerState, STATE_IDLE } from 'events/events';
 import { isValidNumber, isNumber } from 'utils/underscore';
 import { seconds } from 'utils/strings';
 import Providers from 'providers/providers';
 import { StreamType } from '../providers/utils/stream-type';
 import { DefaultProvider, GenericObject, PlaylistItemType, TextTrackLike } from '../types/generic.type';
 import { QualityLevel } from '../providers/data-normalizer';
+import Item from 'playlist/item';
+
+type AutoStart = boolean | 'viewable';
+
+type PlayerModelAttributes = {
+    _destroyed: boolean;
+    audioMode: boolean;
+    autostart: AutoStart;
+    autostartMuted: boolean;
+    bandwidthEstimate: number;
+    bitrateSelection: number | null;
+    captionLabel: string;
+    captionsIndex: number;
+    captionsTrack: TextTrackLike;
+    controlsEnabled: boolean;
+    defaultPlaybackRate: number;
+    dvrSeekLimit: number;
+    flashBlocked: boolean;
+    flashThrottle?: boolean;
+    fullscreen: boolean;
+    instreamMode: boolean;
+    item: number;
+    itemMeta: GenericObject;
+    itemReady: boolean;
+    mediaModel: MediaModel;
+    minDvrWindow: number;
+    mute: boolean;
+    nextUp: Item;
+    playbackRate: number;
+    playlist: PlaylistItemType[];
+    playlistItem: PlaylistItemType | null;
+    playOnViewable: boolean;
+    playRejected: boolean;
+    provider: DefaultProvider;
+    qualityLabel: string;
+    renderCaptionsNatively: boolean;
+    state: InternalPlayerState;
+    streamType: StreamType;
+    supportsPlaybackRate: boolean;
+    volume: number;
+}
+
+interface Model {
+    readonly attributes: Partial<PlayerModelAttributes>;
+    addAttributes(attributes: Partial<PlayerModelAttributes>): void;
+    get<K extends keyof PlayerModelAttributes>(attr: K): PlayerModelAttributes[K];
+    set<K extends keyof PlayerModelAttributes>(attr: K, val: PlayerModelAttributes[K]): void;
+}
 
 // Represents the state of the player
 class Model extends SimpleModel {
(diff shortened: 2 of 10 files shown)


src/js/model/simplemodel.ts
```diff
@@ -1,20 +1,25 @@
 import Events from 'utils/backbone.events';
+import { GenericObject } from '../types/generic.type';
 
+type ModelAttributes = {
+    [name: string]: any;
+}
 
 export default class SimpleModel extends Events {
+    public readonly attributes: ModelAttributes;
```
jnatalzia: 👍

src/js/types/generic.type.ts
```diff
@@ -17,3 +17,5 @@ export type PlayerAPI = any;
 export type PlaylistItemSource = any;
 
 export type DefaultProvider = any;
+
+export type TextTrackLike = any;
```
jnatalzia: i assume this is different than `item/track.ts`?

src/js/utils/backbone.events.ts
```diff
@@ -11,9 +11,38 @@
 
 // Mixin module modified into a class which can be extended
 
-const slice = [].slice;
+type EventsAction = 'on'|'once'|'off'|'trigger';
 
-export default class Events {
+type EventListenerDictionary = {
+    [name: string]: EventListener;
```
jnatalzia: `name: EventsAction`?

src/js/utils/backbone.events.ts
```diff
@@ -22,13 +51,13 @@ export default class Events {
      * @param {any} [context] - The context to apply to the callback's function invocation.
      * @returns {any} `this` context for chaining.
      */
-    on(name, callback, context) {
+    on(name: string | EventListenerDictionary, callback: EventCallback, context?: any): Events {
```
jnatalzia: do we not get these defs for free with the `implements EventsInterface`? i remember @zetagame saying something about this as well

src/js/utils/backbone.events.ts
```diff
@@ -166,14 +198,15 @@ const eventSplitter = /\s+/;
 // Implement fancy features of the Events API such as multiple event
 // names `"change blur"` and jQuery-style event maps `{change: action}`
 // in terms of the existing API.
-function eventsApi(obj, action, name, rest) {
+function eventsApi(obj: Events, action: EventsAction, name: string | EventListenerDictionary | undefined, rest: any[]): name is string {
```
jnatalzia: `name?: string | EventListenerDictionary`?

src/js/types/generic.type.ts
robwalch: It comes primarily from https://github.com/jwplayer/jwplayer/blob/master/src/js/providers/tracks-mixin.js#L232 where most of the time the provider is using native textTracks. It's not from the playlist item. It should be typed once that mixin is typed.

src/js/utils/backbone.events.ts
robwalch: No, this is the map of listeners - ex: `"ready": onReady`.

src/js/utils/backbone.events.ts
robwalch: I get an ESLint `tslint:typedef` error when removing these type definitions.

src/js/utils/backbone.events.ts
robwalch: `name` is optional in `off`. So if you call `object.off()`, then `eventsApi(this, 'off', undefined, [undefined, undefined])` is invoked.

src/js/types/generic.type.ts
jnatalzia: can we add this to maria's ts doc where we're tracking this stuff? https://jwplayer.atlassian.net/wiki/spaces/JW8/pages/607453187/Typescript+Standards

thanks!

robwalch: I don't see any of these in that doc, and they are all tech-debt. Why don't we just comment above each of these type definitions rather than keeping another document?

```
export type CoreModel = any;

export type PlaylistItemType = any;

export type HTMLTemplateString = string;

export type PlayerAPI = any;

export type PlaylistItemSource = any;

export type DefaultProvider = any;

export type Menu = GenericObject;

export type Tooltip = GenericObject;

export type TextTrackLike = TextTrack & {
    _id: string;
    data: Array<any>;
    name: string;
};
```

jnatalzia: For things in one file I think it's easier but my concern with doing comments outside this case is they'd be impossible to find after. Maybe we do `// TS:` so it's easily searchable and scrap the doc?

src/js/controller/model.ts
```diff
@@ -1,13 +1,69 @@
...
+    state: InternalPlayerState;
+    streamType: StreamType;
+    supportsPlaybackRate: boolean;
+    volume: number;
+}
+
+interface Model {
+    readonly attributes: Partial<PlayerModelAttributes>;
+    addAttributes(attributes: Partial<PlayerModelAttributes>): void;
+    get<K extends keyof PlayerModelAttributes>(attr: K): PlayerModelAttributes[K];
```
jnatalzia: whoa 🤯

robwalch: These docs helped:
https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-1.html
https://www.typescriptlang.org/docs/handbook/utility-types.html

robwalch: Also the type checking for this interface only works by naming the interface after the class. When I tried `class Model extends SimpleModel implements ModelInterface` the interface of `SimpleModel` took precedence.

src/js/types/generic.type.ts
robwalch: Isn't this that one file?

jnatalzia: It _should_ be, the doc was meant to catch quirks or problems that might surface beyond having shared untyped generics here but maybe that's overengineering

jnatalzia:
> > how do you feel about updating references in the code to the generic.type's CoreModel to reference your new model type
> 
> I started doing this, but reverted/rebased. We should not import the player model (core.js) into embed modules (jwplayer.js) until we're [using TS 3.8 and can do type-only imports](https://github.com/jwplayer/jwplayer/pull/3631).

ahh yep makes sense,. wonder what the timing is on the babel change. another alternative is to use the TS compiler to go from es6 to es5 then use regular babel preset env and avoid the babel-ts dep altogether to get us there faster

robwalch:
> ah yep makes sense

modified the two core modules used in controller https://github.com/jwplayer/jwplayer/pull/3634/commits/7248cef1205a5fefc5c766afe0d1751ed7eb949b

The plugins one we'll need to import as type, and the other reference is the yet to be typed ViewModel which builds on top of Model.

src/js/utils/backbone.events.ts
robwalch: I think that's the difference with separating your class and interface (where interface has the same name) from implementing an interface. Either way, I removed `EventsInterface` as it wasn't being used.

src/js/controller/events-middleware.ts
```diff
@@ -1,6 +1,7 @@
-import { GenericObject, CoreModel } from 'types/generic.type';
+import { GenericObject } from 'types/generic.type';
+import Model from './model';
```
jnatalzia: did we see if this gets stripped out even without `import type` - little unclear on your previous comment

robwalch: It doesn't matter for modules used in the controller. What we want to avoid is importing model into modules imported in the embed.

src/js/controller/model.ts
zetagame: awesome!

robwalch:
Rebased against TS 3.8 update https://github.com/jwplayer/jwplayer/commit/40b9b524fefe27b4f3537e050199a550791c4cc8

PR #3636, opened 2020-03-14
Type default provider

jnatalzia:
### This PR will...
Add types to the default provider

TS

Comments mark

Nah

#### Addresses Issue(s):

JW8-11027

### Checklist


src/js/providers/default.ts
```diff
@@ -0,0 +1,186 @@
+import { PLAYER_STATE, MEDIA_TYPE } from 'events/events';
+import { GenericObject } from 'types/generic.type';
+
+const noop: () => void = function(): void { /* noop */ };
+const returnFalse: () => boolean = (() => false);
+const getNameResult: { name: string } = { name: 'default' };
+const returnName: () => { name: string } = (() => getNameResult);
+
+type SeekRangeType = {
```
jnatalzia: where should these live?

src/js/providers/default.ts
```diff
@@ -0,0 +1,186 @@
...
+    start: number;
+    end: number;
+};
+
+type SourceType = {
```
jnatalzia: seems like there's probably more that goes into this, should i move to `generic.type.ts`?

jnatalzia:
@robwalch have some questions on this one for you on monday, wondering the best approach to some of this stuff.

src/js/types/generic.type.ts
```diff
@@ -23,3 +23,13 @@ export type Menu = GenericObject;
 export type Tooltip = GenericObject;
 
 export type Localization = GenericObject;
+
+export type SourceObj = {
+    type: string;
+    mimeType: string;
+};
+
+export type SeekRange = {
```
jnatalzia: this one i'm not sure where to house - `source` has a file that will get types but seek ranges are sort of spread throughout. I almost think this generic file might be the best place for em. thoughts?

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+    destroy: () => void; // frees memory
+
+    setVisibility: () => void;
+    setFullscreen: () => void;
+    getFullscreen: () => boolean;
+    supportsFullscreen: () => boolean;
+
+    // If setContainer has been set; this returns the element.
+    //  It's value is used to determine if we should remove the <video> element when setting a new provider.
+    getContainer: () => void;
```
jnatalzia: when these get overridden by the provider in question, they will have return values - should we type those return values as part of those extending interfaces or do it here

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+    getQualityLevels: () => void;
+    getCurrentQuality: () => void;
+    setCurrentQuality: () => void;
+
+    getAudioTracks: () => void;
+
+    getCurrentAudioTrack: () => void;
+    setCurrentAudioTrack: () => void;
+
+    getSeekRange: () => SeekRange;
```
robwalch: I think this module is a better place to define the `SeekRange` type.

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
 import { PLAYER_STATE, MEDIA_TYPE } from 'events/events';
+import { GenericObject, SourceObj, SeekRange } from 'types/generic.type';
+
+const noop: () => void = function(): void { /* noop */ };
+const returnFalse: () => boolean = (() => false);
+const getNameResult: { name: string } = { name: 'default' };
+const returnName: () => { name: string } = (() => getNameResult);
+interface DefaultProvider {
```
robwalch: What do you think about exporting this interface and making all providers implement it?

`export interface Provider`

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
 import { PLAYER_STATE, MEDIA_TYPE } from 'events/events';
+import { GenericObject, SourceObj, SeekRange } from 'types/generic.type';
+
+const noop: () => void = function(): void { /* noop */ };
+const returnFalse: () => boolean = (() => false);
+const getNameResult: { name: string } = { name: 'default' };
+const returnName: () => { name: string } = (() => getNameResult);
```
robwalch: np - new line before `interface` and `class` (or after `const` lists) - maybe we could add a lint rule for this?

src/js/providers/default.ts
robwalch: This works with `setContainer(element)`. See video-actions-mixin for actual implementation.

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+    getContainer: () => void;
+
+    // Sets the parent element; causing provider to append <video> into it
+    setContainer: () => void;
+
+    getName: () => { name: string };
+
+    getQualityLevels: () => void;
+    getCurrentQuality: () => void;
+    setCurrentQuality: () => void;
```
robwalch: Every setter should have at least one argument

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+
+    getAudioTracks: () => void;
+
+    getCurrentAudioTrack: () => void;
+    setCurrentAudioTrack: () => void;
```
robwalch: Every setter should have at least one argument

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+    setFullscreen: () => void;
+    getFullscreen: () => boolean;
+    supportsFullscreen: () => boolean;
+
+    // If setContainer has been set; this returns the element.
+    //  It's value is used to determine if we should remove the <video> element when setting a new provider.
+    getContainer: () => void;
+
+    // Sets the parent element; causing provider to append <video> into it
+    setContainer: () => void;
```
robwalch: missing element argument

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+    stop: () => void;
+    volume: () => void;
+    mute: () => void;
+    seek: () => void;
+    resize: () => void;
+    remove: () => void; // removes from page
+    destroy: () => void; // frees memory
+
+    setVisibility: () => void;
+    setFullscreen: () => void;
```
robwalch: Every setter should have at least one argument

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+    supports: () => boolean;
+    play: () => void;
+    pause: () => void;
+    preload: () => void;
+    load: () => void;
+    stop: () => void;
+    volume: () => void;
+    mute: () => void;
+    seek: () => void;
+    resize: () => void;
```
robwalch: `resize(with, height, stretching?)`

Search in /jwplayer/src/js/providers for implementation.

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+    state?: string;
+    supports: () => boolean;
+    play: () => void;
+    pause: () => void;
+    preload: () => void;
+    load: () => void;
+    stop: () => void;
+    volume: () => void;
+    mute: () => void;
+    seek: () => void;
```
robwalch: missing arg(s)

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+interface DefaultProvider {
+    state?: string;
+    supports: () => boolean;
+    play: () => void;
+    pause: () => void;
+    preload: () => void;
+    load: () => void;
+    stop: () => void;
+    volume: () => void;
+    mute: () => void;
```
robwalch: missing arg(s)

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+const returnName: () => { name: string } = (() => getNameResult);
+interface DefaultProvider {
+    state?: string;
+    supports: () => boolean;
+    play: () => void;
+    pause: () => void;
+    preload: () => void;
+    load: () => void;
+    stop: () => void;
+    volume: () => void;
```
robwalch: missing arg(s)

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+const returnFalse: () => boolean = (() => false);
+const getNameResult: { name: string } = { name: 'default' };
+const returnName: () => { name: string } = (() => getNameResult);
+interface DefaultProvider {
+    state?: string;
+    supports: () => boolean;
+    play: () => void;
+    pause: () => void;
+    preload: () => void;
+    load: () => void;
```
robwalch: item arg

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+const noop: () => void = function(): void { /* noop */ };
+const returnFalse: () => boolean = (() => false);
+const getNameResult: { name: string } = { name: 'default' };
+const returnName: () => { name: string } = (() => getNameResult);
+interface DefaultProvider {
+    state?: string;
+    supports: () => boolean;
+    play: () => void;
+    pause: () => void;
+    preload: () => void;
```
robwalch: item arg

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
 import { PLAYER_STATE, MEDIA_TYPE } from 'events/events';
+import { GenericObject, SourceObj, SeekRange } from 'types/generic.type';
+
+const noop: () => void = function(): void { /* noop */ };
+const returnFalse: () => boolean = (() => false);
+const getNameResult: { name: string } = { name: 'default' };
+const returnName: () => { name: string } = (() => getNameResult);
+interface DefaultProvider {
+    state?: string;
+    supports: () => boolean;
```
robwalch: missing arg

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+    setCurrentQuality: () => void;
+
+    getAudioTracks: () => void;
+
+    getCurrentAudioTrack: () => void;
+    setCurrentAudioTrack: () => void;
+
+    getSeekRange: () => SeekRange;
+
+    setPlaybackRate: () => void;
```
robwalch: rate arg

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+    getPlaybackRate: () => number;
+    getBandwidthEstimate: () => number | null;
+    getLiveLatency: () => number | null;
+    setControls: () => void;
```
robwalch: bool arg (only used in flash provider)

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
 
-const noop = function() {};
-const returnFalse = (() => false);
-const getNameResult = { name: 'default' };
-const returnName = (() => getNameResult);
+    attachMedia: () => void;
+    detachMedia: () => void;
+    init: () => void;
```
robwalch: item arg

src/js/providers/default.ts
```diff
@@ -1,9 +1,68 @@
...
+
+    setState: (state: string) => void;
+
+    sendMediaType: (sources: Array<SourceObj>) => void;
+
+    getDuration: () => number;
+
+    trigger: (evt: string, obj: GenericObject) => void;
```
robwalch: This interface should extend the Events interface. You don't need to redefine trigger, on etc..

src/js/types/generic.type.ts
robwalch: In DefaultProvider is fine. I don't like how the defaults are implemented now, but the idea is all providers extend it so it makes sense for types and _the_ provider interface be defined there.

src/js/providers/default.ts
jnatalzia: 👍 yeah i like this

src/js/providers/default.ts
jnatalzia: oof this is going to take a lot of refactoring of the noops below. gonna dive back in

robwalch: Yeah I hesitated to give you this feedback. On one hand the purpose of this module is to "plug holes" in provider implementations by providing a base no-op implementation. On the other, more ideally, it should be a base class and interface for providers to implement.

jnatalzia: Came up with a better solution using extends to retain type safety but will be typing these funcs as implemented in this pr as well

src/js/providers/default.ts
```diff
@@ -0,0 +1,238 @@
...
+import type { VideoAttachedInt } from 'providers/video-attached-mixin';
+import type Events from 'utils/backbone.events';
+import type PlaylistItem from 'playlist/item';
+
+const noop: () => void = function(): void { /* noop */ };
+const returnFalse: () => boolean = (() => false);
+const getNameResult: { name: string } = { name: 'default' };
+const returnName: () => { name: string } = (() => getNameResult);
+
+export interface ImplementedProvider {
```
jnatalzia, line 14: would love your thoughts here @robwalch  on things we should mandate for external provider integration vs things defined on our internal mixins (resize is a great example - it's defined in the mixin so it does not need to be defined given the union type)

src/js/providers/default.ts
```diff
@@ -0,0 +1,238 @@
...
+
+    seek: (seekPos: number) => void;
+
+    setVisibility: (isVisible: boolean) => void;
+
+    setFullscreen: (isFullscreen: boolean) => void;
+    getFullscreen: () => boolean;
+    supportsFullscreen: () => boolean;
+
+    // to be updated when maria's work merges
```
jnatalzia, line 47: quality levels is underway

src/js/providers/default.ts
robwalch: `ProviderWithMixins` and `ImplementedProvider` (or `ProviderBaseDependantOnMixins`) should be defined in another file, so that `DefaultProvider` only defines the interface required of any provider.

jnatalzia: @robwalch , what do you suggest as far as a diff location? I think having it in default shows the additional arguments that `DefaultProvider` does not enumerate. It also gives indications of what we do on our end to integrate with the base interface which I can see as helpful.

Thoughts?

robwalch: I think the mixins are specific to video element playback and TextTrack handling in our providers, and not relevant to anyone building a provider externally unless they also import the mixins from open source. The mixins may work together and with one of our providers but they should be in no way shape or form relied upon in the player of any provider.

robwalch: The point is to separate what public methods of a provider are used outside of the providers/ modules by the player, from the methods that could be private or protected which are specific to our own provider implementations.

- Types for event objects emitted by providers? Sure put those in DefaultProvider
- Mixin methods used by other mixins or html5 provider? Nobody needs that but us.

src/js/controller/model.ts
```diff
@@ -6,9 +6,10 @@ import { isValidNumber, isNumber } from 'utils/underscore';
 import { seconds } from 'utils/strings';
 import Providers from 'providers/providers';
 import type { StreamType } from 'providers/utils/stream-type';
-import type { DefaultProvider, GenericObject, TextTrackLike, Localization } from 'types/generic.type';
+import type { GenericObject, TextTrackLike, Localization } from 'types/generic.type';
 import type { QualityLevel } from 'providers/data-normalizer';
 import type PlaylistItem from 'playlist/item';
+import type { ProviderWithMixins } from 'providers/default';
```
robwalch, line 12: An external provider would not be a `ProviderWithMixins` but at runtime we will be dealing with external providers that do not use our mixins, so this is confusing.

src/js/providers/bandwidth-monitor.ts
```diff
@@ -1,14 +1,14 @@
 import { BANDWIDTH_ESTIMATE } from 'events/events';
 import { isValidNumber } from 'utils/underscore';
-import type { DefaultProvider } from 'types/generic.type';
+import type { ProviderWithMixins } from './default';
```
robwalch, line 3: This module depends on providers implementing two methods, `trigger` and `getBandwidthEstimate`. It's in src/js/providers so it's only imported by providers when can provide an estimate. It relies in no way on mixins. Based on what you have, it looks like `ImplementedProvider` would be the right type to use here.

src/js/providers/html5.ts
```diff
@@ -41,22 +44,32 @@ const HTML5_BASE_WARNING = 324000;
...
+    currentTime: number;
+    retries: number;
+    maxRetries: number;
+    loadAndParseHlsMetadata: boolean;
+    startDateTime: number;
+    setStartDateTime: (time: number) => void;
+    videoLoad: (this: HTMLVideoElement) => void;
+}
+
+function VideoProvider(this: HTML5Provider, _playerId: string, _playerConfig: GenericObject, mediaElement: HTMLVideoElement): HTML5Provider {
```
robwalch, line 72: With the `this: HTML5Provider` argument isn't the return type `void`?

src/js/providers/html5.ts
```diff
@@ -1012,15 +1028,17 @@ function VideoProvider(_playerId, _playerConfig, mediaElement) {
         return false;
     }
 
-    function clearTimeouts() {
+    function clearTimeouts(): void {
         clearTimeout(_staleStreamTimeout);
         _staleStreamTimeout = -1;
     }
+
+    return this;
```
robwalch, line 1036: We don't need to return `this`. The return type for function invoked as a constructor with the `new` keyword is `void` .

src/js/view/view-model.ts
```diff
@@ -1,8 +1,9 @@
 import SimpleModel from 'model/simplemodel';
 import type Model from 'controller/model';
 import type { MediaModel, PlayerModelAttributes, MediaModelAttributes } from 'controller/model';
-import type { DefaultProvider, GenericObject } from 'types/generic.type';
+import type { GenericObject } from 'types/generic.type';
 import type AdProgramController from 'program/ad-program-controller';
+import type { ProviderWithMixins } from 'providers/default';
```
robwalch, line 6: Use a more generic type for provider like `ImplementedProvider` here.

robwalch: We should strive to remove public access to the provider from models, but this is how they are used (if you search for `getVideo` throughout the projects:

- write to `.src` to reset hls.js state in ad-program-controller
- read-only access to `.setFullscreen` in view.js
- read-only access to `.resize` in view.js
- read-only access to `.getFullScreen` in view.js
- read-only access to `.setContainer` in view.js

The only odd one out is `src`. If we add that to `ImplementedProvider` for now that would be best since a public property in an interface (with a comment) would signal that it is nulled externally and not required by all providers.

src/js/providers/bandwidth-monitor.ts
jnatalzia: Nah `trigger` is part of the `Events` mixin but would also be required for external providers so i'm going to update the types holistically here, i'll use the external facing provider interface which should include the trigger method and leave `ImplementedProvider` (potentially to be renamed) as the provider base for our mixin setup internally

src/js/providers/default.ts
jnatalzia: yeah this is all great feedback, updating with an `ImplementedProvider` which exists as our externally facing provider interface, and building `ProviderWithMixins` for internal use (should be castable back and forth given common interface)

src/js/view/view-model.ts
jnatalzia: :+1: replacing

src/js/view/view-model.ts
```diff
@@ -1,8 +1,9 @@
 import SimpleModel from 'model/simplemodel';
 import type Model from 'controller/model';
 import type { MediaModel, PlayerModelAttributes, MediaModelAttributes } from 'controller/model';
-import type { DefaultProvider, GenericObject } from 'types/generic.type';
+import type { GenericObject } from 'types/generic.type';
```
robwalch, line 4: Is `GenericObject` still used?

src/js/types/generic.type.ts
```diff
@@ -39,3 +37,7 @@ export type MetadataEvent = {
     metadata: GenericObject;
     programDateTime?: string;
 };
+export type SourceObj = {
```
robwalch, line 40: `Source from 'source.ts'`?

src/js/providers/default.ts
```diff
@@ -0,0 +1,262 @@
...
+    getAudioTracks: () => SimpleAudioTrack[];
+
+    getSeekRange: () => SeekRange;
+
+    getPlaybackRate: () => number;
+    getLiveLatency: () => number | null;
+    setControls: () => void;
+    setState: (state: string) => void;
+
+    sendMediaType: (sources: Array<SourceObj>) => void;
```
robwalch, line 70: `Source from 'source.ts'`?

src/js/providers/default.ts
```diff
@@ -0,0 +1,262 @@
...
+    getLiveLatency: () => number | null;
+    setControls: () => void;
+
+    attachMedia: () => void;
+    detachMedia: () => void;
+    init: () => void;
+
+    setState: (state: string) => void;
+
+    sendMediaType: (sources: Array<SourceObj>) => void;
```
robwalch, line 147: `Source from 'source.ts'`?

src/js/providers/tracks-mixin.ts
```diff
@@ -72,19 +79,19 @@ interface TracksMixin extends DefaultProvider {
     getCurrentTextTrack: () => TextTrackLike | void;
     getSubtitlesTrack: () => number;
     addTextTracks: (tracksArray: PlaylistItemTrack[]) => TextTrackLike[];
-    setTextTracks: (tracks: TextTrackList) => void;
-    addTrackListeners: (tracks: TextTrackList) => void;
+    addTrackListeners: (tracks: TextTrackList | TextTrackLike[]) => void;
+    setTextTracks: (tracks: TextTrackList | TextTrackLike[] | null) => void;
```
robwalch, line 83: Where do these take `TextTrackLike[]`? 

That is odd esp for `addTrackListeners` because we're using `TextTrackList.addEventListener` - arrays don't have that.

src/js/providers/html5.ts
```diff
@@ -41,22 +44,32 @@ const HTML5_BASE_WARNING = 324000;
 
 const clearTimeout = window.clearTimeout;
 const _name = 'html5';
-const noop = function () {};
+const noop: () => void = function (): void { /* noop */ };
 
-function _setupListeners(eventsHash, videoTag) {
+function _setupListeners(eventsHash: GenericObject, videoTag: HTMLVideoElement): void {
```
robwalch, line 49: Not super important, but `eventsHash` could be typed something like `[key]: (event?: Event) => void` (excuse the pseudo syntax)

robwalch, line 49: We might allow audio tags soon. Guess we could save the renaming and typing for that to be official:  `media: HTMLMediaElement`

src/js/controller/model.ts
```diff
@@ -62,7 +63,9 @@ export type PlayerModelAttributes = {
     playOnViewable: boolean;
     playReason: PlayReason;
     playRejected: boolean;
-    provider: DefaultProvider;
+    provider?: {
+        name: string;
+    };
```
robwalch, line 68: good catch!

PR #3638, opened 2020-03-18
Utilize WebVTT polyfill internally but respect globals externally

jnatalzia:
### This PR will...
Update our WebVTT Polyfill with the following functionality
* Always utilize polyfill internally when _not_ rendering captions natively
* Only set WebVTT on the global scope if it is not present when the polyfill loads

With these changes, caption styling menu should work in all instances regardless of WebVTT status in the global scope.

If a user uses their own VTT polyfill, none of our caption styling functionality works because in our polyfill we have logic which helps us apply styling to the dynamically created captions elements.

No, but would love input on test cases. I tested:
* Firefox with & w/o `renderCaptionsNatively`
* Firefox with an external WebVTT polyfill loaded in before our polyfill
* Safari with `renderCaptionsNatively` both `true` and `false`

Nope

#### Addresses Issue(s):

JW8-11022

### Checklist

### DIFF
--- a/src/js/polyfills/webvtt.js
+++ b/src/js/polyfills/webvtt.js
@@ -954,9 +954,12 @@ WebVTT.processCues = function (window, cues, overlay, updateBoxPosition) {
     }());
 };
 
-let webVttExport = window.WebVTT;
-if (!webVttExport) {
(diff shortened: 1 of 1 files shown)


jnatalzia:
test this please

PR #3639, opened 2020-03-18
[JW8-11042] Guard against async item tampering in program-controller and ui modules

robwalch:
### This PR will...
Add guards to prevent exceptions in detach logic whenthe instream API and setPlaylistItem() are used externally

Backwards compatibility with custom async playlist item ads integrations. See http://player-develop-test-jenkins.longtailvideo.com/builds/lastSuccessfulBuild/archive/test/public/async-8-12-poc.html

#### Addresses Issue(s):
JW8-11042

### Checklist

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -66,6 +66,9 @@ export default class UI extends Events {
     }
 
     destroy() {
+        if (!this.el) {
+            return;
+        }
         this.off();
         if (USE_POINTER_EVENTS) {
             releasePointerCapture(this);
--- a/test/unit/background-loading-test.js
+++ b/test/unit/background-loading-test.js
@@ -180,12 +180,15 @@ describe('Background Loading', function () {
         it('destroys background media if there is an active foreground item', function () {
             programController._setActiveMedia(mediaController);
             programController.backgroundActiveMedia();
-            programController._setActiveMedia(mediaController);
+            const mockProvider2 = new MockProvider();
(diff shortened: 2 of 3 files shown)


src/js/utils/ui.js
```diff
@@ -66,6 +66,9 @@ export default class UI extends Events {
     }
 
     destroy() {
+        if (!this.el) {
```
zetagame: `this.off` before condition may be safer in case it returns. Really like this change.

jnatalzia:
test this please

src/js/utils/ui.js
robwalch: The issue is in off. So we need this here.

test/unit/background-loading-test.js
```diff
@@ -180,12 +180,15 @@ describe('Background Loading', function () {
...
+            const mockProvider2 = new MockProvider();
+            mockProvider2.video = document.createElement('video');
+            const mediaController2 = new MediaController(mockProvider2, model);
+            programController._setActiveMedia(mediaController2);
             programController._setActiveMedia = sinon.spy();
             mediaPool.recycle = sinon.spy();
 
             programController.restoreBackgroundMedia();
-            expect(programController._setActiveMedia.calledOnce).to.equal(false);
+            expect(programController._setActiveMedia.calledOnce).to.equal(false, 'called _setActiveMedia');
```
jnatalzia: nice!

PR #3641, opened 2020-03-23
Focus player on inactive only after action

jnatalzia:
### This PR will...
Update our focus logic to focus the player when the player is previously focused and the controls blur. _This is currently not a problem in Firefox, but Chrome handles it differently_.

When you click a control in chrome and the controls fade, the player loses focus

Left notes inline

Tested:
* Clicking controls, waiting, focus on
* Clicking controls, clicking elsewhere on the page, focus off
* Hovering over player without clicking anywhere on it, wait for controls fade, focus off
* Same three scenarios in both fullscreen and floating
* Fullscreen, click controls, esc to close player, focus on

Nah

#### Addresses Issue(s):

JW8-10904

### Checklist

### DIFF
--- a/src/js/view/controls/controls.js
+++ b/src/js/view/controls/controls.js
@@ -23,7 +23,6 @@ require('css/controls.less');
 const ACTIVE_TIMEOUT = OS.mobile ? 4000 : 2000;
 // Keys which bypass keyboard shortcuts being off
 const ALWAYS_ALLOWED_KEYS = [27];
-let shouldFocusOnInactive = false;
 
 ErrorContainer.cloneIcon = cloneIcon;
 instances.forEach(api => {
@@ -175,7 +174,7 @@ export default class Controls extends Events {
         const controlbar = this.controlbar = new Controlbar(api, model,
             this.playerContainer.querySelector('.jw-hidden-accessibility'));
         controlbar.on(USER_ACTION, () => {
-            shouldFocusOnInactive = true;
+            this.once('userInactive', () => this.focusPlayerElement(), this);
             this.userActive();
         });
         controlbar.on('nextShown', function (data) {
@@ -309,7 +308,6 @@ export default class Controls extends Events {
                     if (model.get('fullscreen')) {
                         api.setFullscreen(false);
                         this.playerContainer.blur();
-                        shouldFocusOnInactive = false;
                         this.userInactive();
                     } else {
                         const related = api.getPlugin('related');
@@ -417,7 +415,7 @@ export default class Controls extends Events {
 
         // Hide controls when focus leaves the player
         const blurCallback = (evt) => {
-            shouldFocusOnInactive = false;
+            this.off('userInactive', null, this);
             const focusedElement = evt.relatedTarget || document.querySelector(':focus');
             if (!focusedElement) {
                 return;
@@ -616,11 +614,6 @@ export default class Controls extends Events {
         this.showing = false;
         addClass(this.playerContainer, 'jw-flag-user-inactive');
         this.trigger('userInactive');
-
-        if (shouldFocusOnInactive) {
-            this.focusPlayerElement();
-            shouldFocusOnInactive = false;
-        }
     }
 
     addBackdrop() {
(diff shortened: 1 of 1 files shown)


src/js/view/controls/controls.js
```diff
@@ -102,7 +103,7 @@ export default class Controls extends Events {
 
         const touchMode = model.get('touchMode');
 
-        const focusPlayerElement = () => {
+        this.focusPlayerElement = () => {
```
jnatalzia: need this function for `userInactive` below but the reliance on the `model` means we need to define it here or refactor more

src/js/view/controls/controls.js
```diff
@@ -23,6 +23,7 @@ require('css/controls.less');
 const ACTIVE_TIMEOUT = OS.mobile ? 4000 : 2000;
 // Keys which bypass keyboard shortcuts being off
 const ALWAYS_ALLOWED_KEYS = [27];
+let shouldFocusOnInactive = false;
```
jnatalzia: if the user clicks off the controls or never interacts with them, we want to ensure we do not falsely focus the player

jnatalzia: this also controls the other scenarios listed in the pr descript

src/js/view/controls/controls.js
```diff
@@ -305,6 +309,7 @@ export default class Controls extends Events {
                     if (model.get('fullscreen')) {
                         api.setFullscreen(false);
                         this.playerContainer.blur();
+                        shouldFocusOnInactive = false;
```
mamaddox: Does `shouldFocusOnInactive` need to be set to false here if it gets set by the blur eventListener? I honestly don't know.

robwalch: yeah it seems like you might want to do this before `blue` or `setFullscreen` is called.

src/js/view/controls/controls.js
```diff
@@ -610,6 +616,11 @@ export default class Controls extends Events {
         this.showing = false;
         addClass(this.playerContainer, 'jw-flag-user-inactive');
         this.trigger('userInactive');
+
+        if (shouldFocusOnInactive) {
```
robwalch: How about implementing this as a `'userInactive'` listener instead of a flag to avoid side-effects? Ex:
```js
controlbar.on(USER_ACTION, () => {
    this.once('userInactive', () => this.focusPlayerElement(), this);
    this.userActive();
});
```
```js
const blurCallback = (evt) => {
    this.off('userInactive', null, this);
    const focusedElement = evt...
```

jnatalzia: this is a great idea

jnatalzia:
@robwalch using the event handlers worked like a charm, same test cases listed in descript now pass. 

also thanks @mamaddox for the catch on my over-setting of the var, led to a much more consolidated approach

should be good for another review!

src/js/view/controls/controls.js
```diff
@@ -173,7 +173,10 @@ export default class Controls extends Events {
         // Controlbar
         const controlbar = this.controlbar = new Controlbar(api, model,
             this.playerContainer.querySelector('.jw-hidden-accessibility'));
-        controlbar.on(USER_ACTION, () => this.userActive());
+        controlbar.on(USER_ACTION, () => {
+            this.once('userInactive', () => this.focusPlayerElement(), this);
```
robwalch: you don't need the arrow here. With `this` context: `this.once('userInactive', this.focusPlayerElement, this);` will do.  Then you can also call `off` in exactly the same way.

src/js/view/controls/controls.js
```diff
@@ -412,6 +415,7 @@ export default class Controls extends Events {
 
         // Hide controls when focus leaves the player
         const blurCallback = (evt) => {
+            this.off('userInactive', null, this);
```
robwalch: `this.off('userInactive', this.focusPlayerElement, this);`

src/js/view/controls/controls.js
robwalch: You might want to also call `off` before registering `once` in case it is possible for multiple `USER_ACTION` events with fewer `blurCallback`s.

PR #3642, opened 2020-03-23
Cue Marker Class Additions

jnatalzia:
### This PR will...
Add functionality which allows for users to add custom styling to any cues created with `player.addCues`

Example: `jwplayer().addCues([{begin: 1, text: 'test-cue', cueType: 'test'}]);` will add a css class of `jw-cue-type-test` to the cue which can then be independently styled

We need ads specific cue styling

Class name and general feedback as per usual

No, I added notes to the jira ticket around how ads would implement

#### Addresses Issue(s):

JW8-10929

### Checklist

### DIFF
--- a/src/js/view/controls/components/chapters.mixin.js
+++ b/src/js/view/controls/components/chapters.mixin.js
@@ -6,10 +6,11 @@ class Cue {
         this.time = time;
         this.text = text;
         this.el = document.createElement('div');
-        this.el.className = 'jw-cue jw-reset';
+        let cssClasses = 'jw-cue jw-reset';
         if (cueType && typeof cueType === 'string') {
-            this.el.className += ` jw-cue-type-${cueType}`;
+            cssClasses += ` jw-cue-type-${cueType}`;
         }
+        this.el.className = cssClasses;
     }
 
     align(duration) {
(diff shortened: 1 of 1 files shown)


src/js/view/controls/components/chapters.mixin.js
```diff
@@ -2,11 +2,14 @@ import { ajax } from 'utils/ajax';
...
 class Cue {
-    constructor (time, text) {
+    constructor (time, text, cueType) {
         this.time = time;
         this.text = text;
         this.el = document.createElement('div');
         this.el.className = 'jw-cue jw-reset';
+        if (cueType && typeof cueType === 'string') {
+            this.el.className += ` jw-cue-type-${cueType}`;
+        }
```
robwalch: Can you do this to a string variable before setting `className`? Setting `className` more than once could impact performance by triggering repaints or reflow. Ex:
```suggestion
        let cssClasses = 'jw-cue jw-reset';
        if (cueType && typeof cueType === 'string') {
            cssClasses += ` jw-cue-type-${cueType}`;
        }
        this.el.className = cssClasses;
```

jnatalzia: awesome idea

PR #3643, opened 2020-03-24
Type 4 files - Plugins related

jnatalzia:
### This PR will...
Type the following files:

* src/js/playlist/source.js
* src/js/plugins/loader.js
* src/js/plugins/model.js
* src/js/plugins/utils.ts

TS BBY

My PlaylistItemSource typing for sure

Nah

#### Addresses Issue(s):

JW8-11034

### Checklist


src/js/plugins/model.ts
```diff
@@ -0,0 +1,53 @@
...
+    addPlugin: (url: string) => PluginObj;
+    setupPlugin: (url: string) => Promise<PluginObj>;
+    getPlugin: (name: string) => PluginObj;
+    removePlugin: (name: string) => void;
+    getPlugins: () => { [key: string]: PluginObj };
+}
+
+const pluginsRegistered: { [key: string]: PluginObj } = {};
+
+const PluginModel = function(this: PluginModelInt): void {
```
jnatalzia: i moved this into a function vs using a prototype so that i could make all the functions on it required, otherwise since the prototype added them after the function creation i would had to have made them all optional, no bueno

src/js/types/generic.type.ts
```diff
@@ -20,6 +18,10 @@ export type Tooltip = GenericObject;
 
 export type Localization = GenericObject;
 
+export type DrmConfig = GenericObject;
```
jnatalzia: when to type this sucker?

jnatalzia: is now the time?

src/js/playlist/source.ts
```diff
@@ -10,13 +11,34 @@ import { trim, extension } from 'utils/strings';
  * @property {string} label - The quality label to be used with multiple mp4/webm sources.
  */
 
-const Source = function(config) {
+export type PlaylistItemSource = PlaylistItemConfig;
```
jnatalzia: i opted to export this since naming wise it makes more sense, but i think having the internal `config` naming makes sense in terms of the actual application usage

robwalch: That's good, but `type` and `default` properties are not optional on `source`.

jnatalzia: ah nice catch, fixed

src/js/plugins/model.ts
robwalch: Making it a class would be a cleaner refactor than this

src/js/utils/log.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable no-console */
-export const log: () => {} = (typeof console.log === 'function') ? console.log.bind(console) : noop;
+export const log: (msg: string) => {} = (typeof console.log === 'function') ? console.log.bind(console) : noop;
```
robwalch: We don't event use this in the player - only in some ad plugins. So I don't think you need to add the msg arg - (in reality this should be a rest arg since console.log can take any length of arguments.

```
interface Console {
    memory: any;
    assert(condition?: boolean, message?: string, ...data: any[]): void;
    clear(): void;
    count(label?: string): void;
    debug(message?: any, ...optionalParams: any[]): void;
    dir(value?: any, ...optionalParams: any[]): void;
    dirxml(value: any): void;
    error(message?: any, ...optionalParams: any[]): void;
    exception(message?: string, ...optionalParams: any[]): void;
    group(groupTitle?: string, ...optionalParams: any[]): void;
    groupCollapsed(groupTitle?: string, ...optionalParams: any[]): void;
    groupEnd(): void;
    info(message?: any, ...optionalParams: any[]): void;
    log(message?: any, ...optionalParams: any[]): void;
```

src/js/types/generic.type.ts
robwalch: this should be a pass through we define in commercial

src/js/utils/log.ts
jnatalzia: i'll mark it optional for sure - was throwing errors without the arg

src/js/types/generic.type.ts
jnatalzia: cool, will leave here until the TS commercial effort kicks off

src/js/plugins/model.ts
jnatalzia: class-ified and updated

src/js/playlist/source.ts
```diff
@@ -10,14 +11,39 @@ import { trim, extension } from 'utils/strings';
  * @property {string} label - The quality label to be used with multiple mp4/webm sources.
  */
 
-const Source = function(config) {
+export type PlaylistItemSource = {
+    type: string;
+    default: boolean;
+} & PlaylistItemConfig;
```
robwalch: Does this override work? I tried something similar, but did not see type checks enforced on these fields. I had to use `Omit` to remove optional fields so that the one's we're making required are treated that way.

robwalch:
```ts
// OK: type-check ensured `default` is a boolean / required
export type PlaylistFeedSource = Omit<PlaylistMRSSSource, 'default'> & {
    default: boolean;
};

// NO: `default` is still optional based on `PlaylistMRSSSource` in both these examples
export type PlaylistFeedSource = PlaylistMRSSSource & {
    default: boolean;
};

export type PlaylistFeedSource = {
    default: boolean;
} & PlaylistMRSSSource;
```

jnatalzia: i'll use omit to be safe but i found my ide throwing an error on a `PlaylistItemSource` when either of these props were not preset

src/js/utils/log.ts
```diff
@@ -1,5 +1,5 @@
 /* eslint-disable no-console */
-export const log: () => {} = (typeof console.log === 'function') ? console.log.bind(console) : noop;
+export const log: (msg?: string) => {} = (typeof console.log === 'function') ? console.log.bind(console) : noop;
```
robwalch: Oh! `(msg?: string) => {}` is just a type? Well then it should match `console.log`:

`(message?: any, ...optionalParams: any[]) => void`

jnatalzia: fixed

src/js/plugins/loader.ts
```diff
@@ -1,8 +1,15 @@
...
+import type PluginModel from './model';
 
-const PluginLoader = function () {
-    this.load = function (api, pluginsModel, pluginsConfig, model) {
+export interface PluginLoaderInt {
+    load: (api: PlayerAPI, pluginsModel: PluginModel, pluginsConfig: GenericObject, model: SimpleModel) => PluginObj | PlayerError;
+}
+
+const PluginLoader = function (this: PluginLoaderInt): void {
+    this.load = function (api: PlayerAPI, pluginsModel: PluginModel, pluginsConfig: GenericObject, model: SimpleModel): PluginObj | PlayerError {
```
robwalch: Shouldn't the return type be:
`: Promise<PluginObj | PlayerError | void> {`

The `void` is for early `return Promise.resolve()`.

jnatalzia: From MDN: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve

`The Promise.resolve() method returns a Promise object that is resolved with a given value`

jnatalzia: oh I see, yeah fixing

PR #3644, opened 2020-03-24
[JW8-10893] Fix voiceover not announcing volume slider

mamaddox:
### This PR will...
Fix the volume slider not being announced by Voiceover when tabbing through elements.
It also only displays the `jw-horizontal-volume-container` for audio only players. 

Accessibility
#### Addresses Issue(s):

[JW8-10893](https://jwplayer.atlassian.net/browse/JW8-10893)

### Checklist

### DIFF
--- a/src/css/controls/imports/slider.less
+++ b/src/css/controls/imports/slider.less
@@ -213,3 +213,11 @@
         height: 11px;
     }
 }
+
+.jw-horizontal-volume-container {
+    display: none;
+
+    .jw-flag-audio-player & {
+        display: flex;
+    }
+}
--- a/src/js/view/controls/controlbar.js
+++ b/src/js/view/controls/controlbar.js
@@ -113,7 +113,8 @@ export default class Controlbar {
 
             const volumeButtonEl = volumeGroup.element();
             menus.push(volumeGroup);
-            setAttribute(volumeButtonEl, 'role', 'button');
+            // Use 'group' role so that voiceover identifies volume slider and button
+            setAttribute(volumeButtonEl, 'role', 'group');
             _model.change('mute', (model, muted) => {
                 const muteText = muted ? localization.unmute : localization.mute;
                 setAttribute(volumeButtonEl, 'aria-label', muteText);


src/css/controls/imports/slider.less
```diff
@@ -213,3 +213,11 @@
...
     }
 }
+
+.jw-horizontal-volume-container {
+    display: none;
+
+    .jw-flag-audio-player & {
+        display: flex;
+    }
+}
```
mamaddox: Voiceover was actually only detecting the `horizontal-volume-slider` and announcing that as opposed to the `vertical-volume-slider`.

src/js/view/controls/controlbar.js
```diff
@@ -113,7 +113,8 @@ export default class Controlbar {
 
             const volumeButtonEl = volumeGroup.element();
             menus.push(volumeGroup);
-            setAttribute(volumeButtonEl, 'role', 'button');
+            // Use 'group' role so that voiceover identifies volume slider and button
+            setAttribute(volumeButtonEl, 'role', 'group');
             _model.change('mute', (model, muted) => {
```
mamaddox: Because we have the volume slider as a child of the volume button, the only way to get it to be detected is to treat the button as a group.
I think the only way to completely solve this would be to refactor this and separate the button and the slider into two elements. This is the workaround that others who faced similar issues have used as well.

src/css/controls/imports/slider.less
robwalch: How is `jw-horizontal-volume-container` currently hidden - Is it just with `width: 0`?

Can you link to an audio-only test here/ in JIRA (maybe add the right label) so TE is sure to cover it in tests?

mamaddox:
test this please

src/css/controls/imports/slider.less
mamaddox: > How is `jw-horizontal-volume-container` currently hidden - Is it just with `width: 0`?

Yeah. `width: 0` and `tabindex = -1`. Looks like the only time it "opens" is with audio players.
https://github.com/jwplayer/jwplayer/blob/4763f810c473157043d032069611401762f782ab/src/css/controls/flags/audioplayer.less#L78-L85

> Can you link to an audio-only test here/ in JIRA (maybe add the right label) so TE is sure to cover it in tests?

Updated JIRA ticket and added `core` label as well. Not sure if there is another label to add.

robwalch: `@audio_only` is the label that shows up in `.feature` files if you search for audio only test cases. We need to add the label to JIRA. cc @aizaztoppa

PR #3646, opened 2020-03-27
[JW8-11031] Typescript qoe, timer, and config-normalization

mamaddox:
### This PR will...
Type the following 3 files:

src/js/api/config-normalization.js
src/js/api/timer.js
src/js/controller/qoe.js

TSing
#### Addresses Issue(s):

JW8-11031

### Checklist

### DIFF
--- a/src/js/controller/qoe.ts
+++ b/src/js/controller/qoe.ts
@@ -1,4 +1,6 @@
 import {
+    InternalPlayerState,
+    TimeEvent,
     PLAYLIST_ITEM,
     MEDIA_PLAY_ATTEMPT,
     PROVIDER_FIRST_FRAME,
@@ -7,15 +9,28 @@ import {
     MEDIA_VISUAL_QUALITY
 } from 'events/events';
 import Timer from 'api/timer';
+import type Model from 'controller/model';
+import type { MediaModel } from 'controller/model';
+import type { ProgramController } from 'program/program-controller';
+
+type QoeModel = Model & {
+    _qoeItem: Timer & {
+        getFirstFrame: () => number | null;
+    };
+    _triggerFirstFrame: () => void;
+    _onTime: (evt: TimeEvent) => void;
+    _onPlayAttempt: () => void;
+    _onTabVisible?: (modelChanged: Model, activeTab: boolean) => void;
+}
 
 const TAB_HIDDEN = 'tabHidden';
 const TAB_VISIBLE = 'tabVisible';
 
 // This is to provide a first frame event even when
 //  a provider does not give us one.
-const onTimeIncreasesGenerator = (function(callback) {
+const onTimeIncreasesGenerator = (function(callback: () => void): (evt: TimeEvent) => void {
     let lastVal = 0;
-    return function (evt) {
+    return function (evt: TimeEvent): void {
         const pos = evt.position;
         if (pos > lastVal) {
             callback();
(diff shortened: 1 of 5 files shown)


src/js/controller/qoe.ts
```diff
@@ -7,15 +9,28 @@ import {
     MEDIA_VISUAL_QUALITY
 } from 'events/events';
 import Timer from 'api/timer';
+import type Model from 'controller/model';
+import type { MediaModel } from 'controller/model';
+import type { ProgramController } from 'program/program-controller';
+
+type QoeModel = Model & {
```
jnatalzia: Should this be `SimpleModel &`?

src/js/controller/qoe.ts
```diff
@@ -86,30 +101,32 @@ function trackFirstFrame(model, programController) {
...
-        model._qoeItem = new Timer();
-        model._qoeItem.getFirstFrame = function() {
-            const time = this.between(MEDIA_PLAY_ATTEMPT, MEDIA_FIRST_FRAME);
-            // If time between the tab becoming visible and first frame is valid
-            // and less than the time since play attempt, play was not attempted until the tab became visible
-            const timeActive = this.between(TAB_VISIBLE, MEDIA_FIRST_FRAME);
-            if (timeActive > 0 && timeActive < time) {
-                return timeActive;
+        model._qoeItem = {
+            ...Timer(),
```
jnatalzia: Interesting fix here - with JS this _should_ be functionally identical but I’m wondering if we should add whatever component has timer tests to the Jira ticket so we get some automation on the change

src/js/controller/qoe.ts
```diff
@@ -1,4 +1,6 @@
 import {
+    InternalPlayerState,
+    TimeEvent,
```
jnatalzia: Can you `import type` on a separate line for the timer?

src/js/controller/qoe.ts
jnatalzia: Also should this be new Timer()...?

jnatalzia: MIGHT be worth subclassing Timer with qoeItem? Might be more trouble than it’s worth but the more I look at this the less easy it is to quickly “get” for me

mamaddox: Looking at this more, I think it is worth it to subclass because this could get ugly.
This will now be `new Timer()`, but wasn't here because `Timer` is really just a function that returns a Timer so it was complaining.

src/js/controller/qoe.ts
mamaddox: I'm honestly not sure. It looked like the `QoeModel` had all the properties of `model` with a few extra.

mamaddox:
test this please

src/js/controller/qoe.ts
```diff
@@ -7,15 +7,43 @@ import {
...
 
+interface QoeModel extends Model {
+    _qoeItem: QoeItem;
+    _triggerFirstFrame: () => void;
+    _onTime: (evt: TimeEvent) => void;
+    _onPlayAttempt: () => void;
+    _onTabVisible?: (modelChanged: Model, activeTab: boolean) => void;
+}
+
+class QoeItem extends Timer {
```
jnatalzia, line 29: refactor really paid off

mamaddox: Yes it did. It's like night and day. Great suggestion!

PR #3647, opened 2020-03-30
Always include plugin URL in function call

jnatalzia:
### This PR will...
Re-include an arg that is passed to error creation on plugin load failure

Commercial has a function with the _exact same name_ as this one that operates differently and requires that param. When building commercial, it includes that function and not this one in OSS

Nah

Nah

#### Addresses Issue(s):

JW8-11051

### Checklist

### DIFF
--- a/src/js/plugins/utils.ts
+++ b/src/js/plugins/utils.ts
@@ -6,7 +6,8 @@ export const getPluginName = function(url: string): string {
     return url.replace(/^(.*\/)?([^-]*)-?.*\.(js)$/, '$2');
 };
 
-export function getPluginErrorCode(/* pluginURL: string */): number {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function getPluginErrorCode(pluginURL: string): number {
     return 305000;
 }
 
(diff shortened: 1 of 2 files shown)


src/js/plugins/utils.ts
```diff
@@ -6,7 +6,8 @@ export const getPluginName = function(url: string): string {
     return url.replace(/^(.*\/)?([^-]*)-?.*\.(js)$/, '$2');
 };
 
-export function getPluginErrorCode(/* pluginURL: string */): number {
+// eslint-disable-next-line @typescript-eslint/no-unused-vars
+export function getPluginErrorCode(pluginURL: string): number {
```
mamaddox: Do we need to add this back if the commercial method is being used instead? Or is it more so for consistency?

jnatalzia: Since we're passing it in in the OSS loader, typescript's typecheck fails because it's passing an arg to a function that doesn't take any in :(

mamaddox: ahhh. oh typescript.

jnatalzia:
test this please

PR #3650, opened 2020-04-06
AMP Fullscreen

zetagame:
### This PR will...
Allow AMP to handle the fullscreen whenever it is able (When it's an AMP document, and not iOS or Safari)

#### Addresses Issue(s):

JW8-###

### Checklist

### DIFF
--- a/src/js/view/view.js
+++ b/src/js/view/view.js
@@ -321,13 +321,7 @@ function View(_api, _model) {
     this.init = function() {
         this.updateBounds();
 
-        const ampController = _model._model.attributes.ampController;
-
-        if (!ampController || OS.iOS || Browser.safari) {
-            _model.on('change:fullscreen', _fullscreen);
-        } else {
-            _model.on('change:fullscreen', _fullscreenAmp);
-        }
+        _model.on('change:fullscreen', _fullscreen);
         _model.on('change:activeTab', updateVisibility);
         _model.on('change:fullscreen', updateVisibility);
         _model.on('change:intersectionRatio', updateVisibility);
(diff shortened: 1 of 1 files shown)


src/js/view/view.js
```diff
@@ -607,9 +607,11 @@ function View(_api, _model) {
         if (state && _controls && model.get('autostartMuted')) {
             _controls.unmuteAutoplay(_api, model);
         }
+        const AMPController = _model._model.attributes.ampController;
```
jnatalzia, line 610: is there a reason we aren't using `_model.get('ampController')` here?

zetagame: Didn't know all model attribute refs were get-able. Good catch/info

src/js/view/view.js
```diff
@@ -608,8 +608,11 @@ function View(_api, _model) {
             _controls.unmuteAutoplay(_api, model);
         }
 
+        const AMPController = _model.get('ampController');
+        const AMPControlsFullscreen = AMPController && !OS.iOS && !Browser.safari;
+
         if (fullscreenHelpers.supportsDomFullscreen()) {
-            if (state) {
+            if (state && !AMPControlsFullscreen) {
```
robwalch, line 615: If amp controller is a commercial concept then you should create a commercial `fullscreenHelpers` module to handle this logic instead of introducing amp logic here.

zetagame: kk, changing it over now. Originally I wanted to avoid that because that would be a much larger open source change, since those functions are constants scoped in view.init and less DRY (it's the same function but we check if it's AMP and not iOS/Safari before delegating the exit/request to AMP). I'll move them now.

zetagame:
test this please

PR #3652, opened 2020-04-07
Typescript 4 files

jnatalzia:
### This PR will...
TS the following files:

* src/js/utils/playerutils.js
* src/js/view/preview.js
* src/js/view/title.js
* src/js/view/view-model.js

We are ts'ing the player. `view-model` was added because of dependencies in the other files being TS'eg

All of it, I tried following the code paths to figure out the right type of model being used in each location but I don't trust my outcomes. Especially in relation to instreamModel, there is a setter in the `ViewModel` but I could not find a single piece of code that sets that property in either OSS, commercial, or the ads plugins

Nope
#### Addresses Issue(s):

JW8-11054
### Checklist

### DIFF
--- a/src/js/types/generic.type.ts
+++ b/src/js/types/generic.type.ts
@@ -22,10 +22,6 @@ export type DrmConfig = GenericObject;
 
 export type PluginObj = GenericObject;
 
-export type PlaylistItem = GenericObject;
-
-export type AdProgramController = GenericObject;
-
 export type TextTrackLike = TextTrack & {
     _id: string;
     data: Array<any>;
--- a/src/js/types/generic.type.ts
+++ b/src/js/types/generic.type.ts
@@ -22,6 +22,10 @@ export type DrmConfig = GenericObject;
 
 export type PluginObj = GenericObject;
 
+export type PlaylistItem = GenericObject;
+
+export type AdProgramController = GenericObject;
+
 export type TextTrackLike = TextTrack & {
     _id: string;
     data: Array<any>;
--- a/src/js/utils/backbone.events.ts
+++ b/src/js/utils/backbone.events.ts
@@ -27,7 +27,7 @@ type EventListener = {
 }
 
 type EventCallback = {
-    (...args: any): void;
+    (...args: any[]): void;
     _callback?: () => void;
 };
 
--- a/src/js/utils/backbone.events.ts
+++ b/src/js/utils/backbone.events.ts
@@ -27,7 +27,7 @@ type EventListener = {
 }
 
 type EventCallback = {
-    (): void;
+    (...args: any): void;
     _callback?: () => void;
 };
 
@@ -89,7 +89,7 @@ export default class Events {
      * @param {any} [context] - If null, all callbacks with that function will be removed.
      * @returns {any} `this` context for chaining.
      */
-    off(name?: string | EventListenerDictionary, callback?: EventCallback, context?: any): Events {
+    off(name?: string | EventListenerDictionary | null, callback?: EventCallback | null, context?: any): Events {
         if (!this._events || !eventsApi(this, 'off', name, [callback, context])) {
             return this;
         }
@@ -193,7 +193,7 @@ const eventSplitter = /\s+/;
 // Implement fancy features of the Events API such as multiple event
 // names `"change blur"` and jQuery-style event maps `{change: action}`
 // in terms of the existing API.
-function eventsApi(obj: Events, action: EventsAction, name: string | EventListenerDictionary | undefined, rest: any[]): name is string {
+function eventsApi(obj: Events, action: EventsAction, name: string | EventListenerDictionary | undefined | null, rest: any[]): name is string {
     if (!name) {
         return true;
     }
--- a/src/js/utils/playerutils.ts
+++ b/src/js/utils/playerutils.ts
@@ -1,7 +1,12 @@
 import { version } from 'version';
 import { isFileProtocol } from 'utils/validator';
 
-export const getScriptPath = function(scriptName) {
+declare global {
+    const __SELF_HOSTED__: boolean;
+    const __REPO__: string;
+}
+
+export const getScriptPath = function(scriptName: string): string {
     const scripts = document.getElementsByTagName('script');
     for (let i = 0; i < scripts.length; i++) {
         const src = scripts[i].src;
@@ -16,7 +21,7 @@ export const getScriptPath = function(scriptName) {
 };
 
 // Gets the repository location from which modules and plugins are loaded by default
-export const repo = function () {
+export const repo = function (): string {
     if (__SELF_HOSTED__) {
         return getScriptPath('jwplayer.js');
     }
@@ -27,7 +32,7 @@ export const repo = function () {
 };
 
 // Is the player at least a minimum required version?
-export const versionCheck = function (target) {
+export const versionCheck = function (target: string): boolean {
     const tParts = ('0' + target).split(/\W/);
     const jParts = version.split(/\W/);
     const tMajor = parseFloat(tParts[0]);
@@ -42,7 +47,7 @@ export const versionCheck = function (target) {
     return true;
 };
 
-export const loadFrom = function () {
+export const loadFrom = function (): string {
     if (__DEBUG__ || __SELF_HOSTED__) {
         return getScriptPath('jwplayer.js');
     }
--- a/src/js/utils/playerutils.ts
+++ b/src/js/utils/playerutils.ts
@@ -1,11 +1,6 @@
 import { version } from 'version';
 import { isFileProtocol } from 'utils/validator';
 
-declare global {
-    const __SELF_HOSTED__: boolean;
-    const __REPO__: string;
-}
-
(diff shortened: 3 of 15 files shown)


src/js/view/title.ts
```diff
@@ -3,25 +3,40 @@ import {
     replaceInnerHtml
 } from 'utils/dom';
 import { Browser } from 'environment/environment';
+import type { PlaylistFeedItemWithMedia } from 'parsers/jwparser';
```
jnatalzia, line 6: still unclear of the different types here present in the parsers, but i _think_ i selected the right ones? unclear. i think it might be useful to create a union type for all the possible `itemwithmedia` options to import in cases like the below where we have a generic `item` argument.

thoughts @robwalch ?

src/js/view/view-model.ts
```diff
@@ -63,31 +72,34 @@ class PlayerViewModel extends SimpleModel {
         }
     }
 
-    get(attr) {
+    get(attr: keyof PlayerModelAttributes | keyof MediaModelAttributes): any {
         const mediaModel = this._mediaModel;
         if (mediaModel && attr in mediaModel.attributes) {
-            return mediaModel.get(attr);
+            return mediaModel.get(attr as keyof MediaModelAttributes);
```
jnatalzia, line 78: would love a better way of handling this

src/js/view/view-model.ts
```diff
@@ -63,31 +72,34 @@ class PlayerViewModel extends SimpleModel {
...
+        return this._model.get(attr as keyof PlayerModelAttributes);
     }
 
-    set(attr, val) {
+    set(attr: keyof PlayerModelAttributes, val: any): void {
         return this._model.set(attr, val);
     }
 
-    getVideo() {
+    getVideo(): DefaultProvider | null {
```
jnatalzia, line 87: could use a union type here as well for all providers?

jnatalzia:
test this please

src/js/utils/backbone.events.ts
```diff
@@ -27,7 +27,7 @@ type EventListener = {
 }
 
 type EventCallback = {
-    (): void;
+    (...args: any): void;
```
mamaddox, line 30: I think this has to be an array type like `any[]`

robwalch:
> Especially in relation to instreamModel, there is a setter in the ViewModel but I could not find a single piece of code that sets that property in either OSS, commercial, or the ads plugins

What you what you want to look for is model.set('instream', ...). It is an instance of ad-program-controller. It's model represents the playback state of the ad break. Instream (and ad-program-controller) is active for the duration of an ad-break. Inside the view-model,  the instream model's state supersedes that of the player-model.

The program-controller manages playlist items, and the media-controller/media-model/provider trio that handles commands, events and state for each item in a playlist/player's lifecycle. The ad-program-controller is exactly the same but for ad items in an ad-break's lifecycle.

instream-adapter
```js
        // This enters the player into instream mode
        _model.set('instream', _adProgram);
```

view-model
```js
playerModel.on('change:instream', (model, instream) => {
    this.instreamModel = instream ? instream.model : null;
}, this);
```

src/js/utils/playerutils.ts
```diff
@@ -1,7 +1,12 @@
 import { version } from 'version';
 import { isFileProtocol } from 'utils/validator';
 
-export const getScriptPath = function(scriptName) {
+declare global {
+    const __SELF_HOSTED__: boolean;
+    const __REPO__: string;
```
robwalch, line 6: Could we declare all of these compiler replaced arguments in a single place? I like that they are grouped in webpack and karma configs, and would like to see them grouped together for TS as well. This is one case where I wouldn't mind a separate globals.ts(.d?) file, even if that strays from our rule of defining types in the file they are implemented in. Otherwise globals get dropped all over and it's easy to look track of them.

src/js/view/preview.ts
```diff
@@ -1,17 +1,23 @@
 import { style } from 'utils/css';
+import type Model from 'controller/model';
 
-const Preview = function(_model) {
-    this.model = _model;
-    this.image = null;
-};
+class Preview {
```
robwalch, line 4: 👍

src/js/view/title.ts
robwalch: By the time the view get's it, it should a type in the model (e.g.`model.playlist[model.item]` > `playlist: PlaylistItem[];`). The feed or config input has been parsed and normalized via playlist/playlist.js and item.js (which while untyped should be of type `PlaylistItem`.

src/js/view/preview.ts
```diff
@@ -1,17 +1,23 @@
 import { style } from 'utils/css';
+import type Model from 'controller/model';
```
robwalch, line 2: view-model or player-view-model?

If there is _**any**_ place where we are directly using "the" player model in src/view/ it should be called out in a comment, because we want the view-model and it's player-view-model to play go-between and prevent that.

src/js/view/title.ts
```diff
@@ -3,25 +3,40 @@ import {
...
+import type ViewModel from './view-model';
+import type { PlayerViewModel } from './view-model';
+import type Model from 'controller/model';
 
-const Title = function(_model) {
-    this.model = _model.player;
-    this.truncated = _model.get('__ab_truncated') && !Browser.ie;
-};
+class Title {
+    model: PlayerViewModel;
```
robwalch, line 13: 👍 We don't display ad titles, only the playlist item title so use the player properties to ignore change events from the view model when we switch to "instream" (ads) mode.

src/js/view/title.ts
```diff
@@ -3,25 +3,40 @@ import {
...
         style(this.el, { display: '' });
-    },
+    }
 
-    setup: function(titleEl) {
+    setup(titleEl: HTMLElement): void {
         this.el = titleEl;
 
+        if (!this.el) {
+            return;
```
robwalch, line 37: 👍 Darn lazy setups!

src/js/view/title.ts
```diff
@@ -47,9 +65,9 @@ Object.assign(Title.prototype, {
             }
         }
         style(this.el, titleStyle);
-    },
+    }
 
-    playlistItem: function(model, item) {
+    playlistItem(model: ViewModel, item: PlaylistMRSSItemWithMedia| PlaylistFeedItemWithMedia): void {
```
robwalch, line 70: We are listening to changes on the `PlayerViewModel`.

src/js/utils/backbone.events.ts
jnatalzia: great catch

src/js/utils/playerutils.ts
jnatalzia: yeah this is a great idea - will make a sep file

src/js/view/title.ts
jnatalzia: `.player()` only exists on `ViewModel` with the way the classes are currently structured. the superclass of which is `PlayerViewModel` but it would have to be a `ViewModel` passed in here for the code to work as is written

jnatalzia: ah nice catch - ignore my previous comment, thought this was on the constructor

src/js/view/preview.ts
jnatalzia: this is good to know - we are passing in the view model, just verified in the trace

src/js/types/generic.type.ts
```diff
@@ -22,6 +22,10 @@ export type DrmConfig = GenericObject;
 
 export type PluginObj = GenericObject;
 
+export type PlaylistItem = GenericObject;
```
robwalch, line 25: In model.ts, I just did `import type PlaylistItem from 'playlist/item';`. For js imports this type is treated as `any` (unless TS can infer some typing from es6 classes). So, it's basically the same thing and we don't need to make a change later.

jnatalzia: hmmm vs code was throwing an error. will update if possible

src/js/view/view-model.ts
robwalch: This is great. What do you not like?

src/js/view/view-model.ts
robwalch: Base class is good. Most providers are defined outside of this project.

src/js/view/view-model.ts
```diff
@@ -111,11 +123,11 @@ export default class ViewModel extends PlayerViewModel {
         }, this);
     }
 
-    get player() {
+    get player(): PlayerViewModel {
         return this._playerViewModel;
     }
 
-    set instreamModel(instreamModel) {
+    set instreamModel(instreamModel: Model) {
```
robwalch, line 130: It's odd that this is a public setter. It's only used above in the change event handler so really this should be scoped there, but no need to change it now.

src/js/view/view-model.ts
jnatalzia: my understanding was this broke the type safety by manually implying to the compiler which of the union types is present, figured there might be a way to dynamically assign but i agree it's not terrible

PR #3654, opened 2020-04-09
[JW8-11052] Typescript 3 files

pajong:
### This PR will...
Type script the following three files
- `src/js/program/background-media.js`
- `src/js/providers/providers-register.js`
- `src/js/providers/providers-supported.js`

#### Addresses Issue(s):

JW8-11052

### Checklist


src/js/program/background-media.ts
```diff
@@ -1,3 +1,9 @@
+import type Item from 'playlist/item';
```
jnatalzia, line 1: Really confused on how this is pulling a type from a `.js` file. When I tried this locally I ran into a bunch of errors so I resorted to this https://github.com/jwplayer/jwplayer/pull/3652/files#diff-9588e826a5bd96417540cd92b46bc3b6R25

https://github.com/jwplayer/jwplayer/pull/3652/files#diff-751861c9e54ddc75afa6f226a8b9ba2eR18

Your build passed which makes me really confused. Overall though I would say let's avoid `import type`ing non-tsed files and go with a generic placeholder for the time being

src/js/program/background-media.ts
```diff
@@ -12,34 +18,36 @@
...
-export default function BackgroundMedia() {
-    let currentMedia = null;
-    let nextMedia = null;
+export default function BackgroundMedia(): object {
+    let currentMedia: MediaObject | null = null;
+    let nextMedia: MediaObject | null = null;
 
     return Object.defineProperties({
-        setNext(item, loadPromise) {
+        setNext(item: Item, loadPromise: Promise<any>): void {
```
jnatalzia, line 26: this should be `Promise<MediaController>` if i'm reading the JSDOC properly

`Resolves with the mediaController`

@robwalch confirm?

src/js/program/background-media.ts
```diff
@@ -12,34 +18,36 @@
...
+            }
         },
-        clearNext() {
+        clearNext(): void {
             nextMedia = null;
         }
     }, {
         nextLoadPromise: {
-            get() {
+            get(): Promise<any> | null {
```
jnatalzia, line 42: same comment about promise resolution type

src/js/providers/providers-register.ts
```diff
@@ -3,7 +3,7 @@ import ProvidersSupported from 'providers/providers-supported';
 import DefaultProvider from 'providers/default';
 import { find, matches, isFunction, defaults } from 'utils/underscore';
 
-export default function registerProvider(provider) {
+export default function registerProvider(provider: DefaultProvider): void {
```
jnatalzia, line 6: once the other providers are typed this will need to be a union type, thoughts on creating a `type ProviderUnion` in `generic.type.ts` as a placeholder (currently could just be `type ProviderUnion = DefaultProvider`)

src/js/program/background-media.ts
```diff
@@ -12,34 +18,36 @@
  * @property {Promise} nextMedia - A promise representing the media loading in the background. Resolves with the mediaController.
  * @constructor
  */
-export default function BackgroundMedia() {
-    let currentMedia = null;
-    let nextMedia = null;
+export default function BackgroundMedia(): object {
```
mamaddox, line 21: I think it might be worth making a `BackgroundMedia` type to use here instead of `object`.

src/js/program/background-media.ts
```diff
@@ -12,34 +18,36 @@
...
+    let currentMedia: MediaObject | null = null;
```
mamaddox, line 22: Based on the JSDOC it looks like this might actually be a `MediaController` type? 
Though those JSDOC comments are confusing me a little.

src/js/program/background-media.ts
jnatalzia: ignore this comment because of https://github.com/jwplayer/jwplayer/pull/3652#discussion_r407578181

robwalch: Already doing this in model.ts

src/js/program/background-media.ts
robwalch: That's correct. See `background.currentMedia` in program-controller. `BackgroundMedia` is only used in program-controller.

src/js/program/background-media.ts
robwalch: Correct. See program-controller `backgroundLoad` which calls `setNext`.

src/js/program/background-media.ts
```diff
@@ -1,3 +1,9 @@
+import type Item from 'playlist/item';
+
+interface MediaObject {
+    item: Item;
+    loadPromise: Promise<any>;
```
robwalch, line 5: `loadPromise: Promise<MediaController>`

robwalch, line 5: `loadPromise: Promise<MediaController>`

src/js/program/background-media.ts
robwalch: `: Promise<MediaController> | null`

src/js/program/background-media.ts
```diff
@@ -12,34 +18,36 @@
...
                 return nextMedia ? nextMedia.loadPromise : null;
             }
         },
         currentMedia: {
-            get() {
+            get(): MediaObject | null {
                 return currentMedia;
             },
-            set(mediaController) {
+            set(mediaController: MediaObject): void {
```
robwalch, line 50: `currentMedia` is a `MediaController | null`, not a `MediaObject`.

src/js/providers/providers-register.ts
robwalch: only html5 provider is defined this project. The type should be a base class or interface. Custom providers are supported too.

src/js/providers/providers-register.ts
```diff
@@ -1,9 +1,10 @@
 import ProvidersLoaded from 'providers/providers-loaded';
 import ProvidersSupported from 'providers/providers-supported';
 import DefaultProvider from 'providers/default';
+import type VideoProvider from 'providers/html5';
```
jnatalzia, line 4: we should replace this with https://github.com/jwplayer/jwplayer/pull/3636/files#diff-cae9122c8d4f001c56093abab4fae68eR75 when this merged

PR #3657, opened 2020-04-10
[JW8-11053] Typescript play-promise, quality-labels, and network-error-parser

mamaddox:
### This PR will...
Type the following files:
- src/js/providers/utils/network-error-parser.js
- src/js/providers/utils/play-promise.js
- src/js/providers/utils/quality-labels.js

#### Addresses Issue(s):

JW8-11053

### Checklist

### DIFF
--- a/src/js/providers/data-normalizer.ts
+++ b/src/js/providers/data-normalizer.ts
@@ -1,13 +1,16 @@
-import type { GenericObject } from 'types/generic.type';
-
-export type QualityLevel = {
+interface Level {
     label: string;
-    bitrate?: number;
-    width?: number;
     height?: number;
-};
+    width?: number;
+    bitrate?: number;
+}
 
-export type Level = QualityLevel & GenericObject;
+export type QualityLevel = Level;
+
+export interface ProviderLevel extends Omit<Level, 'label'> {
+    label?: string;
+    bandwidth?: number;
+}
 
 export function qualityLevel(level: Level): QualityLevel {
     return {
--- a/src/js/providers/data-normalizer.ts
+++ b/src/js/providers/data-normalizer.ts
@@ -7,7 +7,9 @@ export type QualityLevel = {
     height?: number;
 };
 
-export function qualityLevel(level: QualityLevel & GenericObject): QualityLevel {
+export type Level = QualityLevel & GenericObject;
+
+export function qualityLevel(level: Level): QualityLevel {
     return {
         bitrate: level.bitrate,
         label: level.label,
--- a/src/js/providers/utils/play-promise.ts
+++ b/src/js/providers/utils/play-promise.ts
@@ -1,7 +1,12 @@
+type PlayPromiseError = Omit<DOMException, 'name' | 'code'> & {
+    name: string;
+    code: number;
+};
+
 export default function createPlayPromise(video: HTMLVideoElement): Promise<void> {
-    return new Promise(function(resolve: () => void, reject: (reason: DOMException) => void): void {
+    return new Promise(function(resolve: () => void, reject: (reason: PlayPromiseError) => void): void {
         if (video.paused) {
-            return reject(new DOMException('play() failed.', 'NotAllowedError'));
+            return reject(playPromiseError('NotAllowedError', 0, 'play() failed.'));
         }
         const removeEventListeners = function(): void {
             video.removeEventListener('play', playListener);
@@ -23,12 +28,18 @@ export default function createPlayPromise(video: HTMLVideoElement): Promise<void
             } else {
                 const message = `The play() request was interrupted by a "${e.type}" event.`;
                 if (e.type === 'error') {
-                    reject(new DOMException(message, 'NotSupportedError'));
-                } else {
-                    reject(new DOMException(message, 'AbortError'));
+                    return reject(playPromiseError('NotSupportedError', 9, message));
                 }
+                return reject(playPromiseError('AbortError', 20, message));
             }
         };
         video.addEventListener('play', playListener);
     });
+}
+
+function playPromiseError(name: string, code: number, message: string): PlayPromiseError {
+    const error = new Error(message) as PlayPromiseError;
+    error.name = name;
+    error.code = code;
+    return error;
 }
--- a/src/js/providers/utils/play-promise.ts
+++ b/src/js/providers/utils/play-promise.ts
@@ -1,22 +1,30 @@
-export default function createPlayPromise(video) {
-    return new Promise(function(resolve, reject) {
+type Resolve = () => void;
+
+type Reject = (reason: PlayPromiseError) => void;
+
+type PlayPromiseError = Error & {
+    code: number;
+}
+
+export default function createPlayPromise(video: HTMLVideoElement): Promise<void> {
+    return new Promise(function(resolve: Resolve, reject: Reject): void {
         if (video.paused) {
             return reject(playPromiseError('NotAllowedError', 0, 'play() failed.'));
         }
-        const removeEventListeners = function() {
+        const removeEventListeners = function(): void {
             video.removeEventListener('play', playListener);
             video.removeEventListener('playing', listener);
             video.removeEventListener('pause', listener);
             video.removeEventListener('abort', listener);
             video.removeEventListener('error', listener);
         };
-        const playListener = function() {
+        const playListener = function(): void {
             video.addEventListener('playing', listener);
             video.addEventListener('abort', listener);
             video.addEventListener('error', listener);
             video.addEventListener('pause', listener);
         };
-        const listener = function(e) {
+        const listener = function(e: Event): void {
             removeEventListeners();
             if (e.type === 'playing') {
                 resolve();
@@ -33,8 +41,8 @@ export default function createPlayPromise(video) {
     });
 }
 
-function playPromiseError(name, code, message) {
-    const error = new Error(message);
+function playPromiseError(name: string, code: number, message: string): PlayPromiseError {
+    const error = new Error(message) as PlayPromiseError;
     error.name = name;
     error.code = code;
     return error;
--- a/src/js/providers/utils/play-promise.ts
+++ b/src/js/providers/utils/play-promise.ts
@@ -1,15 +1,11 @@
 type Resolve = () => void;
 
-type Reject = (reason: PlayPromiseError) => void;
-
-type PlayPromiseError = Error & {
-    code: number;
-}
+type Reject = (reason: DOMException) => void;
 
 export default function createPlayPromise(video: HTMLVideoElement): Promise<void> {
     return new Promise(function(resolve: Resolve, reject: Reject): void {
         if (video.paused) {
-            return reject(playPromiseError('NotAllowedError', 0, 'play() failed.'));
+            return reject(new DOMException('play() failed.', 'NotAllowedError'));
         }
         const removeEventListeners = function(): void {
             video.removeEventListener('play', playListener);
@@ -31,19 +27,12 @@ export default function createPlayPromise(video: HTMLVideoElement): Promise<void
             } else {
                 const message = `The play() request was interrupted by a "${e.type}" event.`;
                 if (e.type === 'error') {
-                    reject(playPromiseError('NotSupportedError', 9, message));
+                    reject(new DOMException(message, 'NotSupportedError'));
                 } else {
-                    reject(playPromiseError('AbortError', 20, message));
+                    reject(new DOMException(message, 'AbortError'));
                 }
             }
         };
         video.addEventListener('play', playListener);
     });
-}
-
-function playPromiseError(name: string, code: number, message: string): PlayPromiseError {
-    const error = new Error(message) as PlayPromiseError;
-    error.name = name;
-    error.code = code;
-    return error;
 }
--- a/src/js/providers/utils/play-promise.ts
+++ b/src/js/providers/utils/play-promise.ts
@@ -1,3 +1,4 @@
+// These properties must be redefined because they are readonly on DOMException
 type PlayPromiseError = Omit<DOMException, 'name' | 'code'> & {
     name: string;
     code: number;
--- a/src/js/providers/utils/play-promise.ts
+++ b/src/js/providers/utils/play-promise.ts
@@ -1,9 +1,5 @@
-type Resolve = () => void;
-
-type Reject = (reason: DOMException) => void;
-
 export default function createPlayPromise(video: HTMLVideoElement): Promise<void> {
-    return new Promise(function(resolve: Resolve, reject: Reject): void {
+    return new Promise(function(resolve: () => void, reject: (reason: DOMException) => void): void {
         if (video.paused) {
             return reject(new DOMException('play() failed.', 'NotAllowedError'));
         }
--- a/src/js/providers/utils/quality-labels.ts
+++ b/src/js/providers/utils/quality-labels.ts
@@ -1,4 +1,4 @@
-import type { Level } from 'providers/data-normalizer';
+import type { ProviderLevel } from 'providers/data-normalizer';
 import type { GenericObject } from 'types/generic.type';
 
 type QualityLabels = {
@@ -6,12 +6,12 @@ type QualityLabels = {
 };
 
 // Try and find a corresponding custom label. If there are no custom labels, create one using height, bandwidth, or both
-export function generateLabel(level: Partial<Level>, qualityLabels: QualityLabels | null, redundant: boolean): string {
+export function generateLabel(level: ProviderLevel | undefined, qualityLabels: QualityLabels | null, redundant?: boolean): string {
     if (!level) {
         return '';
     }
     // flash provider uses bitrate instead of bandwidth
-    const bandwidth = parseFloat(level.bitrate || level.bandwidth);
+    const bandwidth = level.bitrate || level.bandwidth || 0;
     // flash provider, in some cases, will create its own label. Prefer it over creating a new label
     return getCustomLabel(qualityLabels, bandwidth) ||
         level.label ||
@@ -19,7 +19,7 @@ export function generateLabel(level: Partial<Level>, qualityLabels: QualityLabel
 }
 
 // Prefer creating a label with height with a fallback to bandwidth. Make a label using both if redundant
-export function createLabel(height: number | undefined, bandwidth: number | undefined, redundant: boolean): string {
+function createLabel(height?: number, bandwidth?: number, redundant?: boolean): string {
     if (!height && !bandwidth) {
         return '';
     }
@@ -44,7 +44,7 @@ export function createLabel(height: number | undefined, bandwidth: number | unde
 
 // Ensures that we're able to find a custom label. As long as there is at least 1 quality label and a defined
 // bandwidth, a quality label will always be found. Return null otherwise
-export function getCustomLabel(qualityLabels: QualityLabels | null, bandwidth?: number): string | null {
+function getCustomLabel(qualityLabels?: QualityLabels | null, bandwidth?: number): string | null {
     let label: string | null = null;
 
     if (bandwidth && qualityLabels) {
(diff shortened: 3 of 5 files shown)


src/js/providers/utils/play-promise.ts
```diff
@@ -1,22 +1,30 @@
-export default function createPlayPromise(video) {
-    return new Promise(function(resolve, reject) {
+type Resolve = () => void;
+
+type Reject = (reason: PlayPromiseError) => void;
+
+type PlayPromiseError = Error & {
```
jnatalzia, line 5: aw man we could use `PlayerError` if we had a `sourceError` prop here. oh well, more potential refactoring that TS is showcasing. good for now

src/js/providers/utils/quality-labels.ts
```diff
@@ -1,24 +1,36 @@
...
     if (!height && !bandwidth) {
         return '';
     }
 
-    const bandwidthString = `${toKbps(bandwidth)} kbps`;
-    let label = bandwidthString;
+    let label = '';
+    let bandwidthString = '';
+
+    if (bandwidth) {
```
jnatalzia, line 30: this potentially fixes a bug too given the required `&&` above, interesting

src/js/providers/utils/quality-labels.ts
```diff
@@ -32,29 +44,31 @@ export function createLabel(height, bandwidth, redundant) {
...
 // Ensures that we're able to find a custom label. As long as there is at least 1 quality label and a defined
 // bandwidth, a quality label will always be found. Return null otherwise
-export function getCustomLabel(qualityLabels, bandwidth) {
-    let label = null;
+export function getCustomLabel(qualityLabels: QualityLabels | null, bandwidth?: number): string | null {
+    let label: string | null = null;
 
     if (bandwidth && qualityLabels) {
         const bandwidths = Object.keys(qualityLabels);
-        const key = parseFloat(bandwidth);
```
jnatalzia, line 40: qq - is this function used elsewhere in non-typed files that don't pass bandwidth in as a number? that could break without this `parseFloat`

src/js/providers/utils/quality-labels.ts
```diff
@@ -32,29 +44,31 @@ export function createLabel(height, bandwidth, redundant) {
...
-    let closest = null;
+export function findClosestBandwidth(bandwidths: string[], targetBandwidth: number): string | null {
+    let closest: string | null = null;
     let smallestDiff = Infinity;
-    let curDiff;
+    let curDiff: number;
 
     if (Array.isArray(bandwidths)) {
-        bandwidths.forEach(function (cur) {
-            curDiff = Math.abs(cur - targetBandwidth);
```
jnatalzia, line 57: this wouldn't have worked previously if it was `bandwidths` was a string array, are we sure on the arg types here? or do the parsing changes in this file make it so bandwidths is always a string array? little confused on the changes in this file

src/js/providers/utils/quality-labels.ts
```diff
@@ -100,7 +114,7 @@ function checkForLevelDuplicates(levels, dupKeys) {
             return false;
         }
         const foundDuplicate = this[key];
```
jnatalzia, line 116: we could remove the `undefined` return value here and have cleaner types by changing this line to `const foundDuplicate = this[key] || false;`

src/js/providers/utils/play-promise.ts
```diff
@@ -33,8 +41,8 @@ export default function createPlayPromise(video) {
     });
 }
 
-function playPromiseError(name, code, message) {
-    const error = new Error(message);
+function playPromiseError(name: string, code: number, message: string): PlayPromiseError {
+    const error = new Error(message) as PlayPromiseError;
```
jnatalzia, line 45: hmmmm - it's tough with the way the code is currently structured but i'm trying to think if there's a quick way to use inheritance here or something similar to avoid the `as`. it's not a deal breaker, i'm fine with this as is but just something to ponder a bit if you'd like

src/js/providers/utils/quality-labels.ts
```diff
@@ -32,29 +44,31 @@ export function createLabel(height, bandwidth, redundant) {
 
 // Ensures that we're able to find a custom label. As long as there is at least 1 quality label and a defined
 // bandwidth, a quality label will always be found. Return null otherwise
-export function getCustomLabel(qualityLabels, bandwidth) {
-    let label = null;
+export function getCustomLabel(qualityLabels: QualityLabels | null, bandwidth?: number): string | null {
```
robwalch, line 47: This function is _only_ used in `generateLabel` so there is no reason to export it.

`qualityLabels` could be made optional (it could be `undefined`, not `null`). Nothing in the config defaults it to `null`.

src/js/providers/utils/quality-labels.ts
```diff
@@ -1,24 +1,36 @@
...
+
 // Try and find a corresponding custom label. If there are no custom labels, create one using height, bandwidth, or both
-export function generateLabel(level, qualityLabels, redundant) {
+export function generateLabel(level: Partial<Level>, qualityLabels: QualityLabels | null, redundant: boolean): string {
     if (!level) {
         return '';
     }
     // flash provider uses bitrate instead of bandwidth
-    const bandwidth = level.bitrate || level.bandwidth;
+    const bandwidth = parseFloat(level.bitrate || level.bandwidth);
```
robwalch, line 14: We could end up with `NaN` here rather than `undefined`. OR it with 0 to ensure a finite number with falsyness:
`const bandwidth = parseFloat(level.bitrate || level.bandwidth) || 0;`

src/js/providers/utils/quality-labels.ts
```diff
@@ -1,24 +1,36 @@
...
     // flash provider, in some cases, will create its own label. Prefer it over creating a new label
     return getCustomLabel(qualityLabels, bandwidth) ||
         level.label ||
         createLabel(level.height, bandwidth, redundant);
 }
 
 // Prefer creating a label with height with a fallback to bandwidth. Make a label using both if redundant
-export function createLabel(height, bandwidth, redundant) {
+export function createLabel(height: number | undefined, bandwidth: number | undefined, redundant: boolean): string {
```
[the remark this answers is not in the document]
robwalch: `createLabel` does not need to be exported. Make sure you search for all uses of a method (and imports) when typing.

src/js/providers/utils/quality-labels.ts
```diff
@@ -32,29 +44,31 @@ export function createLabel(height, bandwidth, redundant) {
...
 // Finds the bandwidth with the smallest difference from the target bandwidth
-export function findClosestBandwidth(bandwidths, targetBandwidth) {
-    let closest = null;
+export function findClosestBandwidth(bandwidths: string[], targetBandwidth: number): string | null {
+    let closest: string | null = null;
     let smallestDiff = Infinity;
-    let curDiff;
+    let curDiff: number;
 
     if (Array.isArray(bandwidths)) {
```
robwalch, line 69: Don't need the `Array.isArray` check anymore- This function is only used above with:
```
const bandwidths = Object.keys(qualityLabels);
     if (bandwidths.length) {
        const closestBandwidth = findClosestBandwidth(bandwidths, toKbps(bandwidth));
```

src/js/providers/utils/quality-labels.ts
```diff
@@ -32,29 +44,31 @@ export function createLabel(height, bandwidth, redundant) {
...
         }
     }
 
     return label;
 }
 
 // Finds the bandwidth with the smallest difference from the target bandwidth
-export function findClosestBandwidth(bandwidths, targetBandwidth) {
-    let closest = null;
+export function findClosestBandwidth(bandwidths: string[], targetBandwidth: number): string | null {
```
robwalch, line 64: Don't need to export.

src/js/providers/utils/quality-labels.ts
```diff
@@ -65,28 +79,28 @@ export function findClosestBandwidth(bandwidths, targetBandwidth) {
...
 }
 
-export function toKbps(bandwidth) {
+export function toKbps(bandwidth: number): number {
     return Math.floor(bandwidth / 1000);
 }
 
 // Use an empty object as the context and populate it like a hash map
-export function hasRedundantLevels(levels) {
+export function hasRedundantLevels(levels: Partial<Level>[]): boolean {
```
robwalch, line 87: Use `Level[]` or (rename to) `ProviderLevel[]`

src/js/providers/utils/quality-labels.ts
```diff
@@ -65,28 +79,28 @@ export function findClosestBandwidth(bandwidths, targetBandwidth) {
...
     if (!Array.isArray(levels)) {
         return false;
     }
     return checkForLevelDuplicates(levels, ['height', 'bitrate', 'bandwidth']);
 }
 
-export function hasRedundantLabels(levels) {
+export function hasRedundantLabels(levels: Partial<Level>[]): boolean {
```
robwalch, line 94: Use `Level[]` or (rename to) `ProviderLevel[]`

src/js/providers/utils/quality-labels.ts
```diff
@@ -65,28 +79,28 @@ export function findClosestBandwidth(bandwidths, targetBandwidth) {
...
         return false;
     }
     return checkForLevelDuplicates(levels, ['label']);
 }
 
-function checkForLevelDuplicates(levels, dupKeys) {
-    return levels.some(function (level) {
-        let key;
+function checkForLevelDuplicates(levels: Partial<Level>[], dupKeys: string[]): boolean {
+    return levels.some(function (this: GenericObject, level: Partial<Level>): boolean | undefined {
```
robwalch, line 102: The type of `this` would be the type of `levels` (`levels.some`). This mutates `levels` which is bad. We should declare a hash/dict at the top and use that instead of `this`:
```ts
function checkForLevelDuplicates(levels: Partial<Level>[], dupKeys: string[]): boolean {
    const duplicates: { [key: string] : boolean ] = {};
...
	    duplicates[key] = true;
      return foundDuplicate;
```

src/js/providers/utils/quality-labels.ts
```diff
@@ -1,24 +1,36 @@
+import type { Level } from 'providers/data-normalizer';
+import type { GenericObject } from 'types/generic.type';
+
+type QualityLabels = {
+    [key: number]: string;
+};
+
 // Try and find a corresponding custom label. If there are no custom labels, create one using height, bandwidth, or both
-export function generateLabel(level, qualityLabels, redundant) {
+export function generateLabel(level: Partial<Level>, qualityLabels: QualityLabels | null, redundant: boolean): string {
```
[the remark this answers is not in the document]
robwalch: This is what an hls.js (v1) level (`hlsjs.levels[]`) looks like:

```js
export class Level {
  public attrs: LevelAttributes;
  public audioCodec?: string;
  public audioGroupIds?: string[];
  public bitrate: number;
  public details?: LevelDetails;
  public fragmentError: boolean = false;
  public height: number;
  public id: number;
  public loadError: number = 0;
  public loaded?: any;
  public name: string | undefined;
  public realBitrate: number = 0;
  public textGroupIds?: string[];
  public url: string[];
  public urlId: number = 0;
  public videoCodec?: string;
  public width: number;
  public unknownCodecs: string[] | undefined;

  constructor (data: LevelParsed) {
    this.url = [data.url];
    this.attrs = data.attrs;
    this.bitrate = data.bitrate;
    this.details = data.details;
    this.id = data.id || 0;
    this.name = data.name;
    this.width = data.width || 0;
    this.height = data.height || 0;
    this.audioCodec = data.audioCodec;
    this.videoCodec = data.videoCodec;
    this.unknownCodecs = data.unknownCodecs;
  }

  get maxBitrate (): number {
    return Math.max(this.realBitrate, this.bitrate);
  }
}
```

For shaka it's `shakaPlayer.getVariantTracks()[]` https://shaka-player-demo.appspot.com/docs/api/shaka.extern.html#.Track
(they also use `bandwidth` rather than `bitrate`)

src/js/providers/utils/quality-labels.ts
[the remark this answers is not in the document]
mamaddox: `createLabel` is exported to be unit tested.

src/js/providers/utils/quality-labels.ts
mamaddox: > This function is _only_ used in `generateLabel` so there is no reason to export it.
> 
Exported to be unit tested as well.

src/js/providers/utils/quality-labels.ts
mamaddox: Also unit tested.

src/js/providers/utils/quality-labels.ts
mamaddox: This function is only called by `generateLabel` or in unit tests. For the unit tests, bandwidth is always a number. I moved the `parseFloat` to `generateLabel` so the bandwidth passed from that will always be a number as well.

src/js/providers/utils/quality-labels.ts
mamaddox: I'm not 100% on how this was working before, but this function is only called by unit tests or `getCustomLabel` above. Both pass in an Object.keys array which returns a string[]?

src/js/providers/utils/quality-labels.ts
mamaddox: I thought the type of `this` would be the {} type that's passed in as the second argument to this `some` function?
Regardless, I prefer the `duplicates` const.

src/js/providers/data-normalizer.ts
```diff
@@ -7,7 +7,9 @@ export type QualityLevel = {
     height?: number;
 };
 
-export function qualityLevel(level: QualityLevel & GenericObject): QualityLevel {
+export type Level = QualityLevel & GenericObject;
```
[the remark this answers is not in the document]
mamaddox: I think it was originally defined this way because of the types of levels you mentioned below (hls, shaka, etc.) and how they have many additional properties.

I'll take another look at where bandwidth is defined.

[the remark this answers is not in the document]
mamaddox: Outside of bandwidthEstimate in the config, from what I can tell, `bandwidth` is only really defined on the provider levels that use it so like Shaka levels, Shaka Audio levels and flash levels (I'm assuming). We never define it.

src/js/providers/utils/quality-labels.ts
robwalch: Does it need to be? It’s a sub-function of one other function so it’s only split out for organization/readability. Ideally it would be inlined and all tests would be on the parent function that we actually use and the inputs and outputs we expect from it.

robwalch: Does it need to be? It’s a sub-function of one other function so it’s only split out for organization/readability. Ideally it would be inlined and all tests would be on the parent function that we actually use and the inputs and outputs we expect from it.

src/js/providers/utils/quality-labels.ts
robwalch: You are right! I didn’t notice the `thisArg` following the callback. It’s fine as is.

test/unit/quality-labels-test.js
```diff
@@ -1,10 +1,6 @@
-import _ from 'utils/underscore';
```
mamaddox, line 1: Great suggestion Rob! I merged all the sub-function unit tests so that they are tested through `generateLabel` instead and no longer need to be exported.

src/js/providers/data-normalizer.ts
```diff
@@ -1,13 +1,18 @@
-import type { GenericObject } from 'types/generic.type';
-
-export type QualityLevel = {
+interface Level {
```
mamaddox, line 1: Open to suggestions for this.
I was taking a look at how `QualityLevel` and the level argument passed into this function vs. the level argument passed into the `quality-labels` functions was used.
From what I could see, they are slightly different. The levels passed into the `quality-labels` functions seem to come directly from the providers, so the types that Rob mentioned in these comments
https://github.com/jwplayer/jwplayer/pull/3657#discussion_r408290654
https://github.com/jwplayer/jwplayer/pull/3657#discussion_r408314156
I used the `ProviderLevel` interface to represent those.

The levels passed into the `qualityLevel` functions, are provider levels that have been slightly normalized. So they may `PlaylistItemSources` or referred to as jwLevels in the hls.js and shaka files.
I used the `Level` interface to represent those.

I think one of the main differences I noticed was that the `Levels` all have labels, which I believe is important for the `QualityLevels` because I think it would be strange if a `QualityLevel` had an optional label.

src/js/providers/utils/play-promise.ts
robwalch: This is a polyfill for `media.play(): Promise<void>`
https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/play
https://html.spec.whatwg.org/multipage/media.html#dom-media-play

The types of errors we are emulating have codes: `NotAllowedError`, `NotSupportedError` of interface`DOMException`...

So I think we shouldn't define our own type - instead use `as DOMException`?

test/unit/quality-labels-test.js
robwalch: Thanks!

src/js/providers/utils/quality-labels.ts
```diff
@@ -1,24 +1,36 @@
+import type { ProviderLevel } from 'providers/data-normalizer';
+import type { GenericObject } from 'types/generic.type';
+
+type QualityLabels = {
+    [key: number]: string;
+};
+
 // Try and find a corresponding custom label. If there are no custom labels, create one using height, bandwidth, or both
-export function generateLabel(level, qualityLabels, redundant) {
+export function generateLabel(level: ProviderLevel | undefined, qualityLabels: QualityLabels | null, redundant?: boolean): string {
```
zetagame, line 9: level?: ProviderLevel

zetagame: nvm. It's to avoid re-ordering right?

src/js/providers/utils/play-promise.ts
mamaddox: Looks like the `code` property is no longer used in favor of just the `name`.
https://developer.mozilla.org/en-US/docs/Web/API/DOMException#Properties

src/js/providers/utils/quality-labels.ts
mamaddox: Actually after taking a look, it looks like that can be since it could be `qualityLabels?`. It looks like the qualityLabels passed in come from the player config, so they could be undefined.
Nice catch!

src/js/providers/utils/quality-labels.ts
jnatalzia: What was the eventual resolution here? just wanna be sure i'm not missing anything

src/js/providers/utils/quality-labels.ts
jnatalzia: 👍

src/js/providers/utils/quality-labels.ts
mamaddox: I think it's safe to assume that bandwidth/bitrate will be a number based on the `level` types that are passed into `generateLabel`. Those types are the direct hls/shaka/flash provider levels and the type definitions are outlined here:
https://github.com/jwplayer/jwplayer/pull/3657#discussion_r408314156
For hls, `bitrate` is a number and same for shaka with `bandwidth`

src/js/providers/utils/play-promise.ts
```diff
@@ -1,41 +1,34 @@
-export default function createPlayPromise(video) {
-    return new Promise(function(resolve, reject) {
+export default function createPlayPromise(video: HTMLVideoElement): Promise<void> {
+    return new Promise(function(resolve: () => void, reject: (reason: DOMException) => void): void {
         if (video.paused) {
-            return reject(playPromiseError('NotAllowedError', 0, 'play() failed.'));
+            return reject(new DOMException('play() failed.', 'NotAllowedError'));
```
robwalch, line 4: I don’t know if it’s safe to use DOMException as a constructor in all browsers. The recommendation was to use it as an interface

src/js/providers/utils/play-promise.ts
```diff
@@ -1,41 +1,34 @@
...
             if (e.type === 'playing') {
                 resolve();
             } else {
                 const message = `The play() request was interrupted by a "${e.type}" event.`;
                 if (e.type === 'error') {
-                    reject(playPromiseError('NotSupportedError', 9, message));
+                    reject(new DOMException(message, 'NotSupportedError'));
                 } else {
-                    reject(playPromiseError('AbortError', 20, message));
+                    reject(new DOMException(message, 'AbortError'));
```
robwalch, line 28: No more error code? Browsers that need this polyfill do not know what those Error types are or what codes to assign

src/js/providers/utils/play-promise.ts
mamaddox: Ah. I see now. Misread that comment. Whoops. 

Issues I run into with using `as DOMException` is that the `code` and `message` are readonly properties that I cannot assign, so I can't do:

> const error = new `Error`(message) as `DOMException`

But `Error` also does not contain the `code` property so I cannot do:

> return `error` as `DOMException`

Why I had originally made that new error type. As I look though, could potentially leverage `PlayerError` as I think Joe had mentioned...

src/js/providers/utils/play-promise.ts
```diff
@@ -1,40 +1,44 @@
-export default function createPlayPromise(video) {
-    return new Promise(function(resolve, reject) {
+type PlayPromiseError = Omit<DOMException, 'name' | 'code'> & {
```
mamaddox, line 1: So I wound up remaking the `PlayPromiseError` type because I think it is important to note that it is a DOMException, but since those properties are readonly, they needed some way to be set.
Open to thoughts.

PR #3661, opened 2020-04-17
[JW8-11074] Handle touch event on iOS10 with controls disabled

robwalch:
### This PR will...
Add workaround that fixes ad clickthough with controls disabled in iOS 10 and lower.

The player's instream ad click through 'touchstart' listener is not dispatched when controls are disabled on on iOS 10. I added the `document.body` "touchstart" listener to investigate which elements on the page were receiving the "touchstart" event, but then everything suddenly worked! I accept this as a valid workaround because I cannot debug iOS 10 in Catalina with Safari 13.1. Since everything else works in Safari 11+, adding a noop handler only in Safari 10 is an acceptable and safe workaround.

No

#### Addresses Issue(s):
JW8-11074

### Checklist

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -21,6 +21,12 @@ const LONG_PRESS_DELAY = 500;
 let longPressTimeout;
 let lastInteractionListener;
 
+if (OS.iOS && OS.version.major < 11) {
+    const body = self.document.body;
+    // When controls are disabled iOS 10 does not dispatch media element touchstart/end events without this line
+    body.ontouchstart = body.ontouchstart || function() {};
+}
+
 export default class UI extends Events {
 
     constructor(element, options) {


src/js/utils/ui.js
```diff
@@ -21,6 +21,12 @@ const LONG_PRESS_DELAY = 500;
 let longPressTimeout;
 let lastInteractionListener;
 
+if (OS.iOS && OS.version.major < 11) {
+    const body = self.document.body;
```
jnatalzia, line 25: where does this `self` keyword get instantiated? is this something that works on ios 10? something that is set in webpack? has self been `this` all along?

src/js/utils/ui.js
```diff
@@ -21,6 +21,12 @@ const LONG_PRESS_DELAY = 500;
...
+    // When controls are disabled iOS 10 does not dispatch media element touchstart/end events without this line
+    body.ontouchstart = body.ontouchstart || function() {};
```
jnatalzia, line 27: should we use `addEventLIstener` here to avoid overwriting other listeners sites might place `ontouchstart`? or does this workaround require setting the function itself

jnatalzia: with the `||` here we should be safe now that i think about it. anything set before this executes is preserved and anything set after will overwrite this no-op

src/js/utils/ui.js
robwalch: https://developer.mozilla.org/en-US/docs/Web/API/Window/self

In hls.js we replaced all references to `window` with `self` and this is one of the changes I want to make in our projects with the headless player. https://github.com/jwplayer/jwplayer-commercial/pull/7252

Using `self` as the global object works in node and in browsers both in the main event loop and in workers.

src/js/utils/ui.js
robwalch: yep. I figured is someone else added `ontouchstart` then we don't need to.

src/js/utils/ui.js
jnatalzia: the more you know - now the `window.top === window.self` check in analytics makes more sense, thanks!

jnatalzia:
test this please

PR #3667, opened 2020-04-20
[JW8-11032] Typescript 3 files

zetagame:
### This PR will...
Typescripts the following files:
- storage.js
- tracks-loader.js
- tracks-helper.js

Also:
- Adds dependency for webpack env types.
- Gates potential type error in dfxp function when argument is null.

TS

- `any` call signature on loader. 
- `itemName` in storage.js. Any way around no-unused-args in this use case?
#### Addresses Issue(s):

JW8-11032

### Checklist


src/js/controller/tracks-loader.ts
```diff
@@ -4,17 +4,22 @@ import { ajax } from 'utils/ajax';
 import { localName } from 'parsers/parsers';
 import srt from 'parsers/captions/srt';
 import dfxp from 'parsers/captions/dfxp';
-import { composePlayerError, convertToPlayerError, ERROR_LOADING_CAPTIONS } from 'api/errors';
+import { composePlayerError, convertToPlayerError, ERROR_LOADING_CAPTIONS, PlayerError } from 'api/errors';
```
robwalch, line 7: `PlayerError` should be imported as a type on a new line:
`import type { PlayerError } from 'api/errors';`

src/js/controller/tracks-loader.ts
```diff
@@ -31,12 +36,17 @@ export function cancelXhr(tracks) {
     }
 }
 
-function convertToVTTCues(cues) {
+function convertToVTTCues(cues: GenericObject[]): VTTCue[] {
```
robwalch, line 39:
```suggestion
function convertToVTTCues(cues: CaptionEntryData[]): VTTCue[] {
```

src/js/controller/tracks-loader.ts
```diff
@@ -31,12 +36,17 @@ export function cancelXhr(tracks) {
...
     // VTTCue is available natively or polyfilled where necessary
     return cues.map(cue => new VTTCue(cue.begin, cue.end, cue.text));
 }
 
-function xhrSuccess(xhr, track, successHandler, errorHandler) {
+function xhrSuccess(
+    xhr: XMLHttpRequest, 
+    track: GenericObject,
```
robwalch, line 46:
```suggestion
    track: PlaylistItemTrack, 
```

src/js/controller/tracks-loader.ts
```diff
@@ -96,8 +107,11 @@ function xhrSuccess(xhr, track, successHandler, errorHandler) {
     }
 }
 
-function loadVttParser() {
-    return require.ensure(['parsers/captions/vttparser'], function (require) {
-        return require('parsers/captions/vttparser').default;
-    }, chunkLoadWarningHandler(301131), 'vttparser');
+function loadVttParser(): any {
```
robwalch, line 110:
```suggestion
function loadVttParser(): Promise<VTTParser> {
```

src/js/controller/tracks-loader.ts
```diff
@@ -64,14 +74,15 @@ function xhrSuccess(xhr, track, successHandler, errorHandler) {
             const responseText = xhr.responseText;
             if (responseText.indexOf('WEBVTT') >= 0) {
                 // make VTTCues from VTT track
-                loadVttParser().then(VTTParser => {
+                loadVttParser().then((VTTParser: (window: Window, decoder?: () => GenericObject) => void): void => {
```
robwalch, line 77: `VTTParser` typing should be temporary until vttparser.js is TSed. Can you put that in types/generic.type for now or below in `loadVttParser(): Promise<(window: Window, decoder?: () => void)>`? It doesn't return an object, we get one with the `new` keyword.

src/js/controller/tracks-loader.ts
```diff
@@ -64,14 +74,15 @@ function xhrSuccess(xhr, track, successHandler, errorHandler) {
...
+                
                     const parser = new VTTParser(window);
                     vttCues = [];
-                    parser.oncue = function(cue) {
+                    parser.oncue = function(cue: GenericObject): void {
```
robwalch, line 81:
```suggestion
                    parser.oncue = function(cue: VTTCue): void {
```

src/js/model/storage.ts
```diff
@@ -1,30 +1,38 @@
...
             if (val) {
                 memo[key] = key !== 'captions' ? serialize(val) : JSON.parse(val);
             }
             return memo;
         }, {});
-    },
-    track(model) {
+    }
+
+    track(model: MediaModel): void {
```
robwalch, line 35:
```suggestion
    track(model: SimpleModel): void {
```

SimpleModel is used until the core (player) Model module is loaded. We don't persist media (playlist item) state.

src/js/parsers/captions/dfxp.ts
```diff
@@ -6,9 +6,10 @@ import type { CaptionEntryData } from './captions.types';
 
 type ParagraphNodes = HTMLCollectionOf<Element & { text?: string }>;
 
-export default function Dfxp(xmlDoc: XMLDocument): CaptionEntryData[] {
+export default function Dfxp(xmlDoc: XMLDocument | null): CaptionEntryData[] {
     if (!xmlDoc) {
         parseError(306007);
+        return [];
```
robwalch, line 12: `parseError` throws so we do not need to return here. Try using `never` below:

```ts
function parseError(code: number): never {
```

src/js/model/storage.ts
```diff
@@ -1,30 +1,38 @@
 import { serialize } from 'utils/parser';
 import ApiSettings from 'api/api-settings';
+import type { GenericObject } from 'types/generic.type';
+import type { MediaModel } from 'controller/model';
 
 let storage = {
-    removeItem: function() {}
+    removeItem: function(itemName: string) {
+        // No unused arguments.
```
mamaddox, line 8: I remember seeing this issue before and this is what we did in plugins/utils.
https://github.com/jwplayer/jwplayer/blob/238ff21f60af85c85943bfcf1e8fc7575fc17ff8/src/js/plugins/utils.ts#L9-L12

src/js/parsers/captions/dfxp.ts
zetagame: can I return parseError then? Need to pass type-check

src/js/model/storage.ts
zetagame: awesome, glad to know that's an option. Thanks for following up!

src/js/parsers/captions/dfxp.ts
robwalch: `never` indicates that the function will not complete.

zetagame: nice, didn't see the screenshot before for some reason. Trying it out now!

src/js/controller/tracks-loader.ts
zetagame: without webpack-env, the name `require` isn't identified and it will throw an error. If it is included (see package.json changes), require.ensure is defined as returning void, so if I don't define as `any` the `then` method isn't supposed to exist.

robwalch: Did you try using `as` for the require.ensure statement?

How does package.json matter? I don’t know what changes you mean.

robwalch: Tech-debt for the near future: we should replace `require.ensure` code splitting imports with `import()`. Harmony syntax is supported by the latest versions of webpack and rollup.

zetagame: yeah, I tried that as well

src/js/controller/tracks-helper.ts
```diff
@@ -1,4 +1,18 @@
-export function createId(track, tracksCount) {
+import type { TextTrackLike } from 'types/generic.type';
+import type { PlaylistItemTrack } from 'playlist/track';
+
+export type Track = (TextTrackLike | PlaylistItemTrack) & {
```
robwalch, line 4: `createLabel` is only used (ignore test) in captions.js. Maybe that's where we should define `Track` or `CaptionsTrack`? Maybe we should just move `createLabel` there too.

src/js/controller/tracks-loader.ts
```diff
@@ -5,16 +5,24 @@ import { localName } from 'parsers/parsers';
...
+import type { PlayerError } from 'api/errors';
+import type { PlaylistItemTrack } from 'playlist/track';
+import type { CaptionEntryData } from 'parsers/captions/captions.types';
+import type VTTParser from 'parsers/captions/vttparser';
 
-export function loadFile(track, successHandler, errorHandler) {
-    track.xhr = ajax(track.file, function(xhr) {
+export function loadFile(
+    track: PlaylistItemTrack, 
+    successHandler: (...args: any[]) => void,
```
robwalch, line 15: Why not `(cues: VTTCue[]) => void` like below?

src/js/controller/tracks-loader.ts
```diff
@@ -96,8 +110,11 @@ function xhrSuccess(xhr, track, successHandler, errorHandler) {
...
-    return require.ensure(['parsers/captions/vttparser'], function (require) {
-        return require('parsers/captions/vttparser').default;
-    }, chunkLoadWarningHandler(301131), 'vttparser');
+function loadVttParser(): any {
+    return require.ensure(['parsers/captions/vttparser'], 
+        function (require: NodeRequire): Promise<VTTParser> {
+            return require('parsers/captions/vttparser').default;
+        }, 
+        chunkLoadWarningHandler(301131), 'vttparser'
+    );
```
robwalch, line 119: you tried?
```suggestion
    ) as Promise<VTTParser> ;
```

src/js/model/storage.ts
```diff
@@ -1,30 +1,36 @@
...
     storage = window.localStorage || storage;
 } catch (e) {/* ignore */}
 
-function Storage(namespace, persistItems) {
-    this.namespace = namespace;
-    this.items = persistItems;
-}
+class Storage {
+    namespace: string;
+    items: any[];
```
robwalch, line 17: `items` is a list of keys
```suggestion
    items: string[];
```

src/js/model/storage.ts
```diff
@@ -1,30 +1,36 @@
...
+    constructor(namespace: string, persistItems: GenericObject[]) {
+        this.namespace = namespace;
+        this.items = persistItems;
+    }
 
-Object.assign(Storage.prototype, {
-    getAllItems() {
+    getAllItems(): GenericObject {
         return this.items.reduce((memo, key) => {
```
robwalch, line 24: `memo: GenericObject, key: string`

src/js/controller/tracks-helper.ts
zetagame: Good call. Removing the exports. We can move it over when caption.js is typescripted.

src/js/controller/tracks-loader.ts
```diff
@@ -96,8 +110,11 @@ function xhrSuccess(xhr, track, successHandler, errorHandler) {
     }
 }
 
-function loadVttParser() {
-    return require.ensure(['parsers/captions/vttparser'], function (require) {
-        return require('parsers/captions/vttparser').default;
-    }, chunkLoadWarningHandler(301131), 'vttparser');
+function loadVttParser(): VTTParser {
```
robwalch, line 113:
```suggestion
function loadVttParser(): Promise<VTTParser> {
```

src/js/parsers/captions/dfxp.ts
```diff
@@ -6,7 +6,7 @@ import type { CaptionEntryData } from './captions.types';
 
 type ParagraphNodes = HTMLCollectionOf<Element & { text?: string }>;
 
-export default function Dfxp(xmlDoc: XMLDocument): CaptionEntryData[] {
+export default function Dfxp(xmlDoc: XMLDocument | null): CaptionEntryData[] {
```
robwalch, line 9: Do we actually call `Dfxp(null)` anywhere? We shouldn't allow that. Adding `null` as an accepted type removes protection from calls that would pass in null.

The `if (!xmlDoc) {` guard is in case it's undefined, but that doesn't mean we should let TypeScript allow undefined or null arguments.

src/js/controller/tracks-loader.ts
zetagame: Woops, something funky happened in rebase. Fixing now

src/js/parsers/captions/dfxp.ts
zetagame: image

zetagame: image

src/js/controller/tracks-loader.ts
zetagame: Alright, added a type to allow this. We aren't really using `require.ensure` in the way specified by the the webpack documentation, so that may explain why `require.ensure` is marked as returning void in the webpack-env types package.

src/js/controller/tracks-loader.ts
zetagame: Yep. Ended up reverting to `any` since there were issues with every trick I've tried to type the return. So far the only thing that works is to cast to unknown before casting to Promise<VTTParser>, which doesn't seem to have an advantage over `any`.

src/js/parsers/captions/dfxp.ts
zetagame: Added a check here to throw an error if responseXML is null

robwalch: I don't see it https://github.com/jwplayer/jwplayer/pull/3667/files#diff-6bb2a789e01fbaf7e5d02917600e01c9L58

src/js/controller/tracks-loader.ts
```diff
@@ -64,14 +80,15 @@ function xhrSuccess(xhr, track, successHandler, errorHandler) {
             const responseText = xhr.responseText;
             if (responseText.indexOf('WEBVTT') >= 0) {
                 // make VTTCues from VTT track
-                loadVttParser().then(VTTParser => {
-                    const parser = new VTTParser(window);
+                loadVttParser().then((VttParser: (window: Window, decoder?: () => void) => void): void => {
```
robwalch, line 83: `loadVttParser().then((VttParser: VTTParser): void => {`

src/js/controller/tracks-loader.ts
```diff
@@ -64,14 +80,15 @@ function xhrSuccess(xhr, track, successHandler, errorHandler) {
...
+                
+                    const parser = new VttParser(window);
```
robwalch, line 85: This isn't being typed a `VTTParser`. The point of importing it is to get the type safety for it here.

PR #3668, opened 2020-04-20
[JW8-11050] Start playlist on a specific index

robwalch:
# This PR will...
Start playlist on a specific index using `config.playlistIndex`. Values are wrapped based on loaded playlist length so negative values count from the end of the playlist.

Allows for developers to setup the player so that playback begins on a specific playlist item. Previously we always started on the first item, without any way to cancel preloading or provider selection based on that item.

https://github.com/jwplayer/jwplayer-commercial/pull/7531 (config schema update)

#### Addresses Issue(s):
JW8-11050

### Checklist

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -28,6 +28,7 @@ export const Defaults = {
     enableShortcuts: true,
     height: 360,
     intl: {},
+    item: 0,
     language: 'en',
     liveTimeout: null,
     localization: en,
@@ -96,6 +97,15 @@ const Config = function(options, persisted) {
     config.mute = !!config.mute;
     config.language = language;
     config.intl = intl;
+
+    const playlistIndex = config.playlistIndex;
+    if (playlistIndex) {
+        config.item = playlistIndex;
+    }
+
+    if (!isNumber(config.item)) {
+        config.item = 0;
+    }
 
     // If autoPause is configured with an empty block,
     // default autoPause.viewability to true.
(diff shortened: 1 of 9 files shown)


src/js/api/config.js
```diff
@@ -28,6 +28,7 @@ export const Defaults = {
     enableShortcuts: true,
     height: 360,
     intl: {},
+    item: 0,
```
zetagame, line 31: Thoughts on replacing `item` with `index` or something similar? When I think of `item` in the player, it's usually always an object.

robwalch: I would like that better too, but I do not want to change any model property names until the next major version.

robwalch: That's why I also push for us to use `PlaylistItem` as the type name or part in any API method related to a playlist item object. `item` or event `Item` is vague and can be anything.

src/js/playlist/playlist.js
```diff
@@ -47,6 +47,15 @@ export function normalizePlaylistItem(model, item, feedData) {
     return formatItem(playlistItem);
 }
 
+export function wrapPlaylistIndex(index, length) {
```
jnatalzia, line 50: can we write a test for this?

robwalch: https://github.com/jwplayer/jwplayer/pull/3668/commits/b22a2754b515bac16fcc76278685a2c387d84852

PR #3669, opened 2020-04-22
Safer iOS 10 tap workaround

robwalch:
### This PR will...
Fix a regression that could cause jwplayer.js to fail to embed when put in the head element. 

Moves a workaround added in #3661 to safer place and avoids an exception if the ui module is run before `document.body` is defined.

#### Addresses Issue(s):
JW8-11086

### Checklist

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -21,12 +21,6 @@ const LONG_PRESS_DELAY = 500;
 let longPressTimeout;
 let lastInteractionListener;
 
-if (OS.iOS && OS.version.major < 11) {
-    const body = self.document.body;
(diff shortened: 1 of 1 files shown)


src/js/utils/ui.js
```diff
@@ -250,6 +244,13 @@ const eventRegisters = {
         initInteractionListeners(ui);
     },
     tap(ui) {
+        if (OS.iOS && OS.version.major < 11) {
```
jnatalzia, line 247: thoughts on a file scoped var to gate whether or not this was added so we only have to do it once? might be overkill given the below is prob fast but could save an ms or two on every tap after the first

PR #3675, opened 2020-04-29
[JW8-11112] HTML5 Provider Safari CC and mediaType event fixes

robwalch:
### This PR will...
- Detect captions in Safari after they have been disabled in MacBook touchbar on a previous page load
- Do not trigger a "mediaType" event of type audio, right before "mediaType" of type video

When subtitles are disabled via the touchbar, media element text tracks are disabled causing our player to ignore them. This also messes with the timing of track change events in relation to media "loadeddata".

Both 608 tracks and videoWidth/Height may not be populated on "loadeddata". Using "canplay"  which fire a little later after enough media is loaded to begin playback ensures we detect text tracks and the content type (video or audio by checking video width and height) correctly.

#### Addresses Issue(s):
JW8-11112

### Checklist

### DIFF
--- a/src/js/providers/tracks-mixin.ts
+++ b/src/js/providers/tracks-mixin.ts
@@ -73,6 +73,7 @@ interface TracksMixin extends DefaultProvider {
     getSubtitlesTrack: () => number;
     addTextTracks: (tracksArray: PlaylistItemTrack[]) => TextTrackLike[];
     setTextTracks: (tracks: TextTrackList) => void;
+    addTrackListeners: (tracks: TextTrackList) => void;
     setupSideloadedTracks: (itemTracks: PlaylistItemTrack[]) => void;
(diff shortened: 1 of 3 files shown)


src/js/providers/tracks-mixin.ts
```diff
@@ -341,7 +342,9 @@ const Tracks: TracksMixin = {
                         continue;
                     }
 
-                    track.mode = mode;
+                    if (mode !== 'disabled' || trackId.indexOf('nativecaptions') !== 0) {
```
jnatalzia, line 345: can i get a quick explainer on why we don't set if the mode is `disabled` for non-native use cases?

robwalch: we can't detect cues in disabled tracks, and we don't consider disabled tracks items in the captions menu.

PR #3683, opened 2020-05-14
[JW8-11061] Fix exceptions on setup due to caption index

pajong:
### This PR will...
1. Send all sided loaded captions to the end of the menu
2. Set `captionsList` before setting the `defaultIndex` for captions

1. Side loaded captions are causing hls VTT captions to load incorrect index. Open for suggestions if there is a better way
2. This was causing exception when the `defaultIndex` from previous selection (before refreshing the browser) was larger than the number of side loaded captions loaded

#### Addresses Issue(s):

JW8-11061

### Checklist

### DIFF
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -35,7 +35,8 @@ const Captions = function(_model) {
                 /* eslint-disable no-loop-func */
                 const track = tracks[i];
                 if (_kindSupported(track.kind) && !_tracksById[track._id]) {
-                    _addTrack(track, true);
+                    track.sideloaded = true;
+                    _addTrack(track);
                     loadFile(track, (vttCues) => {
                         _addVTTCuesToTrack(track, vttCues);
                     }, error => {
@@ -71,7 +72,7 @@ const Captions = function(_model) {
 
             // To avoid duplicate tracks in the menu when we reuse an _id, regenerate the tracks array
             const allTracks = Object.keys(_tracksById).map(id => _tracksById[id]);
-            _tracks = allTracks.filter(track => !track.isSideload).concat(allTracks.filter(track => !!track.isSideload));
+            _tracks = allTracks.filter(track => !track.sideloaded).concat(allTracks.filter(track => !!track.sideloaded));
         }
 
         _setCaptionsList();
@@ -85,11 +86,10 @@ const Captions = function(_model) {
         track.data = vttCues;
     }
 
-    function _addTrack(track, isSideload) {
+    function _addTrack(track) {
         track.data = track.data || [];
         track.name = track.label || track.name || track.language;
         track._id = createId(track, _tracks.length);
-        track.isSideload = isSideload;
 
         if (!track.name) {
             const labelInfo = createLabel(track, _unknownCount);
--- a/src/js/controller/captions.js
+++ b/src/js/controller/captions.js
@@ -35,7 +35,7 @@ const Captions = function(_model) {
                 /* eslint-disable no-loop-func */
                 const track = tracks[i];
                 if (_kindSupported(track.kind) && !_tracksById[track._id]) {
-                    _addTrack(track);
+                    _addTrack(track, true);
                     loadFile(track, (vttCues) => {
                         _addVTTCuesToTrack(track, vttCues);
                     }, error => {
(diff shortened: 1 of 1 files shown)


src/js/controller/captions.js
```diff
@@ -84,10 +85,11 @@ const Captions = function(_model) {
         track.data = vttCues;
     }
 
-    function _addTrack(track) {
+    function _addTrack(track, isSideload) {
```
jnatalzia, line 88: we already have an attribute created for a different fix for distinguishing sideloaded captions, can you use this? https://github.com/jwplayer/jwplayer/pull/3570/files#diff-516afba75eaadc497dca75d8c83b0fe6R557

pajong: Updated, thank you!

src/js/controller/captions.js
```diff
@@ -35,6 +35,7 @@ const Captions = function(_model) {
                 /* eslint-disable no-loop-func */
                 const track = tracks[i];
                 if (_kindSupported(track.kind) && !_tracksById[track._id]) {
+                    track.sideloaded = true;
```
jnatalzia, line 38: do we still need this line? it should be covered by the previous changes right?

robwalch: where is that being set? The lined below depend on `track.sideloaded` being set to true at some point.

jnatalzia: https://github.com/jwplayer/jwplayer/pull/3683#discussion_r426621562

src/js/controller/captions.js
```diff
@@ -152,8 +154,8 @@ const Captions = function(_model) {
     function _setCaptionsList() {
         const captionsList = _captionsMenu();
         if (listIdentity(captionsList) !== listIdentity(_model.get('captionsList'))) {
-            _selectDefaultIndex(_defaultIndex);
             _model.set('captionsList', captionsList);
+            _selectDefaultIndex(_defaultIndex);
```
robwalch, line 158: This change will change event order which we should avoid.

 `_selectDefaultIndex ` calls `_setCurrentIndex` at the end. Can we split these apart so that we maintain current event order?

It would go something like:
```js
            const captionsMenuIndex = _selectDefaultIndex(_defaultIndex); // does not call _setCurrentIndex
            _model.set('captionsList', captionsList);
            _setCurrentIndex(captionsMenuIndex);
```
And below:
```js
this.selectDefaultIndex = (defaultIndex) =>
            const captionsMenuIndex = _selectDefaultIndex(defaultIndex);
            _setCurrentIndex(captionsMenuIndex);
};

Comment (robwalch) [reply src/js/controller/captions.js:158]: hmmmm... that would still change event order. How does the order prevent us from sorting side-loaded last? I'm a little unclear on this.

Comment (pajong) [reply src/js/controller/captions.js:158]: This is fixing another issue, not the side-loaded captions - this line is causing js exception when we have sideloaded captions. Here is the setup that breaks on last successful [player-develop](http://player-develop-test-jenkins.longtailvideo.com/builds/lastSuccessfulBuild/archive/test/harness.html?edition=ads):
```
{
  "file": "http://wowzaec2demo.streamlock.net/vod-multitrack/_definst_/smil:ElephantsDream/elephantsdream2.smil/playlist.m3u8",
  "tracks": [
      {
        "file": "//playertest.longtailvideo.com/assets/os/captions/bunny-en.vtt",
        "label": "Side-loaded",
        "kind": "captions"
      },
      {
            "file": "//playertest.longtailvideo.com/assets/os/captions/bunny-en.srt",
            "kind": "captions",
            "label": "English"
        }
 
  ],
  "analytics": {
    "debug": true
  }
}
```

src/js/controller/captions.js
pajong: The code that you pointed is not reached with this set up, so I had to add it here.

src/js/controller/captions.js
pajong: The problem here is that when the side-loaded captions are first loaded (without the playlistItem captions), the `defaultIndex` is greater than the size of the `captionsList`

src/js/controller/captions.js
jnatalzia: ah! okay makes sense

pajong:
test this please

jnatalzia:
@pajong the automation failures here are due to changes i made in the dash provider, this has since been fixed and you should be good to merge these changes

PR #3694, opened 2020-06-10
[JW8-11181] Check for elementFromPoint with clientX and clientY

pajong:
### This PR will...
- Check the element from `document.elementFromPoint` using `clientX` and `clientY`

- For IE 11, the pointer event `e` has incorrect values of `x` and `y`, causing the `body` of the `document` to be selected as `overElement`, instead of the actual `volumeSlider`
- Using `clientX` and `clientY` to get the position relative to the viewport fixes the issue

#### Addresses Issue(s):

JW8-11181

### Checklist

### DIFF
--- a/src/js/utils/ui.js
+++ b/src/js/utils/ui.js
@@ -311,7 +311,7 @@ const eventRegisters = {
                 if (e.pointerType !== 'touch' && 'x' in e) {
                     // elementFromPoint to handle an issue where setPointerCapture is causing a pointerout event
                     const overElement = document.elementFromPoint(e.x, e.y);
-                    const overElementFromClient = document.elementFromPoint(e.clientX, e.clientY);
+                    const overElementFromClient = 'clientX' in e ? document.elementFromPoint(e.clientX, e.clientY) : null;
                     if (!el.contains(overElement) && !el.contains(overElementFromClient)) {
                         triggerEvent(ui, OUT, e);
                     }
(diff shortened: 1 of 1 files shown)


src/js/utils/ui.js
```diff
@@ -311,7 +311,8 @@ const eventRegisters = {
                 if (e.pointerType !== 'touch' && 'x' in e) {
                     // elementFromPoint to handle an issue where setPointerCapture is causing a pointerout event
                     const overElement = document.elementFromPoint(e.x, e.y);
-                    if (!el.contains(overElement)) {
+                    const overElementFromClient = 'clientX' in e ? document.elementFromPoint(e.clientX, e.clientY) : null;
+                    if (!el.contains(overElement) && !el.contains(overElementFromClient)) {
                         triggerEvent(ui, OUT, e);
                     }
```
robwalch, line 317: I prefer that we avoid calling `elementFromPoint` twice. Each call has to do a lot of dom tree walking under the hood. We can just as easily return early if the first check passes, and skip the second call to `contains` when the event is missing `clientX` to optimize this code path. Feel free to OR `||` these if you only want one return statement:

```suggestion
                    if (el.contains(overElement)) {
                        return;
                    }
                    if ('clientX' in e && el.contains(document.elementFromPoint(e.clientX, e.clientY))) {
                        return;
                    }
                    triggerEvent(ui, OUT, e);
```

robwalch: The alternative is, can we always use `clientX` instead of `x`?

pajong:
test this please

PR #3695, opened 2020-06-10
[JW8-11180] Set track mode to hidden for iOS fullscreen

pajong:
### This PR will...
- Set `track.mode` to `hidden` for iOS native rendered captions

- For fullscreen iOS, setting `track.mode` to `disabled` does not hide the captions already displayed, resulting captions to be stuck while playing ads

#### Addresses Issue(s):

JW8-11180

### Checklist

### DIFF
--- a/src/js/providers/tracks-mixin.ts
+++ b/src/js/providers/tracks-mixin.ts
@@ -1,7 +1,7 @@
 import { loadFile, cancelXhr } from 'controller/tracks-loader';
 import { createId, createLabel } from 'controller/tracks-helper';
 import { parseID3 } from 'providers/utils/id3Parser';
-import { Browser } from 'environment/environment';
+import { Browser, OS } from 'environment/environment';
 import { MEDIA_META_CUE_PARSED, MEDIA_META, WARNING } from 'events/events';
 import { findWhere, each, filter } from 'utils/underscore';
 import type { TextTrackLike, MetadataEvent } from 'types/generic.type';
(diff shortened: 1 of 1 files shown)


src/js/providers/tracks-mixin.ts
```diff
@@ -214,7 +214,7 @@ const Tracks: TracksMixin = {
             // FF does not remove the active cue from the dom when the track is hidden, so we must disable it
```
robwalch, line 214: This comment looks like it needs to be updated.

src/js/providers/tracks-mixin.ts
```diff
@@ -211,12 +211,13 @@ const Tracks: TracksMixin = {
     disableTextTrack(): void {
         const track = this.getCurrentTextTrack();
         if (track) {
-            // FF does not remove the active cue from the dom when the track is hidden, so we must disable it
-            track.mode = 'disabled';
+            // IOS native captions does not remove the active cue from the dom when the track is disabled, so we must hide it
             const trackId = track._id;
-            if (trackId && trackId.indexOf('nativecaptions') === 0) {
+            if ((trackId && trackId.indexOf('nativecaptions') === 0) || (this.renderNatively && OS.iOS)) {
                 track.mode = 'hidden';
```
zetagame, line 217: what does this line do if we set to disabled right after either way?

pajong: As soon as we set to `hidden`, the text track is hidden on iOS. We want to set to disable as per Rob's comment above:

 `Could we hide the track and then disable it? Leaving a track hidden requires that the track continue to be processed (loading and parsing VTT files... etc.). The only subtitles track that should not be disabled is the one being shown.`

pajong:
test this please

PR #3698, opened 2020-06-12
[JW8-11120] [JW8-11119] Add tizen-controls and controls loading

mamaddox:
### This PR will...
This is the basis for tizen-controls in the player. It removes/updates the code used in the first demo app including the following:
* Extracts tizen controls from controls.js into `tizen-controls`
* Adds `tizen-controlbar`
* Removes previously added remote handlers and `isTizenApp` checks that are no longer needed from:
   * menu.js
   * settings-menu.js
   * rightclick.js
   * view.js
* Removes previously added `:not(jw-tizen-app)`css selectors that are no longer needed from:
   * controlbar.less
   * settings-menu.less 
* Updates `core-bundle-loader` and `controls-loader` to load tizen controls when tizen environment is detected
* Removes isTizenApp model attribute in favor of using a tizen environment check
* Adds `tizenApp` to `browser` and `environment` to differentiate between tizen web apps and the tizen browser
*Note: `OS.tizen` is currently used as opposed to `OS.tizenApp` for developing because the tizen simulator uses the tizen browser user agent.

Tizen Demo App

https://github.com/jwplayer/jwplayer-commercial/pull/7657

#### Addresses Issue(s):

[JW8-11120](https://jwplayer.atlassian.net/browse/JW8-11120)
[JW8-11119](https://jwplayer.atlassian.net/browse/JW8-11119)

### Checklist


src/css/controls/flags/tizen.less
```diff
@@ -1,93 +1,239 @@
...
+            border-radius: 0;
+            background-size: auto;
+            box-shadow: none;
+            left: 0;
+            bottom: 0;
+            padding: inherit;
+        }
 
         .jw-text {
-            font-size: @tizen-font-size;
```
jnatalzia, line 81: are these vars still useful at all?

src/js/api/core-bundle-loader.js
```diff
@@ -20,6 +21,10 @@ function selectBundle(model) {
     const polyfills = requiresPolyfills();
     const html5Provider = requiresProvider(model, 'html5');
 
+    if (OS.tizen) {
+        return loadWebCore();
```
jnatalzia, line 25: nice

src/js/controller/controls-loader.js
```diff
@@ -1,19 +1,31 @@
...
-        }, 'jwplayer.controls');
+        if (OS.tizen) {
+            controlsPromise = require.ensure(['view/controls/tizen/tizen-controls'], function (require) {
+                const ControlsModule = require('view/controls/tizen/tizen-controls').default;
+                ControlsLoader.controls = ControlsModule;
+                return ControlsModule;
+            }, function() {
+                controlsPromise = null;
+                chunkLoadWarningHandler(301130)();
+            }, 'jwplayer.controls.tizen');
```
jnatalzia, line 18: this is awesome!

src/js/controller/model.ts
```diff
@@ -285,7 +285,7 @@ class Model extends SimpleModel {
     // Desktop players must have autostart set to viewable
     setAutoStart(autoStart?: AutoStart): void {
         // Always autostart for tizen app
-        if (this.attributes.isTizenApp) {
+        if (OS.tizen) {
```
jnatalzia, line 288: 👍

src/js/view/controls/controls.js
```diff
@@ -467,148 +457,6 @@ export default class Controls extends Events {
         model.set('controlsEnabled', true);
     }
 
-    /* Want to move to controls/tv/tizen-controls.js which extends controls.js */
-    enableTizenControls(api, model) {
```
jnatalzia, line 471: great refactor

src/js/view/controls/tizen/tizen-controlbar.ts
```diff
@@ -0,0 +1,240 @@
...
+        appendChildren(elements.topContainer, topLayout);
+        appendChildren(elements.bottomContainer, bottomLayout);
+        appendChildren(this.el, layout);
+
+        // Initial State
+        elements.play.show();
+        elements.back.show();
+    }
+
+    getRightButton(activeButton: Button, containerLayout: ControlbarElement[]): Button | undefined {
```
jnatalzia, line 118: i like this approach to dynamic button finding, i'm curious your thoughts on a way to potentially create an abstraction of `getRight` and `getLeft`. all of the logic is exactly the same except the for loop condition. it might be more trouble than it's worth but something to explore!

src/js/view/controls/tizen/tizen-controlbar.ts
```diff
@@ -0,0 +1,240 @@
...
+                break;
+            default:
+                break;
+        }
+    }
+
+    setActiveButton(nextButton: Button): void {
+        const currentActiveButton = this.activeButton;
+        if (currentActiveButton === nextButton) {
+            return;
```
jnatalzia, line 220: nice

src/js/view/controls/tizen/tizen-controls.ts
```diff
@@ -0,0 +1,209 @@
...
+import TizenControlbar from 'view/controls/tizen/tizen-controlbar';
+import DisplayContainer from 'view/controls/display-container';
+import NextUpToolTip from 'view/controls/nextuptooltip';
+import SettingsMenu from 'view/controls/components/menu/settings-menu.js';
+import { addClass } from 'utils/dom';
+import { STATE_PLAYING, STATE_PAUSED } from 'events/events';
+import type ViewModel from 'view/view-model';
+import type { PlayerAPI } from 'types/generic.type';
+
+const Controls = require('../controls').default;
```
jnatalzia, line 11: what's the reasoning behind this syntax?

src/js/view/controls/tizen/tizen-controls.ts
```diff
@@ -0,0 +1,209 @@
...
+    playerContainer: HTMLElement;
+    div: HTMLElement | null;
+    backdrop: HTMLElement | null;
+    pauseDisplay: HTMLElement | null;
+    displayContainer: DisplayContainer | null;
+    controlbar: TizenControlbar | null;
+    settingsMenu: SettingsMenu | null;
+    showing: boolean;
+    keydownCallback: ((evt: KeyboardEvent) => void) | null;
+    addControls: any;
```
jnatalzia, line 30: i assume these would get re-typed when controls is typed? or maybe removed?

src/js/view/controls/tizen/tizen-controls.ts
```diff
@@ -0,0 +1,209 @@
...
+        const element = this.context.createElement('div');
+        element.className = 'jw-tizen-controls jw-tizen-reset';
+        this.div = element;
+    
+        const backdrop = this.context.createElement('div');
+        backdrop.className = 'jw-controls-backdrop jw-reset';
+        this.backdrop = backdrop;
+
+        // Pause Display
+        if (!this.pauseDisplay) {
```
jnatalzia, line 61: what are your thoughts on using templating for this similar to how we construct other complex elements? (https://github.com/jwplayer/jwplayer-commercial/blob/master/src/js/templates/dismissButton.js)

src/js/view/controls/tizen/tizen-controls.ts
```diff
@@ -0,0 +1,209 @@
...
+        const localization = model.get('localization');
+        this.settingsMenu = new SettingsMenu(api, model.player, this.controlbar, localization);
+
+        const handleKeydown = (evt: KeyboardEvent) => {
+            if (this.controlbar) {
+                this.controlbar.handleKeydown(evt, this.showing);
+            }
+            switch (evt.keyCode) {
+                case 37: // left-arrow
+                    this.userActive();
```
jnatalzia, line 121: you can remove all these cases and just make it a fall-through (unless out lint rules don't allow that)
```
case 37:
case 39:
...etc:
  this.userActive();
  break;
```

src/js/view/controls/tizen/tizen-controls.ts
```diff
@@ -0,0 +1,209 @@
...
+                    this.removePlayer(api, model);
+                    break;
+                default:
+                    break;
+            }
+        };
+
+        // Trigger backClick when all videos complete      
+        api.on('playlistComplete', () => {
+            this.onBackClick(api, model);
```
jnatalzia, line 173: nice

src/js/view/controls/tizen/tizen-controls.ts
```diff
@@ -0,0 +1,209 @@
...
+        api.setFullscreen(true);
+
+        model.set('controlsEnabled', true);
+    }
+
+    disable(model: ViewModel): void {
+        if (this.keydownCallback) {
+            document.removeEventListener('keydown', this.keydownCallback);
+        }
+        super.disable.apply(this, [model]);
```
jnatalzia, line 191: `super.disable.call(this, model)` stops an extra array from being created 👍

jnatalzia: same comment on the below

src/js/view/controls/tizen/tizen-controlbar.ts
mamaddox: +1 I 100% would love to do this. I spent a little time trying to figure out a solution, but will definitely explore some more.

src/css/controls/flags/tizen.less
mamaddox: good catch. not anymore.

src/js/view/controls/tizen/tizen-controls.ts
mamaddox: These should be removed when controls is typed since they are methods on controls/events. They were throwing a type error when called.

src/js/view/controls/tizen/tizen-controls.ts
mamaddox: Controls uses `require()` because there was an issue when using `import`. Because we resolve the import file paths by checking commercial first and then os, Controls was being imported as CommercialControls, which was then overwriting the TizenControls. Looking into some workarounds, though this may be resolved when controls get typed.

src/js/utils/browser.ts
```diff
@@ -36,6 +36,10 @@ export function isTizen(): boolean {
     return userAgentMatch(/SMART-TV/);
 }
 
+export function isTizenApp(): boolean {
+    return isTizen() && userAgentMatch(/Version\//);
```
robwalch, line 40: Could this be made a little more specific?

src/js/controller/controls-loader.js
```diff
@@ -1,19 +1,31 @@
...
-            chunkLoadWarningHandler(301130)();
-        }, 'jwplayer.controls');
+        if (OS.tizen) {
+            controlsPromise = require.ensure(['view/controls/tizen/tizen-controls'], function (require) {
+                const ControlsModule = require('view/controls/tizen/tizen-controls').default;
+                ControlsLoader.controls = ControlsModule;
+                return ControlsModule;
+            }, function() {
+                controlsPromise = null;
+                chunkLoadWarningHandler(301130)();
```
robwalch, line 17: There should be a different error code per file (so we know it's jwplayer.controls.tizen.js that failed to load).

src/js/view/controls/tizen/tizen-controls.ts
robwalch: Typing won't change import paths. I think the problem is that CommercialControls is not a class so you can't use it as a base for extending.

Why not make CommercialControls a class so that it can extend this class? It does have some functionality related to ads skip button UI that we may need in certain scenarios.

mamaddox: I definitely should have been more clear about typing. What I meant by that was when I had started typing Controls before, I had thought about creating an abstract BaseControls class that could be used by new controls components. So, in this case, TizenControls would import/extend BaseControls as opposed to importing/extending Controls, which I think might solve the issue.

But, from what I could tell, I think the larger issue with this is that CommercialControls instantiates a Controls component.
https://github.com/jwplayer/jwplayer-commercial/blob/master/src/js/view/controls/controls.js#L12
This is what was overriding the TizenControls. Instead of explicitly using the OS Controls here, I tried using the controls component set by the controls-loader (ControlsLoader.controls) and it creates a kind of circular dependency for Tizen. 
TizenControls extends/imports Controls (which is loaded as CommercialControls) and then CommercialControls creates a new instance of TizenControls.
I think converting CommercialControls to a class could possibly result in the same issue since CommercialControls would extend TizenControls and TizenControls would extend the imported CommercialControls.

I'm also curious as to how CommercialControls could be loaded because it seems like importing/using them via OS Controls causes a sort of circular dependency. But, if TizenControls doesn't use OS Controls, CommercialControls wouldn't be loaded from the path resolution. Maybe a commercial controls-loader would be needed and CommercialControls would become like a decorator?

mamaddox: Converting CommercialControls to a class resolved this! @robwalch 🎉

src/js/view/controls/tizen/tizen-controlbar.ts
```diff
@@ -0,0 +1,251 @@
...
+import TimeSlider from 'view/controls/components/timeslider';
+import CustomButton from 'view/controls/components/custom-button';
+import button, { Button } from 'view/controls/components/button';
+import { SimpleTooltip } from 'view/controls/components/simple-tooltip';
+import { toggleClass } from 'utils/dom';
+import type { PlayerAPI } from 'types/generic.type';
+import type ViewModel from 'view/view-model';
+import type NextUpToolTip from 'view/controls/nextuptooltip';
+
+type ControlbarElement = HTMLElement | Button | TimeSlider | any;
```
zetagame, line 12: This is always any

src/js/view/controls/tizen/tizen-controlbar.ts
```diff
@@ -0,0 +1,251 @@
...
+
+function div(classes: string): HTMLElement {
+    const element = document.createElement('div');
+    element.className = `jw-reset ${classes}`;
+    return element;
+}
+
+const appendChildren = (container: HTMLElement, elements: ControlbarElement[]) => {
```
zetagame, line 20: Are we checking for e.element() because this function takes two types of arguments (JS views and HTML elements)? Rather than allowing it to pass the JS view or the element, we should set this in a way so it's always an element being passed in (single responsibility principle).

src/js/view/controls/tizen/tizen-controlbar.ts
mamaddox: Nice catch!

src/js/view/controls/tizen/tizen-controlbar.ts
mamaddox: 👍 This gave me an idea to extract the extract out the elements in a new function to pass to appendChildren.

src/js/utils/browser.ts
mamaddox: It looks like the only difference between the browser and web application user agent is the inclusion of `Version` and it looks like the web app doesn't have `SamsungBrowser`. I think that might be as specific as we can get.

jnatalzia:
test this please

src/css/controls/flags/tizen.less
```diff
@@ -1,93 +1,239 @@
 @import "../../shared-imports/vars.less";
 
 .jw-tizen-app {
-    // Increasing the size of the svg-icons - icon.less
-    .jw-svg-icon {
-        .square(60px);
+    .jw-title {
+        /* Important is required so the title component is never displayed */
+        /* stylelint-disable-next-line declaration-no-important */
+        display: none !important;
```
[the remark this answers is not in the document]
mamaddox: don't actually think this is needed

src/css/controls/imports/settings-menu.less
```diff
@@ -180,7 +180,7 @@
         }
     }
 
-    .jw-breakpoint-7 &:not(.jw-tizen-app &) {
```
robwalch, line 183: Instead of complicating these selectors (making them more specific which could override other selectors), how about having the tizen controls not apply these breakpoint classes? Just apply the breakpoint(s) class you want in tizen controls by overriding the method that is adding them in controls. This is something that in general we should try to do to simplify our selectors.

src/js/view/controls/components/menu/menu.js
```diff
@@ -65,27 +65,6 @@ export default class Menu extends Events {
             const key = sourceEvent && normalizeKey(sourceEvent.key);
             let focusEl;
 
-
-            switch (sourceEvent.keyCode) {
```
robwalch, line 69: Why were these moved to tizen? How are these inputs implemented for desktop now?

src/css/controls/imports/settings-menu.less
mamaddox: I felt the same way which is why I am removing all the `not(.jw-tizen-app)` selectors from the css classes. These were initially added for quick styling in the first app. All the tizen styling is in `tizen.less` now and the fullscreen flag is added in tizen-controls.

src/js/view/controls/components/menu/menu.js
mamaddox: The original implementation was never changed so they still exist for desktop right below. For the first app, this was a quick workaround for navigation through the settings-menu. This is just some cleanup from the first app.

src/css/controls/imports/settings-menu.less
robwalch: Sorry I was commenting on something you fixed, as if it was a new change. ☕☕☕ Nice fix.

PR #3699, opened 2020-06-12
[JW8-11194] Handle async playlist item edge-cases

robwalch:
### This PR will...
Handle cases where the last item or all items are rejected/skipped using `setPlaylistItemCallback`.

Fixes a bug where the player would error when the last item was rejected, instead of going to a complete state or repeating the playlist.

When all items are rejected before setup, then setup should error with new "setupError" code 102700.

#### Addresses Issue(s):
JW8-11194

### Checklist

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -27,8 +27,18 @@ import { PLAYER_STATE, STATE_BUFFERING, STATE_IDLE, STATE_COMPLETE, STATE_PAUSED
 import ProgramController from 'program/program-controller';
 import initQoe from 'controller/qoe';
 import { BACKGROUND_LOAD_OFFSET } from 'program/program-constants';
-import { composePlayerError, convertToPlayerError, getPlayAttemptFailedErrorCode, MSG_CANT_PLAY_VIDEO, MSG_TECHNICAL_ERROR,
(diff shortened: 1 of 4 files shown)


src/js/controller/controller.js
```diff
@@ -924,6 +934,25 @@ Object.assign(Controller.prototype, {
...
 
             return _programController.setActiveItem(wrappedIndex).catch(error => {
+                if (error.code === ASYNC_PLAYLIST_ITEM_REJECTED) {
+                    // If all items were rejected throw. This will fail setup with setupError code 102700
+                    const allSkipped = _programController.asyncItems.reduce((skipped, asyncItem) =>
+                        skipped && asyncItem.skipped, true);
+                    if (allSkipped) {
+                        throw error;
+                    }
+                    // If the last item is rejected we need to stop and complete playlist playback. Otherwise,
```
jnatalzia, line 944: is this true in a recommendations case? if the last item is rejected and we have recs, shouldn't it move onto those or no?

robwalch: We can't because "nextUp" is not updated without a playlist item event. This would require changes to related and nextUp which I am not doing as part of this 2 pointer. (Says so below and in the ticket).

robwalch: Yeah if "nextUp" was the recommendation we'd be fine, but it's not because the "playlistItem" event for the last item is what updates "nextUp" and it's been skipped. The logic that updates "nextUp" needs to be handled either here in the controller and/or in the program-controller to work in sync with this system. Or, we need a new event to signal an item was skipped, to update related.

PR #3705, opened 2020-06-26
[JW8-11180] Change the order back to hide the iOS captions after disabling for FF

pajong:
### This PR will...
- Change the order back to disabling captions for FF before changing to hidden for iOS fullscreen

- The issue is not fixed if we change the order

#### Addresses Issue(s):

JW8-11180

### Checklist


pajong:
@zetagame Your comment was right for this: https://github.com/jwplayer/jwplayer/pull/3695#discussion_r444443453
It looks like I didn't rebuild to test after changing the order - sorry!

PR #3707, opened 2020-06-30
[JW8-11067] Headless Player Updates: Instream

robwalch:
### This PR will...
Update headless player for better compatibility with instream ad playback.
- Remove cyclical `vastAd` object from instream playlist items received from vast plugin
- Update provider volume and mute upon activation
- Do not call `container.removeChild` in headless player, regardless of provider implementation
- Remove `mediaPool.syncMute`  calls from headless player

Allows for headless external providers to play ads as expected using instream.

https://github.com/jwplayer/jwplayer-commercial/pull/7617

#### Addresses Issue(s):
JW8-11067

### Checklist

### DIFF
--- a/src/js/program/media-controller.js
+++ b/src/js/program/media-controller.js
@@ -251,6 +251,10 @@ export default class MediaController extends Events {
 
         this.item = item;
         this.provider.init(item);
+        if (__HEADLESS__) {
+            this.provider.mute(this.model.getMute());
+            this.provider.volume(this.model.get('volume'));
+        }
     }
 
     set audioTrack(index) {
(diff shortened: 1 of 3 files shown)


src/js/program/media-controller.js
```diff
@@ -280,7 +284,12 @@ export default class MediaController extends Events {
                 if (provider.removeFromContainer) {
                     provider.removeFromContainer();
                 } else {
-                    container.removeChild(provider.video);
+                    if (__HEADLESS__) {
```
pajong, line 287: Sorry for nitpick - this can be `else if` and `else`

robwalch: I prefer not to chain headless blocks like this. Keeping it in it's own scope makes it clearer what the headless player does vs normal operation.

PR #3714, opened 2020-07-13
[JW8-11247] Tracking seeking state in video listener mixins

robwalch:
### This PR will...
Set `provider.seeking` to `true` when receiving a `"seeking"` event from the video element. 

The video-event-listener mixin sets `provider.seeking` to `false` on `"seeked"` but expects the provider to set it to` true` when seeking starts. hlsjs and shaka only update `this.seeking` when seeking is performed using the player API. The `seeking` flag is never updated when seeking is initiated externally (by setting `video.currentTime`).

As a result, hlsjs and shaka's `"waiting"` video event handler interpret this event as a stall, because the `seeking` flag is `false` when it should be `true`.

1. Ultimately the seeking/waiting listeners and seek/stall triggers should be implemented in the mixin, normalizing them across providers. We're still triggering JWP "seek" events when seeks are made externally which should be addressed eventually.
2. The video-listener-mixin does not extend the provider interface. Each method in this mixin is an HTMLVideoElement event handler. Providers bind event listeners to themselves using the mixin method's names, but the methods are not added to the provider. (See TS changes) This is important because providers have a `seeking` flag, while `"seeking"` is also the name of a video event and video-listener-mixin method.

No.

#### Addresses Issue(s):
JW8-11247

### Checklist

### DIFF
--- a/src/js/providers/default.ts
+++ b/src/js/providers/default.ts
@@ -7,7 +7,6 @@ import type * as Event from 'events/events';
 import type { TracksMixin, SimpleAudioTrack } from 'providers/tracks-mixin';
 import type { VideoActionsInt } from 'providers/video-actions-mixin';
 import type { VideoAttachedInt } from 'providers/video-attached-mixin';
-import type { VideoListenerInt } from 'providers/video-listener-mixin';
 import type PlaylistItem from 'playlist/item';
 import type { QualityLevel } from 'providers/data-normalizer';
 import type { PlaylistItemSource } from 'playlist/source';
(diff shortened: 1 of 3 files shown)


src/js/providers/default.ts
```diff
@@ -224,7 +223,7 @@ export interface ImplementedProvider extends InternalProvider {
 export type AllProviderEvents = ProviderEvents & ProviderEventNotifications;
 export type AllProviderEventsListener = <E extends keyof AllProviderEvents>(type: E, data: AllProviderEvents[E]) => void;
 
-export type ProviderWithMixins = TracksMixin & VideoActionsInt & VideoAttachedInt & VideoListenerInt & ImplementedProvider & {
+export type ProviderWithMixins = TracksMixin & VideoActionsInt & VideoAttachedInt & ImplementedProvider & {
```
jnatalzia, line 226: makes sense 👍

PR #3716, opened 2020-07-14
[JW8-11240] Copy preload setting from updated playlist item

robwalch:
### This PR will...
Fix copying of "preload" config option from top level to playlist source object. 

Fix support for setting of the preload option in the config and have it applied to sources in the playlist:
```
{ 
    playlist: feedUrl,
    preload:"auto"
}
```
#### Addresses Issue(s):
JW8-11240

### Checklist

### DIFF
--- a/src/js/playlist/playlist.js
+++ b/src/js/playlist/playlist.js
@@ -31,7 +31,7 @@ export function normalizePlaylistItem(model, item, feedData) {
 
     playlistItem.preload = getPreload(item.preload, preload);
 
-    playlistItem.allSources = formatSources(item, model);
+    playlistItem.allSources = formatSources(playlistItem, model);
 
     playlistItem.sources = filterSources(playlistItem.allSources, providers);
 


src/js/playlist/playlist.js
```diff
@@ -31,7 +31,7 @@ export function normalizePlaylistItem(model, item, feedData) {
 
     playlistItem.preload = getPreload(item.preload, preload);
 
-    playlistItem.allSources = formatSources(item, model);
+    playlistItem.allSources = formatSources(playlistItem, model);
```
jnatalzia, line 34: with the `allSources = ` on the left side, this code does not read as something that would mutate the incoming source. I would rather patch the fact that we are mutating the object within `formatSources` rather than attempting to patch this by mutating the object we eventually use elsewhere in the stack. https://github.com/jwplayer/jwplayer/blob/master/src/js/playlist/playlist.js/#L85-L86

Feels like this could just lead to problems in the future

robwalch: Are we muting it? Or are we creating a list or new sources that is filtered below?

robwalch: Ah we are mutating the sources...

robwalch: This doesn't change that, it just passed in the correct item to copy `preload` from, to the sources.

PR #3723, opened 2020-07-24
[JW8-11311] Ignore seeking state when resetting currentTime to 0 while idle

robwalch:
### This PR will...
Ignore seeking state when resetting currentTime to 0 while idle

hls.js and shaka may perform seeks on level errors that should be ignored. Seek or pauses on the video element can change player state.

#### Addresses Issue(s):
JW8-11311

### Checklist


pajong:
Verified the fix locally

PR #3725, opened 2020-07-28
[JW8-11323] Ignore seeking state change while idle

robwalch:
### This PR will...
Ignore seeking state change while idle

So that the player state does not change to paused when starttime is set in the config.

#### Addresses Issue(s):
JW8-11323

### Checklist


esaari:
test this please

PR #3735, opened 2020-08-31
[JW8-11314] Keyboard shortcuts toggle improvements.

zetagame:
### This PR will...
- Improve accessibility functionality of keyboard shortcuts toggle switch
- Fix vertical centering issues with switch
- Fix strange outline on focus
- Improve type-checks on keyboard shortcuts toggle handlers

The functionality of the button was vague when using a screen-reader.

#### Addresses Issue(s):

JW8-11314

### Checklist

### DIFF
--- a/src/css/controls/imports/switch.less
+++ b/src/css/controls/imports/switch.less
@@ -1,7 +1,8 @@
 .jw-shortcuts-container {
     .jw-switch {
         position: relative;
-        display: inline-block;
+        display: flex;
+        align-items: center;
         transition: ease-out 0.15s;
         transition-property: opacity, background;
         border-radius: 18px;
@@ -12,14 +13,14 @@
         cursor: pointer;
         font-size: inherit;
         vertical-align: middle;
+        outline: none;
 
         &.jw-tab-focus {
-            outline: @ui-focus;
+            border: @ui-focus;
         }
 
         .jw-switch-knob {
             position: absolute;
-            top: 2px;
             left: 1px;
             transition: ease-out 0.15s;
             box-shadow: 0 0 10px rgba(0, 0, 0, 0.4);
@@ -29,21 +30,18 @@
             background: @rightclick-shortcuts-secondary-bg;
         }
 
-        &:before,
-        &:after {
+        .jw-switch-disabled,
+        .jw-switch-enabled {
             position: absolute;
-            top: 3px;
             transition: inherit;
             color: @rightclick-shortcuts-primary-text;
         }
 
-        &:before {
-            content: attr(data-jw-switch-disabled);
+        .jw-switch-disabled {
             right: 8px;
         }
 
-        &:after {
-            content: attr(data-jw-switch-enabled);
+        .jw-switch-enabled {
             left: 8px;
             opacity: 0;
         }
@@ -51,11 +49,11 @@
         &[aria-checked="true"] {
             background: #475470;
 
-            &:before {
+            .jw-switch-disabled {
                 opacity: 0;
             }
 
-            &:after {
+            .jw-switch-enabled {
                 opacity: 1;
             }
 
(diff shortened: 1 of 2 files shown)


src/js/view/controls/shortcuts-tooltip.ts
```diff
@@ -116,14 +116,18 @@ export default function (
     };
 
     const documentClickHandler = (e: Event) => {
-        const target = e.target as HTMLElement;
-        if (!/jw-shortcuts|jw-switch/.test(target.className)) {
+        const target: EventTarget | null = e.target;
+        if (target instanceof HTMLElement && !/jw-shortcuts|jw-switch/.test(target.className)) {
```
robwalch, line 120: Can we use `!== null` or simply `target` here?

src/js/view/controls/shortcuts-tooltip.ts
```diff
@@ -116,14 +116,18 @@ export default function (
...
             close();
         }
     };
 
     const toggleClickHandler = (e: Event) => {
-        const toggle = e.currentTarget as HTMLElement;
+        if (!(e.currentTarget instanceof HTMLElement)) {
```
robwalch, line 126: I'd really rather not use runtime `instanceof ` checks. Why the change?

src/js/view/controls/shortcuts-tooltip.ts
robwalch: If `Event` `target` and `currentTarget` are either an element or null, it's much cheaper to just test that it's not `null` or truthy. We don't have a dependency on `HTMLElement` as global, or minor performance impact of looking up the prototype chain.

src/js/view/controls/shortcuts-tooltip.ts
zetagame: as HTMLElement is basically saying "Ignore the check saying this isn't a typesafe operation, no matter what it's an HTMLElement". Since it's based on the target, we would need to check at runtime to know.

robwalch: given that we added the listener to an element, why do we need the check? We didn't need it before TypeScipt, do adding runtime checks to satisfy a build time library is a de-optimization.

src/js/view/controls/shortcuts-tooltip.ts
zetagame: `e.target` is an `EventTarget` when defined, which isn't necessarily an HTMLElement. While instanceof does incur about a 1/24000 seconds cost to check, this is comparable to checking if it has classname and ensures that this is a type-safe function. That being said, if we don't mind reading e.target twice there's a cleaner way to write it. See latest commit @robwalch .

src/js/view/controls/shortcuts-tooltip.ts
zetagame: The casting is preventing a compile time error.

robwalch: You could also check `e.currentTarget.nodeType === 1`, but either way it doesn't relate directly to what this method is meant to do which is to change the `aria-checked` attribute value on the element you attached the listener to which is `shortcutToggleUi`, which should also be the value of `this`.

zetagame: I'm worried about the edge cases, especially in the case of documentClickHandler. `e.currentTarget` might be an element at the time that it's clicked, but removed from the DOM by the time the handler fires. That being said, I do understand not wanting to add work beyond the ticket scope so I've reverted the type-check enhancements.

robwalch: You can set attributes on an element no longer in the DOM. If the point it to modify the element the listener is attached to, then just use `this` instead of `event.currentTarget`.

robwalch (approved):

PR #3736, opened 2020-09-02
[JW8-11126] Add Tizen Error State

mamaddox:
### This PR will...
 - Add a `tizen-error-container` for the tizen app's error state.
 - Extract the `error-container` styling into `error.less`

Tizen App
Have a question about the error state styling that I left a comment about.
Nah
#### Addresses Issue(s):

[JW8-11126](https://jwplayer.atlassian.net/browse/JW8-11126)

### Checklist


mamaddox:
Question about error state styling:
I noticed that there was specific styling in the `error` template that I extracted to `error.less`. I also noticed that there was error styling in `jwplayer/states.less` here:
https://github.com/jwplayer/jwplayer/blob/master/src/css/jwplayer/states.less#L20
The styling is slightly different so is there a reason we have multiple error state stylings? The main difference I noticed is in terms of the background color. The `error` template uses black, but the `state` uses a dark gray.
If we don't need multiple error state stylings, I can merge them together so that all that styling is in one place.

robwalch:
> The styling is slightly different so is there a reason we have multiple error state stylings? 

Hi @mamaddox,

There are two different error containers that the player can render - one pre setup and one post setup/playback. In the event that a setupError occurs before the core controller is loaded or the view is instantiated, jwplayer.js makes a paired down error container to display the setup error.

src/js/api/core-shim.js
```diff
@@ -293,7 +293,7 @@ function setupError(core, api, error) {
         const contextual = model.get('contextual');
         // Remove (and hide) the player if it failed to set up in contextual mode; otherwise, show the error view
         if (!contextual && !__HEADLESS__) {
-            const errorContainer = ErrorContainer(core, playerError);
+            const errorContainer = ErrorContainer(api, core, playerError);
```
robwalch, line 296: Why do we need both the `api` and `core` (controller)? Api is essentially a wrapper for the latter. Events triggered on core/controller are forwarded through the api.

mamaddox: I see. It looks like the `core` argument passed into ErrorContainer is treated as a `Model` in which case, I needed the `api` argument to trigger events for the tizen app controls. Happens in view as well
https://github.com/jwplayer/jwplayer/blob/26feb48b314e47172935d5f24f589b918c5b8f3f/src/js/view/view.js#L708

robwalch: Why is `TizenErrorContainer` a child or owned by ErrorContainer? Since it has more responsibility than the standard error ErrorContainer (handles keyboard input) wouldn't it be better it be better for it to use or replace ErrorContainer when needed?

src/js/view/error-container.ts
```diff
@@ -1,15 +1,28 @@
...
 import { createElement } from 'utils/dom';
 import { style } from 'utils/css';
+import { OS } from 'environment/environment';
+import TizenErrorContainer from 'view/controls/tizen/tizen-error-container';
 import type Model from 'controller/model';
 import type { PlayerError } from 'api/errors';
+import type { PlayerAPI } from 'types/generic.type';
 
-export default function ErrorContainer(model: Model, error: PlayerError): HTMLElement {
+require('css/error.less');
```
robwalch, line 10: why the additional css require here? Injecting CSS is not cheap (it triggers reflow and repaint). The less files are bundled together so that CSS can all be injected once, or twice as needed.

This should be included in either jwplayer.less or less that is imported into a tizen only loaded js module (require.ensure). I think if the tizen app is broken enough to encounter a setup error the standard error container might be fine, and you can deal with navigation in the app, rather than in the player.

src/js/view/error-container.ts
```diff
@@ -1,15 +1,28 @@
...
     const { message, code } = error;
     const html = errorContainerTemplate(model.get('id'), message, model.get('localization').errors.errorCode, code.toString());
+    const element = createElement(html);
+    style(element, {
+        background: 'black'
+    });
+
+    if (OS.tizen) {
+        return TizenErrorContainer(api, element, model.get('localization'));
+    }
```
robwalch, line 22: I don't like ErrorContainer being modified to return TizenErrorContainer adding the additional api dependency. Maybe the keyboard input could be completely separated from the UI?

The only reason the web player implements key input in controls is because it's only supposed to listen for that input when `controls: true` are enabled. If that's not the case for Tizen let's always have those inputs on the tizen player container.

src/js/view/controls/tizen/tizen-error-container.ts
```diff
@@ -0,0 +1,61 @@
...
+    const iconContainer = errorContainer.querySelector('.jw-icon');
+    style(iconContainer, {
+        width: '100px',
+        height: '100px'
+    });
+    if (iconContainer && errorIcon) {
+        iconContainer.appendChild(errorIcon);
+    }
+
+    const prevButton = button('jw-previous', () => api.trigger('backClick'), localization.prev);
```
robwalch, line 24: What is a "backClick" event and why do we need it coming from the player api?

robwalch (changes requested):

src/js/view/error-container.ts
robwalch: So the tizen-controls should require('tizen.less'). tizen.less should not be bundled in controls.less. Otherwise you're adding tizen styles to every player instance instead of only the ones running on tizen loading the tizen controls module.

mamaddox:
Got some great comments and am closing this review in favor of handling the error state in the app as opposed to in the player. That PR can be found here: https://github.com/jwplayer/jwplayer-tizen-app/pull/5.

PR #3740, opened 2020-09-09
JW8-11321: Update back button aria label contextually

nebnitt:
### This PR will...
This PR adds a method for updating the Aria Label of the back button in the captions menu based on what the previous menu viewed was. 

The aria label used to always be "Previous" (or whatever localized word(s) were chosen). Now, it will properly reflect the prior menu that was viewed, e.g., it will be "Previous - Closed Captions" when clicking into a sub-menu beyond the Closed Captions menu.

This PR improves the accessibility of these menus and therefore the player overall.
This is my first PR - please double-check all of it!
Yes - there's a PR in commercial for a new cucumber test. https://github.com/jwplayer/jwplayer-commercial/pull/7692
#### Addresses Issue(s):

https://jwplayer.atlassian.net/browse/JW8-11321

### Checklist

### DIFF
--- a/src/js/view/controls/components/menu/menu.js
+++ b/src/js/view/controls/components/menu/menu.js
@@ -22,6 +22,7 @@ export default class Menu extends Events {
         this.toggle = this.toggle.bind(this);
         this.name = _name;
         this.title = _title || _name;
+        this.localization = _localization;
         this.isSubmenu = !!_parentMenu;
         this.el = createElement(_template(this.isSubmenu, _name));
         this.buttonContainer = this.el.querySelector(`.jw-${this.name}-topbar-buttons`);
@@ -38,7 +39,7 @@ export default class Menu extends Events {
                 this.categoryButton = this.createCategoryButton();
             }
             if (this.parentMenu.parentMenu && !this.mainMenu.backButton) {
-                this.mainMenu.backButton = this.createBackButton(_localization);
+                this.mainMenu.backButton = this.createBackButton(this.localization);
             }
             this.itemsContainer = this.createItemsContainer();
             this.parentMenu.appendMenu(this);
@@ -122,6 +123,12 @@ export default class Menu extends Events {
         );
         prependChild(this.mainMenu.topbar.el, backButton.element());
         return backButton;
+    }
+    setBackButtonAriaLabel(labelText) {
+        const backButtonElement = this.mainMenu.backButton.element();
+        const localizedPrevious = this.localization.prev;
+        const backButtonText = labelText ? localizedPrevious + ' - ' + labelText : localizedPrevious;
+        backButtonElement.setAttribute('aria-label', backButtonText);
     }
     createTopbar() {
         const topbar = createElement(`<div class="jw-reset jw-submenu-topbar"></div>`);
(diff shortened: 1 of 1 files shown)


src/js/view/controls/components/menu/menu.js
```diff
@@ -329,12 +336,14 @@ export default class Menu extends Events {
         if (parentMenu.isSubmenu) {
             parentMenu.el.classList.remove('jw-settings-submenu-active');
             mainMenu.topbar.el.classList.add('jw-nested-menu-open');
-            const menuTitle = mainMenu.topbar.el.querySelector('.jw-settings-topbar-text');
-            menuTitle.setAttribute('name', this.title);
-            menuTitle.innerText = this.title;
+            const menuTitleElement = mainMenu.topbar.el.querySelector('.jw-settings-topbar-text');
```
nebnitt, line 339: I renamed this var to be more accurate.

src/js/view/controls/components/menu/menu.js
```diff
@@ -123,6 +124,12 @@ export default class Menu extends Events {
         prependChild(this.mainMenu.topbar.el, backButton.element());
         return backButton;
     }
+    setBackButtonAriaLabel(labelText) {
+        const backButtonElement = this.mainMenu.backButton.element();
+        const localizedPrevious = this.localization.prev;
+        const backButtonText = labelText ? localizedPrevious + ' - ' + labelText : localizedPrevious;
```
nebnitt, line 130: One question here -  the `labelText` I add is _not_ localized - but I'm not sure there's anything I can do about that, unless we also updated the corresponding localization options.

src/js/view/controls/components/menu/menu.js
```diff
@@ -22,6 +22,7 @@ export default class Menu extends Events {
         this.toggle = this.toggle.bind(this);
         this.name = _name;
         this.title = _title || _name;
+        this.localization = _localization;
```
zetagame, line 25: We should avoid storing this in a property if possible. It's a pretty large JSON file (all of our localized copy), and if it's stored as a property it won't be garbage collected at the end of construction.

src/js/view/controls/components/menu/menu.js
```diff
@@ -123,6 +124,12 @@ export default class Menu extends Events {
         prependChild(this.mainMenu.topbar.el, backButton.element());
         return backButton;
     }
+    setBackButtonAriaLabel(labelText) {
```
zetagame, line 127: Screen readers will always say Button at the end if the aria-role is button. We should instead set the inner text and the aria label to be `labelText` if the argument is passed in, with localization.prev as the fallback value.

src/js/view/controls/components/menu/menu.js
```diff
@@ -329,12 +336,14 @@ export default class Menu extends Events {
...
             parentMenu.el.classList.remove('jw-settings-submenu-active');
             mainMenu.topbar.el.classList.add('jw-nested-menu-open');
-            const menuTitle = mainMenu.topbar.el.querySelector('.jw-settings-topbar-text');
-            menuTitle.setAttribute('name', this.title);
-            menuTitle.innerText = this.title;
+            const menuTitleElement = mainMenu.topbar.el.querySelector('.jw-settings-topbar-text');
+            menuTitleElement.setAttribute('name', this.title);
+            menuTitleElement.innerText = this.title;
+            const parentMenuTitle = parentMenu.title;
+            this.setBackButtonAriaLabel(parentMenuTitle);
```
zetagame, line 343: We should also call this in `open` if it's a nested menu opening.

src/js/view/controls/components/menu/menu.js
nebnitt: Great catch. I'll update this to be just `_localization.prev` as that's all that this class ever needs to access.

src/js/view/controls/components/menu/menu.js
nebnitt: Great call. Testing with Voiceover now & will update accordingly.

nebnitt: Okay, now it says something like "Previous - Subtitle Settings, button" which seems right!

zetagame (approved):
lgtm!

src/js/view/controls/components/menu/menu.js
zetagame: menu.title is already localized, should be good :)

PR #3746, opened 2020-09-11
[JW8-11357] Extend id3 cue end to the start of the next id3 cue group

robwalch:
### This PR will...
Extend id3 cue end to the start of the next id3 cue group

This ensures that id3 metadata events are not skipped because the currentTime passed over a cue without intersecting with the cue start and end time range.

Spanning one ID3 group to the next matches the cue start and end ranges found in Safari when it creates the cues from the same HLS file. It also removes cue time gaps that could be seeked to where no event would fire, when one should. 

#### Addresses Issue(s):
JW8-11357

### Checklist

### DIFF
--- a/src/js/providers/tracks-mixin.ts
+++ b/src/js/providers/tracks-mixin.ts
@@ -14,6 +14,7 @@ type TrackCue = (VTTCue | DataCue | TextTrackCue) & {
 
 type TrackCueParsed = TrackCue & {
     _parsed?: boolean;
+    _extended?: boolean;
 }
 
 type TracksRecord = Record<string, TextTrackLike>;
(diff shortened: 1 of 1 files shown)


src/js/providers/tracks-mixin.ts
```diff
@@ -607,23 +608,48 @@ const Tracks: TracksMixin = {
...
+            if (!cue._extended && !!((cue as DataCue).data || cue.value)) {
                 if (cue.startTime !== startTime || cue.endTime === null) {
+                    previousStart = startTime;
                     startTime = cue.startTime;
-                    cueSets[++dataCueSetIndex] = [];
+                    const previousSet = dataCueSets[dataCueSetIndex];
+                    dataCueSets[++dataCueSetIndex] = [];
+                    parsedDataCueSets[dataCueSetIndex] = [];
+                    // increase id3 cue duration to a minimum of 0.25s up to next id3 cue start to ensure
+                    // "cuechange" event is fired and it is added to activeCues when currentTime intersects cue range
```
mamaddox, line 626: Is this comment more so for cueParsed logic below (Line 640)?

mamaddox (approved):

zetagame (approved):

pajong (approved):

src/js/providers/tracks-mixin.ts
robwalch: `previousCue.endTime = startTime` changes the cue duration. but that block work starts on the next line.

PR #3747, opened 2020-09-21
[JW8-11366] Prevent recursive overlap correction when an empty box is moved to line position

zetagame:
### This PR will...
Prevent recursive overlap correction when an empty box is moved to line position

When moveToLinePosition is called and the stylebox has no height, while loop never exists.

#### Addresses Issue(s):

JW8-11366

### Checklist

### DIFF
--- a/src/js/polyfills/webvtt.js
+++ b/src/js/polyfills/webvtt.js
@@ -694,7 +694,7 @@ function moveBoxToLinePosition(window, styleBox, containerBox, boxPositions, num
 
         // Prevent empty boxes from entering recursive overlap correction.
         if (!b.height || !b.width) {
-            return b;
+            return specifiedPosition;
         }
     
         for (let i = 0; i < axis.length; i++) {
--- a/src/js/polyfills/webvtt.js
+++ b/src/js/polyfills/webvtt.js
@@ -692,6 +692,11 @@ function moveBoxToLinePosition(window, styleBox, containerBox, boxPositions, num
         let percentage = 0; // Lowest possible so the first thing we get is better.
         let bestAxis;
 
+        // Prevent empty boxes from entering recursive overlap correction.
+        if (!b.height || !b.width) {
+            return b;
+        }
+    
         for (let i = 0; i < axis.length; i++) {
             while (b.overlapsOppositeAxis(containerBox, axis[i]) ||
             (b.within(containerBox) && b.overlapsAny(boxPositions))) {


src/js/polyfills/webvtt.js
```diff
@@ -692,6 +692,11 @@ function moveBoxToLinePosition(window, styleBox, containerBox, boxPositions, num
         let percentage = 0; // Lowest possible so the first thing we get is better.
         let bestAxis;
 
+        // Prevent empty boxes from entering recursive overlap correction.
+        if (!b.height || !b.width) {
```
mamaddox, line 696: Just curious, was the issue that because height was not defined, it would sometimes cause the `bottom` property to be not a number when set here and then that would mess with the while loop conditions?
https://github.com/jwplayer/jwplayer/blob/1b99becaacfb93e57fa79ea57c3c439eeb8756c4/src/js/polyfills/webvtt.js#L550

mamaddox (approved):

src/js/polyfills/webvtt.js
zetagame: That's a way of looking at it. The root cause is that some times we're passed a line of captions from the manifest with no text. As far as responsibility goes: HLS.js parses the manifest, our API updates our list of captions tracks, and captionsRenderer renders them, via the webvtt polyfill. 

This was a tricky decision because which solution to pick hinges on whether or not we want to allow caption tracks to have cues without text. I chose this implementation that way, regardless of what path forward we take on how to handle empty text cues, we'll never enter that infinite recursion scenario.

src/js/polyfills/webvtt.js
```diff
@@ -692,6 +692,11 @@ function moveBoxToLinePosition(window, styleBox, containerBox, boxPositions, num
...
+            return b;
```
robwalch, line 697: It may be better/safer to return `specifiedPosition` since prior to these changes this method always returns a new `BoxPosition` instance.

zetagame: If it's an empty div, why waste any processing? If we're running it 

There are pre-existing instances where we return `b`, ie:

```
if (b.within(containerBox)) {
  return b;
}
```

This works because b is already a `BoxPosition` instance at the time findBestPosition is called the first time.

image

robwalch: nothing is being wasted, it's already created on line 691.

robwalch (changes requested):
Output should not strictly match input, and using `specifiedPosition` ensures the return type is the expected type regardless of input.

robwalch (approved):

PR #3908, opened 2021-08-02
[JW8-11938] Sanitize title and description

zetagame:
### This PR will...
Sanitize title and description fields.

Patches an XSS vulnerability.

#### Addresses Issue(s):

JW8-11938

### Checklist

### DIFF
--- a/src/js/utils/dom.js
+++ b/src/js/utils/dom.js
@@ -282,7 +282,7 @@ export function deviceIsLandscape() {
 
 // Removes all unallowed HTML tags/attributes from a template string. Useful for user created HTML.
 export function HTMLSafeString(str) {
-    return str.replaceAll(/&|<|>|"|''/gm, function(i) {
+    return str.replace(/&|<|>|"|''/gm, function(i) {
         return '&#'+i.charCodeAt(0)+';';
-     }).replaceAll(/&#60;(\/?)(b|strong|i|em|p|br|ul|ol|li|h.)&#62;/gmi, '<$1$2>');
+     }).replace(/&#60;(\/?)(b|strong|i|em|p|br|ul|ol|li|h.)&#62;/gmi, '<$1$2>');
 }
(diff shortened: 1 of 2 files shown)


Dawolee (approved):

src/js/utils/dom.js
```diff
@@ -279,3 +279,10 @@ export function deviceIsLandscape() {
 
     return isLandscape || (window.orientation === 90 || window.orientation === -90);
 }
+
+// Removes all unallowed HTML tags/attributes from a template string. Useful for user created HTML.
+export function HTMLSafeString(str) {
+    return str.replaceAll(/&|<|>|"|''/gm, function(i) {
+        return '&#'+i.charCodeAt(0)+';';
+     }).replaceAll(/&#60;(\/?)(b|strong|i|em|p|br|ul|ol|li|h.)&#62;/gmi, '<$1$2>');
```
nebnitt, line 287: Curious where this regex came from and this overall func? Home-grown or perhaps pulled from SO or elsewhere?

src/js/view/title.ts
```diff
@@ -74,10 +75,10 @@ class Title {
             let title = '';
             let description = '';
 
-            if (item.title && model.get('displaytitle')) {
+            if (typeof item.title === 'string' && model.get('displaytitle')) {
```
nebnitt, line 78: Do we definitely need these `typeof` checks here? I'm not sure the implications of adding these and whether it'll result in `title` or `description` being blank in certain cases. Wondering if we can add the `typeof` checks into `HTMLSafeString`. Or maybe it's by design that we want to disallow any non-string values here?

src/js/utils/dom.js
zetagame: homegrown. Can break it down if you want

src/js/view/title.ts
zetagame: Yep. PlayListItem isn't strongly typed (it's imported from JS)

PR #3911, opened 2021-08-05
[JW8-11999] Adds playReason for autostart after control interactions during ad playback

Dawolee:
### This PR will...
Properly set the playReason as `autostart` (assuming player is set to autostart) after interactions with the player controls during ad playback.
After interacting with the controls during ad playback, the playReason was overwritten as `interaction`
#### Addresses Issue(s):

JW8-11999

### Checklist

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -576,6 +576,10 @@ Object.assign(Controller.prototype, {
             const adState = _getAdState();
 
             if (adState) {
+                // prevents the interaction with controls during ad playback to overwrite autostart playReason
+                if (_model.get('autostart') && adState === 'paused') {
+                    _model.set('playReason', 'autostart');
+                }
                 _this._instreamAdapter.off('state', _pauseAd, _this);
                 // this will resume the ad. _api.playAd would load a new ad
                 _api.pauseAd(false, meta);


maniland (approved):

src/js/controller/controller.js
```diff
@@ -576,6 +576,10 @@ Object.assign(Controller.prototype, {
             const adState = _getAdState();
 
             if (adState) {
+                // prevents the interaction with controls during ad playback to overwrite autostart playReason
+                if (_model.get('autostart') && adState === 'paused') {
+                    _model.set('playReason', 'autostart');
```
zetagame, line 581: What happens if I pause an ad and then click play again? Reason shouldn't be autostart if it's manually triggered.

zetagame (approved):

src/js/controller/controller.js
Dawolee: if the player is set to autostart: "true" in the initial setup, it will show "autostart" for playReason after you pause and play during the ad. If the player is set to autostart: "false", the play reason would show "interaction". I figured interacting with the ad playback should not overwrite the playReason for actual VOD playback but lmk if I'm mistaken!

zetagame: I believe you're mistaken. The reason should only be autostart for the initial play, as a play caused by what I'm describing above is triggered by an interaction.

zetagame: Find a way to pass `{ reason: 'autostart' }` when calling this function from the autostart scenario.

zetagame (approved):

PR #3921, opened 2021-09-15
[JW8-12029] Adding 'aria-expanded' attribute to the .jw-icon-settings button

Dawolee:
### This PR will...
Add 'aria-expanded' attribute to the `jw-icon-settings` button
It was not there before.
#### Addresses Issue(s):

JW8-12029

### Checklist

### DIFF
--- a/src/js/view/controls/components/menu/settings-menu.js
+++ b/src/js/view/controls/components/menu/settings-menu.js
@@ -304,12 +304,13 @@ class SettingsMenu extends Menu {
     }
 
     open(evt) {
-        const gearButton = this.controlbar.elements.settingsButton.element();
-        gearButton.setAttribute('aria-expanded', true);
-
         if (this.visible) {
             return;
         }
+
+        const gearButton = this.controlbar.elements.settingsButton.element();
+        gearButton.setAttribute('aria-expanded', true);
+        
         this.el.parentNode.classList.add('jw-settings-open');
         this.trigger('visibility', { visible: true, evt });
         document.addEventListener('click', this.onDocumentClick);
--- a/src/js/view/controls/components/menu/settings-menu.js
+++ b/src/js/view/controls/components/menu/settings-menu.js
@@ -304,6 +304,9 @@ class SettingsMenu extends Menu {
     }
 
     open(evt) {
+        const gearButton = this.controlbar.elements.settingsButton.element();
+        gearButton.setAttribute('aria-expanded', true);
+
         if (this.visible) {
             return;
         }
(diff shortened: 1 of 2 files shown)


src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -304,6 +304,9 @@ class SettingsMenu extends Menu {
     }
 
     open(evt) {
+        const gearButton = this.controlbar.elements.settingsButton.element();
+        gearButton.setAttribute('aria-expanded', true);
```
imbcmdth, line 308: Should this be below the `if (this.visible)`?

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -304,6 +304,9 @@ class SettingsMenu extends Menu {
...
+
         if (this.visible) {
             return;
         }
```
imbcmdth, line 312: There is another `aria-exanded` attribute set on line 316 - should that remain?

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -326,6 +329,8 @@ class SettingsMenu extends Menu {
         // If closed by keypress, focus appropriate element.
```
imbcmdth, line 329: If the `aria-exanded` SHOULD stay on line 316, then should it be removed in this close?

src/js/view/controls/components/menu/settings-menu.js
Dawolee: yep, I just console logged the element and it refers to the `jw-settings-menu` element rather than the icon.

src/js/view/controls/components/menu/settings-menu.js
Dawolee: hmm yeah that does seem like a better spot! Just tested it and it works fine. Will make the suggested change

src/js/view/controls/components/menu/settings-menu.js
Dawolee: whoops did not see this the first time around, but yes! Sorry for the oversight. I did realize that the settings menu `aria-expanded` was not being set to false when collapsed so I will push that up too!

Dawolee:
test this please

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -326,6 +333,11 @@ class SettingsMenu extends Menu {
...
         const key = normalizeKey(evt && evt.sourceEvent && evt.sourceEvent.key);
         const gearButton = this.controlbar.elements.settingsButton.element();
+
```
zetagame, line 336: micro-optimization/nit: move this block to line 325 so we can keep the DOM changes together.

Dawolee:
test this please

Dawolee:
test this please

PR #3922, opened 2021-09-16
Improvement: [JW8-11087] Under specific scenarios enable native html5 element looping

imbcmdth:
### This PR will...
Set the `loop` attribute to `true` when:
1. The configuration has `repeat: true` 
2. There is only one item in the playlist
3. The item has a source that is a video of type `mp4`

Folks want to be able to use a short video that repeats and for the looping to be seamless.

No

No

#### Addresses Issue(s):

JW8-11087

### Checklist

### DIFF
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -39,8 +39,6 @@ export const Defaults = {
     playbackRates: [0.5, 1, 1.25, 1.5, 2],
     renderCaptionsNatively: false,
     repeat: false,
-    // loop is only used on native html5 providers to loop mp4
-    loop: false,
     stretching: 'uniform',
     volume: 90,
     width: 640
--- a/src/js/api/config.js
+++ b/src/js/api/config.js
@@ -39,6 +39,8 @@ export const Defaults = {
     playbackRates: [0.5, 1, 1.25, 1.5, 2],
     renderCaptionsNatively: false,
     repeat: false,
+    // loop is only used on native html5 providers to loop mp4
+    loop: false,
     stretching: 'uniform',
     volume: 90,
     width: 640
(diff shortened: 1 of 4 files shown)


src/js/api/config.js
```diff
@@ -39,6 +39,8 @@ export const Defaults = {
     playbackRates: [0.5, 1, 1.25, 1.5, 2],
     renderCaptionsNatively: false,
     repeat: false,
+    // loop is only used on native html5 providers to loop mp4
+    loop: false,
```
zetagame, line 43: config is specific to anything a user can configure. You can store this attribute on the model without extending config.

src/js/controller/controller.js
```diff
@@ -1310,6 +1310,16 @@ Object.assign(Controller.prototype, {
         } catch (error) {
             return Promise.reject(error);
         }
+
+        // Under very special conditions, we enable the html5 looping attribute
+        if (playlist.length === 1
```
zetagame, line 1315: Is it possible to do this in the HTML5 provider?

imbcmdth: I was trying to keep the provider out of worrying about the playlist level info - to keep responsibilities separate. I know the model is available pretty much everywhere but it feels wrong. I'll try it and see what it's like.

zetagame: The controller is pretty high level for something specific to mp4 sources, especially given the additional constraints (it's essentially for API-level functionality). HTML5 provider's setPlaylistItem function might be a good place to put it. When you say the using the model feels wrong, what do you mean?

esaari:
test this please

zetagame (approved):

PR #3931, opened 2021-10-06
[JW8-12067] Fix error message when display is clicked with controls disabled

Dawolee:
### This PR will...
Fix error message when display is clicked with controls disabled
The displayClick event listener throws a JS error when the player controls are set to false. This does not happen if the controls are on. 
#### Addresses Issue(s):

JW8-12067

### Checklist


imbcmdth (approved):
But I'd like @zetagame to make sure this makes sense since it is partially undoing his previous change.

PR #3773, opened 2020-11-08
[JW8-11109] Fix bug that shows first frame of video before preroll

pajong:
### This PR will...
- Clear video tag source before calling `videotag.load()` to force prime before ad playback

### Why is this Pull Request needed?
- Calling `videotag.load()` caused the first frame of the video content to show, even when we have already started blocking the content by calling `jwplayer().createInstream().init()`

### Are there any points in the code the reviewer needs to double check?
- This bug happens for all ads plugins
- This bug was introduced 2.5 years ago, with this change: https://github.com/jwplayer/jwplayer/commit/d0e768d755b88e07e0be23b9d4c097f33681ff59#diff-15fcddfc5fb5f2333b37e0588e65df209035661bb83f68a5e4d0637c2b260e92R442
- Wasn't sure if there was a better way to force prime the video tag than to call `load()` - if there is, that would be a better solution

#### Addresses Issue(s):
JW8-11109

### Checklist
- [ ] Jenkins builds and unit tests are passing
- [ ] I have reviewed the automated results

PR #3941, opened 2021-11-09
[JW8-11891] Prevent preroll from being interrupted (iOS) when nextup display is clicked

Dawolee:
### This PR will...
Prevent preroll from being interrupted (iOS) when nextup display is clicked
The preroll ad would show a buffer icon before erroring on iOS Safari when the nextup display was clicked.
#### Addresses Issue(s):

JW8-11891

### Checklist

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -622,7 +622,7 @@ Object.assign(Controller.prototype, {
                     if (inInteraction() && !_backgroundLoading) {
                         const video = _model.get('mediaElement');
                         if (_this._instreamAdapter) {
-                            video.src = '';
+                            video.preload = 'none';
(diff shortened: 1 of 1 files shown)


Dawolee:
test this please

Dawolee:
So Jong's code was addressing this bug https://github.com/jwplayer/jwplayer/pull/3773. Seems like setting the `video.src` to an empty string was throwing off the state handling in the player causing the ad interrupt -> error issue here. Jon mentioned what if we set the `preload` to none and this fix actually addresses the current bug and I cannot reproduce Jong's bug either!

zetagame (approved):

jwbrandon (approved):

PR #3947, opened 2021-12-02
chore: update typescript and fix type definitions

jwbrandon:
update typescript and our type definitions


src/js/providers/tracks-mixin.ts
```diff
@@ -8,6 +8,8 @@ import type { TextTrackLike } from 'types/generic.type';
 import type { ProviderEvents, ProviderWithMixins } from './default';
 import type { PlaylistItemTrack } from 'playlist/track';
 
+type DataCue = VTTCue & {data: ArrayBufferLike}
```
jwbrandon, line 11: Wasn't too sure about this type definition, but I think this is correct.

Dawolee (approved):
not as familiar with TS (will learn more on the side) but LGTM!

PR #3959, opened 2022-01-07
[JW8-12177] Prevent playlistComplete from firing when related.autoplay is configured

Dawolee:
### This PR will...
Prevent playlistComplete from firing when related.autoplay is configured
playlistComplete was firing for each playlist item when it should only fire when the whole playlist has played.

For my fix, I added a conditional around when to trigger `PLAYLIST_COMPLETE` in the `_completeHandler` function. I am checking to see that there is no related playlist OR if the current playlist item is the last one before firing the `PLAYLIST_COMPLETE` event.
#### Addresses Issue(s):

JW8-12177

### Checklist

### DIFF
--- a/src/js/controller/controller.js
+++ b/src/js/controller/controller.js
@@ -859,7 +859,9 @@ Object.assign(Controller.prototype, {
                     // This prevents media from replaying when a completed video scrolls into view
                     _model.set('playOnViewable', false);
                     _model.set('state', STATE_COMPLETE);
-                    _this.trigger(PLAYLIST_COMPLETE, {});
+                    if (!_model.get('related') || _model.get('item') === _model.get('playlist').length - 1) {
+                        _this.trigger(PLAYLIST_COMPLETE, {});
+                    }
                 }
                 return;
             }


Dawolee:
test this please

src/js/controller/controller.js
```diff
@@ -859,7 +859,9 @@ Object.assign(Controller.prototype, {
                     // This prevents media from replaying when a completed video scrolls into view
                     _model.set('playOnViewable', false);
                     _model.set('state', STATE_COMPLETE);
-                    _this.trigger(PLAYLIST_COMPLETE, {});
+                    if (!_model.get('related') || _model.get('item') === _model.get('playlist').length - 1) {
```
zetagame, line 862: What if it's a regular playlist that hasn't completed and it's there's no recs (no related)?

Dawolee: If it's a regular playlist, the `complete` event fires but the `playlistComplete` will not fire until the regular playlist finishes its last item. I can see this behavior occurring locally rn with my changes.

PR #3960, opened 2022-01-11
[JW8-12112] <PiP button on Firefox is preventing the next up display to work properly>

jdunningjwp:
### This PR will...
Add opacity to the next up button when using firefox to prevent the PiP button getting clicked through the next up tooltip. 
The PiP button being clicked instead of the next up button ruins the functionality of the next up button.
Nope
Nope
#### Addresses Issue(s):

https://jwplayer.atlassian.net/browse/JW8-12112

### Checklist

### DIFF
--- a/src/css/controls/imports/nextup.less
+++ b/src/css/controls/imports/nextup.less
@@ -165,5 +165,5 @@
 }
 
 .jw-nextup-firefox-pip-fix {
-    background: rgba(38, 38, 38, 1);
+    background: fade(@overlay-background-color, 100%);
 }
--- a/src/css/controls/imports/nextup.less
+++ b/src/css/controls/imports/nextup.less
@@ -163,3 +163,7 @@
         visibility: visible;
     }
 }
+
+.jw-nextup-firefox-pip-fix {
+    background: rgba(38, 38, 38, 1);
+}
--- a/src/js/view/controls/templates/nextup.ts
+++ b/src/js/view/controls/templates/nextup.ts
@@ -1,9 +1,12 @@
 import type { HTMLTemplateString } from 'types/generic.type';
+import { Browser } from 'environment/environment';
 
 export default (header = '', title = '', duration = '', closeAriaLabel = ''): HTMLTemplateString => {
+    let firefoxPipFix = Browser.firefox ? ' jw-nextup-firefox-pip-fix' : '';
+
     return (
-        `<div class="jw-nextup jw-background-color jw-reset">` +
-             `<div class="jw-nextup-tooltip jw-reset">` +
+        `<div class="jw-nextup jw-background-color jw-reset${firefoxPipFix}">` +
+                `<div class="jw-nextup-tooltip jw-reset">` +
                 `<div class="jw-nextup-thumbnail jw-reset"></div>` +
                 `<div class="jw-nextup-body jw-reset">` +
                     `<div class="jw-nextup-header jw-reset">${header}</div>` +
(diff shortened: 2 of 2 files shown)


src/js/view/controls/templates/nextup.ts
```diff
@@ -1,9 +1,12 @@
 import type { HTMLTemplateString } from 'types/generic.type';
+import { Browser } from 'environment/environment';
 
 export default (header = '', title = '', duration = '', closeAriaLabel = ''): HTMLTemplateString => {
+    let firefoxPipFix = Browser.firefox ? ' jw-nextup-firefox-pip-fix' : '';
```
jwbrandon, line 5: Would be good to have a comment in the code explaining why we have this special case for firefox.

jwbrandon, line 5: Might as well use const here since we don't redefine this variable.
```suggestion
    const firefoxPipFix = Browser.firefox ? ' jw-nextup-firefox-pip-fix' : '';
```

src/css/controls/imports/nextup.less
```diff
@@ -163,3 +163,7 @@
         visibility: visible;
     }
 }
+
+.jw-nextup-firefox-pip-fix {
+    background: rgba(38, 38, 38, 1);
```
cesar-rivera, line 168: Wouldn't it be better to use a method like [fade](https://lesscss.org/functions/#color-operations-fade) to only set the opacity and use the `@overlay-background-color` variable instead of using a static value?
I'm not sure if this would affect or override custom themes but that might also be a risk

jdunningjwp: I can change it to
`.jw-nextup-firefox-pip-fix {
    background: fade(@overlay-background-color, 100%);
}`
and it seems to work the same for firefox and chrome

jwbrandon (approved):

Dawolee (approved):

jwbrandon:
test this please

src/css/controls/imports/nextup.less
```diff
@@ -163,3 +163,7 @@
         visibility: visible;
     }
 }
+
+.jw-nextup-firefox-pip-fix {
+    background: fade(@overlay-background-color, 100%);
+}
```
zetagame, line 169: Missing empty line

src/css/controls/imports/nextup.less
```diff
@@ -163,3 +163,7 @@
         visibility: visible;
     }
 }
+
+.jw-nextup-firefox-pip-fix {
+    background: fade(@overlay-background-color, 100%);
```
zetagame, line 168: No need to use fade here. It's a static number. LESS functions are useful for mixins and variables but for a static value it's overkill. Feel free to leave it in as this is more of a nit, but there's no advantage to using fade here.

src/js/view/controls/templates/nextup.ts
```diff
@@ -1,9 +1,13 @@
 import type { HTMLTemplateString } from 'types/generic.type';
+import { Browser } from 'environment/environment';
 
 export default (header = '', title = '', duration = '', closeAriaLabel = ''): HTMLTemplateString => {
+    // When using firefox this adds opacity to stop the firefox PiP button getting clicked instead of the next up div
+    const firefoxPipFix = Browser.firefox ? ' jw-nextup-firefox-pip-fix' : '';
```
zetagame, line 6: You can move this logic to line 32 of `nextuptooltip.js` to avoid adding logic to the template itself to keep things more SRP. It's a nit, so don't consider this a required change.

PR #3964, opened 2022-01-18
[JW-11320]<Teaching Channel 9: No confirmation message is provided it on selecting captions styles 'Reset' button>

jdunningjwp:
### This PR will...
Add confirmation message "Reset Captions Successful" when the reset button in the captions menu is pressed
If the screen reader does not notify anything after selecting "Reset" button non-sighted users will not be able to confirm or notice that the information has been reset
Nope
Nope
#### Addresses Issue(s):

https://jwplayer.atlassian.net/browse/JW8-11320

### Checklist

### DIFF
--- a/src/js/view/controls/components/menu/menu-item.js
+++ b/src/js/view/controls/components/menu/menu-item.js
@@ -45,6 +45,5 @@ export class ResetMenuItem extends MenuItem {
         this.statusEl.setAttribute('role', 'status');
         this.el.appendChild(this.statusEl);
         this.el.setAttribute('aria-controls', 'reset-status-message');
-        console.trace(this.statusEl);
     }
 }
--- a/src/js/view/controls/components/menu/menu-item.js
+++ b/src/js/view/controls/components/menu/menu-item.js
@@ -35,3 +35,16 @@ export class ButtonMenuItem extends MenuItem {
         this.active = false;
     }
 }
+
+export class ResetMenuItem extends MenuItem {
+    constructor(_content, _action, _template = itemButtonTemplate) {
+        super(_content, _action, _template);
+        this.statusEl = document.createElement('span');
+        this.statusEl.id = 'reset-status-message';
+        this.statusEl.style = 'font-size: 0;';
+        this.statusEl.setAttribute('role', 'status');
+        this.el.appendChild(this.statusEl);
+        this.el.setAttribute('aria-controls', 'reset-status-message');
+        console.trace(this.statusEl);
+    }
+}
--- a/src/js/view/controls/components/menu/settings-menu.js
+++ b/src/js/view/controls/components/menu/settings-menu.js
@@ -149,12 +149,23 @@ class SettingsMenu extends Menu {
             this.model.set('captions', newStyles);
         };
         const persistedOptions = model.get('captions');
+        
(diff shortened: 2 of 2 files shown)


src/js/view/controls/components/menu/menu-item.js
```diff
@@ -35,3 +35,15 @@ export class ButtonMenuItem extends MenuItem {
         this.active = false;
     }
 }
+
+export class ResetMenuItem extends MenuItem {
+    constructor(_content, _action, _template = itemButtonTemplate) {
+        super(_content, _action, _template);
+        this.statusEl = document.createElement('span');
+        this.statusEl.id = 'reset-status-message';
```
zetagame, line 43: jw prefix

src/js/view/controls/components/menu/menu-item.js
```diff
@@ -35,3 +35,15 @@ export class ButtonMenuItem extends MenuItem {
...
+        this.statusEl.style = 'font-size: 0;';
```
zetagame, line 44: Avoid using css in js whenever possible

src/js/view/controls/components/menu/menu-item.js
```diff
@@ -35,3 +35,15 @@ export class ButtonMenuItem extends MenuItem {
...
+        this.statusEl.setAttribute('role', 'status');
+        this.el.appendChild(this.statusEl);
+        this.el.setAttribute('aria-controls', 'reset-status-message');
```
zetagame, line 47: This isn't what aria-controls is for, see examples section here: https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-controls

zetagame:
Might want to re-write as an aria-live region that gets passed to anything that needs this sort of alert. See here: https://cccaccessibility.org/web-1/web-developer-tutorials/using-aria-live

zetagame (changes requested):
See comments above.

jwbrandon:
test this please

src/js/view/controls/components/menu/menu-item.js
```diff
@@ -35,3 +35,13 @@ export class ButtonMenuItem extends MenuItem {
         this.active = false;
     }
 }
+
+export class ResetMenuItem extends MenuItem {
```
jwbrandon, line 39: Should this be in its own file? Something like `reset-menu-item`?

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -188,9 +188,15 @@ class SettingsMenu extends Menu {
             });
             captionsSettingsItems.push(resetItem);
             captionsSettingsMenu.setMenuItems(captionsSettingsItems);
+            if (isReset) {
+                resetItem.statusEl.innerHTML = 'Reset Captions Successful';
```
jwbrandon, line 192: Might be better to use [`textContent`](https://developer.mozilla.org/en-US/docs/Web/API/Node/textContent#differences_from_innertext) here and below rather than `innerHTML`

jwbrandon (commented):
This could use a unit test, at least for the ResetMenuItem class. It does seem like it might be very difficult to unit test the actual fix here, but it might still be worth doing.

PR #3966, opened 2022-01-20
[JW8-11322]<Teaching Channel 11: Screen reader does not automatically notify about the changed state after exiting the full screen mode>

jdunningjwp:
### This PR will...
Add a confirmation message "Exit Fullscreen Successful" when you are in fullscreen and press the exit fullscreen button
If the screen reader fails to automatically notify about the changed state after selecting "Exit Fullscreen" button, non-sighted users will not be able to figure out whether any action has been triggered or not
Nope
Nope
#### Addresses Issue(s):

https://jwplayer.atlassian.net/browse/JW8-11322

### Checklist


src/js/view/controls/controlbar.js
```diff
@@ -305,6 +306,15 @@ export default class Controlbar {
         _model.change('fullscreen', (model, val) => {
             const fullscreenElements = [this.elements.fullscreen.element(), this.elements.imaFullscreen.element()];
 
+            if (fullscreenElements[0].getAttribute('aria-label') === 'Exit Fullscreen') {
```
zetagame, line 309: This text localized, so you want to check `val` rather than the label.

Dawolee:
Tried testing this with Mac OS VoiceOver and I'm seeing the message being displayed properly (both text and audio)!

Dawolee (approved):

src/js/view/controls/controlbar.js
```diff
@@ -305,6 +306,15 @@ export default class Controlbar {
...
+                // Uses setTimeout to prevent 'Exit Fullscreen Successful' voice line from being cutoff by other voice lines when using a screen reader
+                setTimeout(() => {
+                    this._exitFullscreenAnnouncer.textContent = 'Exit Fullscreen Successful';
+                }, 1450);
```
jwbrandon, line 313: Do we only need this because we use `aria-live=assertive`? If we were to switch to `aria-live=polite` would that solve the issue of needing a setTimeout? To me this setTimeout is going a bit into the magic number category. Not saying that isn't the right solution, but we will want to make sure that we try other options first. If we do keep the setTimeout around:
1. We need to make sure that we store it on `this` and clear it if we get back into this function faster than the timeout can run. 
2. We will also have to call clearTimeout on it in the `destroy` method, so that it doesn't happen after this object is destroyed and cause issues.

jwbrandon (commented):
I tested this and it does work, but i'm not sure if a timeout is the best way to active what we want here.

src/js/view/controls/controlbar.js
jdunningjwp: I tried setting `aria-live=polite` but the voice line "Exit Fullscreen Successful" gets cut off by another voice line before it finishes. The `setTimeout` is very much a magic number as that is the only way I could get a clear voice line without it getting cut off. I will try and refactor to store it on `this` and add a `clearTimeout` in the `destroy` method.

src/js/view/controls/controlbar.js
```diff
@@ -305,6 +306,15 @@ export default class Controlbar {
...
+            } else if (fullscreenElements[0].getAttribute('aria-label') === 'Fullscreen') {
+                this._exitFullscreenAnnouncer.textContent = '';
```
zetagame, line 315: We should probably add an Enter fullscreen text when it's entered too. Also, is there any reference to what the element is when VoiceOver reads the exit success message? If not we should specify in the message, i.e.: "Player is now fullscreen", "Player is no longer fullscreen"

src/js/templates/player.ts
```diff
@@ -17,6 +17,7 @@ export default (id: string, ariaLabel: string): HTMLTemplateString => {
                 `<div class="jw-hidden-accessibility">` +
                     `<span class="jw-time-update" aria-live="assertive"></span>` +
                     `<span class="jw-volume-update" aria-live="assertive"></span>` +
+                    '<span class="jw-exit-fullscreen-confirmation" aria-live="assertive"></span>' +
```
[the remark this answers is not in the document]
jdunningjwp: On Mac using VoiceOver the "Exit Fullscreen Successful" voice line gets cut off by a line that either says "Space with chrome containing window" on chrome or "Now in JW test harness window" on firefox. I have tried "assertive" and "polite" but it is the same either way.

PR #3969, opened 2022-01-24
chore: switch from grunt to npm scripts

jwbrandon:
### This PR will...
Remove the grunt task runner in favor of using npm.
#### Addresses Issue(s):

JW8-###

### Checklist


karma.conf.js
```diff
@@ -53,12 +65,7 @@ module.exports = function(config) {
     let browsers = [
         'ChromeHeadless',
         'Chrome',
-        'Safari',
```
jwbrandon, line 56: We need to schedule some time internally to get tests running correctly on all the browsers that we want.

zetagame: What's going on with the Safari tests?

package.json
```diff
@@ -79,20 +74,29 @@
...
-    "lint": "npm run type-check && npm run lint:js && npm run lint:tests && npm run lint:styles",
-    "lint:js": "eslint './src/js' --ext .ts --ext .js",
+    "clean": "shx rm -rf ./bin-* && shx mkdir -p ./bin-release ./bin-debug",
+    "build": "npm-run-all -s clean -p build:*",
+    "build:js": "webpack --config webpack.config.js",
+    "build:css": "npm-run-all build:css:less build:css:postcss",
+    "build:css:less": "node scripts/build-css.js",
+    "build:css:postcss": "postcss bin-debug/css/*.css --use autoprefixer --replace",
+    "build:notice": "node scripts/build-notice.js",
+    "start": "npm-run-all -p watch serve",
```
Dawolee, line 84: I tried most of these scripts but I am running into issues with the `start` and `serve` commands:
npm run serve
npm run start

I've tried deleting my node_modules and package-lock.json before running `npm install` but i'm still seeing this behavior on my end.

karma.conf.js
jwbrandon: Since we were always running through grunt safari was never run so our unit tests currently fail for various reasons. Not sure that it's anything to be concerned about, just something we want to rectify eventually.

package.json
jwbrandon: sorry about that I ended up installing `static-server` rather than `serve-static` which is why you get this issue. I am working on a fix right now.

jwbrandon: @Dawolee This should be fixed now. Not entirely sure why we even had a server though there is only an index.html at `localhost:3000/test/manual`

jwbrandon:
test this please

package.json
Dawolee: Okay! Can confirm both commands work now. Not sure about the functionality about the `serve`/`start` here vs in commercial though hmm..

Dawolee (approved):
Otherwise LGTM!

jcholyhead:
I think you might have forgotten the README when you made this change.

jwbrandon:
Good call, I will work on something for that.

PR #3973, opened 2022-02-02
[JW8-12213] Fixing autopause + floating dismiss behavior when ad is not playing

Dawolee:
### This PR will...
Fixing unwanted behavior introduced by autoPause + floating player dismissal. I am now setting a new state on the model `isAdPlaying` which is being toggled by the `instream-adapter` when ad playback starts and ends. I am now checking for `model.get('isAdPlaying')` in the autoPause + floating logic to only prevent the pause when an ad is currently playing.
Playback was not being paused after dismissing floating player. 
#### Addresses Issue(s):

JW8-12213

### Checklist

### DIFF
--- a/src/js/controller/instream-adapter.js
+++ b/src/js/controller/instream-adapter.js
@@ -215,6 +215,7 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
         const { newstate } = event;
         if (newstate === STATE_PLAYING) {
             _controller.trigger(AD_PLAY, event);
+            _model.set('isAdPlaying', true);
         } else if (newstate === STATE_PAUSED) {
             _controller.trigger(AD_PAUSE, event);
         }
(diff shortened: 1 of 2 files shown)


src/js/controller/instream-adapter.js
```diff
@@ -275,6 +276,7 @@ const InstreamAdapter = function(_controller, _model, _view, _mediaPool) {
             data.tag = _options.tag;
         }
         this.trigger(MEDIA_COMPLETE, data);
+        _model.set('isAdPlaying', false);
```
zetagame, line 279: Don't forget outstream adapter. I really think you should be setting this off of api.on('adBreakStart'), and api.on('adComplete') instead to ensure we aren't missing something. Every ad plugin dispatches these events to the player.

src/js/view/view.js
```diff
@@ -548,7 +548,7 @@ function View(_api, _model) {
 
         controls.on('dismissFloating', () => {
             this.stopFloating(true);
-            if (_model.get('autoPause') && !_model.get('autoPause').pauseAds) {
+            if (_model.get('autoPause') && !_model.get('autoPause').pauseAds && _model.get('isAdPlaying')) {
```
zetagame, line 551: _model.get('instream')

zetagame (approved):

jwbrandon (approved):

PR #3976, opened 2022-02-07
chore: sinon update requirements

jwbrandon:
### This PR will...
update sinon and fix some issues with the mocking we do right now
#### Addresses Issue(s):

JW8-###

### Checklist


src/js/utils/browser.ts
```diff
@@ -50,13 +50,18 @@ export const isAndroid = () => userAgentMatch(/Android/i) && !userAgentMatch(/Wi
 
 export const isMobile = () => isIOS() || isAndroid() || userAgentMatch(/Windows Phone/i);
 
-export function isIframe(): boolean {
+export const isIframe = function(): boolean {
+    if (typeof isIframe.mock_ === 'boolean') {
```
jwbrandon, line 54: Wasn't sure of a better way to do this. Ideally we would have a webpack plugin for the test build that allows us to stub functions and return values, but I couldn't find anything.

PR #3979, opened 2022-02-08
[JW8-12182] WebPlayer - When opening any menu inside player, stream is not paused

jdunningjwp:
### This PR will...
add the functionality of pausing the player when any menu is opened when using breakpoint 0 or 1. Closing the menu will continue the player at the same point before opening the menu.  If the player is paused before opening a menu the player will be paused after closing the menu.
This PR is needed because the player should pause while on any menu, and continue playing after closing any menu. On breakpoint 0 or 1 the open menu covers the player fully, so if a user can't see the player, it shouldn't be playing.
Nope
Nope
#### Addresses Issue(s):

https://jwplayer.atlassian.net/browse/JW8-12182

### Checklist

### DIFF
--- a/src/js/view/controls/components/menu/settings-menu.js
+++ b/src/js/view/controls/components/menu/settings-menu.js
@@ -314,6 +314,12 @@ class SettingsMenu extends Menu {
             gearButton.setAttribute('aria-expanded', true);
         }
 
+        if (document.querySelector('.jw-breakpoint-0') || document.querySelector('.jw-breakpoint-1')) {
+            const mediaState = this.model._mediaModel.attributes.mediaState;
+            this.mediaStateWhenOpened = mediaState;
+            this.api.pause()
+        }
+
         this.el.parentNode.classList.add('jw-settings-open');
         this.trigger('visibility', { visible: true, evt });
         document.addEventListener('click', this.onDocumentClick);
@@ -329,6 +335,12 @@ class SettingsMenu extends Menu {
             gearButton.setAttribute('aria-expanded', false);
         }
         
+        if (document.querySelector('.jw-breakpoint-0') || document.querySelector('.jw-breakpoint-1')) {
+            if (this.mediaStateWhenOpened === 'playing') {
+                this.api.play();
+            }
+        }
+
         this.el.setAttribute('aria-expanded', 'false');
         this.el.parentNode.classList.remove('jw-settings-open');
         
--- a/src/js/view/controls/components/menu/settings-menu.js
+++ b/src/js/view/controls/components/menu/settings-menu.js
@@ -7,6 +7,7 @@ import button from 'view/controls/components/button';
 import { cloneIcon } from 'view/controls/icons';
 import { normalizeKey, destroyMenu, selectMenuItem } from './utils';
 import UI from 'utils/ui';
+import { getBreakpoint } from 'view/utils/breakpoint';
 import { 
     nextSibling, 
     previousSibling,
@@ -314,8 +315,8 @@ class SettingsMenu extends Menu {
             gearButton.setAttribute('aria-expanded', true);
         }
 
-        if (document.querySelector('.jw-breakpoint-0') || document.querySelector('.jw-breakpoint-1')) {
-            const mediaState = this.model._mediaModel.attributes.mediaState;
+        if (getBreakpoint(this.model.get('containerWidth')) === 0 || getBreakpoint(this.model.get('containerWidth')) === 1) {
+            const mediaState = this.model.get('state')
             this.mediaStateWhenOpened = mediaState;
             this.api.pause()
         }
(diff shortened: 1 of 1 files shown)


src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -314,6 +314,12 @@ class SettingsMenu extends Menu {
             gearButton.setAttribute('aria-expanded', true);
         }
 
+        if (document.querySelector('.jw-breakpoint-0') || document.querySelector('.jw-breakpoint-1')) {
+            const mediaState = this.model._mediaModel.attributes.mediaState;
+            this.mediaStateWhenOpened = mediaState;
```
jwbrandon, line 319: I think it might be good to add an `_` to the start of this property as we only want to use this internally.

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -314,6 +314,12 @@ class SettingsMenu extends Menu {
             gearButton.setAttribute('aria-expanded', true);
         }
 
+        if (document.querySelector('.jw-breakpoint-0') || document.querySelector('.jw-breakpoint-1')) {
```
jwbrandon, line 317: we should probably use the query selector for the player element here as it's possible to have multiple players on a page. I'm not sure how exactly you get the player element, i will take a look when I get a chance.

zetagame, line 317: There can be multiple players on a page, instead use `getBreakpoint(model.get('containerWidth')`

It's a util in breakpoint.ts

src/js/view/controls/components/menu/settings-menu.js
zetagame: instead use model.get('state')

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -329,6 +335,12 @@ class SettingsMenu extends Menu {
             gearButton.setAttribute('aria-expanded', false);
         }
         
+        if (document.querySelector('.jw-breakpoint-0') || document.querySelector('.jw-breakpoint-1')) {
```
zetagame, line 338: same as first comment, don't use querySelector to check breakpoints

zetagame (changes requested):

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -314,6 +315,12 @@ class SettingsMenu extends Menu {
             gearButton.setAttribute('aria-expanded', true);
         }
 
+        if (getBreakpoint(this.model.get('containerWidth')) === 0 || getBreakpoint(this.model.get('containerWidth')) === 1) {
```
zetagame, line 318: Can reduce the number of calls to model in a couple ways, this one is probably the most performant: 

`if (getBreakPoint(this.model.get('containerWidth')) < 2)`

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -329,6 +336,12 @@ class SettingsMenu extends Menu {
             gearButton.setAttribute('aria-expanded', false);
         }
         
+        if (getBreakpoint(this.model.get('containerWidth')) === 0 || getBreakpoint(this.model.get('containerWidth')) === 1) {
```
zetagame, line 339: `if (getBreakPoint(this.model.get('containerWidth')) < 2)`

src/js/view/controls/components/menu/settings-menu.js
```diff
@@ -314,6 +315,12 @@ class SettingsMenu extends Menu {
             gearButton.setAttribute('aria-expanded', true);
         }
 
+        if (getBreakpoint(this.model.get('containerWidth')) === 0 || getBreakpoint(this.model.get('containerWidth')) === 1) {
+            const mediaState = this.model.get('state')
```
zetagame, line 319: A little redundant since it's only used once below. Might be better to just use this.mediaStateWhenOpened = this.model.get('state');

src/js/view/controls/components/menu/settings-menu.js
jwbrandon: Awesome, this is great to know about!

src/js/view/controls/components/menu/settings-menu.js
jwbrandon: We might want to just store the return value of getBreakpoint above all of the if statements here too.

jwbrandon (commented):
Code LGTM could use a test

jdunningjwp:
Opened a PR with the same changes but in the monorepo and also added tests
https://github.com/jwplayer/jwplayer-commercial/pull/8348

Dawolee (approved):

